PR #37, opened 2015-12-03
Improve travis

ljharb:
This tests more things and improves the travis.yml file. It also runs linting when it runs coverage, which it only does on one node build (currently 4.2, the LTS one).


lelandrichardson:
LGTM

PR #38, opened 2015-12-03
Fix the title for .get() in the docs

horaceko:
to: @lelandrichardson 

`.get()` returns a `ReactElement` and not a `ShallowWrapper`; the title in the docs indicated the former. Let's fix that!

### DIFF
--- a/docs/api/shallow.md
+++ b/docs/api/shallow.md
@@ -26,7 +26,7 @@ describe('<MyComponent />', () => {
     );
     expect(wrapper.contains(<div className="unique" />)).to.be.true;
   });
-  
+
   it('simulates click events', () => {
     const onButtonClick = sinon.spy();
     const wrapper = shallow(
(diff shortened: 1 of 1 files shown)


lelandrichardson:
:+1:

docs/api/shallow.md
```diff
@@ -91,7 +91,7 @@ Returns a string representation of the text nodes in the current render tree.
 #### [`.html() => String`](ShallowWrapper/html.md)
 Returns a static HTML rendering of the current node.
 
-#### [`.get(index) => ShallowWrapper`](ShallowWrapper/get.md)
+#### [`.get(index) => ReactElement`](ShallowWrapper/get.md)
```
ljharb: should the link be updated too?

horaceko: I don't think so. `get()` is still a method in the `ShallowWrapper` API, after all.

ljharb: yes but in case it's updated later, it should probably point to the correct one https://github.com/airbnb/reagent/blob/master/docs/api/ReactWrapper/get.md

horaceko: I defer to @lelandrichardson here. They do look to be completely separate implementations (though they have identical behavior), so I think it does make sense for them to have separate docs.

PR #53, opened 2015-12-05
Support for Stateless Components

lelandrichardson:
to: @ljharb @goatslacker 

Adds support for [stateless functional components](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#stateless-functional-components).

Brought up in https://github.com/airbnb/reagent/issues/45

Functional components cannot be a ref, and thus the "wrapper component" did not work with them.

Note that functional components are only available in react 0.14, so the tests for them are only run for react 0.14...

### DIFF
--- a/src/ReactWrapperComponent.jsx
+++ b/src/ReactWrapperComponent.jsx
@@ -19,10 +19,30 @@ export default class ReactWrapperComponent extends React.Component {
     return new Promise(resolve => this.setState(newProps, resolve));
   }
 
+  getInstance() {
+    const component = this._reactInternalInstance._renderedComponent;
+    const inst = component.getPublicInstance();
+    if (inst === null) {
+      throw new Error(
+        `You cannot get an instance of a stateless component.`
+      );
+    }
+    return inst;
+  }
+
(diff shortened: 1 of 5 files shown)


src/ReactWrapperComponent.jsx
```diff
@@ -19,10 +19,30 @@ export default class ReactWrapperComponent extends React.Component {
...
+        `You cannot get an instance of a stateless component.`
+      );
+    }
+    return inst;
+  }
+
+  getWrappedComponent() {
+    const component = this._reactInternalInstance._renderedComponent;
+    const inst = component.getPublicInstance();
+    if (inst === null) {
```
ljharb: Is there any case where `inst` is falsy but not null? if not, you could `return component.getPublicInstance() || component`? nbd

ljharb:
LGTM - this seems like a minor bump?

src/ReactWrapperComponent.jsx
lelandrichardson: AFAICT its always `null`

PR #57, opened 2015-12-06
Fix code tags in ReactWrapper update docs

silvenon:
http://airbnb.io/enzyme/docs/api/ReactWrapper/update.html

I will try to aggregate these miniature changes in less PRs in the future.


lelandrichardson:
:+1:

PR #59, opened 2015-12-07
Updating the example to import `sinon`

shidhincr:
Just to make sure that `sinon` was properly imported in the example code.

```
it('simulates click events', () => {
    const onButtonClick = sinon.spy();
    const wrapper = shallow(
      <Foo onButtonClick={onButtonClick} />
    );
    wrapper.find('button').simulate('click');
    expect(onButtonClick.calledOnce).to.be.true;
  });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Basic Usage
 
 ```jsx
 import { shallow } from 'enzyme';
+import sinon from 'sinon';
 
 describe('<MyComponent />', () => {
 


simplyianm:
There's more than just Sinon missing. (e.g., Foo, { expect })

lelandrichardson:
Yah some of these examples are sort of just to give a gist of what's going on but aren't necessarily full examples. I don't think it's a bad idea to make them 100% copy-and-paste-functional though, but we'd have to add a lot more to some of these :/

shidhincr:
@lelandrichardson Agree.. I would vote for to make them 100% copy-and-paste-functional, as most of the times users use the sample code from these examples to try out the library/framework.

lelandrichardson:
I'll go ahead and merge this since the PR to rip out sinon is going to take a little while to merge...

Issue #55, opened 2015-12-06
Jasmine support


just-boris:
As it follows from the docs

> Reagent is unopinionated regarding which test runner or assertion library you use, and should be compatible with all major test runners and assertion libraries out there.

I tried to use it with Jasmine and it worked fine until I tried to use `mount`. I got the error.

```
TypeError: before is not a function
```

`before` hook is the Mocha-specific thing, Jasmine has the similar thing named `beforeAll`.

Thanks for the awesome tool! It is really missing piece to my tests, but I can't use it due to lack of Jasmine support.

lelandrichardson:
@just-boris thanks for bringing this up. Can you post the code that is actually causing this to throw?

There are a couple of assumptions made in the library about mocha being present, however these are mostly undocumented features which are used internally at Airbnb but will likely be removed.

Running `mount()` on it's own should not be requiring a `before` function being present. The only thing I can find that would, actually, is the `useSetStateHack()` export?

I could be missing something though. A stack trace would be helpful.

just-boris:
The error has been caused by `describeWithDOM`. This is my usage example: 

```
describeWithDOM('<Component />', function() {
    it('should be rendered without errors', function() {
        const component = mount(<Component />)
    })
});
```

The `describeWithDOM` is necessary wrapper to use `mount`, as it follows from docs.

lelandrichardson:
@just-boris I see.  I think we should make some documentation changes to more accurately reflect what is necessary. There is a good chance that `describeWithDOM` will be removed from the library in a later release, as it is definitely mocha specific.

In order to use `mount`, you need only to have DOM API's present in the global scope.  If you would like to use `jsdom`, as long as you have the DOM loaded in the global scope when you make the call to `mount()`, everything should work.

the `describeWithDOM` function is used in conjunction with mocha for two reasons:

1. It re-instantiates a new DOM before every test so that tests don't have any side effects on other tests
2. It automatically skips the test if it is being run in an environment where `jsdom` cannot be loaded (ie, node < 4).

If you're running the tests in an environment with DOM API's, `mount` should "just work" without the use of `describeWithDOM`.

Let me know if that makes sense?

just-boris:
Yes, I can set up jsdom by myself and it seems to be working. 

[...]

Issue #47, opened 2015-12-04
Webpack build issues


kgoggin:
I'm trying to get enzyme working using a pretty standard webpack/karma setup. Webpack throws a bunch of errors related to the dependencies enzyme imports, including sinon, cheerio, and jsdom. Sinon gives webpack heartburn because it uses it's own `require` (which seems to be documented here: https://github.com/webpack/webpack/issues/304), and then there are a bunch of "Module not found" errors for jsdom and cheerio.

I'm importing enzyme as described in the docs and can't really think of anything else I might be missing, so I'm curious if anyone else is trying to use webpack and running into a similar issue, or if there's just something weird with my configuration. I'd expect to be able to just import it and have it work, but maybe I need to do something with those external libraries?

Here's my test file for reference:

``` javascript
import expect from 'expect';
import { shallow } from 'enzyme';
import React from 'react';
import Button from '../Button';

describe('Button', () => {
    it('should render children passed in', () => {
        const wrapper = shallow(
            <Button><div className="foo" /></Button>
        );

        expect(wrapper.contains(<div className="foo" />)).toBe(true);
    });
});
```

I won't paste the entire error log from webpack because it's pretty long, but here's a sample: 

```

WARNING in ./~/enzyme/~/jsdom/~/acorn/dist/acorn.js
Critical dependencies:
1:478-485 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/enzyme/~/jsdom/~/acorn/dist/acorn.js 1:478-485

WARNING in ./~/enzyme/~/jsdom/~/acorn/dist/walk.js
Critical dependencies:
1:503-510 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
 @ ./~/enzyme/~/jsdom/~/acorn/dist/walk.js 1:503-510

ERROR in ./~/enzyme/build/react-compat.js
Module not found: Error: Cannot resolve module 'react/lib/ExecutionEnvironment' in /Users/agoggin/www/vhosts/hammerhead/node_modules/enzyme/build
 @ ./~/enzyme/build/react-compat.js 22:2-43

ERROR in ./~/enzyme/~/cheerio/index.js
Module not found: Error: Cannot resolve module 'json' in /Users/agoggin/www/vhosts/hammerhead/node_modules/enzyme/node_modules/cheerio
 @ ./~/enzyme/~/cheerio/index.js 11:18-38

[...]

Issue #44, opened 2015-12-04
Remove direct dependency on Sinon


IanVS:
Why are Sinon methods being exported from this package (for example in https://github.com/airbnb/reagent/blob/master/src/index.js#L20)?  This poses a problem for us, as Sinon is throwing an exception on our codebase and we don't necessarily want to use Sinon and would rather have the flexibility to choose our own mocking/stubbing library.  I understand that it's perhaps a convenience to be able to [`useSinon`](https://github.com/airbnb/reagent/blob/master/src/index.js#L57-L60), but It seems outside of the scope of what Reagent is trying to accomplish, and it is trivial to set up Sinon ourselves.  Is there a chance you'll rethink this approach?  I'd really like to be able to use Reagent for testing, but right now this is preventing us from doing so.

ljharb:
It seems reasonable to move the `sinon` stuff into a `reagent-sinon` package which you could use if you wanted that behavior.

mudetroit:
I think it makes a ton of sense to not have sinon as part of the main package. I am curious what direct benefit a user of reagent would get from including sinon with reagent that wouldn't be there if you depended on sinon directly.

Additionally the current implementation makes assumptions about test framework lifecycle methods that might not make sense in every environment.

ljharb:
I'm not sure what assumptions you're talking about - could you please file a separate issue for that?

IanVS:
Not every test framework has a [`beforeEach` and `afterEach`](https://github.com/airbnb/reagent/blob/master/src/index.js#L58-L59).

mudetroit:
@ljharb not certain it is worth an issue. but `before` and `after` while available in most frameworks aren't present in them all. Qunit for example uses setup and teardown.

lelandrichardson:
The use of sinon in this project has been largely left undocumented because of the assumptions it's making about mocha being present (which is a vestige of this being used internally at airbnb).

I agree this should be done as it is somewhat separate from the meat of the library.

just-boris:
+1
I am using this with jasmine, and I don't need an extra library for spies, since I have a built-in one.

Sinewyk:
Considering the related PR was just closed with no comments, what does this mean for this issue ?

Do you still accept work for this or do you have other plans ?

lelandrichardson:
@Sinewyk this is still going to happen!  I just closed the PR because this branch was doing some mono-repo things that we've decided not to do.

This is definitely planned for 2.0!

[...]

PR #60, opened 2015-12-08
Rip out sinon & mocha into separate modules

lelandrichardson:
to: @ljharb @goatslacker 

This refactor addresses several issues regarding enzyme working in a variety of environments, and
the library just generally not making any assumptions about the environment that tests will be
run in.

For the most part, this amounts to:

- remove direct dependency on jsdom
- remove direct dependency on sinon
- remove assumed dependency on mocha

Additionally, this change moves the code that was removed into separate modules that are
defined in the packages folder. In this case we have added two modules: "enzyme-mocha" and
"enzyme-sinon".

This design will allow us to have one repo where all issues, code, and documentation
regarding enzyme are centralized, but modules are still separated at the NPM level. This
will allow us to make atomic releases across the modules more easily, as well as keep
all of the discussion in one place.

Left to do for this to be mergable is:

- [ ] Add a "guides" section in the docs explaining how to use enzyme in different environments
- [ ] Build a proper npm script to run an npm publish in the proper order for each module
- [ ] Add more meaningful tests for packages

Related Issues:

- Fixes https://github.com/airbnb/enzyme/issues/55
- Fixes https://github.com/airbnb/enzyme/issues/47
- Fixes https://github.com/airbnb/enzyme/issues/44


README.md
```diff
@@ -84,19 +89,11 @@ Read the full [API Documentation](/docs/api/shallow.md)
...
 ```jsx
-import {
-  describeWithDOM,
-  mount,
-  spyLifecycle,
-} from 'enzyme';
-
-describeWithDOM('<Foo />', () => {
+import React from 'react';
+import sionon from 'sinon';
```
kjbekkelund: typo "sionon"

docs/api/guides/mocha.md
```diff
@@ -0,0 +1,21 @@
...
+```jsx
+import { mount } from 'enzyme';
+import { describeWithDOM } from 'enzyme-mocha';
+
+describe('<Foo />', () => {
+
+  it('calls componentDidMount', () => {
+    spy(Foo.prototype, 'componentDidMount');
+    const wrapper = mount(<Foo />);
+    expect(Foo.prototype.componentDidMount.calledOnce).to.be.true;
```
ljharb: should this be `expec(foo).toHaveBeenCalled.once` or something similar?

docs/api/mount.md
```diff
@@ -4,8 +4,11 @@ Full DOM rendering is ideal for use cases where you have components that may int
 or may require the full lifecycle in order to fully test the component (ie, `componentDidMount` 
 etc.)
 
-Full DOM rendering  depends on a library called [jsdom](https://github.com/tmpvar/jsdom) which is
-essentially a headless browser implemented completely in JS.
+Full DOM rendering  requires that a full DOM API be available at the global scope. This means that
```
ljharb: extra space after "rendering"

package.json
```diff
@@ -9,11 +9,13 @@
...
-    "lint": "eslint src/**",
-    "test": "mocha --compilers js:babel/register --recursive src/**/__tests__/*.js",
+    "lint": "eslint src/** packages/**/src/**",
     "check": "npm run lint && npm run test:all",
-    "build": "babel src --out-dir build",
-    "test:watch": "mocha --compilers js:babel/register --recursive src/**/__tests__/*.js --watch",
+    "build": "npm run build:core && npm run build:packages",
+    "build:core": "babel src --out-dir build",
+    "build:packages": "bash ./scripts/build-packages.sh",
+    "test": "mocha --compilers js:babel/register test/*.js packages/**/test/*.js",
```
ljharb: since packages have their own tests, this should probably run `npm test` in that directory, rather than baking subpackage-specific paths into the root package.json.

in other words, the packages should be as separable as possible, such that the decision to put them in a separate repo or not can be made and changed without impacting much.

packages/enzyme-mocha/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "mocha",
+    "enzyme"
+  ],
+  "author": "Leland Richardson <leland.richardson@airbnb.com>",
+  "license": "MIT",
+  "dependencies": {
+    "jsdomify": "^1.0.2"
+  },
+  "peerDependencies": {
+    "react": "0.13.x || 0.14.x"
```
ljharb: this should have a peerDep of enzyme itself

packages/enzyme-sinon/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "Make Sinon and Enzyme play nice",
+  "main": "build/index.js",
+  "repository": "https://github.com/airbnb/enzyme/tree/master/packages/enzyme-mocha",
+  "keywords": [
+    "sinon",
+    "enzyme"
+  ],
+  "dependencies": {
+    "sinon": "^1.17.2"
+  },
```
ljharb: this should have a peerDep of enzyme itself

scripts/build-packages.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env zsh
+# this cookie was getting too long for a JSON prop
+# builds the Phoenix yavascripst from src to lib
+
+# die on error
+setopt -e
+
+for pkg in packages/*/ ; do
+  babel ${pkg}/src --out-dir ${pkg}/build
```
ljharb: it'd be nicer imo if this ran `npm run build` or the equivalent in each folder, so it can delegate to its npm run-scripts

ljharb:
> This design will allow us to have one repo where all issues, code, and documentation
regarding enzyme are centralized, but modules are still separated at the NPM level.

I'm not sure I agree that we'll see this as valuable. Issues with mocha-enzyme *don't belong* on the main enzyme repo, because not all people looking at enzyme will care about mocha nor want to see them. Similarly, people looking at enzyme-mocha issues won't want to see enzyme-jasmine issues. "non-centralized documentation" isn't a problem one way or the other since things can just link to each others' docs.

> This will allow us to make atomic releases across the modules more easily, as well as keep
all of the discussion in one place.

I'm not sure how cd-ing to different dirs and running `npm publish` is different whether those dirs are inside the main repo, or in separate repos. What discussions do you anticipate in a separate-repo world that won't belong in the main repo, or in any of the separate repos, and thus won't have an obvious place to be held?

docs/api/guides/mocha.md
lelandrichardson: @ljharb it would be if we were using the sinon-chai plugin, but that would be assuming yet-another-library for this example, and should maybe be avoided?

ljharb: hm, i guess i assumed that "enzyme-mocha" would have mocha-specific matchers, but perhaps that would require another module altogether. nvm then :-)

packages/enzyme-mocha/package.json
lelandrichardson: So I was going to add it and realized... it doesn't actually need enzyme.  I'm not sure we want to have a dependency that isn't real?

It begs the question about whether or not these small little utilities are even worth it... but I think we will be able to add more over time.

ljharb: it does depend on a global `describe` however :-/ so maybe it does need mocha?

I think it makes sense for a package called "foo-bar" to require foo and bar, even if they're not actually required. The goal is to prevent user mistakes, and when would someone want to install enzyme-mocha without both enzyme and mocha?

packages/enzyme-mocha/src/index.js
```diff
@@ -0,0 +1,33 @@
...
+  jsdomify.create();
+}
+
+export function jsdomTeardown() {
+  if (!jsdomify) return;
+  jsdomify.destroy();
+}
+
+export function describeWithDOM(a, b) {
+  describe('(uses jsdom)', () => {
```
ljharb: what happens when this is imported outside a test context, and then cached with the global `describe` set to `undefined`? I think perhaps this export may need to either take `describe` as an argument, or, actually be a function that takes `describe` and returns `describeWithDOM`?

packages/enzyme-mocha/src/index.js
```diff
@@ -0,0 +1,33 @@
...
+    if (typeof jsdom === 'function') {
+      beforeEach(jsdomSetup);
```
ljharb: these globals are also implicitly depended on

docs/api/guides/mocha.md
lelandrichardson: I'm not sure what you mean, but those matchers are more "sinon-specific matchers" than "mocha-specific matchers"

packages/enzyme-mocha/src/index.js
lelandrichardson: we aren't caching describe anywhere? Every time `describeWithDOM` is called, it uses the `describe` that is in the global namespace.

ljharb: the function retains a lexical reference to the one that's in the global namespace when the function is defined, it's not freshly looked up. do you mean this to be `global.describe` then?

Issue #49, opened 2015-12-05
Add support for passing context


laurakelly:
I have a component in my app that has a router passed to it via [context](https://facebook.github.io/react/docs/context.html): 

`const contextTypes = {
  router: React.PropTypes.func.isRequired,
};`

Reagent tests currently do not support passing in a context and will throw this error:

``` 
Error: expected zero console.warn calls, got 1:
      warn(Warning: Failed Context Types: Required context `router` was not specified in `MyComponent`.) at warning (node_modules/react/lib/warning.js:48:15)
    at Context.<anonymous> (spec/javascripts/_failTestsOnErrors.js:21:23)
```

It would be great to be able to pass context into `shallow`.

lelandrichardson:
I think this would be a great addition! The only remaining question is a question of what the right API is.

I propose that we add a second argument to `mount`, `shallow`, and `render` which is simply an object named `options` in order to be most flexible with future changes.

`options` could have a `context` property that, if defined, would be passed in as context to the parent component.

```jsx
const context = { router: mockReactRouter() };
const wrapper = shallow(<MyComponent />, { context });
// ...
```

cc @ljharb @goatslacker thoughts?

blainekasten:
That is the same API, `TestUtils.shallowRenderer().render` uses. So I would second that and stay consist when possible. :+1:

lelandrichardson:
@blainekasten to be clear, I'm recommending that on object hash with a `context` prop be passed in as a second argument to `shallow` and the like, NOT that the context object itself be the second argument. The latter, I believe, is the current API for the shallowRenderer

blainekasten:
My bad. Why that approach? What other keys would be used and why?

lelandrichardson:
My thinking is that it allows other "options" be passed in as well in the future, rather than just having a bunch of arguments at random slots of the function signature.

One such thing we discussed internally was the ability to have an option for `mount` which indicated the node it mounts onto actually be attached to the document body:

```jsx
const wrapper = mount(<Foo />, { attach: true });
```

Which could come in handy if you're wanting to test a component that uses DOM api's such as `elem.getBoundingClientRect();` that require the DOM nodes be attached to a document.

lelandrichardson:
@laurakelly context support is now available on enzyme v1.2.0

laurakelly:
@lelandrichardson Great, thank you! I'll try it out

havenchyk:
[...]

PR #62, opened 2015-12-09
Added context support

lelandrichardson:
to: @ljharb @goatslacker 

Adds support for context with `mount` and `shallow`.

The API looks like:

```jsx
const wrapper = mount(<Component />, { context });
```

```jsx
const wrapper = shallow(<Component />, { context });
```

Additionally, there is a `.setContext(newContext)` prototype method on each wrapper class.

Fixes https://github.com/airbnb/enzyme/issues/49

Note: this will constitute a minor change (new feature, backwards compatible)


docs/api/ReactWrapper/setContext.md
```diff
@@ -0,0 +1,55 @@
...
+expect(wrapper.text()).to.equal('foo');
+wrapper.setContext({ name: 'bar' });
+expect(wrapper.text()).to.equal('bar');
+wrapper.setContext({ name: 'baz' });
+expect(wrapper.text()).to.equal('baz');
+```
+
+#### Common Gotchas
+
+- `.setContext()` can only be used on a wrapper that was initially created with a call to `mount()`
```
ljharb: Is this guaranteed, ie, will it throw an exception? I don't see any tests that cover this case.

docs/api/ShallowWrapper/setContext.md
```diff
@@ -0,0 +1,55 @@
...
+expect(wrapper.text()).to.equal('foo');
+wrapper.setContext({ name: 'bar' });
+expect(wrapper.text()).to.equal('bar');
+wrapper.setContext({ name: 'baz' });
+expect(wrapper.text()).to.equal('baz');
+```
+
+#### Common Gotchas
+
+- `.setContext()` can only be used on a wrapper that was initially created with a call to `mount()`
```
ljharb: the example immediately above has a wrapper initially created with a call to shallow, not mount

src/ReactWrapperComponent.jsx
```diff
@@ -8,45 +8,75 @@ import React, { PropTypes } from 'react';
...
+
+    getInitialState() {
+      return {
+        props: this.props.props,
+        context: this.props.context,
+      };
+    },
+
+    setChildProps(props) {
+      return new Promise(resolve => this.setState({ props }, resolve));
```
ljharb: not part of this PR, but why doesn't `this.setState` return a promise rather than taking a callback? is that to mirror react's API?

docs/api/ReactWrapper/setContext.md
lelandrichardson: good call. will add some tests.

docs/api/ShallowWrapper/setContext.md
lelandrichardson: good catch.

src/ReactWrapperComponent.jsx
lelandrichardson: because `this.setState()` can or maybe will be an asynchronous call... but it doesn't seem to be in react 0.13 or 0.14...  originally I had set up the API in such a way that it returned a promise, but it doesn't appear to be needed in most cases (and is thus not used right now). This promise code could be removed (maybe separate PR)

ljharb: I like everything returning a Promise :-) it's future-compatible with async/await!

lelandrichardson:
@ljharb tests added. comments addressed.

ljharb:
Changes LGTM - @goatslacker should weigh in on the context stuff tho since he's got more experience with it than I

lelandrichardson:
Looking at the coverage info with the line numbers mismatched is a PITA. I'm going to submit a followup PR that fixes istanbul line numbers, as well as improve coverage where needed

src/ReactWrapperComponent.jsx
```diff
@@ -8,45 +8,75 @@ import React, { PropTypes } from 'react';
...
-    if (inst === null) {
-      return component;
-    }
-    return inst;
+  if (options.context && node.type.contextTypes) {
+    // For full rendering, we are using this wrapper component to provide context if it is
+    // specified in both the options AND the child component defines `contextTypes` statically.
+    // In that case, we define both a `getChildContext()` function and a `childContextTypes` prop.
+    Object.assign(spec, {
+      childContextTypes: node.type.contextTypes,
```
goatslacker: You're mutating spec here. Is that ok? Seems like a bad idea

goatslacker: Yeah it's ok.

goatslacker:
LGTM

src/ReactWrapper.js
```diff
@@ -54,10 +54,12 @@ export default class ReactWrapper {
     }
 
     if (!root) {
+      const ReactWrapperComponent = createWrapperComponent(nodes, options);
       this.component = renderIntoDocument(
         <ReactWrapperComponent
           Component={nodes.type}
           props={nodes.props}
+          context={options.context}
           />
```
blainekasten: not due to this diff, but the airbnb styleguide shows this closing wrapper to be back 2 spaces. https://github.com/airbnb/javascript/tree/master/react#alignment

goatslacker: tru dat

kujon:
Super useful PR. May I ask what is the reason for `render` not having this functionality?

Issue #64, opened 2015-12-10
id selector does not seem to work


clouddra:
Was trying to test a simple piece of code:
```javascript
const renderShallow = shallow(<div id='ttt' className='ttt'>hello</div>);
console.log(renderShallow.find('#ttt')); // not found
console.log(renderShallow.find('.ttt')); // found
```
and it returns an empty array when i tried to use an id selector. It works fine on class selectors.

clouddra:
Just to add, I had the following configuration in my karma config. Not sure if it affects for shallow rendering.
```
externals: {
        'jsdom': 'window',
        'cheerio': 'window',
        'react/lib/ExecutionEnvironment': true,
}
```

lelandrichardson:
Hi @clouddra,  can you confirm that the following code works?

```jsx
const wrapper = shallow(
  <div>
    <div id='ttt' className='ttt'>hello</div>
  </div>
);
console.log(wrapper.find('#ttt').length); // 1
```

There may be some edge-case behavior when the id you are looking for is on the root node... especially if you're using React 0.14, where shallow rendering of a DOM Element as opposed to a composite component is behavior shimmed by enzyme rather than built in to the shallow renderer.

lelandrichardson:
Confirmed the issue and fixed. Will publish to npm later today when this lands.

lelandrichardson:
@clouddra  The fix is now available in v1.2.0

clouddra:
@lelandrichardson That was fast! Tested it and it works fine. Thanks a lot for your help.

PR #65, opened 2015-12-10
[bugfix] nodeHasId has issue with 0.14 in some cases

lelandrichardson:
to: @ljharb @goatslacker 

The spec that was added in this PR would succeed in React 0.13 but fail in React 0.14 due to some changes in behavior from the two. This had been abstracted away for some selectors (like class name) through the correct `propsOfNode(node)` function, but the `nodeHasId` function was still prone to these issues. 

Thanks to @clouddra for pointing out.

Fixes https://github.com/airbnb/enzyme/issues/64

### DIFF
--- a/src/ShallowTraversal.js
+++ b/src/ShallowTraversal.js
@@ -63,8 +63,7 @@ export function parentsOfNode(node, root) {
 }
 
 export function nodeHasId(node, id) {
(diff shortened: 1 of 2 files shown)


ljharb:
LGTM

PR #70, opened 2015-12-12
Support Property Selector

blainekasten:
Per discussion in #4 and continued discussion here. This PR adds support for querying react properties, and not html attributes. The following syntax works:

* `component.find('input[type="text"][value="1"])`
* `component.find('label[htmlFor="button"]')`
* `component.find('div[data-wrapper]')`

### DIFF
--- a/docs/api/selector.md
+++ b/docs/api/selector.md
@@ -12,6 +12,24 @@ follows:
 - class syntax (`.foo`, `.foo-bar`, etc.)
 - tag syntax (`input`, `div`, `span`, etc.)
 - id syntax (`#foo`, `#foo-bar`, etc.)
+- prop syntax (`[htmlFor="foo"]`, `[bar]`, `[baz=1]`, etc.);
+
+**Note -- Prop selector**
+Strings, numeric literals and boolean property values are supported for prop syntax
+in combination of the expected string syntax. For example, the following
+is supported:
+
+```js
+const wrapper = mount(
+  <div>
+    <span foo={3} bar={false} title="baz" />
+  </div>
+)
+
+wrapper.find('[foo=3]')
+wrapper.find('[bar=false]')
+wrapper.find('[title="baz"]')
+```
 
 Further, enzyme supports combining any of those supported syntaxes together to uniquely identify a
 single node.  For instance:
@@ -19,6 +37,7 @@ single node.  For instance:
 ```css
 div.foo.bar
 input#input-name
+label[foo=true]
 ```
 
 Are all valid selectors in enzyme.  At this time, however, any contextual CSS selector syntax that
--- a/src/MountedTraversal.js
+++ b/src/MountedTraversal.js
@@ -1,10 +1,14 @@
 import {
+  coercePropValue,
   nodeEqual,
+  propsOfNode,
   isSimpleSelector,
   splitSelector,
   selectorError,
+  selectorType,
   isCompoundSelector,
   AND,
+  SELECTOR,
 } from './Utils';
 import {
   isDOMComponent,
@@ -63,6 +67,26 @@ export function instHasType(inst, type) {
     default:
       return false;
   }
+}
+
+export function instHasProperty(inst, propKey, stringifiedPropValue) {
+  if (!isDOMComponent(inst)) return false;
+  const node = getNode(inst);
+  const nodeProps = propsOfNode(node);
+  const nodePropValue = nodeProps[propKey];
+
+  const propValue = coercePropValue(stringifiedPropValue);
+
+  // intentionally not matching node props that are undefined
+  if (nodePropValue === undefined) {
+    return false;
+  }
+
+  if (propValue) {
+    return nodePropValue === propValue;
+  }
+
+  return nodeProps.hasOwnProperty(propKey);
 }
 
 // called with private inst
@@ -166,15 +190,22 @@ export function buildInstPredicate(selector) {
       if (isCompoundSelector.test(selector)) {
         return AND(splitSelector(selector).map(buildInstPredicate));
       }
-      if (selector[0] === '.') {
-        // selector is a class name
-        return inst => instHasClassName(inst, selector.substr(1));
-      } else if (selector[0] === '#') {
-        // selector is an id name
-        return inst => instHasId(inst, selector.substr(1));
+
+      switch (selectorType(selector)) {
+        case SELECTOR.CLASS_TYPE:
+          return inst => instHasClassName(inst, selector.substr(1));
+        case SELECTOR.ID_TYPE:
+          return inst => instHasId(inst, selector.substr(1));
+        case SELECTOR.PROP_TYPE:
+          const propKey = selector.split(/\[([a-zA-Z\-\:]*?)(=|\])/)[1];
+          const propValue = selector.split(/=(.*?)]/)[1];
+
+          return node => instHasProperty(node, propKey, propValue);
+        default:
+          // selector is a string. match to DOM tag or constructor displayName
+          return inst => instHasType(inst, selector);
       }
-      // selector is a string. match to DOM tag or constructor displayName
-      return inst => instHasType(inst, selector);
+      break;
 
     default:
       throw new TypeError('Expecting a string or Component Constructor');
--- a/src/ShallowTraversal.js
+++ b/src/ShallowTraversal.js
@@ -1,12 +1,16 @@
 import React from 'react';
 import {
+  coercePropValue,
   propsOfNode,
   isSimpleSelector,
   splitSelector,
   selectorError,
   isCompoundSelector,
+  selectorType,
   AND,
+  SELECTOR,
 } from './Utils';
+
 
 export function childrenOfNode(node) {
   if (!node) return [];
@@ -66,6 +70,24 @@ export function nodeHasId(node, id) {
   return propsOfNode(node).id === id;
 }
 
+
+export function nodeHasProperty(node, propKey, stringifiedPropValue) {
+  const nodeProps = propsOfNode(node);
+  const propValue = coercePropValue(stringifiedPropValue);
+  const nodePropValue = nodeProps[propKey];
+
+  if (nodePropValue === undefined) {
+    return false;
+  }
+
+  if (propValue) {
+    return nodePropValue === propValue;
+  }
+
+  return nodeProps.hasOwnProperty(propKey);
+}
+
+
 export function nodeHasType(node, type) {
   if (!type || !node) return false;
   if (!node.type) return false;
@@ -86,20 +108,31 @@ export function buildPredicate(selector) {
       if (isCompoundSelector.test(selector)) {
         return AND(splitSelector(selector).map(buildPredicate));
       }
-      if (selector[0] === '.') {
-        // selector is a class name
-        return node => hasClassName(node, selector.substr(1));
-      } else if (selector[0] === '#') {
-        // selector is an id name
-        return node => nodeHasId(node, selector.substr(1));
+
+      switch (selectorType(selector)) {
+        case SELECTOR.CLASS_TYPE:
+          return node => hasClassName(node, selector.substr(1));
+
+        case SELECTOR.ID_TYPE:
+          return node => nodeHasId(node, selector.substr(1));
+
+        case SELECTOR.PROP_TYPE:
+          const propKey = selector.split(/\[([a-zA-Z\-]*?)(=|\])/)[1];
+          const propValue = selector.split(/=(.*?)\]/)[1];
+
+          return node => nodeHasProperty(node, propKey, propValue);
+        default:
+          // selector is a string. match to DOM tag or constructor displayName
+          return node => nodeHasType(node, selector);
       }
-      // selector is a string. match to DOM tag or constructor displayName
-      return node => nodeHasType(node, selector);
+      break;
+
 
     default:
       throw new TypeError('Expecting a string or Component Constructor');
   }
 }
+
 
 export function getTextFromNode(node) {
   if (node === null || node === undefined) {
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -102,12 +102,12 @@ export function withSetStateAllowed(fn) {
 }
 
 export function splitSelector(selector) {
-  return selector.split(/(?=\.)/);
+  return selector.split(/(?=\.|\[.*\])/);
 }
 
 export function isSimpleSelector(selector) {
   // any of these characters pretty much guarantee it's a complex selector
-  return !/[~\s\[\]:>]/.test(selector);
+  return !/[~\s:>]/.test(selector);
 }
 
 export function selectorError(selector) {
@@ -116,8 +116,25 @@ export function selectorError(selector) {
   );
 }
 
-export const isCompoundSelector = /[a-z]\.[a-z]/i;
+export const isCompoundSelector = /([a-z]\.[a-z]|[a-z]\[.*\])/i;
 
+const isPropSelector = /^\[.*\]$/;
+
+export const SELECTOR = {
+  CLASS_TYPE: 0,
+  ID_TYPE: 1,
+  PROP_TYPE: 2,
+};
+
+export function selectorType(selector) {
+  if (selector[0] === '.') {
+    return SELECTOR.CLASS_TYPE;
+  } else if (selector[0] === '#') {
+    return SELECTOR.ID_TYPE;
+  } else if (isPropSelector.test(selector)) {
+    return SELECTOR.PROP_TYPE;
+  }
+}
 
 export function AND(fns) {
   return x => {
@@ -127,4 +144,27 @@ export function AND(fns) {
     }
     return true;
   };
+}
+
+export function coercePropValue(propValue) {
+  // can be undefined
+  if (propValue === undefined) {
+    return propValue;
+  }
+
+  // if propValue includes quotes, it should be
+  // treated as a string
+  if (propValue.search(/"/) !== -1) {
+    return propValue.replace(/"/g, '');
+  }
+
+  const numericPropValue = parseInt(propValue, 10);
+
+  // if parseInt is not NaN, then we've wanted a number
+  if (!isNaN(numericPropValue)) {
+    return numericPropValue;
+  }
+
+  // coerce to boolean
+  return propValue === 'true' ? true : false;
 }
--- a/src/__tests__/ReactWrapper-spec.js
+++ b/src/__tests__/ReactWrapper-spec.js
@@ -122,6 +122,100 @@ describeWithDOM('mount', () => {
       expect(wrapper.find(Foo).type()).to.equal(Foo);
     });
 
+    it('should find component based on a react prop', () => {
+      const wrapper = mount(
+        <div>
+          <span htmlFor="foo" />
+        </div>
+      );
+
+      expect(wrapper.find('[htmlFor="foo"]')).to.have.length(1);
+      expect(wrapper.find('[htmlFor]')).to.have.length(1);
+    });
+
+    it('should compound tag and prop selector', () => {
+      const wrapper = mount(
+        <div>
+          <span htmlFor="foo" />
+        </div>
+      );
+
+      expect(wrapper.find('span[htmlFor="foo"]')).to.have.length(1);
+      expect(wrapper.find('span[htmlFor]')).to.have.length(1);
+
+    });
+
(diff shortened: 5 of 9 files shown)


src/Utils.js
```diff
@@ -128,3 +136,46 @@ export function AND(fns) {
     return true;
   };
 }
+
+export const attributePropMap = {
```
goatslacker: curious is there an npm module we can extract these from?

blainekasten: I did find one. The only potential problem is that react supports some non-standard attributes, which aren't covered in the package I found, and likely wouldn't be covered by other packages.

docs/api/selector.md
```diff
@@ -12,6 +12,16 @@ follows:
 - class syntax (`.foo`, `.foo-bar`, etc.)
 - tag syntax (`input`, `div`, `span`, etc.)
 - id syntax (`#foo`, `#foo-bar`, etc.)
+- attr syntax(`[foo="bar"]`, `[foo]`, etc.);
+
+A note on attribute selectors: For compatibilty, you can use either the
+react prop name, or the native html selector. For example, the following
```
ljharb: I still see this as a horribly bad API decision. `[]` means "HTML attributes", not "React component props" - the two aren't the same thing. Wouldn't the code be a lot simpler to use a different syntax instead of smashing two things into one existing syntax?

ljharb: In addition, this is a decision that will be almost impossible to unmake - whereas introducing a different syntax now won't in any way prevent bloating the attribute selector syntax later (and just using the same code) if that's what we decide we want.

blainekasten: cc @lelandrichardson 

I'm open for changing either way.
@ljharb what about the context of different rendering layers. Think ReactNative, ReactHardware. In those contexts, do we just say the `[]` isn't for them? I think leland had a good point that the `[]` selector was built for XML, not explicitly the DOM. so in a ReactNative/Hardware standpoint, the Props *are* the attributes of the "rendered" xml

ljharb: as I understand it (@lelandrichardson can confirm) enzyme was originally built to replicate much of jQuery's API, which is why the querySelectorAll syntax is used for the selector string. If we're going to be deviating from it, for React Native or otherwise, that's a bigger discussion - does React Native itself use selector syntax?

src/MountedTraversal.js
```diff
@@ -65,6 +67,33 @@ export function instHasType(inst, type) {
...
+   * we could gain from the `getAttribute` `HTMLElement` api, but
+   * for compatibility between `shallow` and `mount` We'll stick
+   * prop look up with a compatiblity check for the attr/react
+   * prop naming.
+   */
+  if (!node.props) { // React 0.13
+    instProps = node._store.props;
+  } else {
+    instProps = node.props;
+  }
```
lelandrichardson: there should be a `propsOfNode()` function that abstracts this away... it's also important to check for falsy nodes as well where this would probably throw.

blainekasten: :+1:

docs/api/selector.md
hartzis: @ljharb I can understand your hesitation with the API decision. We solved something similar in the testing library [skin-deep](https://github.com/glenjamin/skin-deep/blob/master/test/test.js#L562) with `tree.subTreeLike('*', {foo: 'bar'})`.

We wanted to query for specific components using props, so we could use specific testing properties like `data-testref='foo'` that didn't interfere with our ui teams use of classnames.

cc @lelandrichardson @blainekasten

blainekasten: I started on an implementation to use seperate syntax (`{}` for props, `[]` for html attr). It seems feasible. In shallow renders, I'm basically faking it by mapping the html attr selector to the matching react prop since I don't get an actual render. I'm just waiting on word from @lelandrichardson if that is the way we want this togo.

blainekasten: I just pushed up the changes that @ljharb requested. Hopefully this can help the conversation move along.

looking for feedback from @lelandrichardson and @ljharb

src/MountedTraversal.js
```diff
@@ -166,15 +194,27 @@ export function buildInstPredicate(selector) {
...
+
+    switch (selectorType(selector)) {
+    case SELECTOR.CLASS_TYPE:
       return inst => instHasClassName(inst, selector.substr(1));
-    } else if (selector[0] === '#') {
-      // selector is an id name
+    case SELECTOR.ID_TYPE:
       return inst => instHasId(inst, selector.substr(1));
+    case SELECTOR.ATTR_TYPE:
+      const attrKey = selector.split(/\[([a-z\-\:]*?)(=|\])/)[1];
```
0xR: are we sure attributes can't have capitals?

blainekasten: Good catch. That is actually a bug now that I think about it. svg introduced a lot of camelCase attributes, like `preserveAspectRatio`

src/MountedTraversal.js
```diff
@@ -166,15 +194,27 @@ export function buildInstPredicate(selector) {
...
+      const attrValue = selector.split(/"(.*?)"/)[1];
+
+      return node => instHasAttribute(node, attrKey, attrValue);
+    case SELECTOR.PROP_TYPE:
+      const propKey = selector.split(/\{([a-zA-Z]*?)(=|\})/)[1];
```
blainekasten: this should probably also include `-`'s for `data-*` props

lelandrichardson:
wow, thanks for all this work!  I'm going to try and take a closer look tonight after we finish up this hackathon. Looks great!

lelandrichardson:
Hey @blainekasten ,

I hate to give you more work to do... but i was hoping you could change this back to the original `[prop]` syntax. Jordan and I discussed the pros/cons of each syntax and we agreed that handling just `[prop]` and getting rid of `{prop}` syntax.

To recap, the syntax we would like is:

- `[prop="foo"]` gets nodes with `prop="foo"`
- `[foo]` gets nodes that have a `foo` prop defined
- there will be no special mapping between props and html attributes such as `htmlFor` => `for`.

src/Utils.js
smacker: how about take it from react itself? You can import HTMLDOMPropertyConfig and SVGDOMPropertyConfig from renderers/dom/shared. I think it's better, because when react team add new property it starts working here without any changes.

blainekasten:
No worries. I want to see this land. I can do that. I agree with that API also honestly. Glad to have it sorted out. I'll finish it tomorrow :+1:

src/Utils.js
blainekasten: This will be going away entirely actually.

blainekasten:
So. This is done. But coverage keeps getting worse as I add more tests... Any tips @lelandrichardson ?

lelandrichardson:
wow, good work on the tests! Istanbul is becoming a bit of a pain since we are using babel... I need to figure out how to get istanbul working with the original source lines so that it makes more sense. just ignore for now if the percentage goes down.

I'm curious of a couple of things:

1. What's the behavior of numeric proptypes? Seems like something that we could also use the string selector for (or not?). Either way, the expectations should be tested.

```jsx
const wrapper = shallow(<div><Foo bar={2} /></div>);
expect(wrapper.find('[bar="2"]')).to.have.length(1);
```
Does the above pass?  (I'm not necessarily saying it should)

src/Utils.js
```diff
@@ -116,7 +116,26 @@ export function selectorError(selector) {
   );
 }
 
-export const isCompoundSelector = /[a-z]\.[a-z]/i;
+export const isCompoundSelector = /([a-z]\.[a-z]|[a-z]\[.*\])/i;
```
lelandrichardson: will this always work?  what about `[foo][bar]`?

src/__tests__/ReactWrapper-spec.js
```diff
@@ -122,6 +122,53 @@ describeWithDOM('mount', () => {
...
+
+      expect(wrapper.find('[data-foo="bar"]')).to.have.length(1);
+      expect(wrapper.find('[data-foo]')).to.have.length(1);
+    });
+
+    it('should find components with multiple matching props', () => {
+      const onChange = () => {};
+      const wrapper = mount(
+        <div>
+          <span htmlFor="foo" onChange={onChange}preserveAspectRatio="xMaxYMax" />
```
lelandrichardson: missing whitespace?

lelandrichardson:
Also, what is the expected behavior for falsy but present props?

- does `<div foo="" />` match `[foo]`?
- does `<Foo visible={false} />` match `[visible]`?

src/ShallowTraversal.js
```diff
@@ -66,6 +69,18 @@ export function nodeHasId(node, id) {
   return propsOfNode(node).id === id;
 }
 
+
+export function nodeHasProperty(node, propKey, propValue) {
+  const nodeProps = propsOfNode(node);
+
+  if (propValue) {
+    return nodeProps[propKey] === propValue;
```
lelandrichardson: i guess my questions boil down to: should this be `==` or `===`?

src/ShallowTraversal.js
```diff
@@ -66,6 +69,18 @@ export function nodeHasId(node, id) {
...
+  }
+
+  return nodeProps.hasOwnProperty(propKey);
```
lelandrichardson: and, should this be `nodeProps.hasOwnProperty(propKey)` or `!!nodeProps[propKey]`

src/Utils.js
blainekasten: It does, I tested just tested that exact scenario, but also this test effectively proves it also, https://github.com/airbnb/enzyme/pull/70/files#diff-e08e795cdfa46e03ea97df132b57d75aR167

blainekasten:
Ahh good catch. I didn't think about non-string values.. I'll have to work on this a bit more.

Empty string currently works fine:

`<div foo="" />` matches `[foo]` or `[foo=""]`
`<Foo visible={false} />` DOES match `[visible]`, not sure if that's the right call or not?

I guess in the situation of non string prop values, I would think dropping the quotes would be the way to do it.

So `<Foo visible={false} />` could match `[visible=false]` but not `[visible="false"]`

src/ShallowTraversal.js
blainekasten: I haven't ran into a situation where `hasOwnProperty` didn't work. Can you think of a test case?

src/ShallowTraversal.js
blainekasten: That doesn't work for situations like `'false'` or `'true'`

lelandrichardson:
@ljharb what do you think? How should empty values be handled?  should `[foo]` match `<div foo="" />`?

I'm trying to think pragmatically, and it seems to me that if i look for `[visible]`, I do NOT want `<Foo visible={false} />` to match...

src/ShallowTraversal.js
blainekasten: In fact, I just tested this with the new changes and it doesn't work in this situation.

```js
wrapper = <div foo={false} />

wrapper.find('[foo=false]')
```
In that situation that turns into 
```js
!!wrapper['foo'] // false
```
which then says it doesn't have the prop, though it does. So hasOwnProperty seems to be the way to go.

lelandrichardson: this is basically the `foo={false}` question. If it has a `foo` property but it's set to something falsy, do we still want to return true?

src/ShallowTraversal.js
lelandrichardson: yah... so we may need some special case logic put in here. let's try and figure out what the corner cases are and what the desired behavior is...

ljharb: `===`. A prop of `3` and a prop of `'3'` are different and shouldn't be selected in the same group.

blainekasten:
If we want to stick to `querySelector` syntax, a query of `[foo]` only looks for a node that has that attribute, irrelevant of the value.

Some new code that I have you could do this.

`[visible=false]` if you want the non-visible things.

src/Utils.js
lelandrichardson: My testing shows that it doesnt?  `div[foo][bar]` matches but `[foo][bar]` does not

ljharb:
That's a very good question - since this is a string-based API, it will already only work for a subset of prop values (and won't work for `Symbol` props). I think we can just draw a line and say that for this limited form of `findWhere`/`filterWhere`, `[foo]` should not match falsy values, and `[foo=${bar}]` should mean that a `bar` of a non-string value should never match anything.

ljharb:
Alternatively, I'd be totally fine with saying that `[foo]` matches any non-`undefined` value - and that might be more in line with people's expectations.

lelandrichardson:
I think I like `[foo]` matching non-undefined-values.

That just leaves the non-string-based values...  in this case we could say that things have to be `===` to each other...  (so `[foo="2"]` would NOT match `<Foo foo={2} />`).

ljharb:
agreed. I think non-string values is for `xWhere` to handle.

lelandrichardson:
Let's also add two tests describing its behavior for `<Foo ref="foo" />` and `<Foo key={1} />` since those are not treated the same as other props in react...

blainekasten:
Feeling a little lost by the string of comments. Can you drop me a few examples of what tests you'd want to pass/fail?

blainekasten:
I don't think adding support for `ref` makes much sense since React is moving away from the string ref api. It's almost guaranteed that in the coming months/year that react won't support a string for ref and only accept a fn.

lelandrichardson:
Agreed. I don't want to support it... I just want to add a test that asserts that it's not supported so we know that that doesn't change at some point

lelandrichardson:
Here's a list of JSX => selector => match/ not match:

Without values:
- `<Foo foo="bar" />` => `[foo]` => match
- `<Foo foo={null} />` => `[foo]` => match
- `<Foo foo={2} />` => `[foo]` => match
- `<Foo />` => `[foo]` => not match
- `<Foo foo={undefined} />` => `[foo]` => not match

With values:
- `<Foo foo="bar" />` => `[foo="bar"]` => match
- `<Foo foo="foo" />` => `[foo="bar"]` => not match
- `<Foo foo="2" />` => `[foo="2"]` => match
- `<Foo foo={2} />` => `[foo="2"]` => not match

Then, special react props: (I'm not strong feeling on these, but I think we should document how it behaves right now)
- `<Foo ref="foo" />` => `[ref]` => not match
- `<Foo ref="foo" />` => `[ref="foo"]` => not match
- `<Foo key="foo" />` => `[key]` => not match
- `<Foo key="foo" />` => `[key="foo"]` => not match

hartzis:
I'm really excited to see support for attribute/prop selectors, but it feels like this is coding to the wrong abstraction. Under the hood JSX is javascript and this solution seems right if this was html.

Instead of a CSS selector why not supply an object of props to search for?
`<Foo foo="bar" />` => `expect(wrapper.find({foo:"bar"}))` => match
`<Foo foo="bar" blah="sheep" />` => `expect(wrapper.find({foo:"bar"}))` => match?

lelandrichardson:
@hartzis That's a great idea! I would love to see that syntax opened as a PR. I don't think this string-based selector syntax collides with anything object-based though, so the two should be able to work in harmony.

blainekasten:
So you only want string based matching? You wouldn't want something like this?


`<Foo foo={2} />` => `[foo="2"]` => not match
`<Foo foo={2} />` => `[foo=2]` => match

I think it'd be nice to support non-string lookups. I understand this deviates from qSA but that API is dealing with DOM where everything is stringified, but react props are not. So I think this could work for this situation. What do you think?

src/Utils.js
blainekasten: Maybe test again? I literally just tested both those ideas and they both matched.

ljharb:
I could be convinced for the string-based API to go with either a) only allow string values, or b) allow any non-symbol primitive (ie, booleans, null, undefined, numbers, strings with explicit quotes). What I wouldn't want is `[foo=3]` to match a string `"3"`, nor `[foo="3"]` to match a number `3`.

@lelandrichardson, any thoughts?

blainekasten:
cc @lelandrichardson for continued discussion.

lelandrichardson:
I agree with @ljharb that a 2 with quotes shouldn't match a numerical 2. I'd be open to letting other primitives match the non-quoted literal versions of them (ie, `[foo=3]` or `[foo=true]`). @hartzis proposed solution of `.find({foo: 3})` and `.find({foo: true})` is probably better suited for these things, and extends beyond primitives... BUT, you can't do something like `input.foo-bar[boo="bam"]` with an API like that. Though once you are getting that complicated of a selector, the `.findWhere()` API is likely your best choice.

For this PR, if we don't want to spend time implementing the primitives `[foo=3]`, we can leave it undefined behavior and throw an unsupported selector error similar to what we do now for the attribute selectors, and decide whether we want to support it later.  It's up to you...  I will merge this in either way.

blainekasten:
I have already pulled off the non-quoted version once, so I'll just rollback to that point and make sure it's all working. I'll get to it this afternoon.

Thanks for the feedback.

lelandrichardson:
@blainekasten thanks for bearing with us for this whole conversation!

blainekasten:
No prob! It's a pretty big api addition. Always gotta make sure it serves the right direction.

blainekasten:
@lelandrichardson This is ready to go now! Thanks

lelandrichardson:
@blainekasten one last request: can you squash your commits before I merge?

ljharb:
@blainekasten a rebase would be preferred over a merge

blainekasten:
Yeah. I'm trying to rebase. I always struggle through the squashing process...

ljharb:
sadly github doesn't make it easy with a rebase button. on your branch, `git fetch source; git rebase source/master` where `source` is the git remote for this repo should do it tho?

blainekasten:
Yeah.. it is too hard. I am not following exactly for that suggestion. 

I tried `git rebase -i HEAD~n` and squashed the previous commits.. but it didn't squash them haha

ljharb:
Alternatively, you could make a brand new branch locally, recreate it manually by checking things out or cherry picking from the original branch, and then, force push the new local branch to the original remote branch, which will update the PR.

lelandrichardson:
on your branch, figure out the total # of commits `n` that you have that you want to squash.  Run `git rebase -i HEAD~n`. In the menu, leave the first commit as `pick <sha> <message>`. For all of the following commits, change it to `squash <sha> <message>`.  Submit that screen (`:wq` if you're using vim), then in the next screen submit that screen as well, unless you want to change the commit message, which you can do.

After that, run `git push -f`

blainekasten:
Are you guys happy enough with 2 squashedish commits? haha. I can fix it if you want. I just get so worried when I do this that I'll break something.

blainekasten:
Oh boom. I got this down now. Alright, I think we're good to :)

src/__tests__/Utils-spec.js
```diff
@@ -196,4 +200,48 @@ describe('Utils', () => {
...
+    });
+
+    it('returns number if passed a stringified number', () => {
+      expect(coercePropValue('1')).to.be.equal(1);
+      expect(coercePropValue('0')).to.be.equal(0);
+    });
+
+    it('returns a boolean if passed a stringified bool', () => {
+      expect(coercePropValue('true')).to.be.true;
+      expect(coercePropValue('false')).to.be.false;
```
lelandrichardson: whoops. something got merged into master that prevents these types of tests. You need to change these to `.to.equal(true)` instead. Our linter prevents it the use of these now. You can run `npm run lint` to ensure everything is fixed.

blainekasten: when I run the linter, those aren't reported. But I changed them, we'll see if this gets it passing

ljharb: @blainekasten you may need to fetch and rebase again - there's been some merges today.

blainekasten:
This should be ready to go now. Thanks

lelandrichardson:
hey @blainekasten it looks like it has some merge conflicts. Some stuff got merged in that must be conflicting. Can you rebase once more and resolve whatever conflicts show up?  Thanks!

blainekasten:
Alright. Should be good again. Hopefully it can get merged before something else conflicts again :)

lelandrichardson:
Awesome work. Thank you!

blainekasten:
Thanks for merging! No problem. I'm excited about this project. Thanks for
starting it

PR #72, opened 2015-12-14
Use a consistent form for the messages in the example tests

jonathanp:


lelandrichardson:
:+1: there are other tests that are using "should" still though, right?

PR #74, opened 2015-12-15
Added test for find by display name

marlonbernardes:
According to the docs, it should be possible to `find` an element based on its display name. I simply added a test to assure that.


marlonbernardes:
I'm not sure why the coverage decreased, since I only added one test to an existing feature.

pasviegas:
@marlonbernardes very strange indeed ;P hahaha

lelandrichardson:
The coverage is messed up because of babel... feel free to ignore it.

Also, thanks! This is great. Not sure why i never had this covered :/

PR #77, opened 2015-12-17
Map lowercase mouse events in #simulate

### LINKED ISSUE #29: `.simulate('mouseenter')` doesn't work, but `.simulate('mouseEnter')` does
It might make sense to support both?

jackfranklin:
React camelcases a bunch of events that JS has in all lower case. This
PR adds a map of them so people can call `simulate` with either the
React version or the native event version.

Fixes #29.

Thought I'd take a stab at the fix - let me know what you think!

### DIFF
--- a/src/ReactWrapper.js
+++ b/src/ReactWrapper.js
@@ -15,6 +15,7 @@ import {
   Simulate,
   findDOMNode,
 } from './react-compat';
+import { mapNativeEventNames } from './Utils';
 
 /**
  * Finds all nodes in the current wrapper nodes' render trees that match the provided predicate
@@ -306,7 +307,8 @@ export default class ReactWrapper {
    */
   simulate(event, ...args) {
     this.single(n => {
-      const eventFn = Simulate[event];
+      const mappedEvent = mapNativeEventNames(event);
+      const eventFn = Simulate[mappedEvent];
       if (!eventFn) {
         throw new TypeError(`ReactWrapper::simulate() event '${event}' does not exist`);
       }
--- a/src/__tests__/ReactWrapper-spec.js
+++ b/src/__tests__/ReactWrapper-spec.js
@@ -457,6 +457,42 @@ describeWithDOM('mount', () => {
       expect(wrapper.simulate.bind(wrapper, 'invalidEvent'))
         .to.throw(TypeError, "ReactWrapper::simulate() event 'invalidEvent' does not exist");
     });
+
+    describe('Normalizing JS event names', () => {
+      it('should convert lowercase events to React camelcase', () => {
+        const spy = sinon.spy();
+        class Foo extends React.Component {
+          render() {
+            return (
+              <a onDoubleClick={spy}>foo</a>
+            );
+          }
+        }
+
+        const wrapper = mount(<Foo />);
+
+        wrapper.simulate('doubleclick');
+        expect(spy.calledOnce).to.equal(true);
+      });
+
+      describeIf(!REACT013, 'normalizing mouseenter', () => {
+        it('should convert lowercase events to React camelcase', () => {
+          const spy = sinon.spy();
+          class Foo extends React.Component {
+            render() {
+              return (
+                <a onMouseEnter={spy}>foo</a>
+              );
+            }
+          }
+
+          const wrapper = mount(<Foo />);
+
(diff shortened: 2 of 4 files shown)


src/ReactWrapper.js
```diff
@@ -294,6 +295,7 @@ export default class ReactWrapper {
    * @returns {ReactWrapper}
    */
   simulate(event, ...args) {
+    event = mapNativeMouseEvent(event);
```
ljharb: this approach seems both brittle and only will work on mouse events. There's lots of other kinds of events, like "dragstart", "touchstart", "keypress", etc that this could impact.

What about an explicit whitelist - just an object mapping lowercase names to react-cased ones?

jackfranklin: That thought had crossed my mind - happy to do that instead :+1:

mattpage:
+1 

I've been bit by this.

```
something.simulate('mouseover');
```
Generates this not-so-helpful error:
```
TypeError: Cannot read property 'apply' of undefined
      at ReactWrapper.<anonymous> (node_modules/enzyme/build/ReactWrapper.js:351:44)
      at ReactWrapper.single (node_modules/enzyme/build/ReactWrapper.js:739:17)
      at ReactWrapper.simulate (node_modules/enzyme/build/ReactWrapper.js:350:12)
```

jackfranklin:
@ljharb just updated it - let me know what you think!

ljharb:
This looks great as far as covering lowercase names - it would be nice to also have a nicer, more explicit error message when the event name wasn't valid (#79) - but that doesn't have to happen as part of this PR.

The tests appear to be failing in most (but not all) of the React 0.13 builds, though.

jackfranklin:
@ljharb happy to tackle #79 in a follow up PR :+1: 

I found this from the React 0.14 release notes:

> Add-Ons: When using the test utils, Simulate.mouseEnter and Simulate.mouseLeave now work.

Also see: https://github.com/facebook/react/issues/1297

So it seemed that `mouseEnter` and `mouseLeave` just didn't work pre 0.14. I'm tempted to just swap the test over to using a different event to avoid this?

I could also in this or a follow up PR add an error message that's shown in the user is React 0.13 and they try to use `mouseEnter` or `mouseLeave`.

Let me know what you think is best!

ljharb:
@jackfranklin Could you conditionally have `mouseEnter`/`mouseLeave` to the object based on which react version is being used?

jackfranklin:
@ljharb :+1: will make that change now

src/__tests__/ReactWrapper-spec.js
```diff
@@ -332,6 +332,43 @@ describeWithDOM('mount', () => {
...
+              return (
+                <a onMouseEnter={spy}>foo</a>
+              );
+            }
+          }
+
+          const wrapper = mount(<Foo />);
+
+          wrapper.simulate('mouseenter');
+          expect(spy.calledOnce).to.be.true;
```
blainekasten: You'll need to pull from master and changes these assertions to be `to.equal(true)`. A change was merged that causes a linting error with these types of assertions!

src/__tests__/Utils-spec.js
```diff
@@ -200,6 +201,7 @@ describe('Utils', () => {
 
   });
 
+<<<<<<< HEAD
```
ljharb: merge conflict?

jackfranklin:
@ljharb I've just tidied this up with a rebase and everything looks fine locally - I'm not entirely sure why the last Travis rebuild said test coverage had dropped so significantly mind. I think if I can fix that ( might need a little pointer in the right direction ) I hope this is now finally in a state where it could be merged.

jackfranklin:
Also I'll squish the commits down when you're happy :+1:

ljharb:
Awesome - I wouldn't worry too much about the coverage, it's still a bit flaky for some reason. This LGTM (pending a rebase) but I'll let @lelandrichardson look it over first.

lelandrichardson:
:+1: this looks great. if you rebase, i'll merge this in and push out a release this afternoon.

lelandrichardson:
@jackfranklin can you rebase + squash so I can get this in? thanks!

jackfranklin:
@lelandrichardson done!

Issue #78, opened 2015-12-17
Update to babel 6


tmcw:
Ideally enzyme would update to babel 6 so that it can be compatible with applications built in babel 6 land and their configuration files.

ljharb:
Why would the babel version used matter, since the transpiled version is what's published, and `node_modules` shouldn't ever be transpiled?

goatslacker:
We should probably upgrade regardless but not for that reason.

lelandrichardson:
PR Merged

PR #81, opened 2015-12-18
Update to babel 6 #78

lecstor:
- tests pass with React 13 & 14
- npm run travis produces coverage
- npm link works

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["airbnb"]
+}
--- a/package.json
+++ b/package.json
@@ -10,12 +10,12 @@
     "version": "npm run build",
     "clean": "rimraf build",
     "lint": "eslint src/**",
-    "test": "mocha --compilers js:babel/register --recursive src/**/__tests__/*.js",
+    "test": "mocha --compilers js:babel-core/register --recursive src/**/__tests__/*.js",
     "check": "npm run lint && npm run test:all",
     "build": "babel src --out-dir build",
-    "test:watch": "mocha --compilers js:babel/register --recursive src/**/__tests__/*.js --watch",
-    "test:describeWithDOMOnly": "mocha --compilers js:babel/register --recursive src/**/__tests__/describeWithDOM/describeWithDOMOnly-spec.js",
-    "test:describeWithDOMSkip": "mocha --compilers js:babel/register --recursive src/**/__tests__/describeWithDOM/describeWithDOMSkip-spec.js",
+    "test:watch": "mocha --compilers js:babel-core/register --recursive src/**/__tests__/*.js --watch",
+    "test:describeWithDOMOnly": "mocha --compilers js:babel-core/register --recursive src/**/__tests__/describeWithDOM/describeWithDOMOnly-spec.js",
+    "test:describeWithDOMSkip": "mocha --compilers js:babel-core/register --recursive src/**/__tests__/describeWithDOM/describeWithDOMSkip-spec.js",
     "test:all": "npm run react:13 && npm test && npm run test:describeWithDOMOnly && npm run test:describeWithDOMSkip && npm run react:14 && npm test && npm run test:describeWithDOMOnly && npm run test:describeWithDOMSkip",
     "react:clean": "rimraf node_modules/react node_modules/react-dom node_modules/react-addons-test-utils",
     "react:13": "npm run react:clean && npm i react@0.13",
@@ -25,7 +25,7 @@
     "docs:build": "npm run docs:prepare && gitbook build",
     "docs:watch": "npm run docs:prepare && gitbook serve",
     "docs:publish": "npm run docs:clean && npm run docs:build && cd _book && git init && git commit --allow-empty -m 'update book' && git fetch https://github.com/airbnb/enzyme.git gh-pages && git checkout -b gh-pages && git add . && git commit -am 'update book' && git push https://github.com/airbnb/enzyme.git gh-pages --force",
-    "travis": "istanbul cover _mocha -- --compilers js:babel/register --recursive src/**/__tests__/*.js"
+    "travis": "babel-istanbul cover --report html _mocha -- --recursive src/**/__tests__/*.js"
   },
   "repository": {
     "type": "git",
@@ -55,8 +55,12 @@
     "underscore": "^1.8.3"
   },
   "devDependencies": {
-    "babel": "^5.8.21",
+    "babel-cli": "^6.3.17",
+    "babel-core": "^6.3.21",
     "babel-eslint": "^4.1.4",
+    "babel-istanbul": "^0.5.9",
+    "babel-preset-airbnb": "^1.0.1",
+    "babel-register": "^6.3.13",
     "chai": "^3.2.0",
     "coveralls": "^2.11.4",
     "eslint": "^1.10.3",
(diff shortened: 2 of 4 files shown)


lecstor:
hrmm bugga. iojs-v3.3 `TypeError: Object.assign is not a function`
I will have to come back to this later.

ljharb:
@lecstor `Object.assign` isn't available natively in all node versions.

package.json
```diff
@@ -51,8 +51,15 @@
...
-    "babel": "^5.8.21",
+    "babel-cli": "^6.3.17",
+    "babel-core": "^6.3.21",
     "babel-eslint": "^4.1.4",
+    "babel-istanbul": "^0.5.9",
+    "babel-preset-es2015": "^6.3.13",
+    "babel-preset-react": "^6.3.13",
+    "babel-preset-stage-0": "^6.3.13",
+    "babel-register": "^6.3.13",
+    "babel-runtime": "^6.3.19",
```
ljharb: It seems very unfortunate to require `babel-runtime` in a reusable module. Why is this necessary?

src/index.js
```diff
@@ -3,7 +3,8 @@ import Sinon from 'sinon';
 import ReactWrapper from './ReactWrapper';
 import ShallowWrapper from './ShallowWrapper';
 import { onPrototype } from './Utils';
-import { renderToStaticMarkup } from './react-compat';
+import reactCompat from './react-compat';
+const {renderToStaticMarkup} = reactCompat;
```
ljharb: please preserve the spaces inside the object destructuring brackets

lelandrichardson: curious, but why were these changed from imports to imports + destructure?  Is there value to doing so? is this just a style preference?

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  presets: ["react", "es2015", "stage-0"]
```
goatslacker: Use this preset instead https://github.com/airbnb/babel-preset-airbnb

package.json
lecstor: thanks, you are correct, it is not required and will be removed.

src/index.js
lecstor: how easily we come to rely on linting in the editor.. I will update eslint-config-airbnb and fix linting.

Babel6 has changed the way bare objects are exported for compatability reasons meaning the properties of the exported object are actually in a `default` property.

```
TypeError: (0 , _reactCompat.renderToStaticMarkup) is not a function
      at render (index.js:8:30)
      at matchesRender (ReactWrapper-spec.js:6:9)
      at Context.<anonymous> (ReactWrapper-spec.js:540:7)
```

The alternative would be to use require:
`const { renderToStaticMarkup } = require('./react-compat').default;`

I'm not aware of any other ES6 options.

lecstor:
`babel-plugin-transform-object-assign` fixes the iojs issues. I guess Babel5 was doing this by default.
What is the preference here?
- add the plugin to enzyme
- add the plugin to the airbnb preset
- replace the single use of `Object.assign` in ReactWrapperComponent.jsx with an iojs compatible equivalent

ljharb:
@lecstor the single use could be replaced with `object.assign` usage which would be compatible with even node 0.4 if necessary - that seems simplest rather than adding another transform.

package.json
```diff
@@ -47,20 +47,25 @@
   "license": "MIT",
   "dependencies": {
     "cheerio": "^0.19.0",
+    "object-assign": "^4.0.1",
```
ljharb: please use `object.assign` instead of `object-assign` - it's more spec-compliant.

src/index.js
ljharb: ES6 imports indeed can't be destructured inline, per spec - babel 5 violates the spec in this regard. babel 6 violates it in a different way.

Since the ES6 loader spec doesn't yet exist, ES6 modules don't yet exist, so technically, nobody should be using import/export anywhere under any circumstances. Since a number of people have chosen to use the syntax anyways, a number of compromises have to be made depending on the situation. This is appropriate here.

lecstor:
sorry about that. I haven't needed to use it before (obviously 8).
Is the `shim` method ok? there appear to be a few ways to use it..

ljharb:
Since this isn't a top-level app, only a reusable module, modifying the global environment in any way is unacceptable - so please don't use the `shim` method, just use the module as a pure function.

lecstor:
done. I think. I'm not entirely sure if that is the correct usage. Reading the source it seems it should be fine, but the readme doesn't mention that usage style at all. The examples all call shim or getPolyfill, and assign the result to shimmedAssign mostly, but use `assign` as the function (from the first example) so I found them a little confusing.

ljharb:
@lecstor that looks great, thanks! The readme on https://www.npmjs.com/package/object.assign has the examples under "most common usage".

lecstor:
excellent, thank you @ljharb

src/MountedTraversal.js
```diff
@@ -6,13 +6,16 @@ import {
...
-import {
+import reactCompat from './react-compat';
+const {
   isDOMComponent,
   isCompositeComponent,
   isCompositeComponentWithType,
   isElement,
   findDOMNode,
-} from './react-compat';
+} = reactCompat;
```
lelandrichardson: was this changed for a reason?

src/__tests__/ReactWrapper-spec.js
```diff
@@ -506,18 +506,20 @@ describeWithDOM('mount', () => {
...
       }
       matchesRender(<Foo items={['abc', 'def', 'hij']} />);
       matchesRender(
-        <Foo items={[
-          <i key={1}>abc</i>,
-          <i key={2}>def</i>,
-          <i key={3}>hij</i>,
-        ]} />
+        <Foo items=
+          {[
```
lelandrichardson: this indentation seems strange. was this required by the linter or something?

src/MountedTraversal.js
ljharb: babel 5 incorrectly allows destructuring on a default import - the spec does not.

lelandrichardson: this makes me sad :/

goatslacker: Why don't we fix this in `react-compat.js`?

Just `export` all these individually.

lelandrichardson: :+1: i agree that would be better

src/__tests__/ReactWrapper-spec.js
lecstor: it does, doesn't it.. but yes, it seemed the nicest way I could make the linter happy.

src/MountedTraversal.js
lecstor: PR updated accordingly.

src/__tests__/ReactWrapper-spec.js
ljharb: you want this:
```
<Foo
  items={[
    …
  ]}
/>
```

lecstor: ok, ta, will update when I get a chance.

lelandrichardson:
Hey @lecstor, we are wanting to merge this now....

Can you rebase and squash your commits?  Thanks!

lecstor:
Hi @lelandrichardson, I had to make a final change to update the test commands to use `babel-core/register`. Should be suitably rebased and squashed now.

Issue #94, opened 2015-12-29
Allow support for Node.JS <= v0.12.9 


chriscartlidge:
A lot of companies still use older versions of Node.JS for a variety of reasons.

That being said the optional dependancy on jsdom (^6.1.0) requires us to be using Node.JS v4 or greater. 

My recommendation is to use jsdom (~3.1.2) which is allow support for a greater number of previous versions of Node.JS. 

I have ran through the source code and also check the tests whilst using this version of the jsdom dependency and everything looks OK.

I'll submit a PR and link to this ticket for your consideration. (https://github.com/airbnb/enzyme/pull/95)

blainekasten:
Looks like this can be closed as the PR is merged.

ljharb:
Closed via #95.

PR #95, opened 2015-12-29
CHANGED: The jsdom dependency to allow a range of supported versions.

chriscartlidge:
This allows versions of node <= 0.12.0 to use this project as jsdom 4.X > require node 4.X >

Issue: https://github.com/airbnb/enzyme/issues/94

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "react": "0.13.x || 0.14.x"
   },
   "optionalDependencies": {
-    "jsdom": "^6.1.0",
+    "jsdom": "^3.1.2 || ^5.6.1 || ^6.1.0 || ^7.2.2",
     "mocha-jsdom": "^1.0.0"
   }
 }


package.json
```diff
@@ -67,7 +67,7 @@
     "react": "0.13.x || 0.14.x"
   },
   "optionalDependencies": {
-    "jsdom": "^6.1.0",
+    "jsdom": "~3.1.2",
```
ljharb: Please use `^`, instead of `~` - older nodes are fine, but older npms don't even work properly with the registry, so there's no reason to support them.

chriscartlidge:
@ljharb Sorted.

ljharb:
So, this will actually prohibit v6, and only work with v3 - what would be ideal is confirming exactly which major version lines of jsdom enzyme works with, and including all of them in the semver range - for example, `^3.1.2 || ^6.1.0`, if it works with just those two, etc.

chriscartlidge:
@ljharb So it looks like we work against all of the jsdom versions apart from v4. I have updated the semver to take this into consideration.

package.json
```diff
@@ -67,7 +67,7 @@
     "react": "0.13.x || 0.14.x"
   },
   "optionalDependencies": {
-    "jsdom": "^6.1.0",
+    "jsdom": "^3.1.2 || ^5.6.1 || ^7.2.2",
```
ljharb: could this also include `|| ^6.1.0` if it works with 6.x?

chriscartlidge:
@ljharb Sorted.

lelandrichardson:
:+1: I think this is good to go

PR #96, opened 2015-12-29
Minor Doc Fixes for ShallowWrapper.{forEach, simulate}

### LINKED ISSUE #88: simulate example doesn't make sense, nor does it work
http://airbnb.io/enzyme/docs/api/ShallowWrapper/simulate.html

I'm trying to get a grip on how to test my React components. Specifically, I feel like a proper test would be to assert some state, call a function that calls `setState`, and re-assert the state. The below example uses an inline function, which confuses me more.. is there no such possibility to directly call the function without "simulating" an event? Furthermore, the simulate example in the docs doesn't seem to make sense:

```
  constructor(props) {
    super(props);
    this.state = { count: 0 };
  }
  render() {
    const { count } = this.state;
    return (
      <div>
        <div className={`clicks-${count}`}>
          {count} clicks
        </div>
        <a onClick={() => this.setState({ count: count + 1 })}>
          Increment
        </a>
      </div>
    );
  }
}

expect(wrapper.find('.clicks-0').length).to.equal(1);
wrapper.simulate('click');
expect(wrapper.find('.clicks-1').length).to.equal(1);
```

I don't understand `simulate('click')` here.. doesn't it need to find the anchor tag first to click it?

erikthedeveloper:
- Fix typos in ShallowWrapper/forEach docs example
- Fix ShallowWrapper/simulate docs example. Was simulating on wrong node (Fixes #88)

### DIFF
--- a/docs/api/ShallowWrapper/forEach.md
+++ b/docs/api/ShallowWrapper/forEach.md
@@ -30,7 +30,7 @@ const wrapper = shallow(
 );
 
 wrapper.find('.foo').forEach(function (node) {
-  expect(s.hasClass('foo')).to.be true;
+  expect(node.hasClass('foo')).to.be.true;
 });
 ```
 
--- a/docs/api/ShallowWrapper/forEach.md
+++ b/docs/api/ShallowWrapper/forEach.md
@@ -30,7 +30,7 @@ const wrapper = shallow(
 );
 
 wrapper.find('.foo').forEach(function (node) {
-  expect(node.hasClass('foo')).to.be.true;
+  expect(node.hasClass('foo')).to.equal(true);
 });
 ```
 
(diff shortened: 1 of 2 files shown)


docs/api/ShallowWrapper/forEach.md
```diff
@@ -30,7 +30,7 @@ const wrapper = shallow(
 );
 
 wrapper.find('.foo').forEach(function (node) {
-  expect(s.hasClass('foo')).to.be true;
+  expect(node.hasClass('foo')).to.be.true;
```
ljharb: would you mind changing this to `.to.equal(true)`? Our styleguide prohibits noop expressions which includes `.to.be.true`

ljharb: Thanks!

ljharb:
LGTM

PR #101, opened 2016-01-06
Enable no-unused-expressions ESLint rule

### LINKED ISSUE #97: Convert all usages of noop mocha expressions to function matchers
There are a few instances in our [doc examples](https://github.com/airbnb/enzyme/blob/38ae4b22a78ae3570b2c35c4071602037e2b0c9c/docs/api/ReactWrapper/every.md#examples), and our [tests](https://github.com/airbnb/enzyme/blob/36b38ed9e41fe4744ab883f8de2430e494ce04fc/src/__tests__/ReactWrapper-spec.js#L64-L65) which currently violate our [styleguide](https://github.com/airbnb/javascript). This also forces the relevant [eslint rule](https://github.com/airbnb/enzyme/blob/9ced15fe03a3b45ab176f5f97e9b224480bfeb6c/src/__tests__/.eslintrc#L4) to be disabled.

A great first task would be removing this rule override, and fixing our tests and docs to use, for example, `.to.equal(true)` instead of `.to.be.true`. (One reason this form is superior, is that `expect(false).to.be.truthy`, which many might think is a valid matcher, will silently pass tests as a no-op, because `truthy` isn't a thing).

(Per https://github.com/airbnb/enzyme/commit/d00c1422ba8559e6f41f9e3ff9cbd38cf9925c4a#commitcomment-15192634 and #96)

iancmyers:
This enables the `no-unused-expressions` ESLint rule and fixes the corresponding violations. In fixing the errors I stumbled upon two tests that were broken. After discussing with @lelandrichardson, I've inverted the test case as the current behavior is the desired behavior.

Fixes #97

### DIFF
--- a/docs/api/ReactWrapper/parent.md
+++ b/docs/api/ReactWrapper/parent.md
@@ -13,7 +13,7 @@ Returns a wrapper with the direct parent of the node in the current wrapper.
 
 ```jsx
 const wrapper = mount(<ToDoList />);
-expect(wrapper.find('ul').parent().is('div')).to.be.true;;
+expect(wrapper.find('ul').parent().is('div')).to.equal(true);
 ```
 
 #### Related Methods
--- a/docs/api/ShallowWrapper/parent.md
+++ b/docs/api/ShallowWrapper/parent.md
@@ -2,18 +2,15 @@
 
 Returns a wrapper with the direct parent of the node in the current wrapper.
 
-
 #### Returns
 
 `ShallowWrapper`: A new wrapper that wraps the resulting nodes.
-
-
 
 #### Examples
 
 ```jsx
 const wrapper = shallow(<ToDoList />);
-expect(wrapper.find('ul').parent().is('div')).to.be.true;;
+expect(wrapper.find('ul').parent().is('div')).to.equal(true);
 ```
 
 #### Related Methods
--- a/src/__tests__/ReactWrapper-spec.js
+++ b/src/__tests__/ReactWrapper-spec.js
@@ -28,7 +28,7 @@ describeWithDOM('mount', () => {
       expect(wrapper.text()).to.equal('foo');
     });
 
-    it('throws if context is passed in but contextTypes is missing', () => {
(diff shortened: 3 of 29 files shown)


lelandrichardson:
LGTM

docs/api/ReactWrapper/parent.md
```diff
@@ -13,7 +13,7 @@ Returns a wrapper with the direct parent of the node in the current wrapper.
 
 ```jsx
 const wrapper = mount(<ToDoList />);
-expect(wrapper.find('ul').parent().is('div')).to.be.true;;
+expect(wrapper.find('ul').parent().is('div')).to.equal(true);;
```
ljharb: double semicolon

docs/api/ShallowWrapper/parent.md
```diff
@@ -13,7 +13,7 @@ Returns a wrapper with the direct parent of the node in the current wrapper.
 
 ```jsx
 const wrapper = shallow(<ToDoList />);
-expect(wrapper.find('ul').parent().is('div')).to.be.true;;
+expect(wrapper.find('ul').parent().is('div')).to.equal(true);;
```
ljharb: also double semicolon

src/__tests__/ReactWrapper-spec.js
```diff
@@ -36,7 +36,7 @@ describeWithDOM('mount', () => {
       });
 
       const context = { name: 'foo' };
-      expect(() => mount(<SimpleComponent />, { context })).to.throw;
+      expect(() => mount(<SimpleComponent />, { context })).to.not.throw(Error);
```
ljharb: we should make an eslint rule to standardize on `not.to` or `to.not` :-p

ljharb:
LGTM, only the 2 nonblockers of the pre-existing double semicolons (those statements are REALLY terminated)

iancmyers:
@ljharb Went ahead and removed those extra semicolons.

PR #106, opened 2016-01-08
FIX #80: Added options to jsdom constructor

### LINKED ISSUE #80: When running with wallaby.js the mocha-jsdom setup breaks often
After running some tests enzyme rejects to run dom tests with the error message:
```
"before all" hook: mocha-jsdom: already a browser environment, or mocha-jsdom invoked twice. use 'skipWindowCheck' to disable this check.
```

KnisterPeter:
Added optional options to jsdom constructor for example
to configure skipWindowCheck.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -19,10 +19,14 @@ try {
 
 export let sinon = Sinon.sandbox.create();
 
-export function describeWithDOM(a, b) {
+export function describeWithDOM(a, options, b) {
+  if (typeof b == 'undefined') {
+    b = options;
+    options = undefined;
+  }
   describe('(uses jsdom)', () => {
     if (typeof jsdom === 'function') {
-      jsdom();
+      jsdom(options);
       describe(a, b);
     } else {
       // if jsdom isn't available, skip every test in this describe context


src/index.js
```diff
@@ -19,10 +19,14 @@ try {
 
 export let sinon = Sinon.sandbox.create();
 
-export function describeWithDOM(a, b) {
+export function describeWithDOM(a, options, b) {
+  if (typeof b == 'undefined') {
+    b = options;
```
ljharb: reassigning to arguments deopts in v8, and our linter forbids it.

ljharb:
It makes me sad to have a function signature like this, but I can't think of a better way to do it, since it'd be a bit awkward to have `describeWithDOM('message', func, options)`.

Perhaps another export, `describeWithJSDOMOptions` or something?

vesln:
@ljharb while i understand the value of getting up and running quickly, i truly believe enzyme should stay away from trying to help people setup jsdom/mocha, imo it's way out of scope.

the jsdom setup is one thing, the other is `describe` really is just one of the interfaces... what about `suite`, `testCase`, mocha's export styles etc.

a "preset" package that bundles things up, as planned in the next major version, is fine but the approach needs to be reconsidered (eg. assuming that people will use only mocha's bdd interface).

imo, any changes related to jsdom/mocha until then should not be merged, but rather a documentation on how to setup jsdom should be provided. the `describeWithDOM` util can be provided as an example.

just my 2 cents

KnisterPeter:
No problem for me.
The last fallback could be to use an own custom describe. It is not that hard :-)

PR #109, opened 2016-01-10
Provides a descriptive error when simulating non-existent events

### LINKED ISSUE #79: Better error if you try to simulate a non-existing event
Running

```
input.simulate('keyDown', { which: val });
```

Produces

```
TypeError: Cannot read property 'apply' of undefined
      at ReactWrapper.<anonymous> (node_modules/enzyme/build/ReactWrapper.js:351:44)
      at ReactWrapper.single (node_modules/enzyme/build/ReactWrapper.js:739:17)
      at ReactWrapper.simulate (node_modules/enzyme/build/ReactWrapper.js:350:12)
```

Because there's an unchecked reference to an event type. Instead, enzyme should give an informative message.

cambridgemike:
Fixes #79

### DIFF
--- a/src/ReactWrapper.js
+++ b/src/ReactWrapper.js
@@ -294,7 +294,14 @@ export default class ReactWrapper {
    * @returns {ReactWrapper}
    */
   simulate(event, ...args) {
-    this.single(n => Simulate[event](findDOMNode(n), ...args));
+    this.single(n => {
+      const eventFn = Simulate[event];
+      if (!eventFn) {
+        throw new TypeError(`ReactWrapper::simulate() event '${event}' does not exist`);
+      }
+
+      eventFn(findDOMNode(n), ...args);
+    });
     return this;
   }
 
--- a/src/__tests__/ReactWrapper-spec.js
+++ b/src/__tests__/ReactWrapper-spec.js
@@ -341,7 +341,6 @@ describeWithDOM('mount', () => {
 
     });
 
-
     it('should pass in event data', () => {
       const spy = sinon.spy();
       class Foo extends React.Component {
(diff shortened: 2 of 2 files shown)


src/ReactWrapper.js
```diff
@@ -294,7 +294,14 @@ export default class ReactWrapper {
    * @returns {ReactWrapper}
    */
   simulate(event, ...args) {
-    this.single(n => Simulate[event](findDOMNode(n), ...args));
+    this.single(n => {
+      const eventFn = Simulate[event];
+      if (eventFn !== undefined) {
+        eventFn(findDOMNode(n), ...args);
+      } else {
+        throw new Error(`ReactWrapper::simulate() event '${event}' does not exist`);
```
ljharb: might be nice to throw a `TypeError` here

cambridgemike: Good idea, I'll update

ljharb:
LGTM - the error type thrown is identical, so this is clearly a patch

src/ReactWrapper.js
```diff
@@ -294,7 +294,14 @@ export default class ReactWrapper {
...
    */
   simulate(event, ...args) {
-    this.single(n => Simulate[event](findDOMNode(n), ...args));
+    this.single(n => {
+      const eventFn = Simulate[event];
+      if (eventFn !== undefined) {
+        eventFn(findDOMNode(n), ...args);
+      } else {
+        throw new TypeError(`ReactWrapper::simulate() event '${event}' does not exist`);
+      }
```
blainekasten: nit: Only 1 of these clauses will ever be hit at a time making the `else` superfluous. What about just doing early termination when it is undefined. Something more like:

```js
const eventFn = ...;
if (!eventFn) {
  throw Error
}

eventFn(...);
```

Writing this way ends up being a little clearer because the only clause is for if the `eventFn` is undefined. Otherwise a reader can assume that block doesn't exist and read over it. Like I said, it's nit, but simplifies

blainekasten:
You'll need to squash your commits before it'll be merged.

Issue #105, opened 2016-01-08
Object property selector


blainekasten:
From @hartzis:

I'm really excited to see support for attribute/prop selectors, but it feels like this is coding to the wrong abstraction. Under the hood JSX is javascript and this solution seems right if this was html.

Instead of a CSS selector why not supply an object of props to search for?
`<Foo foo="bar" />` => `expect(wrapper.find({foo:"bar"}))` => match
`<Foo foo="bar" blah="sheep" />` => `expect(wrapper.find({foo:"bar"}))` => match?

lelandrichardson:
Added in 1.4

PR #110, opened 2016-01-11
add support for object property selector

hartzis:
@blainekasten, thanks for quoting me in #105, gave me the drive to attempt to add the support. Let me know what you guys think.

I also just happened to notice while I tried to follow code patterns adding functionality that there seems to be some opportunity to reuse some code between `MountedTraversal` and `ShallowTraversal`. There was definitely some `copy & paste` I did :)

### DIFF
--- a/src/MountedTraversal.js
+++ b/src/MountedTraversal.js
@@ -1,3 +1,5 @@
+import { isEmpty } from 'underscore';
+import isSubset from 'is-subset';
 import {
   coercePropValue,
   nodeEqual,
@@ -177,6 +179,12 @@ export function parentsOfInst(inst, root) {
   return pathToNode(inst, root).reverse();
 }
 
+export function instMatchesObjectProps(inst, props) {
+  if (!isDOMComponent(inst)) return false;
+  const node = getNode(inst);
+  return isSubset(propsOfNode(node), props);
+}
+
 export function buildInstPredicate(selector) {
   switch (typeof selector) {
     case 'function':
@@ -207,8 +215,16 @@ export function buildInstPredicate(selector) {
       }
       break;
 
+    case 'object':
+      if (!Array.isArray(selector) && selector !== null && !isEmpty(selector)) {
+        return node => instMatchesObjectProps(node, selector);
+      }
+      throw new TypeError(
+        'Enzyme::Selector does not support an array, null, or empty object as a selector'
+      );
+
     default:
-      throw new TypeError('Expecting a string or Component Constructor');
+      throw new TypeError(`Enzyme::Selector expects a string, object, or Component Constructor`);
   }
 }
 
--- a/src/ShallowTraversal.js
+++ b/src/ShallowTraversal.js
@@ -1,4 +1,6 @@
 import React from 'react';
+import { isEmpty } from 'underscore';
+import isSubset from 'is-subset';
 import {
   coercePropValue,
   propsOfNode,
@@ -95,6 +97,10 @@ export function nodeHasType(node, type) {
   return node.type.name === type || node.type.displayName === type;
 }
 
+export function nodeMatchesObjectProps(node, props) {
+  return isSubset(propsOfNode(node), props);
+}
+
 export function buildPredicate(selector) {
   switch (typeof selector) {
     case 'function':
@@ -127,9 +133,16 @@ export function buildPredicate(selector) {
       }
       break;
 
+    case 'object':
+      if (!Array.isArray(selector) && selector !== null && !isEmpty(selector)) {
+        return node => nodeMatchesObjectProps(node, selector);
+      }
+      throw new TypeError(
+        'Enzyme::Selector does not support an array, null, or empty object as a selector'
+      );
 
     default:
-      throw new TypeError('Expecting a string or Component Constructor');
+      throw new TypeError(`Enzyme::Selector expects a string, object, or Component Constructor`);
   }
 }
 
--- a/src/__tests__/ReactWrapper-spec.js
+++ b/src/__tests__/ReactWrapper-spec.js
@@ -261,6 +261,58 @@ describeWithDOM('mount', () => {
       expect(() => wrapper.find('.foo .foo')).to.throw(Error);
     });
 
+    it('should support object property selectors', () => {
+      const wrapper = mount(
+        <div>
+          <input data-test="ref" className="foo" type="text" />
+          <input data-test="ref" type="text"/>
+          <button data-test="ref" prop={undefined} />
+          <span data-test="ref" prop={null} />
+          <div data-test="ref" prop={123} />
+          <input data-test="ref" prop={false} />
+          <a data-test="ref" prop />
+        </div>
+      );
+      expect(wrapper.find({ a: 1 })).to.have.length(0);
+      expect(wrapper.find({ 'data-test': 'ref' })).to.have.length(7);
+      expect(wrapper.find({ className: 'foo' })).to.have.length(1);
+      expect(wrapper.find({ prop: undefined })).to.have.length(1);
+      expect(wrapper.find({ prop: null })).to.have.length(1);
+      expect(wrapper.find({ prop: 123 })).to.have.length(1);
+      expect(wrapper.find({ prop: false })).to.have.length(1);
+      expect(wrapper.find({ prop: true })).to.have.length(1);
+    });
+
+    it('should support complex and nested object property selectors', () => {
+      const testFunction = () => {};
+      const wrapper = mount(
+        <div>
+          <span more={[{ id: 1 }]} data-test="ref" prop onChange={testFunction} />
+          <a more={[{ id: 1 }]} data-test="ref" />
+          <div more={{ item: { id: 1 } }} data-test="ref" />
+          <input style={{ height: 20 }} data-test="ref" />
+        </div>
+      );
+      expect(wrapper.find({ 'data-test': 'ref' })).to.have.length(4);
+      expect(wrapper.find({ more: { a: 1 } })).to.have.length(0);
+      expect(wrapper.find({ more: [{ id: 1 }] })).to.have.length(2);
+      expect(wrapper.find({ more: { item: { id: 1 } } })).to.have.length(1);
+      expect(wrapper.find({ style: { height: 20 } })).to.have.length(1);
+      expect(wrapper
+        .find({ more: [{ id: 1 }], 'data-test': 'ref', prop: true, onChange: testFunction })
+      ).to.have.length(1);
+    });
+
+    it('should throw when given empty object, null, or an array', () => {
+      const wrapper = mount(
+        <div>
+          <input className="foo" type="text" />
+        </div>
+      );
+      expect(() => wrapper.find({})).to.throw(Error);
+      expect(() => wrapper.find([])).to.throw(Error);
+      expect(() => wrapper.find(null)).to.throw(Error);
+    });
   });
 
   describe('.findWhere(predicate)', () => {
--- a/docs/api/ReactWrapper/find.md
+++ b/docs/api/ReactWrapper/find.md
@@ -47,7 +47,11 @@ const wrapper = mount(<MyComponent />);
 expect(wrapper.find('Foo')).to.have.length(1);
 ```
 
-
+Object Property Selector:
+```jsx
+const wrapper = mount(<MyComponent />);
+expect(wrapper.find({prop: 'value'})).to.have.length(1);
+```
 
 #### Related Methods
 
--- a/docs/api/ShallowWrapper/find.md
+++ b/docs/api/ShallowWrapper/find.md
@@ -44,6 +44,11 @@ const wrapper = shallow(<MyComponent />);
 expect(wrapper.find('Foo')).to.have.length(1);
 ```
 
+Object Property Selector:
+```jsx
+const wrapper = shallow(<MyComponent />);
+expect(wrapper.find({prop: 'value'})).to.have.length(1);
+```
 
 
 #### Related Methods
--- a/docs/api/selector.md
+++ b/docs/api/selector.md
@@ -6,7 +6,7 @@ one of the following three categories:
 
 ### 1. A Valid CSS Selector
 
-Enzyme supports a subset of valid CSS selectors to find nodes inside a render tree. Support is as 
+Enzyme supports a subset of valid CSS selectors to find nodes inside a render tree. Support is as
 follows:
 
 - class syntax (`.foo`, `.foo-bar`, etc.)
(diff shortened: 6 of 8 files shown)


src/ShallowTraversal.js
```diff
@@ -95,6 +96,10 @@ export function nodeHasType(node, type) {
   return node.type.name === type || node.type.displayName === type;
 }
 
+export function nodeHasProps(node, props) {
```
blainekasten: This function name clashes quite a lot with the `nodeHasProperty` function. I think it should probably be more explicit so it's known that it's meant for the object selector.

Something like `nodeMatchesObjectProps` ?

src/MountedTraversal.js
```diff
@@ -207,8 +214,14 @@ export function buildInstPredicate(selector) {
...
 
+    case 'object':
+      if (!isArray(selector) && !isNull(selector)) {
+        return node => instHasProps(node, selector);
+      }
+      throw new TypeError('An array or null is not supported as a selector');
+
     default:
-      throw new TypeError('Expecting a string or Component Constructor');
+      throw new TypeError('Expecting a string, object, or Component Constructor');
```
blainekasten: Maybe not for this PR, maybe for this one. It'd be nice if these `TypeError`s was prefaced with something like `Enzyme::SelectorError ...` 

similar to https://github.com/airbnb/enzyme/blob/master/src/ShallowWrapper.js#L107

src/MountedTraversal.js
```diff
@@ -207,8 +214,14 @@ export function buildInstPredicate(selector) {
       }
       break;
 
+    case 'object':
+      if (!isArray(selector) && !isNull(selector)) {
```
ljharb: do we really need to use an `isNull` or `isArray` function? `Array.isArray` is built into the language, and `=== null` both avoids function call overhead and is equally clear if not more so.

src/ShallowTraversal.js
hartzis: I like that name a lot more!

src/MountedTraversal.js
hartzis: Nice, I'll remove those functions.

src/MountedTraversal.js
hartzis: I'll throw this in since this PR adds another TypeError.

hartzis:
Updated with suggested edits.

src/MountedTraversal.js
```diff
@@ -177,6 +178,12 @@ export function parentsOfInst(inst, root) {
   return pathToNode(inst, root).reverse();
 }
 
+export function instMatchesObjectProps(inst, props) {
+  if (!isDOMComponent(inst)) return false;
+  const node = getNode(inst);
+  return !isEmpty(props) && isMatch(propsOfNode(node), props);
```
blainekasten: the `!isEmpty` check seems redudant. On L318, you are ensuring that `props` can't be null. 

So it seems like the `isEmpty` call is just a perf concern to make sure users aren't doing `find({})`, but that should be the minority case. 

So you are actually just adding complexity for the typical case of `find({foo: bar})`.

Ultimately, I think the `isEmpty` call is superflous and negatively affecting performance in a very small way.

ljharb: I don't see a line 318 in this diff.

what does `isMatch(propsOfNode(node), {})` do?

blainekasten: L218. My bad. 

isMatch compares keys and values of two objects to be equal

ljharb: right, so, if `propsOfNode(node)` are empty, and `props` are empty, then as-is, it returns false - with your change, it returns true.

Which is desired?

blainekasten: That is true. But how often is somebody going to look for something with an empty object? That just seems like a design against the wrong use-case. And I guess maybe somebody does want to search for nodes that don't have any props? which `find({})` could support? feels awkward though. I guess I could see either way here.

ljharb: First and foremost it needs to be correct, and performance is the least important (not that it should be ignored). The API should do something sensible in all cases.

Thinking about it, however, I think that if, in fact, `propsOfNode` returns an empty object, then `instMatchesObjectProps` *should* return true. So, perhaps removing the isEmpty call is the right thing.

blainekasten: I lean that way also. Or it should throw an error. 

Is there a use case for finding nodes that don't have any props?

blainekasten: Actually, i'm leaning more towards throwing an error for an empty object.

I don't think it should be encouraged to use a selector with an empty objects. There are better ways to find nodes and a more descriptive selector should be used. 

If I looked at tests and saw `wrapper.find({})`. It'd be quite confused on what it was trying to find.

hartzis: I too would lean towards throwing an error. `isEmpty` currently exists because an empty object matches all nodes, and that is probably not what a user would want.

ljharb: That seems reasonable (to throw an error). @lelandrichardson, thoughts?

lelandrichardson: if they are actually passing in an empty object, then matching all nodes actually seems like the reasonable response to me?

I agree this doesn't seem like a normal use case, although I can imagine some code that builds up some object dynamically, and in some cases might build up to nothing...

I don't feel strongly about this, but I don't see the harm in having `{}` match all nodes. It seems like the expected behavior, despite being a "corner-case" type input?

blainekasten: > I agree this doesn't seem like a normal use case, although I can imagine some code that builds up some object dynamically, and in some cases might build up to nothing...

Exactly. But I would say that's a bug in their code. If they are dynamically building an object, and it gets to an empty state. An error should be thrown, so they can make sure they are testing what they want.

> I don't feel strongly about this, but I don't see the harm in having {} match all nodes. It seems like the expected behavior, despite being a "corner-case" type input?

Would you also want to see `find('[]')` to match every node? Because that is the same API effectively. 

Comparing it to the `'[]'` api, I would think it should only find components with props that match the specified key/values.

lelandrichardson: It seems its 3-against-1, so I'm happy to go with the approach of throwing an error for this case.

lelandrichardson:
This looks great. Quick question, though:

`isMatch` makes a strict equality check on the props of the object passed in. This means that the following will not work:

```jsx
const wrapper = shallow(<div style={{ height: 20 }} />);
expect(wrapper.is({ style: { height: 20 } })).to.equal(true);
```

Do we potentially want to implement our own matcher?  We may even want to have special flattening for the `style` prop.

hartzis:
Yeah `style` poses an interesting problem. Also just general nested objects `{ prop: { id: 1 } }`.

`isMatch` was used because `underscore` is already included in the library.

[is-subset](https://www.npmjs.com/package/is-subset) is used by [skin-deep](https://github.com/glenjamin/skin-deep) to compare prop objects. What about using `is-subset`?

ljharb:
I also made [is-equal](https://www.npmjs.com/package/is-equal) which works a bit more reliably than `assert.deepEqual` and is used by [expect](https://www.npmjs.com/package/expect). It doesn't work for subsets though.

src/MountedTraversal.js
hartzis: I'll update this after we discuss the best solution to the nested objects below.

hartzis:
@ljharb Just took a look at `is-equal`, very nice library, but i think you're right, since it doesn't work for subsets I decided to give `is-subset` a whirl. 

Let me know how you guys feel about this possible solution.

src/__tests__/ReactWrapper-spec.js
```diff
@@ -261,6 +261,58 @@ describeWithDOM('mount', () => {
...
+          <a more={[{ id: 1 }]} data-test="ref" />
+          <div more={{ item: { id: 1 } }} data-test="ref" />
+          <input style={{ height: 20 }} data-test="ref" />
+        </div>
+      );
+      expect(wrapper.find({ 'data-test': 'ref' })).to.have.length(4);
+      expect(wrapper.find({ more: { a: 1 } })).to.have.length(0);
+      expect(wrapper.find({ more: [{ id: 1 }] })).to.have.length(2);
+      expect(wrapper.find({ more: { item: { id: 1 } } })).to.have.length(1);
+      expect(wrapper.find({ style: { height: 20 } })).to.have.length(1);
```
hartzis: @lelandrichardson This works now with `is-subset`, what do you think?

src/ShallowTraversal.js
```diff
@@ -127,9 +133,16 @@ export function buildPredicate(selector) {
       }
       break;
 
+    case 'object':
+      if (!Array.isArray(selector) && selector !== null && !isEmpty(selector)) {
+        return node => nodeMatchesObjectProps(node, selector);
+      }
+      throw new TypeError(
+        'Enzyme::Selector does not support an empty object, array, or null as a selector'
```
ljharb: perhaps `null, or an empty object or array`? Will empty strings also trigger this?

src/ShallowTraversal.js
```diff
@@ -127,9 +133,16 @@ export function buildPredicate(selector) {
...
+      );
 
     default:
-      throw new TypeError('Expecting a string or Component Constructor');
+      throw new TypeError(`Enzyme::Selector expects a string, object, or Component Constructor`);
```
ljharb: does it expect an array also, per the above?

src/ShallowTraversal.js
lelandrichardson: an empty string should go through the path that string based selectors go... in which case i think it does throw a different error but i dont remember for sure.

lelandrichardson:
I was trying to think about what the most sensible API for this is...  I think there are 2 questions:

1. Since we are effectively doing "subset" matching behavior one level down for props, should we continue to match things as subsets at the sub-prop level?
2. Since style has special built-in behavior, do we want to handle it differently by pre-flattening the styles array for matching?

Question 2 is sort of a special case of Question 1... it just goes a bit further. At the moment I feel like the answer to 2 is "yes", and if 2 is "yes" I think 1 has to be yes or else the API becomes inconsistent.

As a result, I think the following API would be ideal:

- `{ foo: 'bar' }` matches `<Foo foo="bar" />`
- `{ user: { id: 1 } }` matches `<Foo user={{ id: 1, name: 'Bob' }} />`
- `{ foo: 'bar', qoo: 'qux' }` matches `<Foo foo="bar" qoo="qux" />`
- `{ style: { width: 20 } }` matches `<Foo style={{ width: 20, height: 20 }} />`
- `{ style: { width: 20 } }` matches `<div style={[{ width: 20 }, { height: 20 }]} />`
- `{ foo: 'bar', qoo: 'qux' }` does not match `<Foo foo="bar" />`
- `{ style: { width: 20 } }` does not match `<Foo style={[{ width: 20 }, { height: 20 }]} />`

Let me know if any of you guys disagree, but I feel like that last rule is probably the right behavior. Basically, I'm thinking we will only "flatten" styles when the node type is a string (ie, it's a DOM node).

This could end up kind of annoying though, because some people could have a test set up where their render method somewhere along the lines has something like:

```jsx
<input style={[styles.input, error && styles.inputError]} ... />
```

Then they have a test somewhere that asserts that the inputError styles are applied to the input. With the proposed API above, this works swimmingly BUT... at some point there might be a refactor like the following:

```jsx
<Input style={[styles.input, error && styles.inputError]} ... />
```

At this point, this test will end up failing (assuming a shallow render), but its still behaving properly. Now writing the same test becomes a lot more difficult.

I guess the question comes down to whether or not we want to treat the `style` prop as a special case for all components, or just for DOM components where we know that they behave in this way?  I think there are arguments for both cases...

ljharb:
If it matches subsets all the way down, how would I go about *not* matching a subset, ie, only doing strict matching?

I think that the only special cases should ever be on DOM components - `style` is just a random prop on anything else.

hartzis:
>I think that the only special cases should ever be on DOM components - style is just a random prop on anything else.

I agree with this. @ljharb. I also haven't thought about strict matching. Maybe hold off till users are asking for it? Then maybe an `options` object passed as second argument with the `selector`?

@lelandrichardson I'm having trouble understanding `<div style={[{ width: 20 }, { height: 20 }]} />`. This doesn't seem to currently be a thing in v0.14, it looks like it might be something [someday](https://github.com/facebook/react/issues/3081)?

But the array for `style` is currently working for [React Native Components](https://facebook.github.io/react-native/docs/style.html), but they have to be Components `<Text/>` not dom elements/nodes `<div/>`. And then the trouble is in 'regular react' passing that array for `style` just becomes a prop of style whose value is an array.

example fiddle https://jsfiddle.net/n38pwsdc/

So i could see this:
* `{ style: { width: 20 } }` matches `<Foo style={[{ width: 20 }, { height: 20 }]} />`
* `{ style: { width: 20 } }` does not match `<div style={[{ width: 20 }, { height: 20 }]} />`

but even this is weird because the top one truly only works for react native.  

If we just keep it as is just looking for subsets of props, then the user has to know what to expect since they are testing. No flattening needed on our end.

src/ShallowTraversal.js
hartzis: array gets caught by the `typeof` is `'object'`, then errors out there

i can also update the verbiage on the object error if you would like

lelandrichardson:
LOL. @hartzis you're completely right. I've just been working way too much in react native recently...  we don't need to do anything specific for `style` then at this point.

As far as strict matching goes... I agree that maybe we just wait on this.  As a result, I think the current state of this PR is what we all agree on?

src/ShallowTraversal.js
lelandrichardson: Arguably, you could fence for each one of these scenarios separately and throw an error for each one. Might be more helpful to the user. Not a big deal though.

hartzis: I can't tell if you'd like me to just throw this in?

and if so is it best to just if..else if.. else if.. else it?
```
if (array) TypeError
else if (null) TypeError
else if (empty {}) TypeError
else return isMatch
```

hartzis:
no worries. let me know about those errors and i'll get it updated asap to get this merged in

src/ShallowTraversal.js
lelandrichardson: yeah, although since you are throwing or returning, there is no need for the `else` in each case

src/ShallowTraversal.js
ljharb: ah yes i misread the wording. it doesn't support null, or an array, or an empty object - i read "empty" as applying to all three. perhaps reverse the list on line 141?

hartzis: nice catch, updated

blainekasten:
LGTM.

lelandrichardson:
LGTM.

PR #124, opened 2016-01-17
Add `equals` method to ShallowWrapper.

### LINKED ISSUE #122: Add an .equals() method which accepts JSX
based on the conversation in https://github.com/airbnb/enzyme/issues/116 I think adding a `.equals(<Foo />);` method would be a reasonable addition to the API.

```jsx

const MyComponent = ({ type }) => (
<Page type={type}>
  <Header>
    My Page Title
  </Header>
  <Body>
    My Page Content
  </Body>
</Page>
);

describe('test JSX output', () => {
  it('should be done', () => {
    const component = shallow(<MyComponent type="custom" />);
    expect(component.equals(
<Page type="custom">
  <Header>
    My Page Title
  </Header>
  <Body>
    My Page Content
  </Body>
</Page>)).to.equal(true);
  });
});
```

NOTE: looking at `.contains()` should make this pretty straight-forward to implement...

themouette:
Fixes #122

### DIFF
--- a/src/ShallowWrapper.js
+++ b/src/ShallowWrapper.js
@@ -206,6 +206,22 @@ export default class ShallowWrapper {
   }
 
   /**
+   * Whether or not a given react element exists in the shallow render tree.
+   *
+   * Example:
+   * ```
+   * const wrapper = shallow(<MyComponent />);
+   * expect(wrapper.contains(<div className="foo bar" />)).to.equal(true);
+   * ```
+   *
+   * @param {ReactElement} node
+   * @returns {Boolean}
+   */
+  equals(node) {
+    return this.single(() => nodeEqual(this.node, node));
+  }
+
+  /**
    * Finds every node in the render tree of the current wrapper that matches the provided selector.
    *
    * @param {String|Function} selector
--- a/src/__tests__/ShallowWrapper-spec.js
+++ b/src/__tests__/ShallowWrapper-spec.js
@@ -85,6 +85,54 @@ describe('shallow', () => {
 
   });
 
+  describe('.equals(node)', () => {
+
+    it('should allow matches on the root node', () => {
+      const a = <div className="foo" />;
+      const b = <div className="foo" />;
+      const c = <div className="bar" />;
+      expect(shallow(a).equals(b)).to.equal(true);
+      expect(shallow(a).equals(c)).to.equal(false);
+    });
+
+    it('should NOT allow matches on a nested node', () => {
+      const wrapper = shallow(
+        <div>
+          <div className="foo" />
+        </div>
+      );
+      const b = <div className="foo" />;
+      expect(wrapper.equals(b)).to.equal(false);
+    });
+
+    it('should match composite components', () => {
+      class Foo extends React.Component {
+        render() { return <div />; }
+      }
+      const wrapper = shallow(
+        <div>
+          <Foo />
+        </div>
+      );
+      const b = <div><Foo /></div>;
+      expect(wrapper.equals(b)).to.equal(true);
+    });
+
+    it('should not expand `node` content', () => {
+      class Bar extends React.Component {
+        render() { return <div />; }
+      }
+
+      class Foo extends React.Component {
+        render() { return <Bar />; }
+      }
+
+      expect(shallow(<Foo />).equals(<Bar />)).to.equal(true);
+      expect(shallow(<Foo />).equals(<Foo />)).to.equal(false);
+    });
+
+  });
+
   describe('.find(selector)', () => {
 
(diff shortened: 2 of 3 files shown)


src/__tests__/ShallowWrapper-spec.js
```diff
@@ -85,6 +85,41 @@ describe('shallow', () => {
...
+      const wrapper = shallow(
+        <div>
+          <Foo />
+        </div>
+      );
+      const b = <div><Foo /></div>;
+      expect(wrapper.equals(b)).to.equal(true);
+    });
+
+  });
```
lelandrichardson: can we add 2 tests to this?

say we have a composite componente `<Foo />` which renders <Bar />`...

```
expect(shallow(<Foo />).equals(<Bar />)).to.equal(true);
expect(shallow(<Foo />).equals(<Foo />)).to.equal(false);
```

lelandrichardson:
This is great. Thanks for contributing!

One thing I'd like us to consider:

Crazy thought: What if we had this be an extension of the `.is(selector)` API instead?

The issue in this case would be the fact that this would actually conflict with #110 unless we had some way of differentiating a JSX element from an object that just looks like a JSX element...

thoughts? cc @blainekasten @ljharb @hartzis

`.equals()` might be the way to go...

lelandrichardson:
cc @vesln for how we might want to use this in chai-enzyme.

src/__tests__/ShallowWrapper-spec.js
themouette: done

themouette:
`.is(<Node />)` is appealing, but it would require similar API in `ReactWrapper`, and I'm wondering how this would behave:

``` javascript
const Bar = () => <p />;
const Foo = () => <div><Bar /></div>;

// obviously this returns true:
render(<Foo />).is(<div><p /></div>);
// should this returns true as well?
render(<Foo />).is(<div><Bar /></div>);
```

blainekasten:
Thanks for the cc @lelandrichardson. I'm personally not that excited about this API in general. My thoughts are that this gives the wrong intention to testing components.

What I think should be the primary goal in testing react components, is that given a certain set of state/props/context, an expected HTML render is asserted. Not an expected JSX call. At that point, you're just duplicating code and making sure it doesn't change. But the purpose of tests is to assert results, not implementations. Because I don't care if code get's refactored in a different way as long as the end result is the same.

If I used this method, I could see somebody doing something like this:

```jsx
const component = render(
  <div><b/></div>
);

component.equals(<div><b/></div>) // true
```

But then, say a refactor happens to this
```jsx
Foo = <b/>
const component = render(
  <div><Foo/></div>
);
```

now all of a sudden, the result is the exact same, but my tests have failed. So I just think this encourages the wrong habits. I see the benefit of this to make sure something is rendered that may be dynamic based on a component attribute, but that should be tested in other ways I think.

vesln:
@lelandrichardson :+1:

vesln:
@lelandrichardson it should be quite straight forward for chai-enzyme to implement.

`expect(xyz).to.equal(xyz)`

hartzis:
Great explanation @blainekasten. 

Though I don't agree completely that the primary goal of testing react components needs to be html output. Using shallow render you're able to keep the scope of a test to a single component and make assertions about props being passed and what nodes/components will be 'rendered'.

I could see this being useful is for shallow render testing. I also like @vesln's idea, but it does tie you to chai.

lelandrichardson:
I agree with @hartzis . While I think this API could be used in a way that I wouldn't encourage (as you describe), I think the same could be said for most of the APIs on enzyme. This provides the ability for someone to describe what a specific sub-tree of the resulting output looks like, and I think that is powerful.

After giving it some thought, I think this is fine to be on `.equals()` instead of `.is()` as well.

As a result, this PR is ready to merge.

src/ShallowWrapper.js
```diff
@@ -206,6 +206,22 @@ export default class ShallowWrapper {
...
+   * ```
+   * const wrapper = shallow(<MyComponent />);
+   * expect(wrapper.contains(<div className="foo bar" />)).to.equal(true);
+   * ```
+   *
+   * @param {ReactElement} node
+   * @returns {Boolean}
+   */
+  equals(node) {
+    return nodeEqual(this.node, node);
```
lelandrichardson: can you wrap this inside a `return this.single(() => nodeEqual(this.node, node));`

themouette:
rebased onto master.

src/ShallowWrapper.js
themouette: done

PR #126, opened 2016-01-19
describeWithDOM enhancement

nfpiche:
Let me know how this looks, wasn't 100% sure on taking describeWithDOM out of index.js


src/__tests__/Utils-spec.js
```diff
@@ -13,9 +13,11 @@ import {
   mapNativeEventNames,
 } from '../Utils';
 import {
-  describeWithDOM,
   mount,
 } from '../';
+import {
+	describeWithDOM,
```
ljharb: s/\t/  /g please

.gitignore
```diff
@@ -29,5 +29,8 @@ node_modules
 # Jetbrains IDEs
 .idea
 
+# vim swp files
```
ljharb: just like .idea above, this also seems like something that can't hurt, but should be in your global gitignore if you use vim, rather than spilling out into every project you contribute to :-/

src/index.js
```diff
@@ -9,28 +9,8 @@ import { renderToStaticMarkup } from './react-compat';
...
-try {
-  require('jsdom'); // could throw
-  jsdom = require('mocha-jsdom');
-} catch (e) {
-  // jsdom is not supported...
-}
-
 export let sinon = Sinon.sandbox.create();
 
-export function describeWithDOM(a, b) {
```
ljharb: the only problem is that this makes it a breaking change - it'd be nice if you re-exported the now-extracted `describeWithDOM`, making this PR a minor change only

ljharb:
Other than my comments, this looks awesome.

src/__tests__/Utils-spec.js
nfpiche: woops, thought I caught those, will update

.gitignore
nfpiche: I actually took a note from the above .idea line and thought it would be fine to add it in, will remove

src/index.js
nfpiche: Will do!

.gitignore
ljharb: no worries either way, i'd prefer neither but i don't care that much

src/__tests__/Utils-spec.js
lelandrichardson: @ljharb  eslint should fail on these, right?

ljharb: theoretically, but i haven't gotten around to these rules yet, so i don't know how they're configured :-)

ljharb:
This LGTM! It'd be a semver-minor.

src/__tests__/Utils-spec.js
nfpiche: This is actually why I thought I got them all, the linter wasn't yelling at me anymore, so it doesn't appear to be catching them everywhere, because I was getting warnings in other places

src/describeWithDOM.js
```diff
@@ -0,0 +1,47 @@
...
+      describe.skip(a, b);
+    } else {
+      // if jsdom isn't available, skip every test in this describe context
+      describe.skip(a, b);
+    }
+  });
+}
+
+describeWithDOM.only = only;
+describeWithDOM.skip = skip;
```
vesln: in order to eliminate duplication, you could do smth like: (psuedo code)

```js
function testSuite(register, title, fn) {
  describe('(uses jsdom)', () => {
    if (typeof jsdom === 'function') {
      jsdom();
      register(a, b);
    } else {
      // if jsdom isn't available, skip every test in this describe context
      describe.skip(a, b);
    }
  });
}

function only(title, fn) {
  testSuite(describe.only, title, fn);
}

function skip(title, fn) {
  testSuite(describe.skip, title, fn);
}

function describeWithDOM(title, fn) {
  testSuite(describe, title, fn);
}
```

vesln:
As I stated in different PRs, I think we should take care of other mocha interfaces as well (out of scope for this pr obviously), but @nfpiche might be interested in exploring

lelandrichardson:
For context, `describeWithDOM` will be removed from enzyme entirely in the 2.0 release. The idea behind it (refreshing the global document for each test) is inherently incompatible with React and causes lots of issues. We will be recommending a single document be used for your entire test-suite moving forward.

That said, I'm fine with merging this in for 1.x releases

package.json
```diff
@@ -14,7 +14,9 @@
     "check": "npm run lint && npm run test:all",
     "build": "babel src --out-dir build",
     "test:watch": "mocha --compilers js:babel/register --recursive src/**/__tests__/*.js --watch",
-    "test:all": "npm run react:13 && npm test && npm run react:14 && npm test",
+    "test:describeWithDOMOnly": "mocha --compilers js:babel/register --recursive src/**/__tests__/describeWithDOM/describeWithDOMOnly-spec.js",
+    "test:describeWithDOMSkip": "mocha --compilers js:babel/register --recursive src/**/__tests__/describeWithDOM/describeWithDOMSkip-spec.js",
+    "test:all": "npm run react:13 && npm test && npm run test:describeWithDOMOnly && npm run test:describeWithDOMSkip && npm run react:14 && npm test && npm run test:describeWithDOMOnly && npm run test:describeWithDOMSkip",
```
lelandrichardson: i'm confused. Why exactly were these tests needed to be added separately?

nfpiche: We didn't technically need it for .skip, but I felt like having it be consistent would be the way to go.  We do need a separate run for the .only tests though, otherwise it will skip the rest of the test suite.  Happy to change things up a bit to be less verbose if you have an idea though.

lelandrichardson: I suppose that's fine. I'm not 100% sure we even need to add tests for describeWithDOM, but since they're already added it's better than nothing.

PR #129, opened 2016-01-20
Property selector: Throw when using non-literal

### LINKED ISSUE #127: .find('[type=text]') behaves like .find('[type]')
If I have a component that renders a form that has one input with `type=text` on it and a few others that have `type` attributes with other values, `.find('[type=text'])` will give me all the nodes that have a `type` attribute. I would normally expect to get just the one that has a `type` attribute with a value of `"text"`, or see an error. To fix this in my test, I am quoting the value of the attribute: `.find('[type="text"]')`.

blainekasten:
Fixes #127 by throwing an error when a non-literal value is supplied to the property selector.

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -168,8 +168,17 @@ export function coercePropValue(propValue) {
     return numericPropValue;
   }
 
-  // coerce to boolean
-  return propValue === 'true' ? true : false;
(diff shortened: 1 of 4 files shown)


blainekasten:
Closing in favor of #130

PR #130, opened 2016-01-20
Fixed confusing prop selector behavior. Added helpful warning

### LINKED ISSUE #127: .find('[type=text]') behaves like .find('[type]')

lelandrichardson:
to: @lencioni @ljharb @blainekasten 

When users were passing in selectors with a non-quoted string such as `[type=foo]` it was causing undesired behavior, matching any component that had the `type` prop.  This refactors the logic so that if the string is unquoted (and not an acceptable numeric literal or boolean), we throw a helpful error indicating that it should be quoted.

Fixes #127; Closes #129

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -149,19 +149,21 @@ export function AND(fns) {
   };
 }
 
-export function coercePropValue(propValue) {
+export function coercePropValue(propName, propValue) {
   // can be undefined
   if (propValue === undefined) {
     return propValue;
   }
 
+  const trimmedValue = propValue.trim();
+
   // if propValue includes quotes, it should be
   // treated as a string
-  if (propValue.search(/"/) !== -1) {
-    return propValue.replace(/"/g, '');
+  if (/^(['"]).*\1$/.test(trimmedValue)) {
+    return trimmedValue.slice(1, -1);
   }
 
-  const numericPropValue = parseInt(propValue, 10);
+  const numericPropValue = +trimmedValue;
 
   // if parseInt is not NaN, then we've wanted a number
   if (!isNaN(numericPropValue)) {
@@ -169,7 +171,14 @@ export function coercePropValue(propValue) {
   }
 
   // coerce to boolean
-  return propValue === 'true' ? true : false;
+  if (trimmedValue === 'true') return true;
+  if (trimmedValue === 'false') return false;
+
+  // user provided an unquoted string value
+  throw new TypeError(
+    `Enzyme::Unable to parse selector '[${propName}=${propValue}]'. ` +
+    `Perhaps you forgot to escape a string? Try '[${propName}="${trimmedValue}"]' instead.`
+  );
 }
 
 export function mapNativeEventNames(event) {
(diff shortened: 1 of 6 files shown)


src/Utils.js
```diff
@@ -169,7 +170,14 @@ export function coercePropValue(propValue) {
   }
 
   // coerce to boolean
-  return propValue === 'true' ? true : false;
+  if (propValue === 'true') return true;
+  if (propValue === 'false') return false;
+
+  // user provided an unquoted string value
+  throw new Error(
```
ljharb: this should probably be a `TypeError`

src/Utils.js
```diff
@@ -157,11 +158,11 @@ export function coercePropValue(propValue) {
 
   // if propValue includes quotes, it should be
   // treated as a string
-  if (propValue.search(/"/) !== -1) {
-    return propValue.replace(/"/g, '');
+  if (/^(['"]).*\1$/.test(propValue.trim())) {
```
ljharb: we should cache `propValue.trim()` so we don't have to call it twice.

lelandrichardson: I thought we don't care about performance? :grin:

src/Utils.js
```diff
@@ -149,7 +149,8 @@ export function AND(fns) {
   };
 }
 
-export function coercePropValue(propValue) {
+export function coercePropValue(propValue, propName) {
```
blainekasten: every other internal function for this has had the signature of `fn(key, value) {}` maybe swapping these two for consistency sake?

lelandrichardson: :+1: sounds raisinable

src/Utils.js
blainekasten: I also like what's going on this [PR](https://github.com/airbnb/enzyme/pull/110/files#diff-c33afeb87a3143ade6f458848655083bR141) with prefacing these errors with `Enzyme::Selector`

lelandrichardson: :+1: i know, i know. my laziness is getting the best of me.

src/Utils.js
ljharb: producing a value once, and operating on that, rather than baking in the assumption that it will be idempotent when called twice, helps both clarity and performance :-p

lencioni: Alternatively to caching, we could do this by modifying the regex a little.

```js
const matches = propValue.match(/^\s*(['"])(.*)\1\s*$);
if (matches) {
  return matches[2];
}
```

or something like that

lencioni: Although I don't know if that's actually faster, so I would go with whatever is clearest (which is likely not the regex, lol).

lencioni:
:+1:

PR #133, opened 2016-01-22
Possible error in example.

dbillingham:
I think in this example the className attribute should contain the name prop rather than the state name.


vesln:
:+1:

lelandrichardson:
Thanks!

PR #135, opened 2016-01-24
Fix shallowwrapper not calling events with lowercase names

benstokoe:
When using the ShallowHelper you could not use lowercase event names such as `keydown`. This is because the propsFromEvent function is expecting the camelcased version.
To fix run the event through mapToNativeEventNames.


lelandrichardson:
awesome! Thank for helping out.

ljharb:
LGTM, thanks!

PR #140, opened 2016-01-26
Mention chai-enzyme in readme

ayrton:
Mentions [chai-enzyme](https://github.com/producthunt/chai-enzyme) in the readme.

If you think the wording is wrong, feel free to suggest any other text and I'll be happy to change (or move to a different place).

We could even go stronger in wording if we want to push users more towards the usage of the library:

> For convenience functions implementated as Chai.js custom assertions install [chai-enzyme](https://github.com/producthunt/chai-enzyme).

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ compatible with all major test runners and assertion libraries out there. The do
 examples for enzyme use [mocha](https://mochajs.org/) and [chai](http://chaijs.com/), but you
 should be able to extrapolate to your framework of choice.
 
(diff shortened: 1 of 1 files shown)


lelandrichardson:
I've been meaning to add this myself!

Perhaps we should also link the words `Chai.js` to http://chaijs.com/?

ayrton:
@lelandrichardson I can add it 😄, reason why I didn't was because it was already linked in the paragraph above

lelandrichardson:
Ah, fair enough. I'll go ahead and merge this then. Thanks!

PR #148, opened 2016-01-29
fix #145 `contains` does not fail on string input

### LINKED ISSUE #145: `contains` does not fail on string input
The API for `contains()` clearly indicates that it should be provided a node, and yet if a string is provided instead, it seems that it will return `true`, no matter the string.  I would have expected an error to be thrown, or at least that `false` would be returned.

smacker:
Also I added a check for number:

```
    it('should work with numbers', () => {
      const wrapper = shallow(<div>1</div>);

      expect(wrapper.contains(1)).to.equal(true);
      expect(wrapper.contains(2)).to.equal(false);
    });
```

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -64,6 +64,7 @@ export function nodeEqual(a, b) {
   if (a === b) return true;
   if (!a || !b) return false;
   if (a.type !== b.type) return false;
+
   const left = propsOfNode(a);
   const leftKeys = Object.keys(left);
   const right = propsOfNode(b);
@@ -80,7 +81,12 @@ export function nodeEqual(a, b) {
       return false;
     }
   }
-  return leftKeys.length === Object.keys(right).length;
+
+  if (typeof a !== 'string') {
+    return leftKeys.length === Object.keys(right).length;
+  }
+
+  return false;
 }
 
 // 'click' => 'onClick'
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -82,7 +82,7 @@ export function nodeEqual(a, b) {
     }
   }
 
-  if (typeof a !== 'string') {
+  if (typeof a !== 'string' && typeof a !== 'number') {
     return leftKeys.length === Object.keys(right).length;
   }
 
--- a/src/__tests__/ShallowWrapper-spec.js
+++ b/src/__tests__/ShallowWrapper-spec.js
@@ -83,6 +83,20 @@ describe('shallow', () => {
       expect(wrapper.contains(b)).to.equal(true);
     });
 
+    it('should work with strings', () => {
+      const wrapper = shallow(<div>foo</div>);
+
+      expect(wrapper.contains('foo')).to.equal(true);
+      expect(wrapper.contains('bar')).to.equal(false);
+    });
+
+    it('should work with numbers', () => {
+      const wrapper = shallow(<div>1</div>);
+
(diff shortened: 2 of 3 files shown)


src/__tests__/ShallowWrapper-spec.js
```diff
@@ -83,6 +83,20 @@ describe('shallow', () => {
...
     });
 
+    it('should work with strings', () => {
+      const wrapper = shallow(<div>foo</div>);
+
+      expect(wrapper.contains('foo')).to.equal(true);
+      expect(wrapper.contains('bar')).to.equal(false);
+    });
+
+    it('should work with numbers', () => {
```
ljharb: can you add a counter-test here, that in `<div>1</div>`, `wrapper.contains('1')` is `false`?

smacker: oh. Actually I did wrong comparison with numbers. Fixed it.

src/Utils.js
```diff
@@ -80,7 +81,12 @@ export function nodeEqual(a, b) {
       return false;
     }
   }
-  return leftKeys.length === Object.keys(right).length;
+
+  if (typeof a !== 'string' && typeof a !== 'number') {
+    return leftKeys.length === Object.keys(right).length;
```
lelandrichardson: @ljharb I just noticed this, but should we use a lib for `Object.keys` instead of assuming it's available?

src/__tests__/ShallowWrapper-spec.js
```diff
@@ -83,6 +83,21 @@ describe('shallow', () => {
...
+      const wrapper = shallow(<div>{1}</div>);
+
+      expect(wrapper.contains(1)).to.equal(true);
+      expect(wrapper.contains(2)).to.equal(false);
+      expect(wrapper.contains('1')).to.equal(false);
+    });
```
lelandrichardson: can we add another test or two that make sure this works properly for multiple nested levels?


```jsx
const wrapper = shallow(
  <div>
    <div>
      <div>{5}</div>
    </div>
  <div>
);

expect(wrapper.contains(5)).to.equal(true);
expect(wrapper.contains(<div>{5}</div>)).to.equal(true);
```

src/Utils.js
blainekasten: I think it's pretty safe to assume it's available.

`Object.keys` has really wide support. On the browser side you'd have to use less than IE9, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys

On node side, it's been included since at least 0.10. I'm not sure exactly when it was added to V8, but I think being in 0.10 makes it more than safe to use natively

ljharb: It's part of ES5 - node 0.8 and higher, and IE 9 and higher have it.

However, if you want to support older engines (which I strongly recommend!) you can use the https://npmjs.com/object-keys module :-)

lelandrichardson: I suppose that since react requires ES5 and above, it is safe to conclude that at least an es5-shim will be loaded in any environment enzyme will be used...

ljharb: SGTM

smacker:
Tests added. It's safe to assume Object.keys is available, if you run it inside IE8, you definitely have some polyfills. es5-shrink or babel-polyfill, because otherwise almost nothing will work.

PR #149, opened 2016-01-29
Make correct output for number children in debug

roadhump:
Now we have strange `<undefined />` tag instead of child of number type. This PR should fix it.

### DIFF
--- a/src/Debug.js
+++ b/src/Debug.js
@@ -43,8 +43,8 @@ function propsString(node) {
 }
 
 export function debugNode(node, indentLength = 2) {
+  if (typeof node === 'string' || typeof node === 'number') return escape(node);
   if (!node) return '';
-  if (typeof node === 'string') return escape(node);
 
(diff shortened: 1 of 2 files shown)


src/Debug.js
```diff
@@ -44,7 +44,7 @@ function propsString(node) {
 
 export function debugNode(node, indentLength = 2) {
   if (!node) return '';
```
lelandrichardson: While i'm looking at this, it seems like this should be
```
if (!node && node !== 0) return '';
```

Since react will stringify `0` to `'0'`

lelandrichardson:
This looks good to me. Mind implementing my one comment + corresponding test for bonus points??? :smiley:

lelandrichardson:
Thanks for contributing, btw!

roadhump:
@lelandrichardson Fixed issue with zero, thank you

PR #150, opened 2016-01-29
Make correct output for component name in debug

roadhump:
Composite components without `displayName` property are output as `<Component />` now. Debug can rely of `name` property of component if it is a class or a function and output it as component name.


lelandrichardson:
Awesome! Thanks.

PR #155, opened 2016-02-02
Added mount/unmount APIs to ReactWrapper

lelandrichardson:
These APIs only work if you are using a single document across tests, rather than using the `describeWithDOM` method which recycles the global document. This is because React caches a reference to the global document under the hood, which we can not do anything about unfortunately.


lelandrichardson:
@ljharb or @goatslacker PTAL. I'm adding documentation right now. will update with that in a sec.

lelandrichardson:
Also important to note: This PR changes our jsdom test strategy to load a single document for the entire test suite, which is the strategy we will be recommending moving forward (in enzyme 2.0 and onward)

ljharb:
LGTM

PR #156, opened 2016-02-02
Add render method to ReactWrapper

lelandrichardson:
to: @goatslacker @ljharb 

Adds render method. pretty simple.


ljharb:
LGTM

PR #158, opened 2016-02-02
Add .debug() method to ReactWrapper

### LINKED ISSUE #128: .debug() is not on the mount API
It is useful to use `.debug()` when using the shallow API, but it is missing from the mount API. It would be nice to add this convenient method to aid in debugging.

lelandrichardson:
to: @lencioni @ljharb @goatslacker 

Adds a debug method to the ReactWrapper. Mirrors the API and implementation of the method on ShallowWrapper.

Fixes #128


src/Debug.js
```diff
@@ -63,3 +73,61 @@ export function debugNode(node, indentLength = 2) {
 export function debugNodes(nodes) {
   return nodes.map(debugNode).join('\n\n\n');
 }
+
+export function debugInst(inst, indentLength = 2) {
+  if (typeof inst === 'string' || typeof inst === 'number') return escape(inst);
+  if (!inst) return '';
```
ljharb: just want to confirm - these two lines will not do anything special for `true`, or for Symbols.

lelandrichardson: hmm.... not that I'm aware of.

I don't actually know what calling this on `<div>{true}</div>` would do, but it should do whatever react itself does...

No idea how or why or what symbols would be doing getting called in this function....?

lelandrichardson: just checked and `<div>{true}</div>` gets rendered as `<div />`

src/Debug.js
```diff
@@ -63,3 +73,61 @@ export function debugNode(node, indentLength = 2) {
...
+
+  // do stuff with publicInst
+  const currentElement = inst._currentElement;
+  const type = typeName(currentElement);
+  const props = propsString(currentElement);
+  const children = [];
+  if (isDOMComponent(publicInst)) {
+    const renderedChildren = renderedChildrenOfInst(inst) || childrenOfNode(currentElement);
+    let key;
+    for (key in renderedChildren) {
```
lencioni: Instead of 

```js
let key;
for (key in renderedChildren) {
```

why not

```js
for (const key in renderedChildren) {
```

?

Or, can we use `Object.values` yet?

lencioni: Or would `Array.from` work here? I'm not really familiar with what `_renderedChildren` looks like.

src/__tests__/Debug-spec.js
```diff
@@ -188,4 +189,92 @@ describe('debug', () => {
...
+      expect(mount(<Bar id="2" />).find(Foo).debug()).to.eql(
+        `<Foo baz="bax">
+  <div className="foo">
+    <span>
+      Foo
+    </span>
+  </div>
+</Foo>`);
+    });
+  });
```
lencioni: It would be nice to see an example where a component is rendering `this.props.children` somewhere.

src/ReactWrapper.js
```diff
@@ -699,4 +702,13 @@ export default class ReactWrapper {
     }
     return new ReactWrapper(node, this.root);
   }
+
+  /**
+   * Returns an html-like string of the shallow render for debugging purposes.
```
lencioni: nit: HTML

src/__tests__/Debug-spec.js
lelandrichardson: good point. I'll add that.

src/Debug.js
lelandrichardson: Object.values might work. This ends up not being an array. Not really sure why though. A lot of this code started out from source copied from react core.

src/Debug.js
ljharb: In that case, wouldn't we want `if (inst === true) return ''`? or am i misunderstanding something.

what about `<div>{Symbol('foo')}</div>`?

src/Debug.js
ljharb: You can pull in https://npmjs.com/object.values if you want to use it ;-)

vesln:
:+1: we can leverage it in chai-enzyme cc @ayrton

src/Debug.js
lelandrichardson: @ljharb `<div>{Symbol('foo')}</div>` also renderes an empty div. I think this function isn't being used exactly how you imagine it.  Though i'm not sure I event 100% understand it... lol

PR #159, opened 2016-02-02
Rip out sinon, mocha, and jsdom dependencies

### LINKED ISSUE #44: Remove direct dependency on Sinon

### LINKED ISSUE #55: Jasmine support

### LINKED ISSUE #66: Jest support documentation
In order for this to work with jest, jest needs to know about some libraries specifically not to be mocked. It'd be nice if you could document this somewhere. This is specifically what I had to do to get tests running:

`/package.json`
```json
{
  "jest": {
    "unmockedModulePathPatterns": [
      "enzyme",
      "cheerio",
      "htmlparser2",
      "sinon",
      "..."
    ]
  }
}
```

This is an exhaustive list of needed items
- [ ] enzyme
- [ ] cheerio
- [ ] htmlparser2
- [ ] sinon
- [ ] underscore
- [ ] lodash  `dep of cheerio, needed for render method`
- [ ] domhandler `dep of htmlparser2, from cheerio`
- [ ] object.assign
- [ ] define-properties
- [ ] function-bind
- [ ] object-keys

### LINKED ISSUE #73: jsdom dependency?
I'm just curious as to why `jsdom` is required for the full rendering. I understand the need for a DOM, but what about in environments in which a DOM already exists? In our current architecture, we're running our tests directly in the browser via Karma.

I do see that `jsdom` is an optional dependency, and that its require statement is wrapped in a try/catch. Does this mean that it would still work if you're using some alternative DOM? If so, then just the docs are misleading.

### LINKED ISSUE #75: How do you use mount/jsdom without mocha?
Hi. I've loaded jsdom like this:

```
jsdom.env({
    html: '<html><body></body</html>',
    done: () => runTests()
});
```

And just to be through I tried const document = jsdom.env... and const window = jsdom.env ...

The runTests method had this statement: mount(<Text content={content} />);

But ever time I run it, I get the following error: "It looks like you called `mount()` without a jsdom document being loaded. ' + 'Make sure to only use `mount()` inside of a `describeWithDOM(...)"

Your documentation says I don't need to use describeWithDOM if I'm not using mocha.

### LINKED ISSUE #80: When running with wallaby.js the mocha-jsdom setup breaks often

### LINKED ISSUE #104: Mounting with jsdom on tape-based testing frameworks
I'm currently using [ava](https://github.com/sindresorhus/ava) to execute unit tests, and it's based on the tape system. From what I gathered, enzyme requires `describe` to exist in order to do full jdsom mounting, and that's not a construct available in ava.

Just wondering if there's currently a workaround or perhaps plans to support this in the future.

Thanks!

### LINKED ISSUE #132: Is it possible to run Enzyme tests in a browser?
I like the library a lot, but I was wondering if it's possible to run tests written with enzyme in a browser? Say I use the full DOM rendering mode which uses jsdom, then run the test in Chrome. Will it use Chrome's DOM API instead of jsdom? 
I think it would be beneficial to still be able to run tests in a real browser, surely there can be subtle differences between jsdom and the real thing.

Thanks

lelandrichardson:
This refactor addresses several issues regarding enzyme working in a variety of environments, and
the library just generally not making any assumptions about the environment that tests will be
run in.

For the most part, this amounts to:

- remove direct dependency on jsdom
- remove direct dependency on sinon
- remove assumed dependency on mocha

In addition to this, I would like to create several "example" projects that are some basic boilerplate
to getting enzyme up and running with some combination of bundler/test runner/etc.  These projects
can end up being devDependencies to enzyme and we can run their tests as part of our tests, which will
ensure that changes we make to enzyme will be compatible with environments we claim to support moving
forward.

Lastly, as a matter of organization, tests have been moved from `src/__tests__/*` to `tests/*`.

Left to do for this to be mergable is:

  [ ] Add a "guides" section in the docs explaining how to use enzyme in different environments
  [ ] Add example projects as dev dependencies, include their tests in enzyme's test script

NOTE: This will be a semver-MAJOR change because of the removal of `describeWithDOM` and `sinon` etc. The main `mount`, `shallow`, and `render` APIs have no breaking changes.

Fixes #132 
Fixes #104 
Fixes #80 
Fixes #75 
Fixes #73 
Fixes #66 
Fixes #55 
Fixes #44


docs/guides/mocha.md
```diff
@@ -0,0 +1,21 @@
...
+```jsx
+import { mount } from 'enzyme';
+import { describeWithDOM } from 'enzyme-mocha';
+
+describe('<Foo />', () => {
+
+  it('calls componentDidMount', () => {
+    spy(Foo.prototype, 'componentDidMount');
+    const wrapper = mount(<Foo />);
+    expect(Foo.prototype.componentDidMount.calledOnce).to.be.true;
```
ljharb: these examples shouldn't use the getter assertions - `.to.equal(true)`?

lelandrichardson: true. I'm going to be rewriting these examples anyway :)

ljharb:
LGTM overall, pending the one comment - but before we merge this, we should cut a (hopefully) final release on the 1.x line with everything that's in master

test/Debug-spec.js
```diff
@@ -4,10 +4,10 @@ import {
   spaces,
   indent,
   debugNode,
-} from '../Debug';
-import { mount } from '../';
-import { describeWithDOM, itIf } from './_helpers';
-import { REACT013 } from '../version';
+} from '../src/Debug';
+import { mount } from '../src/';
+import { describeWithDOM,itIf } from './_helpers';
```
blainekasten: nit: space should be added after comma

docs/api/mount.md
```diff
@@ -4,8 +4,11 @@ Full DOM rendering is ideal for use cases where you have components that may int
 or may require the full lifecycle in order to fully test the component (ie, `componentDidMount`
 etc.)
 
-Full DOM rendering  depends on a library called [jsdom](https://github.com/tmpvar/jsdom) which is
-essentially a headless browser implemented completely in JS.
+Full DOM rendering requires that a full DOM API be available at the global scope. This means that 
+it must be run in an environment that at least "looks like" a browser environment. If you do not 
+want to run your tests inside of a browswer, the recommended approach to using `mount` is to depend
```
smacker: typo. browser.

lelandrichardson:
@ljharb Made some documentation updates. PTAL.

lelandrichardson:
Main thing left is i'm going to add the referenced example repos in a special test script so we can be reactive regarding regressions in these environments

lelandrichardson:
Update: Got the "Example" projects pulled in as dev-deps and running those tests is now part of CI. I just have to add in the other projects (only mocha and karma are included ATM)

.travis.yml
```diff
@@ -1,30 +1,23 @@
...
   - "0.12"
-  - "0.11"
   - "0.10"
 before_install:
   - 'if [ "${TRAVIS_NODE_VERSION}" != "0.9" ]; then case "$(npm --version)" in 1.*) npm install -g npm@1.4.28 ;; 2.*) npm install -g npm@2 ;; esac ; fi'
   - 'if [ "${TRAVIS_NODE_VERSION}" != "0.6" ] && [ "${TRAVIS_NODE_VERSION}" != "0.9" ]; then npm install -g npm; fi'
 before_script: "sh install-relevant-react.sh"
 script:
-  - 'if [ "${TRAVIS_NODE_VERSION}" = "4.2" ]; then npm run lint && npm run travis ; elif [ "${TRAVIS_NODE_VERSION}" = "0.12" ]; then npm run travis ; else npm test ; fi'
+  - 'if [ -z "$REACT" ] && [ "${TRAVIS_NODE_VERSION}" = "4" ]; then npm run test:env -- "${EXAMPLE}" ; elif [ -z "$REACT" ]; then echo "Test Skipped" ; elif [ "${TRAVIS_NODE_VERSION}" = "4" ]; then npm run lint && npm run travis ; elif [ "${TRAVIS_NODE_VERSION}" = "0.12" ]; then npm run travis ; else npm test ; fi'
```
ljharb: we should also make sure `npm test` locally runs all the `npm run test:env` scripts so that devs don't miss them

PR #162, opened 2016-02-03
[Fix] `{ShallowWrapper,ReactWrapper}#{type,props}()` should work with `null` nodes

### LINKED ISSUE #113: Cannot read property 'type' of null on ''.type()
If returning an empty string/null inside a react component and trying to call the type() function on that node will result in the following  error: Cannot read property 'type' of null. This can be reproduced when trying to the use the shallow rendering filterWhere() API method. 
Please see the below code snippet:

```javascript
const MyComp = ({foo}) => {
    return (
          <div>
                 <MyOtherComp />
                 {foo !== 0 && (
                      <MyAnotherComp />
                  )}
          </div>
    );
});

it('should rock the world', () => {
    const wrapper = shallow(<MyComp foo={0});
    const result = wrapper.findWhere(n => (
          n.type() ===  MyAnotherComp
    );
   expect('nothing because "Cannot read property \'type\' of null" will occur');
});
```
In my test I want to check if the value of a specific prop is for e.g. {bar} then component should not be rendered.
Are there any workarounds for this issue ?
Thanks !

ljharb:
Fixes #113.

^ @lelandrichardson, @goatslacker

### DIFF
--- a/src/ReactWrapper.js
+++ b/src/ReactWrapper.js
@@ -21,6 +21,7 @@ import {
   mapNativeEventNames,
   containsChildrenSubArray,
   propsOfNode,
+  typeOfNode,
 } from './Utils';
 import {
   debugInsts,
@@ -485,10 +486,7 @@ export default class ReactWrapper {
    * @returns {String|Function}
    */
   type() {
-    return this.single((n) => {
-      const node = getNode(n);
-      return node ? node.type : null;
-    });
+    return this.single(n => typeOfNode(getNode(n)));
   }
 
   /**
--- a/src/ReactWrapper.js
+++ b/src/ReactWrapper.js
@@ -20,6 +20,7 @@ import {
 import {
   mapNativeEventNames,
   containsChildrenSubArray,
+  propsOfNode,
 } from './Utils';
 import {
   debugInsts,
@@ -381,7 +382,7 @@ export default class ReactWrapper {
    * @returns {Object}
    */
   props() {
-    return this.single(n => getNode(n).props || {});
+    return this.single(propsOfNode);
   }
 
   /**
@@ -484,7 +485,10 @@ export default class ReactWrapper {
    * @returns {String|Function}
    */
   type() {
-    return this.single(n => getNode(n).type);
+    return this.single((n) => {
+      const node = getNode(n);
+      return node ? node.type : null;
+    });
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/ReactWrapper.js
```diff
@@ -381,7 +381,10 @@ export default class ReactWrapper {
    * @returns {Object}
    */
   props() {
-    return this.single(n => getNode(n).props || {});
+    return this.single((n) => {
+      const node = getNode(n);
+      return node ? node.props || {} : {};
```
lelandrichardson: i think in the Utils file there is a `propsOfNode` function that we should probably be using instead of this. It deals with things like null being passed in.

src/ReactWrapper.js
```diff
@@ -484,7 +487,10 @@ export default class ReactWrapper {
    * @returns {String|Function}
    */
   type() {
-    return this.single(n => getNode(n).type);
+    return this.single((n) => {
+      const node = getNode(n);
+      return node ? node.type : null;
```
lelandrichardson: :+1:

src/ReactWrapper.js
ljharb: Sounds good - will that work the same in shallow and mount?

src/__tests__/ShallowWrapper-spec.js
```diff
@@ -484,6 +484,53 @@ describe('shallow', () => {
       expect(spy.args[3][0].hasClass('bux')).to.equal(true);
     });
 
+    describeIf(!REACT013, 'stateless functional components', () => {
+      it('finds nodes', () => {
```
lelandrichardson: these tests don't really seem to correspond with the behavior that was changed in this PR.... right?

ljharb: the first "it" block is just a sanity check (and passed prior to my change) - the second one is the one that changed in this PR (and failed prior to my change)

same in the ReactWrapper specs

jedwards1211:
@ljharb I'm wondering -- why would we want wrappers with null nodes to get passed to `findWhere` anyway?  I'm assuming there are still other methods that won't work with null nodes (I stumbled on this problem with `.text()`).  Wouldn't it be more robust to simply not pass wrappers with null nodes?

ljharb:
That's a good point, and may be an alternative approach.

jedwards1211:
Yeah I was going to post an issue about `.text()` in `.findWhere()` but I could just post an issue about null nodes instead

ljharb:
Let's do that - findWhere and filterWhere etc perhaps shouldn't return `null` nodes? It's worth a discussion.

jedwards1211:
Okay, sounds good

jedwards1211:
And thanks for this awesome project!

PR #170, opened 2016-02-07
fixed setProps execption swallowing issue and added regression test

psfrankie:
Let me know if there's a problem w this!


lelandrichardson:
LGTM. @ljharb I'm thinking this is a patch-level change as it's fixing a bug. Thoughts?

ljharb:
I agree. I'm not entirely sure why the promise was there in the first place tho - is there some benefit it provided (beyond swallowing errors) that might make this a breaking change?

PR #171, opened 2016-02-07
add option for childContextTypes of `ReactWrapper`

### LINKED ISSUE #144: Pass a context to rendered Component's child?
I am using `enzyme.mount` to render a Component with some child Components and I need to pass a context to these child Components. I can pass a context with `options.context` to the Component but I would like to pass something else to Component's children. 

The options object in `mount` would be good place to do this, however, it is currently not possible. I know I can create my own wrapper with childContextTypes, but for better readability and simplicity of tests, to do this directly with `enzyme.mount` would be amazing.

What do you think?

jakubzitny:
Fixes #144.

### DIFF
--- a/src/ReactWrapperComponent.jsx
+++ b/src/ReactWrapperComponent.jsx
@@ -71,12 +71,18 @@ export default function createWrapperComponent(node, options = {}) {
     },
   };
 
-  if (options.context && node.type.contextTypes) {
+  if (options.context && (node.type.contextTypes || options.childContextTypes)) {
     // For full rendering, we are using this wrapper component to provide context if it is
-    // specified in both the options AND the child component defines `contextTypes` statically.
+    // specified in both the options AND the child component defines `contextTypes` statically
+    // OR the merged context types for all children (the node component or deeper children) are
+    // specified in options parameter under childContextTypes.
     // In that case, we define both a `getChildContext()` function and a `childContextTypes` prop.
+    let childContextTypes = node.type.contextTypes;
+    if (options.childContextTypes) {
+      childContextTypes = options.childContextTypes;
+    }
     objectAssign(spec, {
-      childContextTypes: node.type.contextTypes,
+      childContextTypes,
       getChildContext() {
         return this.state.context;
       },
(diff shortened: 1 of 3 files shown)


src/ReactWrapperComponent.jsx
```diff
@@ -71,12 +71,18 @@ export default function createWrapperComponent(node, options = {}) {
...
+  if (options.context && (node.type.contextTypes || options.childContextTypes)) {
     // For full rendering, we are using this wrapper component to provide context if it is
-    // specified in both the options AND the child component defines `contextTypes` statically.
+    // specified in both the options AND the child component defines `contextTypes` statically
+    // OR the merged context types for all children (the node component or deeper children) are
+    // specified in options parameter under childContextTypes.
     // In that case, we define both a `getChildContext()` function and a `childContextTypes` prop.
+    let childContextTypes = node.type.contextTypes;
+    if (options.childContextTypes) {
+      childContextTypes = options.childContextTypes;
```
lelandrichardson: should we merge `node.type.contextTypes` and `options.childContextTypes`?

jakubzitny: Yeah, it's probably better. I'll update this, however, I think there should be better automatic fix to this issue.

lelandrichardson:
@jakubzitny I suppose you mean by traversing the tree and finding the required context types of the descendants?

I'm really not sure if this is possible. At the time this code executes, the `node` is only the output of `React.createElement()` on the root node you are mounting. Traversing it's rendered subtree would require actually rendering, which we can't do until this component is created.

It's possible I'm not fully grokking the way context is working, but i'm unsure how react itself solves this if this is how context works.

jakubzitny:
Yeah I meant something like that. I know it's not really possible without rendering. :(

I was thinking about just looking at the contextTypes of all children somehow. Isn't it really possible to get the children of a component without rendering somehow? By looking at the `render` function, or just rendering it shallowly, then getting the children from `props().children` and then rendering shallowly their children, or something like that?

The thing is that in React if you do this, you always specify both `childContextTypes` and also `getChildContext`. So it is not handled automatically. But for the test cases with enzyme it'd be great if it was automatic, so you wouldn't need to specify that many things and making the tests longer..

---

Btw I pushed a fixup for merging both childContextTypes. I will rebase it if you want to accept this PR.

lelandrichardson:
Going to go ahead and merge this in. I think this is the best way to do this at this point. Open to PRs for alternatives though.

blainekasten:
Is there any specific reason this wasn't built into the shallow wrapper API?

jakubzitny:
@blainekasten there is no need for it in ShallowWrapper..

PR #174, opened 2016-02-10
Prevent null or false nodes from being passed into tree traversal

### LINKED ISSUE #166: null nodes are passed to findWhere()/filterWhere() if tree contains stateless components
If the render tree contains stateless components (a.k.a. pure function components), wrappers with `null` `node`s are passed to `findWhere()` and (I'm assuming) `filterWhere()`.

```js
  const MyComp = () => (<span>hello</span>);

  it("shouldn't throw at least", () => {
    const wrapper = mount(<MyComp/>);
    const result = wrapper.findWhere(n => n.text() === 'hello');
  });
```

I can't imagine anything useful we could do with a null node passed to `findWhere`, plus this was unexpected to me, so I think it would be more helpful to avoid passing null nodes to `findWhere`.

As an aside, this is not the first awkwardness since stateless components were introduced that makes me want to avoid using them...

lelandrichardson:
cc: @ljharb 

Fixes #166

### DIFF
--- a/src/ShallowTraversal.js
+++ b/src/ShallowTraversal.js
@@ -28,7 +28,9 @@ export function hasClassName(node, className) {
 }
 
 export function treeForEach(tree, fn) {
-  fn(tree);
+  if (tree !== null && tree !== false) {
+    fn(tree);
+  }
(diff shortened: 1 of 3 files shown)


ljharb:
LGTM!

PR #176, opened 2016-02-11
Update karma.md

debugwand:
webpack example and browserify example were the wrong way round


lelandrichardson:
:facepalm: Thanks!

Issue #179, opened 2016-02-12
Problem running with jest


mauriciosoares:
Hey guys, I'm doing some tests with enzyme in my Jest environment, and I got a problem when I ran the test command...

```
TypeError: <path>/src/shared/components/fieldset/__tests__/fields.js:
<path>/node_modules/enzyme/build/index.js:
<path>/node_modules/enzyme/build/ReactWrapper.js:
<path>/node_modules/enzyme/build/Debug.js:
<path>/node_modules/object.values/index.js:
<path>/node_modules/object.values/implementation.js:
<path>/node_modules/es-abstract/es7.js:
<path>/node_modules/es-abstract/es6.js: Cannot convert undefined or null to object
```

I believe this is cause jest is mocking the `object.values` and `es-abstract` modules, after I put them in the `unmockedModulePathPatterns` array it started working flawlessly.

Should these be put in the [docs](http://airbnb.io/enzyme/docs/guides/jest.html) as well?

Cheers.

lelandrichardson:
@mauriciosoares yes, please!

blainekasten:
It should also be noted this is a bug due to npm3. Users of npm2 won't face this issue. It's only due to jest not knowing how to mock the flat tree structure right now. This will go away when their haste2 implementation gets merged. at that point it will only need the `enzyme` module to be put in there.

mauriciosoares:
I've created a PR that adds these modules in the docs.

nice to know @blainekasten, I'll follow this up.

PR #181, opened 2016-02-12
Update Jest docs

mauriciosoares:
This commit adds 2 modules that should not be mocked by default in jest.

closes: https://github.com/airbnb/enzyme/issues/179

### DIFF
--- a/docs/guides/jest.md
+++ b/docs/guides/jest.md
@@ -22,7 +22,9 @@ several modules to be unmocked in your `package.json`:
     "object.assign",
     "define-properties",
     "function-bind",
-    "object-keys"
+    "object-keys",
+    "object.values",
+    "es-abstract"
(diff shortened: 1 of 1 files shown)


lelandrichardson:
:+1: I'll have to update the example projects to have this as well.

docs/guides/jest.md
```diff
@@ -22,7 +22,9 @@ several modules to be unmocked in your `package.json`:
     "object.assign",
     "define-properties",
     "function-bind",
-    "object-keys"
+    "object-keys",
+    "object.values",
```
ljharb: might as well add object.entries too?

lelandrichardson: @ljharb we don't use `object.entries` though... why would we need to add it to this doc?

PR #189, opened 2016-02-15
Switch to lodash

existentialism:


lelandrichardson:
@ljharb any issues you can think of with the changes of `object.values` and `object.assign`?

ljharb:
Yes, we shouldn't be using lodash when we don't have to be. @existentialism please only convert existing underscore uses to lodash.

existentialism:
@ljharb @lelandrichardson reverted those changes!

lelandrichardson:
Cool. LGTM!

PR #196, opened 2016-02-16
Ensure that `ShallowWrapper` and `ReactWrapper` handle a null child correctly

### LINKED ISSUE #52: Document how to test for null/false
I would like to add to the docs a good way to test for `null`, but I'm not sure which one is it.

This is the component:

```jsx
import React, { PropTypes } from 'react';

const Component = (props) => {
  return props.returnNull ? null : <div />;
};

Component.propTypes = {
  returnNull: PropTypes.bool
};

export default Component;
```

I want to test if setting the flag to `true` will return `null`:

```jsx
import React from 'react';
import { shallow } from 'reagent';
import { expect } from 'chai';
import Component from './Component';

describe('<Component />', () => {
  it('returns null if the flag is on', () => {
    const wrapper = shallow(
      <Component returnNull={true} />
    );
    expect(wrapper.text()).to.equal('');
  });
});
```

What is the correct way to do it? Would it perhaps be better if `wrapper.type()` returned `null` instead of throwing?

ljharb:
Fixes #52 .

### DIFF
--- a/src/ShallowWrapper.js
+++ b/src/ShallowWrapper.js
@@ -310,7 +310,7 @@ export default class ShallowWrapper {
    * @returns {String}
    */
   html() {
-    return this.single(renderToStaticMarkup);
+    return this.single(n => this.type() === null ? null : renderToStaticMarkup(n));
   }
 
   /**
@@ -321,7 +321,7 @@ export default class ShallowWrapper {
    * @returns {CheerioWrapper}
    */
   render() {
-    return cheerio.load(this.html()).root();
+    return this.type() === null ? cheerio() : cheerio.load(this.html()).root();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/ShallowWrapper.js
```diff
@@ -321,7 +321,7 @@ export default class ShallowWrapper {
    * @returns {CheerioWrapper}
    */
   render() {
-    return cheerio.load(this.html()).root();
+    return this.type() === null ? cheerio() : cheerio.load(this.html()).root();
```
lelandrichardson: did the previous implementation of this throw an exception in this case?  is `cheerio()` what we want to return?

ljharb: yes, previously it threw `TypeError: Cannot read property 'parent' of null` from inside cheerio.

`cheerio()` and `cheerio(null)` and `cheerio('')` all return an empty cheerio wrapper - i just chose the simplest of these

lelandrichardson:
LGTM.

PR #198, opened 2016-02-16
Root-level file exports

### LINKED ISSUE #195: Allow for specific imports: ie, `require('enzyme/mount')` and `require('enzyme/shallow')`
This API would allow for people who just want to work with shallow or mount etc. to not fight with their bundlers.

Ideally, `cheerio` could only be in the require path for `enzyme/render`. This may not be possible though, given that we have the `ShallowWrapper::render` API now. :/

Either way, would be a good addition.

ljharb:
This PR enables `require('enzyme/mount')`, `'enzyme/shallow'`, etc.

Fixes #195

### DIFF
--- a/ReactWrapper.js
+++ b/ReactWrapper.js
@@ -1,0 +1,1 @@
+module.exports = require('./build/ReactWrapper').default;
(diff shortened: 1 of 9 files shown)


lelandrichardson:
This LGTM.

PR #199, opened 2016-02-16
Update doc with context property

brendangibson:


docs/api/ShallowWrapper/setContext.md
```diff
@@ -32,7 +32,7 @@ const SimpleComponent = React.createClass({
 ```
 ```jsx
-const wrapper = shallow(<SimpleComponent />, { context });
```
ljharb: this is already valid ES6 concise property shorthand syntax.

PR #205, opened 2016-02-18
Remove broken link

kmkr:


ljharb:
I don't think this link is meant to be broken. We should probably fix it instead of removing it.

lelandrichardson:
Actually, this fix is correct. I was going to do a jasmine guide, but there's not really anything specific for jasmine that needs to be done. It's more the bundlers + environments that are the issue.

ljharb:
Seems like it'd still be useful to have a guide there that says "nothing needs to be done!" so Jasmine users don't think their tool is omitted.

lelandrichardson:
@ljharb fair enough. PRs welcome :)

I think there are a couple of the `enzyme-example-*` projects that are using jasmine already. we could potentially rename them to include jasmine in the name... although the names of these projects are going to get a bit crazy :)

Issue #23, opened 2015-11-16
Add nested CSS Selector support


lelandrichardson:
For example:

```
wrapper.find('.foo .bar')
```

badeball:
Hi, @lelandrichardson 

I've actually been thinking for a while about creating a CSS selector engine for the virtual DOM of React. Particularly because I recently released a DOM-agnostic XPath [implementation](https://github.com/badeball/xpath-react) that also works on it (which I've personally have found to be very useful for the purpose of testing). Perhaps this can be of use to you?

blainekasten:
Closing issue as PR has been merged! 🎉

PR #217, opened 2016-02-26
Contextual css selector

dustinsanders:
Issue #23.

Added support for descendent, >, ~, and +.

### DIFF
--- a/src/ComplexSelector.js
+++ b/src/ComplexSelector.js
@@ -1,0 +1,125 @@
+import split from 'lodash/split';
+
+export default class ComplexSelector {
+  constructor(buildPredicate, findWhereUnwrapped, childrenOfNode) {
+    this.buildPredicate = buildPredicate;
+    this.findWhereUnwrapped = findWhereUnwrapped;
+    this.childrenOfNode = childrenOfNode;
+  }
+
+  getSelectors(selector) {
+    const cleaned = selector.replace(/\s{2,}/g, ' ');
+    const selectors = split(cleaned, ' ');
+    return selectors.reduce((list, sel) => {
+      if (sel === '+' || sel === '~') {
+        const temp = list.pop();
+        list.push(sel, temp);
+        return list;
+      }
+
+      list.push(sel);
+      return list;
+    }, []);
+  }
+
+  handleSelectors(selectors, wrapper) {
+    const recurseSelector = (offset, fn, pre) => {
+      const predicate = pre || this.buildPredicate(selectors[offset]);
+      const nextWrapper = this.findWhereUnwrapped(wrapper, predicate, fn);
+      const nextSelectors = selectors.slice(offset + 1);
+      return this.handleSelectors(nextSelectors, nextWrapper);
+    };
+
+    const buildSiblingPredicate = (first, second) => {
+      const firstPredicate = this.buildPredicate(first);
+      const secondPredicate = this.buildPredicate(second);
+
+      return (child) => {
+        if (firstPredicate(child)) {
+          return (sibling) => secondPredicate(sibling);
+        }
+
+        return false;
+      };
+    };
+
+    let predicate;
+    let selectSiblings;
+
+    if (selectors.length) {
+      switch (selectors[0]) {
+        case '>':
+          return recurseSelector(1, this.treeFilterDirect());
+        case '+':
+          predicate = buildSiblingPredicate(selectors[1], selectors[2]);
+          selectSiblings = (children, pre, results, idx) => {
+            const adjacent = children[idx + 1];
+            if (pre(adjacent)) { results.push(adjacent); }
+          };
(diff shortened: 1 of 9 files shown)


src/ComplexSelector.js
```diff
@@ -0,0 +1,128 @@
...
+      return (child) => {
+        if (firstPredicate(child)) {
+          return (sibling) => secondPredicate(sibling);
+        }
+
+        return false;
+      };
+    };
+
+    let { predicate, selectSiblings } = {};
```
ljharb: this seems like a pretty sketchy way to avoid `let predicate; let selectSiblings;`

src/ComplexSelector.js
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    const predicate = this.buildPredicate(selector);
+    return this.findWhereUnwrapped(wrapper, predicate);
+  }
+
+  treeFilterDirect() {
+    return (tree, fn) => {
+      const results = [];
+      this.childrenOfNode(tree).forEach(child => {
```
ljharb: do we not provide a `filter` function here, rather than using forEach + a closure?

dustinsanders: Before in each of the Wrapper classes the `findWhereUnwrapped` always used `treeFilter`, which worked for looping through all nodes, but not for other use cases. That's why I need `treeFilterDirect` and `treeFindSiblings`

src/ComplexSelector.js
dustinsanders: I usually do this when defining multiple variables to undefined, but `let predicate; let selectSiblings;` is also fine

src/ReactWrapper.js
```diff
@@ -269,8 +270,12 @@ export default class ReactWrapper {
    * @returns {ReactWrapper}
    */
   find(selector) {
-    const predicate = buildInstPredicate(selector);
-    return findWhereUnwrapped(this, predicate);
+    const complexSelector = new ComplexSelector(
+      buildInstPredicate,
+      findWhereUnwrapped,
+      childrenOfInst
+    );
```
blainekasten: Is there a good argument for this to be a class instead of just pure functions?

src/ShallowWrapper.js
```diff
@@ -35,8 +36,8 @@ import {
  * @param {Function} predicate
```
blainekasten: missing jsdoc declaration for new argument. Same with in other Wrapper

blainekasten:
Does this only support 1 level of nested selectors? Will it be _extra_ hard to add infinite depth to nested selectors later on if/when wanted?

e.g.,
```js
find('.foo + div > span')

**Comment by dustinsanders** [comment]
@blainekasten I just ran the following

```js
it.only('.foo + div > span', () => {
    const wrapper = shallow(
      <div>
        <div className="foo" />
        <div>
          <span />
        </div>
      </div>
    );

    expect(wrapper.find('.foo + div > span').length).to.equal(1);
  });
```

and it passed. Based on how it is written it should support nesting, but there should be more testing around the nested selectors.

src/ReactWrapper.js
dustinsanders: Originally I was currying it but it didn't seem as clean.

```js
complexSelector(buildInstPredicate, findWhereUnwrapped, childrenOfInst)(selector, this);
```

I guess it could all be one function, but I wanted to separate initialization from the logic.

blainekasten: Ultimately it's up to the maintainers. They might see this argument as moot. But it could make maintenance easier when all selector engines follow a similar flow. Previously that was the case until here.

Additionally, it is heavier to instantiate an object for every selector. But I don't they worry too much about the perf for this lib.

dustinsanders:
See changes made based off of suggestions

src/ComplexSelector.js
```diff
@@ -0,0 +1,126 @@
+import reduce from 'lodash/reduce';
```
borisyankov: Array.prototype.reduce is available in ES5

dustinsanders:
@borisyankov good call, using native reduce in latest commit

koba04:
any progress on this?

dustinsanders:
I think I've addressed all the concerns, but am open to other changes

ljharb:
@dustinsanders would you mind rebasing this on top of latest master, and down to a smaller number of conceptually atomic commits? It's got merge conflicts atm.

dustinsanders:
@ljharb I have rebased and squashed to one commit

ljharb:
@dustinsanders looks like this needs a fresh rebase - sorry this has taken so long.

dustinsanders:
@ljharb this has been freshly rebased

lelandrichardson:
@dustinsanders this looks awesome. great work.

Issue #134, opened 2016-01-22
hasClass does not work on root ReactWrapper


amccloud:
I think this is because `MountedTraversal#instHasClassName` short circuits to false if the instance does not pass the `isDOMComponent` check. Could we relax the check?

Example test case (React 0.14):

```javascript
const shallowComponent = shallow(<Video src="example.mp4" />);
const mountComponent = mount(<Video src="example.mp4" />);

// From ShallowTraversal
// Works!
assert.equal(shallowComponent.node.props.className, 'Video');

// Works!
assert.isTrue(shallowComponent.hasClass('Video'));

// Fails :( because isDOMComponent(mountComponent.node) is false
// Would otherwise pass
assert.isTrue(mountComponent.hasClass('Video'));

// From MountedTraversal (without isDOMComponent check)
// Works!
assert.equal(findDOMNode(mountComponent.node).className, 'Video');

// Works! ... but seems unnecessary
assert.isTrue(mountComponent.find('.Video').hasClass('Video'));
```

lelandrichardson:
hmmm. @amccloud this is an interesting side effect of how both react + enzyme is implemented. The react render tree is sort of a two-dimensional graph, where-as in shallow it is one-dimensional.  It's hard to know what the right implementations are.

I need to look at what the side effects of this are, but I suspect relaxing the isDOMComponent requirement will end up doubling the number of results getting returned for class-based selection...

```jsx
const Bar = () => (
  <div className="bar" />
);
const Foo = () => (
  <div><Bar /></div>
);

const wrapper = mount(<Foo />);
```

```jsx
// with your suggested change:
wrapper.find('.bar');
// <Bar />
// <div className="bar" />
```
```jsx
// right now it does this...
wrapper.find('.bar');
// <div className="bar" />
```

dozoisch:
I have the exact same issue. Was working with < 1.6. I've also tested on 2.0.0, same issue.

Here is a minimal example to repro

```js
const TestComponent = ({ children, className }) => (
  <div className={className}/>
);

describe('TestComponent', () => {
  it('mount allows us to set custom props', () => {
    const wrapper = mount(<TestComponent className="test" />);
    assert(wrapper.hasClass('test')); // false
  });
  it('shallow allows us to set custom props', () => {
    const wrapper = shallow(<TestComponent className="test" />);
    assert(wrapper.hasClass('test')); // true, as expected
  });
});
```

Note: `.props()` returns an empty object.

[...]

PR #220, opened 2016-02-29
Fix props() and debug() for SFCs

ryanalane:
Mounted wrappers for Stateless Function Components now return expected values for props() and debug(), where before they returned empty values. (addresses #134)

### DIFF
--- a/test/ReactWrapper-spec.js
+++ b/test/ReactWrapper-spec.js
@@ -466,7 +466,7 @@ describeWithDOM('mount', () => {
         expect(foundNotSpan).to.have.length(0);
       });
 
-      it('should return props object when debug() is called', () => {
+      it('should return props object when props() is called', () => {
         const SFC = function SFC({ selector }) {
           return (
             <div data-foo={selector}>Test SFC</div>
@@ -476,6 +476,18 @@ describeWithDOM('mount', () => {
         const selector = 'blah';
         const wrapper = mount(<SFC data-foo={selector} />);
         expect(wrapper.props()).to.deep.equal({ 'data-foo' : selector });
+      });
+
+      it.only('should return shallow rendered string when debug() is called', () => {
+        const SFC = function SFC({ data }) {
+          return (
+            <div data-foo={data}>Test SFC</div>
+          );
+        };
+
+        const content = 'blah';
+        const wrapper = mount(<SFC data={content} />);
+        expect(wrapper.debug()).to.equal('<SFC data="' + content + '">\n  <div data-foo="' + content + '">\n    Test SFC\n  </div>\n</SFC>');
       });
     });
 
--- a/test/ReactWrapper-spec.js
+++ b/test/ReactWrapper-spec.js
@@ -465,6 +465,18 @@ describeWithDOM('mount', () => {
         ));
         expect(foundNotSpan).to.have.length(0);
       });
+
+      it('should return props object when debug() is called', () => {
+        const SFC = function SFC({ selector }) {
+          return (
+            <div data-foo={selector}>Test SFC</div>
+          );
+        };
+
+        const selector = 'blah';
+        const wrapper = mount(<SFC data-foo={selector} />);
+        expect(wrapper.props()).to.deep.equal({ 'data-foo' : selector });
+      });
     });
 
     it('should not pass in null or false nodes', () => {
--- a/test/ReactWrapper-spec.js
+++ b/test/ReactWrapper-spec.js
@@ -475,7 +475,7 @@ describeWithDOM('mount', () => {
 
         const content = 'blah';
         const wrapper = mount(<SFC data={content} />);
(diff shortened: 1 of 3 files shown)


blainekasten:
You should add some tests around this.

ryanalane:
Thanks, added.

test/ReactWrapper-spec.js
```diff
@@ -465,6 +465,30 @@ describeWithDOM('mount', () => {
...
+
+      it('should return props object when props() is called', () => {
+        const SFC = function SFC({ selector }) {
+          return (
+            <div data-foo={selector}>Test SFC</div>
+          );
+        };
+
+        const selector = 'blah';
+        const wrapper = mount(<SFC data-foo={selector} />);
```
blainekasten: this is moot, as it doesn't really test the component itself, but you are implementing the SFC incorrectly. You are passing it `{data-foo: selector}` when the deconstructor in the props is looking for something more like `{selector: ...}`. Might be worth just having this be correct to prevent any sort of misunderstanding

ryanalane: I'm afraid I am not clear on what you mean.

As far as implementing the SFC correctly goes, do you mean passing properties that make sense? (I admit that was a sloppy use of an argument named `selector` I did there) Would the SFC used in lines 482-486 fit the bill?:
``` 
        const SFC = function SFC({ data }) {
           return (
             <div data-foo={data}>Test SFC</div>
           );
         };
```
And then as far as testing the component itself goes, what would be the right approach? Happy to make the changes, just want to understand before.

ryanalane: No worries, thank you. :) SFC updated.

test/ReactWrapper-spec.js
```diff
@@ -465,6 +465,30 @@ describeWithDOM('mount', () => {
...
+      it('should return props object when props() is called', () => {
+        const SFC = function SFC({ data }) {
+          return (
+            <div data-foo={data}>Test SFC</div>
+          );
+        };
+
+        const content = 'blah';
+        const wrapper = mount(<SFC data={content} />);
+        expect(wrapper.props()).to.deep.equal({ 'data' : content });
```
ljharb: tiny nit: `{ data: content }` - ie, remove the quotes and the space before the colon

ryanalane: :+1:

lelandrichardson:
I think this is actually a breaking change, and a difference in the intent of the `.props()` method.

Before `.props()` would return `{ "data-foo": "blah" }`, and now it returns `{ data: "blah" }` because it's getting the props passed into the shallow rendered component, rather then the props of the root level node.

The former doesn't really have any actual value in the context of a test, because you're the one who just rendered the component (by passing it into the wrapper).

ljharb:
@lelandrichardson is your statement true for all kinds of components, or just SFCs?

ryanalane:
@lelandrichardson I think I understand -- so in this case, to compute an accurate return value of `props()`, it seems we need a way of traversing (or at least accessing) the SFC node structure shallow-ly (to get `{data-foo: "blah"}` out of the `<div>`) even when using a full wrapper from `mount()`, correct? Because the `mount()` wrappers use internal React instances to return properties/DOM attribute values, and SFCs don't have React instances by design--if I understand correctly.

Should full-rendering SFCs even be supported in this case?

mdotasia:
I just wanted to add that fixing `debug` for stateless functional components is really crucial. I was trying out [`redux`](https://github.com/reactjs/redux) and [`react-redux`](https://github.com/reactjs/react-redux) and was using `debug` to figure out how `redux` works under the hood.

The problem seems to be that the `redux` approach makes use of a lot of stateless functional components. And `debug` does not seem to support them. E.g. in one example `debug` displayed this

```
<Provider store={{...}}>
  <Connect(EventList) />
</Provider>
```

`Connect` is from `react-redux` and it wraps stateless functional component `EventList`, which should be an entire list of events. At first I thought I had wired up something incorrectly. But when I actually debugged the code I noticed that the entire list was actually there and it dawned on me that the culprit might be `debug`.

That's a bit unfortunate since the last thing I suspected was that `debug` might be broken. This was quite a time sink for me.

ljharb:
@mdotasia does the `Connect` wrapper set the `displayName` of the wrapped element to `Connect(EventList)`?

mdotasia:
Referring to my previous comment further up: `Provider` is a real component containing real component `Connect` with `displayName` set to `Connect(EventList)`. `Connect` then contains `StatelessComponent` `EventList` with only `name` set to `EventList` and no `displayName` set.

lelandrichardson:
@ryanalane I apologize - I misunderstood this PR. This is indeed a bug. Would you like to rebase your PR so we can merge it?  I've got the changes in my own branch which I will PR and merge if not.

PR #226, opened 2016-03-02
Changed tests to highlight a bug with contains, fixed the bug in Utils/containsChildrenSubArray

CurtisHumphrey:
moved checker to another line to make linter happy

### DIFF
--- a/test/ReactWrapper-spec.js
+++ b/test/ReactWrapper-spec.js
@@ -138,13 +138,18 @@ describeWithDOM('mount', () => {
         <div>Goodbye</div>,
       ];
 
-      const passes = [
+      const passes1 = [
         <span>Hello</span>,
         <div>Goodbye</div>,
       ];
+      const passes2 = [
+        <div>Goodbye</div>,
+        <span>More</span>,
+      ];
 
       expect(wrapper.contains(fails)).to.equal(false);
-      expect(wrapper.contains(passes)).to.equal(true);
+      expect(wrapper.contains(passes1)).to.equal(true);
+      expect(wrapper.contains(passes2)).to.equal(true);
     });
 
   });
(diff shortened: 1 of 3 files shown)


test/ReactWrapper-spec.js
```diff
@@ -139,8 +139,8 @@ describeWithDOM('mount', () => {
       ];
 
       const passes = [
-        <span>Hello</span>,
         <div>Goodbye</div>,
+        <span>More</span>,
```
ljharb: Could you include both tests (with both orderings, i mean) to ensure we don't regress in the future?

CurtisHumphrey:
Sure I added both cases as passes1 and passes2 @ljharb

ljharb:
Thanks! hmm - i can't find anything in the documentation that talks about whether order should or shouldn't matter if you pass an array of nodes. However, the usage of `some` means that `shallow(<A />).contains([<A />, <B />])` would be expected to pass (which could use a note in the docs) - so I'd expect `shallow(<A />).contains([<B />, <A />])` to pass as well.

@CurtisHumphrey it'd be awesome if you could rebase this on top of latest master - if not, no worries.

@lelandrichardson what do you think about updating the docs to explicitly mention the "any" semantics as well as that ordering doesn't matter?

CurtisHumphrey:
@ljharb I just rebased

lelandrichardson:
@ljharb I think you may have misread this function.  If the render tree contains the entire array being passed in, then it matches. This means that order does matter.  The bug was that this was only working properly if the array in the tree *started* with this sub-array.

Thanks for catching this, @CurtisHumphrey !

lelandrichardson:
@CurtisHumphrey do you mind rebasing and squashing, and I'll merge this in? Thanks!

ljharb:
@lelandrichardson ah, thanks for clarifying. I definitely misread the test, but looking at it again, what you say is indeed represented in the tests :-)

CurtisHumphrey:
@lelandrichardson rebased and squashed

Issue #225, opened 2016-03-02
react-addons-test-utils used, but not in package.json


blainekasten:
[react-compat#L73](https://github.com/airbnb/enzyme/blob/master/src/react-compat.js#L73) requires `react-addons-test-utils`

But it is not specified in the [package.json](https://github.com/airbnb/enzyme/blob/master/package.json)

Should it be a peerDep or something? This caused me a bit of confusion.

ljharb:
See the comment on https://github.com/airbnb/enzyme/blob/master/src/react-compat.js#L70-L72

PR #228, opened 2016-03-03
Add warning about implicit dependency on react-addons-test-utils

thatjessicakelly:
A warning like this should hopefully help with #225.

Not exactly sure what the text should say.

### DIFF
--- a/src/react-compat.js
+++ b/src/react-compat.js
@@ -70,7 +70,16 @@ if (REACT013) {
   // require them here through this node module. The bummer is that we are not able
   // to list this as a dependency in package.json and have 0.13 work properly.
   // As a result, right now this is basically an implicit dependency.
-  TestUtils = require('react-addons-test-utils');
+  try {
+    TestUtils = require('react-addons-test-utils');
+  } catch (e) {
+    console.error(
+      'react-addons-test-utils is an implicit dependency in order to support react@0.13-14. ' +
+      'Please add the appropriate version to your devDependencies. ' +
+      'See https://github.com/airbnb/enzyme/issues/225'
+    );
+    throw e;
+  }
 
   // Shallow rendering changed from 0.13 => 0.14 in such a way that
   // 0.14 now does not allow shallow rendering of native DOM elements.
--- a/src/react-compat.js
+++ b/src/react-compat.js
@@ -62,7 +62,19 @@ if (REACT013) {
     return TestUtils.renderIntoDocument(node);
   };
 } else {
-  const ReactDOM = require('react-dom');
+  let ReactDOM;
+
+  try {
+    ReactDOM = require('react-dom');
+  } catch (e) {
+    console.error(
+      'react-dom is an implicit dependency in order to support react@0.13-14. ' +
+      'Please add the appropriate version to your devDependencies. ' +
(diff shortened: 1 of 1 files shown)


lelandrichardson:
Hmm. This is a good idea. we should do this for all of the implicit dependencies. We could link to the README of the repo for more context as well

src/react-compat.js
```diff
@@ -70,7 +70,16 @@ if (REACT013) {
...
-  TestUtils = require('react-addons-test-utils');
+  try {
+    TestUtils = require('react-addons-test-utils');
+  } catch (e) {
+    console.error(
+      'react-addons-test-utils is an implicit dependency in order to support react@0.13-14. ' +
+      'Please add the appropriate version to your devDependencies. ' +
+      'See https://github.com/airbnb/enzyme/issues/225'
+    );
+    throw e;
```
ljharb: tbh this should probably `process.exit(1)` rather than rethrowing the error

ljharb:
How many other implicit non-peer deps are there?? ideally we have as close to zero as possible

src/react-compat.js
lelandrichardson: I think error is better. Enzyme can be run in browser-based test runners

ljharb: ah, very true, this is fine then

lelandrichardson:
We are in this situation because of 0.13 and 0.14 compatibility...

If you are using React 0.13, then you are all good. `react` is all you need, and react is a dep.
If you are using React 0.14, then you need to have `react`, `react-dom`, and `react-addons-test-utils`.

Since `react-dom` and `react-addons-test-utils` are packages that cannot be installed alongside react 0.13, we need to have implicit dependencies.

ljharb:
I wonder if making them `optionalDependencies` would be any better - i'm not sure how that field interacts with npm

lelandrichardson:
@ljharb optionalDependencies won't work. We tried that first, and shit blew up if you tried to use enzyme with 0.13

ljharb:
ah right, i'd forgotten. in that case this try/require/catch pattern is probably the cleanest way forward - maybe we could put all of those in one file (either one group or one per weird-dep) that re-exports either the libs or `undefined`, and require *that* module where they're needed?

thatjessicakelly:
Let me know how you want me to handle it and I'm happy to update this

lelandrichardson:
@ljharb i feel like there is already enough misdirection with the `react-compat` file. This file is already dealing with the differences between 0.13 and 0.14, so putting this logic here I think is sane.

@thatjessicakelly, do you think you could just update your PR to do the same thing for `react-dom` in addition to `react-addons-test-utils`?

ljharb:
Makes sense.

thatjessicakelly:
Done. I also linked to the README

capaj:
@thatjessicakelly you keep using that word "warning", yet you throw an error. Interesting.

ljharb:
@capaj that tone is not appreciated. Conceptually it's a warning, so by a technical definition of English, it's correct. You will not do well playing the condescending pedantry game here.

capaj:
> condescending pedantry game here
ok :grinning:

Issue #223, opened 2016-03-01
describeWithDOM still mentioned in mount()


danez:
When calling mount() without jsdom being initialized, it still suggests using describeWithDOM(), although it was removed. :)

https://github.com/airbnb/enzyme/blob/master/src/ReactWrapper.js#L65

PR #229, opened 2016-03-03
Dom not found error to explain that jsdom needs to be initialised before using mount

ldabiralai:
This solves #223 

Not sure if we want to be linking to the docs from the code though?

What are peoples thoughts on this?

### DIFF
--- a/src/ReactWrapper.js
+++ b/src/ReactWrapper.js
@@ -61,8 +61,7 @@ export default class ReactWrapper {
   constructor(nodes, root, options = {}) {
     if (!global.window && !global.document) {
       throw new Error(
-        'It looks like you called `mount()` without a jsdom document being loaded. ' +
-        'Make sure to only use `mount()` inside of a `describeWithDOM(...)` call. '
+        'It looks like you called `mount()` without a global document being loaded.'
       );
     }
 


lelandrichardson:
I agree this message should be changed, but we probably don't need to mention jsdom. Mount will work fine if it has a global document, which could be provided by a browser or a library like jsdom.

Maybe the message should be:
```
\`mount()\` can only be used when a global `document` exists.
```

ljharb:
(due to a linter fix which only permits backticks when there's interpolation, this will need to be rebased)

src/ReactWrapper.js
```diff
@@ -61,8 +61,8 @@ export default class ReactWrapper {
   constructor(nodes, root, options = {}) {
     if (!global.window && !global.document) {
       throw new Error(
-        'It looks like you called `mount()` without a jsdom document being loaded. ' +
-        'Make sure to only use `mount()` inside of a `describeWithDOM(...)` call. '
+        'It looks like you called \`mount()\` without a jsdom document being loaded. ' +
+        'Make sure to only use \`mount()\` after initialising jsdom (see http://airbnb.io/enzyme/docs/guides/jsdom.html)'
```
ljharb: no need to escape the backticks in a single-quoted string

src/ReactWrapper.js
```diff
@@ -62,7 +62,7 @@ export default class ReactWrapper {
     if (!global.window && !global.document) {
       throw new Error(
         'It looks like you called `mount()` without a jsdom document being loaded. ' +
-        'Make sure to only use `mount()` inside of a `describeWithDOM(...)` call. '
+        'Make sure to only use `mount()` after initialising jsdom (see http://airbnb.io/enzyme/docs/guides/jsdom.html)'
```
ljharb: i agree with @lelandrichardson here - enzyme works in browsers too, so it's not about jsdom as it is about having a global `document`.

lelandrichardson:
@ldabiralai do you mind squashing your commits into one commit and rebasing?

ldabiralai:
Thanks for the feedback both, I've made the changes and squashed the commits into one.

lelandrichardson:
:+1:

PR #230, opened 2016-03-03
Found and fix a mount issue with find

CurtisHumphrey:
Found and fix a mount issue with find when using a named component e.g. find('Foo')

Added test case to highlight the issue

fixed someone else's lint issues too

### DIFF
--- a/src/MountedTraversal.js
+++ b/src/MountedTraversal.js
@@ -11,6 +11,7 @@ import {
   isCompoundSelector,
   AND,
   SELECTOR,
+  nodeHasType,
 } from './Utils';
 import {
   isDOMComponent,
@@ -62,8 +63,7 @@ export function instHasId(inst, id) {
 export function instHasType(inst, type) {
   switch (typeof type) {
     case 'string':
-      return isDOMComponent(inst) &&
-        inst.tagName.toUpperCase() === type.toUpperCase();
+      return nodeHasType(getNode(inst), type);
     case 'function':
       return isCompositeComponentWithType(inst, type);
     default:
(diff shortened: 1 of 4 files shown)


ljharb:
Thanks - could you squash these two commits?

CurtisHumphrey:
@ljharb Ok squashed

lelandrichardson:
@CurtisHumphrey thanks for finding this bug!  I'm not sure we should be using the `ShallowTraversal` helpers with `mount`, though.  Can you pull in `MountedTraversal` instead and use the `instHasType` helper function and see if that works?

CurtisHumphrey:
@lelandrichardson The instHasType helper in MountedTraversal is where the problem was. The same logic from ShallowTraversal's nodeHasType solved the problem and I was trying to keep it dry.  I did not find anything in MountedTraversal that did the same check

lelandrichardson:
@CurtisHumphrey we should be adding code to the `instHasType` helper, then. We should try not to mix the logic between MountedTraversal and ShallowTraversal.

CurtisHumphrey:
@lelandrichardson So you would prefer copy any paste? If so I can do that. Or I guess we could move the function into the utils file

CurtisHumphrey:
@lelandrichardson ok, I move the function into util, rebased and squashed

Aweary:
@CurtisHumphrey it looks like you might have added some Sublime project files to your commit.

CurtisHumphrey:
@aweary, remove them oops.

lelandrichardson:
Looks legit. Thanks!

PR #231, opened 2016-03-04
remove strange assignment

istarkov:
There is no need in this line `documentRef = document;`


lelandrichardson:
You're right. I think that was a botched copy/paste job. Thanks!

PR #232, opened 2016-03-04
Documentation, remove unused code

istarkov:
Remove unused 
```
const passes = [		
  <span>Hello</span>,		
  <div>Goodbye</div>,		
];
```


lelandrichardson:
Good catch.

PR #233, opened 2016-03-04
Documentation, Alphabetize API Reference function links in Table of Contents

staticinstance:
Makes it easier to scan for functions


lelandrichardson:
Seems reasonable. I was originally trying to group them semantically, but the order in that case is not always very clear. This may be more obvious.

@ljharb do you have any thoughts?

ljharb:
I tend to prefer semantically, because people will use find-in-page to find what they want. Alphabetically seems fine too though, and less subjective.

lencioni:
I generally prefer alphabetical because it is a simple rule that is easy for producers and consumers to follow.

ljharb:
@staticinstance would you mind squashing these 3 commits to one?

ljharb:
@staticinstance that added merge commits rather than being rebased down to a single commit :-/

staticinstance:
Thanks, I'll work on fixing that.

staticinstance:
It should be fixed now.

lelandrichardson:
Thanks, @staticinstance !

staticinstance:
Thanks for merging :+1:

PR #240, opened 2016-03-09
Add compatibility for React 15 RC1

iancmyers:
This adds compatibility for React 15 RC1. The most painful/icky change is in retrieving the internal instance, which has a random number appended to its key.

There's also a hack to work around a bug in React 15 where warnings on ref and key are throwing for nodes with string types.


src/MountedTraversal.js
```diff
@@ -124,7 +133,7 @@ export function childrenOfInstInternal(inst) {
     }
     return children;
   } else if (
-    REACT014 &&
+    (!REACT013) &&
```
ljharb: nbd, but i don't think we need parens around `!foo` here or elsewhere

src/MountedTraversal.js
```diff
@@ -20,10 +20,11 @@ import {
   isElement,
   findDOMNode,
 } from './react-compat';
-import { REACT013, REACT014 } from './version';
+import { REACT013 } from './version';
 
 export function internalInstance(inst) {
-  return inst._reactInternalInstance;
+  return inst._reactInternalInstance ||
+    inst[Object.keys(inst).find(key => key.match(/^__reactInternalInstance\$/))];
```
ljharb: `.find` is ES6 - you'll need `.indexOf` or `.filter` here (and throughout) to keep it ES5 - or else use http://npmjs.com/array.prototype.find :-)

test/ReactWrapper-spec.js
```diff
@@ -449,14 +449,14 @@ describeWithDOM('mount', () => {
 
         const selector = 'blah';
         const wrapper = mount(<SFC selector={selector} />);
-        const foundSpan = wrapper.findWhere(n => (
-          n.type() === 'span' && n.props()['data-foo'] === selector
-        ));
+        const foundSpan = wrapper.findWhere(n => {
+          return n.type() === 'span' && n.props()['data-foo'] === selector
```
ljharb: i think the styleguide/linter requires the paren form?

lencioni:
LGTM

src/Utils.js
```diff
@@ -17,6 +17,13 @@ export function propsOfNode(node) {
   if (node && node._reactInternalComponent && node._reactInternalComponent._currentElement) {
     return (node._reactInternalComponent._currentElement.props) || {};
   }
+  if (REACT15 && node) {
+    const instanceKey = Object.keys(node).find(key => key.match(/^__reactInternalInstance\$/));
```
lelandrichardson: since we're doing this in more than one place, it's probably best if we pull this out into a helper method that also makes it more semantically clear.

test/ReactWrapper-spec.js
```diff
@@ -1632,7 +1632,44 @@ describeWithDOM('mount', () => {
...
+          return (
+            <div className="in-bar">
+              <Foo />
+            </div>
+          );
+        }
+      }
+      const wrapper = mount(<Bar />);
+      expect(wrapper.html()).to.equal(
+        `<div data-reactroot="" class="in-bar"><div class="in-foo"></div></div>`
```
lelandrichardson: similar to how we used to remove the `data-reactid` attributes, we should maybe also remove the `data-reactroot` attribute as well. Makes the return value of this function more consistent actually, and can prevent the need for us to make these tests different per version

lelandrichardson:
wow, this is great!  Thanks for working on this.

iancmyers:
I've updated this PR to address all of the feedback.

install-relevant-react.sh
```diff
@@ -11,3 +11,7 @@ fi
 if [ "$REACT" = "0.14" ]; then
     npm run react:14
 fi
+
+if [ "$REACT" = "15" ]; then
+    npm run react:14
```
ljharb: `s/14/15`?

test/ReactWrapper-spec.js
```diff
@@ -7,7 +7,7 @@ import {
   ReactWrapper,
 } from '../src/';
 import sinon from 'sinon';
-import { REACT013 } from '../src/version';
+import { REACT013, REACT15 } from '../src/version';
```
ljharb: it looks like you brought in `itIf` and `REACT15` but aren't using them anywhere?

test/ReactWrapper-spec.js
```diff
@@ -1667,6 +1667,8 @@ describeWithDOM('mount', () => {
         `<div class="in-foo"></div>`
       );
     });
+
+
```
lelandrichardson: i think the linter gets mad at this whitespace as well

src/Utils.js
```diff
@@ -7,9 +7,18 @@ import {
 } from './react-compat';
 import {
   REACT013,
-  REACT014,
+  REACT15,
 } from './version';
 
+function internalInstanceKey(node) {
+  return Object.keys(node).filter(key => key.match(/^__reactInternalInstance\$/))[0];
```
ljharb: Per https://travis-ci.org/airbnb/enzyme/jobs/114854317 - pre-ES6, `Object.keys` throws on primitive values. You can either use the `object-keys` module, or do `Object.keys(Object(node))` (but that runs decently slower).

ljharb:
LGTM :+1:

src/Utils.js
```diff
@@ -7,16 +7,31 @@ import {
 } from './react-compat';
 import {
   REACT013,
-  REACT014,
+  REACT15,
 } from './version';
 
+function internalInstanceKey(node) {
+  return Object.keys(Object(node)).filter(key => key.match(/^__reactInternalInstance\$/))[0];
```
spicyj: Really?

cartoon drawing of a sad fox

ljharb: Accessible is public, Ben. ¯\\\_(ツ)_/¯

spicyj: Next time I'm making it nonenumerable and monkey-patching Object.getOwnPropertyNames (and Reflect.ownKeys I guess).

ljharb: @spicyj also every goPD and ownKeys in every realm that i could possibly create!

PR #243, opened 2016-03-09
Update jest.md for Jest 0.9

cpojer:
We finally fixed the npm3 module resolution and we do transitive unmocking of dependencies now.

"react-dom" and "react-addons-test-utils" should be unmocked too, however they aren' part of either react or enzyme so I guess we can remove them from this list too.

### DIFF
--- a/docs/guides/jest.md
+++ b/docs/guides/jest.md
@@ -1,7 +1,7 @@
 # Using Jest with Enzyme
 
 If you are using Jest with enzyme and using Jest's "automocking" feature, you will need to mark
-several modules to be unmocked in your `package.json`:
+react and enzyme to be unmocked in your `package.json`:
 
 ```js
 /* package.json */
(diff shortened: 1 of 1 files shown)


lencioni:
This looks okay to me. Is it worth mentioning that this applies to Jest 0.9+ and keeping some information around for folks using older versions?

cpojer:
I added a note for 0.9 and previously.

cpojer:
Should we suggest unmocking react-addons-test-utils by default?

lelandrichardson:
It seems like enzyme, react, react-dom, and react-addons-test-utils should be unmocked by default?

cpojer:
yes, but if enzyme is requiring them and enyzme itself is unmocked, they won't be mocked (that was a 0.9 fix together with npm3). They'll only be mocked if they are not unmocked by default and required directly. Does that explanation make sense?

lelandrichardson:
@cpojer I think so. I'll go ahead and merge this as is, then.

PR #247, opened 2016-03-10
Fixes version of React in Browserify guide

Danwhy:
The 'Using Enzyme with Browserify' guide specifies that you should make browserify ignore a different set of files if using React 0.13, but these are the files that should be ignored if using 0.14 onwards.


ljharb:
We still support v0.13, so while it's great to add a v0.14 section, we should make sure that the information is accurate for 0.13 as well.

lelandrichardson:
This is accurate. Thanks :+1:

PR #251, opened 2016-03-13
Prevent falsy nodes from being counted as children

nishitaniyuki:
Like [#151](https://github.com/airbnb/enzyme/issues/151)

I have a problem like below.

```js
test('component should have no children', t => {
    const component = shallow(
        <div>
            {undefined}
            {undefined}
        </div>
    );
    t.same(component.children().length, 0) // This should be success, but fail since length is 1.
});
```

I expect `null`, `undefined` and `false` should not be counted as a child node.
What do you think??

### DIFF
--- a/src/ShallowTraversal.js
+++ b/src/ShallowTraversal.js
@@ -19,7 +19,11 @@ export function childrenOfNode(node) {
   if (!node) return [];
   const maybeArray = propsOfNode(node).children;
   const result = [];
-  React.Children.forEach(maybeArray, child => result.push(child));
+  React.Children.forEach(maybeArray, child => {
+    if (child !== null && child !== false && typeof child !== 'undefined') {
+      result.push(child);
+    }
+  });
   return result;
 }
 
@@ -29,7 +33,7 @@ export function hasClassName(node, className) {
 }
 
 export function treeForEach(tree, fn) {
-  if (tree !== null && tree !== false) {
+  if (tree !== null && tree !== false && typeof tree !== 'undefined') {
     fn(tree);
   }
   childrenOfNode(tree).forEach(node => treeForEach(node, fn));
--- a/test/ShallowWrapper-spec.js
+++ b/test/ShallowWrapper-spec.js
@@ -1019,6 +1019,31 @@ describe('shallow', () => {
       expect(wrapper.children().length).to.equal(0);
     });
 
+    it('should skip the falsy children', () => {
+      const wrapper = shallow(
+        <div>
+          <div>
+            {false}
+            {[false, false]}
+            <p>foo</p>
+          </div>
+          <div>
+            {undefined}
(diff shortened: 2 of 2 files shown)


src/ShallowTraversal.js
```diff
@@ -18,8 +21,17 @@ import {
...
+      if (isArray(value)) {
+        return recursiveWithout(value);
+      }
+      return value;
+    });
+
+  const filtered = isArray(maybeArray) ? recursiveWithout(maybeArray) : maybeArray;
   const result = [];
-  React.Children.forEach(maybeArray, child => result.push(child));
+  React.Children.forEach(filtered, child => result.push(child));
```
lelandrichardson: a simpler solution here might be to just check and see if `child` is `false`, `null`, or `undefined` before pushing it into the result?  The recursive nature of children will be taken care of in `React.Children.forEach` in this case.

lelandrichardson:
Thanks for working on this. I believe we omit these nodes from being passed into methods such as `.findWher()`, so omitting them here feels more consistent.

ljharb:
We omit `null` nodes, but not usually all falsy nodes - what about the empty string and `NaN`?

lelandrichardson:
@ljharb i'm not too worried about `NaN`, but we should have defined behavior for all things that React considers non-renderable: `false`, `null`, and `undefined`.

We explicitly don't traverse `false` and `null`. It looks like we might (incorrectly) handle undefined though:

https://github.com/airbnb/enzyme/blob/master/src/ShallowTraversal.js#L32

nishitaniyuki:
> i'm not too worried about NaN, but we should have defined behavior for all things that React considers non-renderable: false, null, and undefined.
I agree.

> We explicitly don't traverse false and null. It looks like we might (incorrectly) handle undefined though: https://github.com/airbnb/enzyme/blob/master/src/ShallowTraversal.js#L32
I think this also should skip `undefined`

src/ShallowTraversal.js
nishitaniyuki: Exactly.
Can I rewrite & send PR again?? (I'm new to github ...)

lelandrichardson: @nishitaniyuki yep! Update your code and then push the changes to github. this PR should update automatically.

When you update your code, it's best for a change like this if you can make all of your changes into a single atomic commit. Let me know if you need any extra help with this.

lelandrichardson: FYI, you can change the code and then use `git commit --amend` to combine the changes into the latest commit so that it stays as a single commit.

nishitaniyuki: I've combined the changes, thank you!
Please check it.

lelandrichardson:
@nishitaniyuki great, thanks!

PR #255, opened 2016-03-15
Fix props and debug methods for SFCs

### LINKED ISSUE #134: hasClass does not work on root ReactWrapper

lelandrichardson:
to: @ljharb 
cc: @ryanalane

This is essentially a copy of #220 . (There's a line or two different and that PR currently has conflicts).

Apologies to @ryanalane for stealing his git thunder :/

Fixes #134 
Closes #220

### DIFF
--- a/src/Debug.js
+++ b/src/Debug.js
@@ -88,7 +88,7 @@ export function debugInst(inst, indentLength = 2) {
   const publicInst = inst.getPublicInstance();
 
   if (typeof publicInst === 'string' || typeof publicInst === 'number') return escape(publicInst);
-  if (!publicInst) return '';
+  if (!publicInst && !inst._renderedComponent) return '';
 
   // do stuff with publicInst
   const currentElement = inst._currentElement;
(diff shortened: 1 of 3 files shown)


ljharb:
LGTM

ryanalane:
@lelandrichardson No problem! Just glad the work was helpful. :)

PR #256, opened 2016-03-15
Fix context method for SFCs

### LINKED ISSUE #180: .context method not working with stateless components.
Hey guys, I'm trying to use the `.context` method from the **Full DOM rendering API** in a stateless component and I'm getting the following error:

```
  - Error: You cannot get an instance of a stateless component.
        at getInstance (node_modules/enzyme/build/ReactWrapperComponent.js:81:17)
        at ReactWrapper.instance (node_modules/enzyme/build/ReactWrapper.js:143:31)
        at ReactWrapper.<anonymous> (node_modules/enzyme/build/ReactWrapper.js:601:25)
        at ReactWrapper.single (node_modules/enzyme/build/ReactWrapper.js:1066:19)
        at ReactWrapper.context (node_modules/enzyme/build/ReactWrapper.js:600:29)
        at Spec.<anonymous> (src/shared/components/fieldset/__tests__/fields.js:155:27)

```

I copied the code from `enzyme` tests, and updated it to support context:

```js
const Foo = ({ foo }, {_}) => (
  <div>
    <div className="bar">bar</div>
    <div className="qoo">{foo}</div>
  </div>
);

Foo.contextTypes = {
  _: React.PropTypes.func
};

const wrapper = mount(<Foo foo="qux" />, {
  context: {
    _: (e) => e
  }
});
console.log(wrapper.context('_'));
```

Basically I'm trying to pass a context for my stateless component, if I don't run `wrapper.context('_')` than no errors are thrown. But when I use it then the error occurs.

I don't know if that's the expected behaviour, or if that's happening with othre methods as well.

Cheers

lelandrichardson:
Fixes #180

### DIFF
--- a/src/ReactWrapper.js
+++ b/src/ReactWrapper.js
@@ -123,6 +123,9 @@ export default class ReactWrapper {
    * @returns {ReactComponent}
    */
   instance() {
+    if (this.root !== this) {
+      throw new Error('ReactWrapper::instance() can only be called on the root');
+    }
     return this.component.getInstance();
   }
 
--- a/src/ReactWrapperComponent.jsx
+++ b/src/ReactWrapperComponent.jsx
@@ -45,7 +45,7 @@ export default function createWrapperComponent(node, options = {}) {
       const component = this._reactInternalInstance._renderedComponent;
       const inst = component.getPublicInstance();
       if (inst === null) {
-        throw new Error('You cannot get an instance of a stateless component.');
+        return component._instance;
       }
       return inst;
     },
--- a/test/ReactWrapper-spec.js
+++ b/test/ReactWrapper-spec.js
@@ -75,6 +75,28 @@ describeWithDOM('mount', () => {
       expect(wrapper.context().name).to.equal(context.name);
       expect(wrapper.context('name')).to.equal(context.name);
     });
+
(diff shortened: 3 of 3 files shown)


src/ReactWrapper.js
```diff
@@ -123,6 +123,9 @@ export default class ReactWrapper {
    * @returns {ReactComponent}
    */
   instance() {
+    if (this.root !== this) {
+      throw new Error('ReactWrapper::instance() can only be called on the root');
```
ljharb: is there any test for this throw?

src/ReactWrapperComponent.jsx
```diff
@@ -45,7 +45,7 @@ export default function createWrapperComponent(node, options = {}) {
       const component = this._reactInternalInstance._renderedComponent;
       const inst = component.getPublicInstance();
       if (inst === null) {
-        throw new Error('You cannot get an instance of a stateless component.');
+        return component._instance;
```
ljharb: what exactly *is* `component._instance` in the case of a stateless component?

src/ReactWrapper.js
lelandrichardson: not explicitly...  though it would have thrown before (in a different, less informative way) anyway. This method on anything but the root doesn't make a whole lot of sense

PR #257, opened 2016-03-15
Fix children() for stateless children

poluyanov:
Children method did not work with nested stateless component. It always returns [null]
This patch fixes it


lelandrichardson:
Awesome. Thanks!

I'm getting the feeling that we *really* need to beef up our test suite for stateless functional components. All kinds of SFC-related bugs coming through :(

poluyanov:
@lelandrichardson thank you for enzyme!

PR #269, opened 2016-03-18
Adding Pinterest to organizations using Enzyme

zackargyle:
We love Enzyme here at Pinterest. Adding our name to the list of organizations! Nice work!


ljharb:
Thanks!

PR #271, opened 2016-03-19
Move React 15 support tests from RC1 to RC2

iancmyers:
This changes the React 15 support scripts from using RC1 to using RC2. The docs have been updated as well.


ljharb:
:+1:

PR #272, opened 2016-03-19
Make enzyme compatible with all React 15 Release Candidates

nikgraf:
Did a similar change here: https://github.com/facebook/draft-js/pull/179


lelandrichardson:
:+1:

package.json
```diff
@@ -80,6 +80,6 @@
     "sinon": "^1.15.4"
   },
   "peerDependencies": {
-    "react": "0.13.x || 0.14.x || 15.* || 15.0.0-rc.2"
+    "react": "0.13.x || 0.14.x || 15.* || ^15.0.0-rc"
```
ljharb: semver ranges don't actually match against prerelease versions, so this is not a good change.

ljharb: hm, nevermind - it seems this does work. sorry about that

nikgraf:
@lelandrichardson do you plan to publish a new release after this change? I would happy to avoid these warning …

screen shot 2016-03-20 at 13 52 53

lelandrichardson:
@nikgraf yep!  sorry, was going to do it yesterday but got distracted. I'll do it before the end of the day :+1:

nikgraf:
cool, thx @lelandrichardson

PR #274, opened 2016-03-20
Fix issue finding nested SFCs by constructor in mount

### LINKED ISSUE #262: wrapper.find() cannot retrieve stateless components
I'm trying to find rendered stateless components in a wrapper using `name`/`displayName` as selector (when using `mount`). For normal components `find` always returns selected components as expected, but for stateless components it returns nothing, even though they are there.

In [`findAllInRenderedTreeInternal`](https://github.com/airbnb/enzyme/blob/7efeda383c3c914f038ec0005ca493ed591b9d47/src/MountedTraversal.js#L236-L289) at [line 249](https://github.com/airbnb/enzyme/blob/7efeda383c3c914f038ec0005ca493ed591b9d47/src/MountedTraversal.js#L249) there is this a call

```
var publicInst = inst.getPublicInstance();
```

which calls `getPublicInstance` of `ReactCompositeComponent.js`:

```
  getPublicInstance: function () {
    var inst = this._instance;
    if (inst instanceof StatelessComponent) {
      return null;
    }
    return inst;
  },
```

Obviously, `find` will fail for stateless components since `getPublicInstance` always returns `null`. I'm wondering why in line 249 further up it is necessary to call `getPublicInstance` since there should be other ways to retrieve the element for testing.

I'm fairly new to `React` and `enzyme` so I am not sure if I am missing something.

iancmyers:
This fixes an issue where calling `.find` with an SFC constructor/function would fail to return instances of the the SFC in the tree.

This seems like a round-about way of getting at this. Is there a better way?

Fixes #262

### DIFF
--- a/src/MountedTraversal.js
+++ b/src/MountedTraversal.js
@@ -5,6 +5,7 @@ import {
   coercePropValue,
   nodeEqual,
   propsOfNode,
+  isFunctionalComponent,
   isSimpleSelector,
   splitSelector,
   selectorError,
@@ -60,12 +61,17 @@ export function instHasId(inst, id) {
   return instId === id;
 }
 
+function isFunctionalComponentWithType(inst, func) {
+  return isFunctionalComponent(inst) && getNode(inst).type === func;
+}
+
 export function instHasType(inst, type) {
   switch (typeof type) {
     case 'string':
       return nodeHasType(getNode(inst), type);
     case 'function':
-      return isCompositeComponentWithType(inst, type);
+      return isCompositeComponentWithType(inst, type) ||
+        isFunctionalComponentWithType(inst, type);
     default:
       return false;
   }
@@ -249,8 +255,7 @@ function findAllInRenderedTreeInternal(inst, test) {
     const internal = internalInstance(inst);
     return findAllInRenderedTreeInternal(internal, test);
   }
-
-  const publicInst = inst.getPublicInstance();
+  const publicInst = inst.getPublicInstance() || inst._instance;
   let ret = test(publicInst) ? [publicInst] : [];
   const currentElement = inst._currentElement;
   if (isDOMComponent(publicInst)) {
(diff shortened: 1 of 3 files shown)


src/Utils.js
```diff
@@ -19,6 +19,10 @@ export function internalInstance(inst) {
     inst[internalInstanceKey(inst)];
 }
 
+export function isFunctionalComponent(inst) {
+  return inst.constructor.name === 'StatelessComponent';
```
ljharb: will `inst.constructor` always exist?

iancmyers:
I've updated this with a different (correct-er) solution and another test.

src/Utils.js
iancmyers: Provided what is passed to the function is an instance, yes, I think it will have a constructor.

src/MountedTraversal.js
```diff
@@ -249,8 +255,7 @@ function findAllInRenderedTreeInternal(inst, test) {
     const internal = internalInstance(inst);
     return findAllInRenderedTreeInternal(internal, test);
   }
-
-  const publicInst = inst.getPublicInstance();
+  const publicInst = inst.getPublicInstance() || inst._instance;
```
ljharb: should the `_instance` stuff be moved into `getPublicInstance`?

iancmyers: I believe `.getPublicInstance` is a private method in React itself, which always returns `null` for SFCs, thus we fall back to `._instance`.

ljharb: ah ok, if it's a react method then we certainly can't fix that :-) LGTM

src/Utils.js
lelandrichardson: let's maybe fence this with `inst && inst.constructor &&` anyway? I intend on using this in several places, and it seems possible for us to pass null in at some point...

iancmyers: @lelandrichardson @ljharb I've constructed a fence!

lelandrichardson:
@iancmyers think you can rebase?

PR #275, opened 2016-03-20
Add options to ShallowWrapper#shallow

ncuillery:
I have just enhanced the `ShallowWrapper#shallow` to support the same `options` arg as the top level `shallow` method.

It is useful when your component is wrapped by an high-order component like the ones provided in popular Flux implementations : 
- http://fluxible.io/addons/connectToStores.html
- https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options

For example :
```javascript
class MyComponent {
  [...]
}
MyComponent.contextTypes = {
  foo: React.PropTypes.string
}

export default connectToStore(MyComponent, [], () => {
  [...]
})
```

Then you can easily shallow render `MyComponent`:
```javascript
const context = {foo: 'bar'};
const wrapper = shallow(MyComponent).shallow({ context })
```

I added the same context tests of `shallow` method to `ShallowWrapper#shallow`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -49,7 +49,7 @@
     * [setContext(context)](/docs/api/ShallowWrapper/setContext.md)
     * [setProps(nextProps)](/docs/api/ShallowWrapper/setProps.md)
     * [setState(nextState)](/docs/api/ShallowWrapper/setState.md)
-    * [shallow()](/docs/api/ShallowWrapper/shallow.md)
+    * [shallow([options])](/docs/api/ShallowWrapper/shallow.md)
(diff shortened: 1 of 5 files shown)


lelandrichardson:
@ncuillery seems pretty reasonable. Looks like your branch is out of date though. You mind rebasing and fixing any conflicts?

Thanks.

ncuillery:
Done !

ncuillery:
Lint errors due to recent https://github.com/airbnb/enzyme/issues/276

I'll fix that

PR #279, opened 2016-03-23
Add second() selector

HeeL:
Along with `.first()` and `.last()`, this PR adds a functionality for `.second()` for reducing matched nodes. 

Don't get me wrong, if someone will do a PR with `.third()` or `.before_last()` - I will be totally against this kind of methods. In my mind, `.second()` is nice to have method for many people.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -92,6 +92,7 @@
     * [reduceRight(fn[, initialValue])](/docs/api/ReactWrapper/reduceRight.md)
     * [ref(refName)](/docs/api/ReactWrapper/ref.md)
     * [render()](/docs/api/ReactWrapper/render.md)
+    * [second()](/docs/api/ReactWrapper/second.md)
     * [setContext(context)](/docs/api/ReactWrapper/setContext.md)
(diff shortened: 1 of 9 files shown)


HeeL:
You guys have a linting error. Idk if I have to fix it with this PR  or should it be a separated one: https://travis-ci.org/airbnb/enzyme/jobs/118067424#L2112

lelandrichardson:
I'm not really convinced of the value of this method?  is `.at(1)` really that hard to type or read?

HeeL:
@lelandrichardson, the value is pretty much the same as from `.first` method. The only difference is that `.first` is used more often

ljharb:
@HeeL that's a pretty big difference. We don't need to go [this far](http://apidock.com/rails/Array/forty_two) :-)

HeeL:
@ljharb , `.forty_two` is more like a joke in rails and there is no methods between `.fifth` and `.forty_two`. 

Ok, its up to you, feel free to close this PR if it seems so useless to you

ljharb:
Thanks for the contribution! I think for now let's strive to keep the API minimal.

PR #296, opened 2016-04-06
Fix pathToNode with a sibling node which has children

koba04:
Currently, the following test is failed.

```js
const wrapper = shallow(
  <div className="bax">
    <div className="foo">
      <div className="bar">
        <div className="baz" />
        <div>
          <div />
        </div>
      </div>
    </div>
  </div>
);

expect(wrapper.find('.baz').parent().hasClass('bar')).to.equal(true);
```

This is because `pathToNode` returns trees which are including the sibling node.
This PR makes `pathToNode` return only nodes including the current node.

Thanks

### DIFF
--- a/src/ShallowTraversal.js
+++ b/src/ShallowTraversal.js
@@ -49,15 +49,20 @@ export function treeFilter(tree, fn) {
   return results;
 }
 
+function pathFilter(path, fn) {
+  return path.filter(tree => treeFilter(tree, fn).length !== 0);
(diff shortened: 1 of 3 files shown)


lelandrichardson:
Thanks for helping out. This is a great fix.

Do you mind rebasing before I merge it in? Thanks!

koba04:
@lelandrichardson Thank you! rebased.

PR #299, opened 2016-04-07
Add tap() into ShallowWrapper and ReactWrapper

koba04:
This is similar to underscore's `tap()`.

http://underscorejs.org/#tap

## motivation

When I'm testing complex React Element tree using method chains, I sometimes would like to inject a debug code between the method chains.

```js
wrapper
.find('.foo')
.findWhere(n => n.text() === 'bar')
// ???
.at(0)
```

* Using `tap()`

```js
wrapper
.find('.foo')
.findWhere(n => n.text() === 'bar')
.tap(n => console.log(n.debug())
.at(0)
```

* without `tap()`

```js
wrapper
.find('.foo')
.findWhere(n => n.text() === 'bar')
.map(n => {
  console.log(n.debug());
  return n;
})
.at(0)
```

I think this is not an essential feature, but it would be nice if enzyme has it.
How do you think about this? :+1:? :-1:?

Thanks.

### DIFF
--- a/test/ReactWrapper-spec.js
+++ b/test/ReactWrapper-spec.js
@@ -1764,6 +1764,22 @@ describeWithDOM('mount', () => {
     });
   });
 
+  describe('.tap()', () => {
+    it('should call the passed function with current ShallowWrapper and returns itself', () => {
+      const spy = sinon.spy();
+      const wrapper = mount(
+        <ul>
+          <li>xxx</li>
+          <li>yyy</li>
+          <li>zzz</li>
+        </ul>
+      ).find('li');
+      const result = wrapper.tap(spy);
+      expect(spy.calledWith(wrapper)).to.equal(true);
+      expect(result).to.equal(wrapper);
+    });
+  });
+
   describe('attachTo option', () => {
     it('should attach and stuff', () => {
       class Foo extends React.Component {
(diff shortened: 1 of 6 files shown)


test/ReactWrapper-spec.js
```diff
@@ -1761,6 +1761,24 @@ describeWithDOM('mount', () => {
...
+      const spy = sinon.spy();
+      const wrapper = mount(
+        <ul>
+          <li>xxx</li>
+          <li>yyy</li>
+          <li>zzz</li>
+        </ul>
+      );
+      const result = wrapper.find('li').tap(spy);
+      expect(spy.calledOnce).to.equal(true);
```
ljharb: `expect(spy.calledWith(wrapper)).to.equal(true)` might be a stronger assertion?

ljharb:
I think this is a pretty useful feature to be sure.

test/ReactWrapper-spec.js
koba04: Thanks! I fixed it.

lelandrichardson:
I like this as well. Never knew about this w/ underscore.

@koba04 do you mind rebasing before we merge in?  Thanks.

koba04:
@lelandrichardson Thank you! rebased.

Issue #302, opened 2016-04-08
Issue after upgrading to React 15


toddw:
After upgrading to React 15, I started getting the following error with enzyme.

```
ERROR in ./~/enzyme/build/react-compat.js
Module not found: Error: Cannot resolve module 'react/addons' in /projects/test/node_modules/enzyme/build
 @ ./~/enzyme/build/react-compat.js 37:16-39
```

toddw:
Testing using 77e6bdc17ad52aefd9ccdc82d4197bd0709c95aa and I do no longer have the issue… Patiently waiting for new publish :smile:

ljharb:
hmm - that's only ran in the `REACT013` code path - that commit shouldn't make a difference. On which SHA can you reproduce the issue?

toddw:
Well then I wonder if pulling from git and telling it to build myself resolved it. I had the issue using 2.2.0, then I switched it from 2.2.0 to the latest commit, cd'd into the directory `npm install` then rand `npm run build` myself and things worked. I'll try that with 2.2.0, now I'm not sure if it is the version or some weird npm thing

toddw:
Pulling from the v2.2.0 tag directly, `npm install && npm run build` in that folder also works. Let me try 2.2.0 again via npm instead of github

toddw:
Going back to the `npm install enzyme` version and the problem is back.

toddw:
after installing enzyme 2.2.0 I get the error again but if I cd into `node_modules/enzyme` and run `npm install && npm run build` then it resolves it. I'll see if rolling back to an older version resolves it, I'm starting to think this has less to do with the latest version of React but actually from updating enzyme

toddw:
Rolled my react version to 0.14.8 and it works again.

The issue is only happening when using React version 15.0.0
It is resolved if I cd into `node_modules/enzyme` and run `npm install && npm run build`. Maybe there is a flag in the publish script causing an issue?

ljharb:
after upgrading `React`, can you run `npm install enzyme` in your project dir?

toddw:
When I do that, the error message comes back. Only resolved if I manually cd into the folder and build it myself

ljharb:
that's very strange. which version of enzyme?

toddw:
2.2.0 and I also tried 2.0.0

toddw:
not a problem if I'm using 0.14.8

mordrax:
[...]

PR #304, opened 2016-04-08
Documentation about react 15 and webpack usage

vslinko:
ref https://github.com/airbnb/enzyme/issues/302


docs/guides/webpack.md
```diff
@@ -15,6 +15,7 @@ labeled as "external", which means they will be ignored:
```
tarrencev: Sentence above should be updated to read

> ... the following four files are ...

ljharb: we should probably just say "the following files" to avoid having to update the number every time.

vslinko: done

PR #311, opened 2016-04-11
Remove babelrc from npm package.

cpojer:
I have a case where I need to compile node_modules with babel but because enzyme ships with its own babelrc, it breaks my babel setup. I don't think there is a reason why the babelrc needs to be part of the package, so let's remove it, maybe? :)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -16,6 +16,9 @@ coverage
 # Grunt intermediate storage (http://gruntjs.com/creating-plugins#storing-task-files)
 .grunt
 
+# babel configuration
+.babelrc
+
 # node-waf configuration
 .lock-wscript
 


ljharb:
I prefer the babel configuration to be included, so those who wish to use the original source can have access to the transpilation settings for it.

Are you using Webpack? I believe it's a webpack bug that babelrc files inside node_modules (which tools shouldn't be reaching inside to transpile) break things.

I'd consider moving the config to package.json, but not omitting it from the package entirely, especially to work around a tooling bug.

lelandrichardson:
@ljharb This is a common problem with the react-native ecosystem. The recommended "fix" is to not include .bablerc in the npm package, though I think that's a pretty backwards fix also.

@cpojer can you tell us more about your use case? :)

cpojer:
@lelandrichardson the use case in this example was on react-native (trying to make some stuff work ;) ). Many people there ship npm packages uncompiled – so we have to compile everything, even node_modules. enzyme was tripping up because the preset wasn't installed locally. I figured I'll send a PR instead of just creating an issue, because that makes it easier for you :)

I was able to work around it by blacklisting enzyme from the preprocessor, but still think you probably don't need to ship the .babelrc as is.

lelandrichardson:
We don't need to ship babelrc with the package, but it does seem like a pretty backwards fix. React Native is pretty backwards in this regard... compiling node_modules is just something a bad idea IMO. We can fix it for enzyme, but it's just "fixed" until the next package you depend on ends up publishing a babelrc that doesn't work for you.

I actually have a node_modules script that recurses through my node_modules dir and removes all of the babelrc files inside node_modules, and just run that whenever i run into this issue.... is it a good idea? Not really... but probably not any worse than having every library remove it :)

cpojer:
Yes, I agree that this should ideally be fixed in react-native and I don't quite understand why they chose to advice this to plugin developers. Nothing I can do to fix this immediately, however, so this is what we'll have to live with for a bit :)

SimenB:
Some discussion here: facebook/react-native#4062

lelandrichardson:
@ljharb while I don't like that this is somthing library authors are expected to do, I think we should go ahead and merge this. People using enzyme with react native is a significant use-case we want to allow, and if people want to compile their react-native tests with the packager, this will hit them.

PR #317, opened 2016-04-12
Add `dblclick` to properly map the native event.

### LINKED ISSUE #310: Documentation for simulate is minimal and dblclick event is named doubleclick
I tried to simulate a double click like this:
```javascript
component.find('[id="purse"]').parent().simulate('dblclick');
```
Because the documentation for [simulate](http://airbnb.io/enzyme/docs/api/ShallowWrapper/simulate.html) seems to suggest dom events are used.
I'm pretty sure (although could be wrong) that the dom event for double click is [dblclick](https://developer.mozilla.org/en-US/docs/Web/Events/dblclick).

Then I looked through the repo and found mentions of `doubleclick`, and doing this:
```javascript
component.find('[id="purse"]').parent().simulate('doubleclick');
```
works.

So it would be good to have a list of simulated events other than click because that's in every tut and example but there's no documentation about the list of events that can be simulated.
Does [this](https://github.com/airbnb/enzyme/blob/699ec7e39560a68c198ecf80b59d177d003fc869/src/Utils.js#L230) have something to do with it?

ljharb:
Fixes #310.

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -227,7 +227,8 @@ export function mapNativeEventNames(event) {
     keyup: 'keyUp',
     keypress: 'keyPress',
     contextmenu: 'contextMenu',
-    doubleclick: 'doubleClick',
+    dblclick: 'doubleClick',
+    doubleclick: 'doubleClick', // kept for legacy. TODO: remove with next major.
     dragend: 'dragEnd',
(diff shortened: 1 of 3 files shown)


lelandrichardson:
👍

Issue #34, opened 2015-11-20
`.setProps()` doesn't trigger componentDidUpdate lifecycle method


yzimet:
When using `shallow`, the `.setProps()` method doesn't trigger the componentDidUpdate lifecycle method. However, `.setState()` does trigger it. It seems like they both should (or both shouldn't).
##### simple example

``` jsx
describe('MyComponent', () => {
  let spy;

  before(function() {
    spy = sinon.spy(MyComponent.prototype, 'componentDidUpdate');
  });

  after(function() {
    spy.restore();
  });

  // fails
  it('calls componentDidUpdate when setting props', () => {
    const wrapper = shallow(<MyComponent color="yellow" />);
    wrapper.setProps({ 'color': 'green' });
    expect(spy.calledOnce).to.be.true;
  });

  // passes
  it('calls componentDidUpdate when setting state', () => {
    const wrapper = shallow(<MyComponent color="yellow" />);
    wrapper.setState({ foo: 'bar' });
    expect(spy.calledOnce).to.be.true;
  });
});
```

blainekasten:
FWIW, `setProps` is being deprecated. 

https://github.com/facebook/react/pull/5588/files
http://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#new-deprecations-introduced-with-a-warning

lelandrichardson:
@blainekasten thanks for that info. Funny - I didn't even know `.setProps()` existed as a real API.

The deprecation doesn't affect us in this case, since `.setProps()` is achieved using `setState()` of a parent wrapper component, anyway.

CurtisHumphrey:
At the moment I'm doing 
``` javascript
const prevProp = wrapper.props()
wrapper.setProps({
  current_step: 2,
})
// forces an update call to componentDidUpdate - https://github.com/airbnb/enzyme/issues/34
wrapper.instance().componentDidUpdate(prevProp)
```

anorudes:
+1, how correctly to solve the problem?)

Aweary:
@yzimet does this occur if you use `mount` as well, or is it unique to `shallow`?

erikthedeveloper:
I'm also performing similar hacks to what @CurtisHumphrey mentioned above.

I wonder what the preferred way of achieving this is, if not with `setProps`? I have yet to dig in too deep myself, but this is something I am definitely interested in contributing to. It looks like https://github.com/airbnb/enzyme/pull/318 is on a good track.

[...]

PR #318, opened 2016-04-12
Support all Component Lifecycle methods in ShallowRenderer

koba04:
This is a feature request which is supporting all React Component Lifecycle methods.
This is including some breaking changes.

This implementation is working on progress, and no (tests is passed)

Does this make sense?
If you are interested in this PR, please give me your feedbacks.
I'd like to work on it 😄 

Thanks.

### DIFF
--- a/test/ShallowWrapper-spec.js
+++ b/test/ShallowWrapper-spec.js
@@ -2634,7 +2634,7 @@ describe('shallow', () => {
         expect(result.state('count')).to.equal(1);
       });
 
-      it('should provoke an another render to call setState in componentWillUpdate twice', () => {
+      it('should provoke an another render to call setState twice in componentWillUpdate', () => {
         const spy = sinon.spy();
         class Foo extends React.Component {
           constructor(props) {
@@ -2663,7 +2663,7 @@ describe('shallow', () => {
         expect(result.state('count')).to.equal(1);
       });
 
-      it('should provoke an another render to call setState in componentDidUpdate twice', () => {
+      it('should provoke an another render to call setState twice in componentDidUpdate', () => {
         const spy = sinon.spy();
         class Foo extends React.Component {
           constructor(props) {
@@ -2792,7 +2792,7 @@ describe('shallow', () => {
         expect(spy.args).to.deep.equal([['render'], ['shouldComponentUpdate']]);
       });
 
-      it('should provoke an another render to call setState in componentWillUpdate twice', () => {
+      it('should provoke an another render to call setState twice in componentWillUpdate', () => {
         const spy = sinon.spy();
         class Foo extends React.Component {
           constructor(props) {
@@ -2822,7 +2822,7 @@ describe('shallow', () => {
         expect(result.state('count')).to.equal(1);
       });
 
-      it('should provoke an another render to call setState in componentDidUpdate twice', () => {
+      it('should provoke an another render to call setState twice in componentDidUpdate', () => {
         const spy = sinon.spy();
         class Foo extends React.Component {
           constructor(props) {
@@ -2953,7 +2953,7 @@ describe('shallow', () => {
         expect(spy.args).to.deep.equal([['render'], ['shouldComponentUpdate']]);
       });
 
-      it('should provoke an another render to call setState in componentWillUpdate twice', () => {
+      it('should provoke an another render to call setState twice in componentWillUpdate', () => {
         const spy = sinon.spy();
         class Foo extends React.Component {
           constructor(props) {
@@ -2988,7 +2988,7 @@ describe('shallow', () => {
         expect(result.state('count')).to.equal(1);
       });
 
-      it('should provoke an another render to call setState in componentDidUpdate twice', () => {
+      it('should provoke an another render to call setState twice in componentDidUpdate', () => {
         const spy = sinon.spy();
         class Foo extends React.Component {
           constructor(props) {
--- a/test/ShallowWrapper-spec.js
+++ b/test/ShallowWrapper-spec.js
@@ -2597,6 +2597,7 @@ describe('shallow', () => {
         expect(
           spy.calledWith({ foo: 'bar' }, { foo: 'baz' }, { foo: 'state' }, { foo: 'context' })
         ).to.equal(true);
+        expect(spy.callCount).to.equal(1);
       });
 
       it('should call componentWillUpdate', () => {
@@ -2808,6 +2809,7 @@ describe('shallow', () => {
         expect(
           spy.calledWith({ foo: 'props' }, { foo: 'bar' }, { foo: 'baz' }, { foo: 'context' })
         ).to.equal(true);
+        expect(spy.callCount).to.equal(1);
       });
 
       it('should call componentWillUpdate', () => {
@@ -2992,6 +2994,7 @@ describe('shallow', () => {
         );
         wrapper.setContext({ foo: 'baz' });
         expect(spy.calledWith({ foo: 'bar' }, { foo: 'baz' })).to.equal(true);
+        expect(spy.callCount).to.equal(1);
       });
 
       it('should call componentWillUpdate', () => {
--- a/test/ShallowWrapper-spec.js
+++ b/test/ShallowWrapper-spec.js
@@ -2486,65 +2486,138 @@ describe('shallow', () => {
     });
 
     context('updating props', () => {
+
       it('should call componentWillReceiveProps', () => {
         const spy = sinon.spy();
+
         class Foo extends React.Component {
-          componentWillReceiveProps(nextProps) {
-            spy(this.props, nextProps);
+          componentWillReceiveProps(nextProps, nextContext) {
+            spy(this.props, nextProps, nextContext);
           }
           render() {
             return <div>foo</div>;
           }
         }
+        Foo.contextTypes = {
+          foo: React.PropTypes.string,
+        };
+
+        const wrapper = shallow(<Foo foo="bar" />, { context: { foo: 'context' } });
+        wrapper.setProps({ foo: 'baz' });
+        expect(
+          spy.calledWith({ foo: 'bar' }, { foo: 'baz' }, { foo: 'context' })
+        ).to.equal(true);
+      });
+
+      it('is able to call setState in componentWillReceiveProps', () => {
+
+        class Foo extends React.Component {
+          constructor(...args) {
+            super(...args);
+            this.state = {
+              foo: 'bar',
+            };
+          }
+          componentWillReceiveProps() {
+            this.setState({ foo: 'baz' });
+          }
+          render() {
+            return <div>foo</div>;
+          }
+        }
+
         const wrapper = shallow(<Foo foo="bar" />);
         wrapper.setProps({ foo: 'baz' });
-        expect(spy.calledWith({ foo: 'bar' }, { foo: 'baz' })).to.equal(true);
+        expect(wrapper.state('foo')).to.equal('baz');
       });
+
       it('should call shouldComponentUpdate', () => {
         const spy = sinon.spy();
+
         class Foo extends React.Component {
-          shouldComponentUpdate(nextProps) {
-            spy(this.props, nextProps);
+          constructor(...args) {
+            super(...args);
+            this.state = {
+              foo: 'state',
+            };
+          }
+          shouldComponentUpdate(nextProps, nextState, nextContext) {
+            spy(this.props, nextProps, nextState, nextContext);
             return true;
           }
           render() {
             return <div>foo</div>;
           }
         }
-        const wrapper = shallow(<Foo foo="bar" />);
+        Foo.contextTypes = {
+          foo: React.PropTypes.string,
+        };
+
+        const wrapper = shallow(<Foo foo="bar" />, { context: { foo: 'context' } });
         wrapper.setProps({ foo: 'baz' });
-        expect(spy.calledWith({ foo: 'bar' }, { foo: 'baz' })).to.equal(true);
+        expect(
+          spy.calledWith({ foo: 'bar' }, { foo: 'baz' }, { foo: 'state' }, { foo: 'context' })
+        ).to.equal(true);
       });
+
       it('should call componentWillUpdate', () => {
         const spy = sinon.spy();
+
         class Foo extends React.Component {
-          componentWillUpdate(nextProps) {
-            spy(this.props, nextProps);
+          constructor(...args) {
+            super(...args);
+            this.state = {
+              foo: 'state',
+            };
+          }
+          componentWillUpdate(nextProps, nextState, nextContext) {
+            spy(this.props, nextProps, nextState, nextContext);
           }
           render() {
             return <div>foo</div>;
           }
         }
-        const wrapper = shallow(<Foo foo="bar" />);
+        Foo.contextTypes = {
+          foo: React.PropTypes.string,
+        };
+
+        const wrapper = shallow(<Foo foo="bar" />, { context: { foo: 'context' } });
         wrapper.setProps({ foo: 'baz' });
-        expect(spy.calledWith({ foo: 'bar' }, { foo: 'baz' })).to.equal(true);
+        expect(
+          spy.calledWith({ foo: 'bar' }, { foo: 'baz' }, { foo: 'state' }, { foo: 'context' })
+        ).to.equal(true);
       });
+
       it('should call componentDidUpdate', () => {
         const spy = sinon.spy();
         class Foo extends React.Component {
-          componentDidUpdate(prevProps) {
-            spy(prevProps, this.props);
+          constructor(...args) {
+            super(...args);
+            this.state = {
+              foo: 'state',
+            };
+          }
+          componentDidUpdate(prevProps, prevState, prevContext) {
+            spy(prevProps, this.props, prevState, prevContext);
           }
           render() {
             return <div>foo</div>;
           }
         }
-        const wrapper = shallow(<Foo foo="bar" />);
+        Foo.contextTypes = {
+          foo: React.PropTypes.string,
+        };
+
+        const wrapper = shallow(<Foo foo="bar" />, { context: { foo: 'context' } });
         wrapper.setProps({ foo: 'baz' });
-        expect(spy.calledWith({ foo: 'bar' }, { foo: 'baz' })).to.equal(true);
+        expect(
+          spy.calledWith({ foo: 'bar' }, { foo: 'baz' }, { foo: 'state' }, { foo: 'context' })
+        ).to.equal(true);
       });
+
       it('should cancel rendering when Component returns false in shouldComponentUpdate', () => {
         const spy = sinon.spy();
+
         class Foo extends React.Component {
           shouldComponentUpdate() {
             return false;
@@ -2559,6 +2632,7 @@ describe('shallow', () => {
             return <div>foo</div>;
           }
         }
+
         const wrapper = shallow(<Foo foo="bar" />);
         wrapper.setProps({ foo: 'baz' });
         expect(spy.callCount).to.equal(0);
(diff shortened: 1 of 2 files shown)


lelandrichardson:
@koba04 This is really cool work. I had briefly considered something similar to this a while ago but never pursued it.  I like this idea in general, but let me voice a couple of things concerning me:

1. refs need to work correctly, and I don't think they will out of the box. We would probably have to write some code to handle this as well.
2. context will also have to work, which it looks like is not included in this PR yet.
3. does this work how you'd expect when `this.setState` is called from with inside the lifecycle methods? I didn't see any tests covering this yet
4. we have to ensure that the lifecycle methods are called properly, in the proper order, and with the proper level of asynchrony.  Is `componentDidMount` called in the same tick that `componentWillMount` is called in? I *think* it is, but we should make sure. 

This PR would be a big breaking change and would push us to 3.0. As a result, I might choose to sit on it for some amount of time, as we had a couple of other things planned for 3.0. This is great work though and would be a welcome addition.

koba04:
Thank you for your great feedbacks :smile:

> refs need to work correctly, and I don't think they will out of the box. We would probably have to write some code to handle this as well.

yeah, refs doesn't work in this implementation because Shallow Rendering doesn't support it yet.
I think supporting refs is the other issue. But I'm looking for the way refs works well.

> context will also have to work, which it looks like is not included in this PR yet.

right. I'll add tests for context and try to work it well if it doesn't work.

> does this work how you'd expect when this.setState is called from with inside the lifecycle methods? I didn't see any tests covering this yet

I'll add tests for when setState is called in lifecycle methods. I guess it works well except batchingUpdate feathers.
In lifecycle methods, setState calls are batched.
I guess it might be able to support it using `ReactDOM.unstable_batchedUpdates`.

> we have to ensure that the lifecycle methods are called properly, in the proper order, and with the proper level of asynchrony. Is componentDidMount called in the same tick that componentWillMount is called in? I think it is, but we should make sure.

I guess so. I'll confirm the original behaviors of React.

Currently, this PR is including very minimum implementations.
I'll consider the implementation makes more precisely using React internal APIs like `ReactUpdates`.

jwbay:
This would be fantastic. My shallow rendering tests are chock full of x.instance().componentDidMount(). Would ref support be a blocker for this PR getting released considering the underlying shallow renderer from React just doesn't support them?

koba04:
I think ref support is not a blocker for this PR.

Because it is a limitation of Shallow rendering. Current `shallow` implementation is not supporting it too.

I might be able to add the refs support in this PR. But it would be better to implement it in React Shallow rendering instead of including it in enzyme.

This PR is in an early stage. I have to solve some issues and write some test cases.

Aweary:
@koba04 any progress on this since? I'm super interested in seeing this merged

koba04:
@Aweary 
I think this PR has the following problems.

- [ ] Support Refs. (must?)
- [x] Support Context
- [ ] Support to call setState in Lifecycle methods. (I already created the PR for this #342.)
- [ ] Ensure that the lifecycle methods are called properly, in the proper order, and with the proper level of asynchrony. (I have to confirm this but I guess it is implemented properly already.)

I think this can be merged if #342 is merged and I've ensured that the lifecycle methods are called properly.

Aweary:
@koba04 

Is #342 required for `setState` to work in specific lifecycle methods? It seems like `setState` works in `componentWillReceiveProps` as is, so what does batching provide that enables `setState` in other methods?

Most lifecycle methods are invoked in [`ReactCompositeComponent.updateComponent`](https://github.com/facebook/react/blob/master/src/renderers/shared/reconciler/ReactCompositeComponent.js#L797) with `componentDidMount` being enqueued in a transaction and  `componentWillMount` being called in `performInitialMount`. How accurate does the asynchrony need to be? I think if we just follow this diagram from the React source we should be mostly fine.

```

/**
 * ------------------ The Life-Cycle of a Composite Component ------------------
 *
 * - constructor: Initialization of state. The instance is now retained.
 *   - componentWillMount
 *   - render
 *   - [children's constructors]
 *     - [children's componentWillMount and render]
 *     - [children's componentDidMount]
 *     - componentDidMount
 *
 *       Update Phases:
 *       - componentWillReceiveProps (only called if parent updated)
 *       - shouldComponentUpdate
 *         - componentWillUpdate
 *           - render
 *           - [children's constructors or receive props phases]
 *         - componentDidUpdate
 *
 *     - componentWillUnmount
 *     - [children's componentWillUnmount]
 *   - [children destroyed]
 * - (destroyed): The instance is now blank, released by React and ready for GC.
 *
 * -----------------------------------------------------------------------------
 */
```

koba04:
@Aweary Thank you for your feedbacks!

I think the following cases should be batching updates.

* `componentWillMount` and `componentDidMount` when a component is mounted.
  * https://jsfiddle.net/koba04/s3pmr7yz/1/

* `componentWillReceiveProps` and `componentWillUpdate` and `componentDidUpdate` when a component is updated.
  * https://jsfiddle.net/koba04/ruaz8eu2/ (componentWillUpdate)
  * https://jsfiddle.net/koba04/y4syhyyL/1/ (componentDidUpdate)
  * https://jsfiddle.net/koba04/3yg25mwm/1/ (componentWillReceiveProps)

For supporting the above, I'm going to use #342 .
The test cases are https://github.com/airbnb/enzyme/pull/318/commits/224d320e9ca33708ff89a60c8c6fbe427fc8d58e

Aweary:
Got it, so that PR should make `setState` work as it does internally, which in turn allows the lifecycle methods to behavior as we'd expect 👍 

As far as execution order goes this seems generally in line with `ReactCompositeComponent`, so `ref` support seems to be the major blocking item for this specific PR.

koba04:
Yeah, I'm considering to support Refs. I think I have two options for it.

* Simulating Refs behaviors in enzyme.
* Adding Refs support in React Shallow.

erikthedeveloper:
This is great to see someone ( @koba04 ) taking this on! We have been also performing similar hacks (i.e. `instance.componentDidUpdate(prevProps)` etc...) all over our tests. I'm not 100% clear on what steps remain here, but I am open to contribute.

lelandrichardson:
The reason `ref` support is a big deal is actually part of a broader issue than just refs. The problem is that  there are certain lifecycle methods where the existence of backing DOM nodes is expected (and thus, `React.findDOMNode` should return something, and `refs` should return instances.

Historically `shallow` has been something that systematically stayed away from these lifecycle methods because when you start expecting backing instances of child components to exist, the whole point of `shallow` goes out the window.

Adding in full lifecycle support for components that *don't* use `refs` or `React.findDOMNode` makes a lot of sense, but probably *only* for these cases.

If we could somehow make this API an opt-in feature, then I don't think support for refs would be needed at all.

One thing we could do starting off is provide it as an "experimental" flag:

```
const wrapper = shallow(<Foo />, { lifecycleExperimental: true })
```
or something like that.

@koba04 what are your thoughts?

ljharb:
SECRET_LIFECYCLE_DO_NOT_USE_OR_YOU_WILL_BE_FIRED

lelandrichardson:
more like EXPERIMENTAL_LIFECYCLE_DO_NOT_GET_MAD_AT_US_IF_BEHAVIOR_CHANGES :)

koba04:
@lelandrichardson 

> One thing we could do starting off is provide it as an "experimental" flag:
> const wrapper = shallow(<Foo />, { lifecycleExperimental: true })

I think it is a good first step for supporting lifecycle methods!
It would be nice if this is added with the experimental flag at first.

I agree with supporting `ref` is a big deal.
It's a limitation of ShallowRendering.

https://facebook.github.io/react/docs/test-utils.html#shallow-rendering

I think supporting `ref` is an issue of the `shallow` rather than lifecycle methods in `shallow`.
When we use `shallow`, we can't use `ref` in your component's `render` method even now.

I'd like to take this too, but I'd like to take as an another issue.

koba04:
> Historically shallow has been something that systematically stayed away from these lifecycle methods because when you start expecting backing instances of child components to exist, the whole point of shallow goes out the window.

Right, Supporting `ref` needs to instantiate child components, which is including to call lifecycle methods of child components.
If the child component has `ref` for using its child component instances, should we instantiate it?

I think it's no longer **shallow**...

We need to discuss how `ref` should behave in `shallow`.

koba04:
interesting.

```
Ben has been prototyping a full React test renderer that can render everything both on React Native and DOM.
Unlike shallow renderer, it is stateful and renders deeply.
```

https://github.com/reactjs/core-notes/blob/master/2016-05/may-12.md#snapshot-testing

jwbay:
Looks like there might be another renderer for Enzyme to wrap 😅 

As described, I don't think the new test renderer would make shallow rendering any less valuable. If there was a switch or option to turn off deep rendering, though, *that* would be interesting.

Aweary:
If anything it seems to be closer to `mount`, though I guess we'll see 👍 it might make things a lot easier for us. It'd be nice if we could see what progress has been made with it, or what the API might look like.

@spicyj how do you think your new renderer might affect enzyme?

spicyj:
Yes, it's like `mount` except you don't need jsdom and it would also work for RN components without a host environment. It just outputs JSON.

Aweary:
@spicyj very interesting, thanks for chiming in. JSON should be just about as easy as it gets for us in terms of consuming/wrapping the API. Sounds like a great middle-ground between `shallow` and `mount`, looking forward to seeing more 👍

koba04:
@Aweary @lelandrichardson

I've implemented `lifecycleExperimental` flag for enabled this feature.

To be clear, the following tables are differences of behavior by the flag.

## mounting phase

| lifecycleExperimental  | true | false |
| ------------ | ---- | ---- |
| componentWillMount | ◯ | ◯ |
| componentDidMount  | ◯ | × |

## updating phase

### setState

| lifecycleExperimental  | true | false |
| ------------ | ---- | ---- |
|shouldComponentUpdate|◯|◯|
|componentWillUpdate|◯|◯|
|componentDidUpdate|◯|◯|

### setProps and setContext

| lifecycleExperimental  | true | false |
| ------------ | ---- | ---- |
|shouldComponentUpdate|◯|◯|
|componentWillReceiveProps|◯|◯|
|componentWillUpdate|◯|◯|
|componentDidUpdate|◯|×|

## unmounting phase

| lifecycleExperimental  | true | false |
| ------------ | ---- | ---- |
|componentWillUnmount|◯|◯|

nfcampos:
@koba04 could we have a test asserting the order in which lifecycle hooks are called, especially those related to updates?

koba04:
@nfcampos 
Thank you for your review! I've updated tests for asserting the order.

test/ShallowWrapper-spec.js
```diff
@@ -2455,6 +2455,575 @@ describe('shallow', () => {
...
+        expect(spy.args).to.deep.equal(
+          [
+            [
+              'shouldComponentUpdate',
+              { foo: 'bar' }, { foo: 'baz' },
+              { foo: 'state' }, { foo: 'state' },
+              { foo: 'context' },
+            ],
+            [
+              'componentWillReceiveProps',
```
nfcampos: `componentWillReceiveProps` should be called before `shouldComponentUpdate` (see this test for an example https://github.com/facebook/react/blob/c0ecde687ae3e599dced83a111e57ed49897b0b9/src/renderers/shared/stack/reconciler/__tests__/ReactComponentLifeCycle-test.js#L565)

test/ShallowWrapper-spec.js
```diff
@@ -2455,6 +2455,575 @@ describe('shallow', () => {
...
+            return true;
+          }
+          componentWillUpdate(nextProps, nextState, nextContext) {
+            spy('componentWillUpdate', this.props, nextProps, this.state, nextState, nextContext);
+          }
+          componentDidUpdate(prevProps, prevState, prevContext) {
+            spy('componentDidUpdate', prevProps, this.props, prevState, this.state, prevContext);
+          }
+          render() {
+            return <div>foo</div>;
```
nfcampos: we should add the `spy()` call here in `render` to assert that it is called between `componentWillUpdate` and `componentDidUpdate`

test/ShallowWrapper-spec.js
```diff
@@ -2455,6 +2455,575 @@ describe('shallow', () => {
...
+            this.setState({ count: this.state.count + 1 });
+            this.setState({ count: this.state.count + 1 });
+          }
+          render() {
+            spy();
+            return <div>{this.props.foo}</div>;
+          }
+        }
+        const result = shallow(<Foo />, { lifecycleExperimental: true });
+        result.setProps({ name: 'bar' });
```
nfcampos: let's add `expect(spy.callCount).to.equal(1);` before this line to make it clearer that the `setProps` call is calling `render` only once

test/ShallowWrapper-spec.js
```diff
@@ -2455,6 +2455,575 @@ describe('shallow', () => {
...
+              this.setState({ count: this.state.count + 1 });
+            }
+          }
+          render() {
+            spy();
+            return <div>{this.props.foo}</div>;
+          }
+        }
+        const result = shallow(<Foo />, { lifecycleExperimental: true });
+        result.setProps({ name: 'bar' });
```
nfcampos: `expect(spy.callCount).to.equal(1);` before this line

test/ShallowWrapper-spec.js
```diff
@@ -2455,6 +2455,575 @@ describe('shallow', () => {
...
+        expect(result.state('count')).to.equal(1);
+        expect(spy.callCount).to.equal(2);
```
nfcampos: this line (or the title of this test) should have a comment saying that the fact that `callCount` is `2` asserts that `setState` inside `componentWillReceiveProps` does not provoke another render.

test/ShallowWrapper-spec.js
```diff
@@ -2455,6 +2455,575 @@ describe('shallow', () => {
...
+          }
+          render() {
+            spy();
+            return <div>{this.props.foo}</div>;
+          }
+        }
+        const result = shallow(<Foo />, { lifecycleExperimental: true });
+        result.setProps({ name: 'bar' });
+        expect(result.state('count')).to.equal(1);
+        expect(spy.callCount).to.equal(3);
```
nfcampos: this line (or the title of this test) should have a comment saying that the fact that `callCount` is `3` asserts that `setState` inside `componentWillReceiveProps` *does* provoke another render.

test/ShallowWrapper-spec.js
```diff
@@ -2455,6 +2455,575 @@ describe('shallow', () => {
...
+              /* eslint-enable react/no-did-update-set-state */
+            }
+          }
+          render() {
+            spy();
+            return <div>{this.props.foo}</div>;
+          }
+        }
+        const result = shallow(<Foo />, { lifecycleExperimental: true });
+        result.setProps({ name: 'bar' });
```
nfcampos: same comments for this test as the one right above

test/ShallowWrapper-spec.js
```diff
@@ -2455,6 +2455,575 @@ describe('shallow', () => {
...
+            return true;
+          }
+          componentWillUpdate(nextProps, nextState, nextContext) {
+            spy('componentWillUpdate', this.props, nextProps, this.state, nextState, nextContext);
+          }
+          componentDidUpdate(prevProps, prevState, prevContext) {
+            spy('componentDidUpdate', prevProps, this.props, prevState, this.state, prevContext);
+          }
+          render() {
+            return <div>foo</div>;
```
nfcampos: let's add a `spy` call to `render` here as well

test/ShallowWrapper-spec.js
```diff
@@ -2455,6 +2455,575 @@ describe('shallow', () => {
...
+        const spy = sinon.spy();
+        class Foo extends React.Component {
+          constructor(props) {
+            super(props);
+            this.state = {
+              foo: 'bar',
+            };
+          }
+          shouldComponentUpdate() {
+            return false;
```
nfcampos: there should be a `spy()` call here to make sure we are actually executing this method

nfcampos:
@koba04 thanks for your work on this! I've left a few more comments, other than that this looks ready to me 👍

nfcampos:
@koba04 I didn't leave any comments on the `updating context` block of texts but the same comments I made on the other tests apply to those as well

test/ShallowWrapper-spec.js
ljharb: and `expect(spy).to.have.property('callCount', 1)` to have a nicer failure message ;-)

koba04:
@nfcampos 
Thank you for your great feedbacks!
I've fixed them 😄

test/ShallowWrapper-spec.js
```diff
@@ -3056,7 +3056,7 @@ describe('shallow', () => {
         }
       }
       shallow(<Foo />, { lifecycleExperimental: false });
-      expect(spy.calledOnce).to.equal(false);
+      expect(spy.called).to.equal(false);
```
ljharb: `expect(spy).to.have.property('callCount', 0)` will have a nicer failure message

koba04: Thank you 😄  I've fixed it!

lelandrichardson:
LGTM. Awesome work. Let's get this out behind the flag and see what happens. I'll try to make sure ppl at Airbnb are aware of it.

lelandrichardson:
Am I right in that this change is *NOT* a breaking change, because it is behind the experimental flag? That was the whole point, right?

thatguynamedandy:
Apologies but I think this may have had backwards compatibility issues, I have a simplified test case from our codebase which started failing post upgrade.

The following now fails where it used to pass..
```js
      class Test extends Component {
        shouldComponentUpdate() { return false; }
        render() { return <div />; }
      }

      component = shallow(<Test  />);

      instance = component.instance();
      sandbox.spy(instance, 'shouldComponentUpdate');
      component.setProps({});
      assert(instance.shouldComponentUpdate.calledOnce);
```

If I add the lifecycleExperimental to shallow, it fixes the issue, I'm not sure if this has been a misuse by us or wether something else is going on.
```js
{ lifecycleExperimental: true }
```

koba04:
@lelandrichardson 
It's a bug. I'm going to fix this bug as soon as possible. Sorry...

koba04:
@thatguynamedandy It's fixed in 2.4.1. Thank you for reporting it!

Aweary:
Thanks for all the great work @koba04 🎉

blainekasten:
Did we ever remove the `lifecycleExperimental` flag and just make it part of the standard flow?

koba04:
I've created a PR for React supports all lifecycle methods in ShallowRender.

* https://github.com/facebook/react/pull/7912

If the PR is merged, we can remove this implementation and enable this without `lifecycleExperimental` flag.

ljharb:
@koba04 we'd still need to support React 0.13, 0.14, and v15, so we'd have to think carefully about how we maintain compat across all React versions.

koba04:
@ljharb Sure. That PR has a breaking change, so I think these are possible to exist together by using the `REACT16` variable.

nevir:
Does https://github.com/facebook/react/pull/9426 allow it to be enabled by default?

koba04:
@nevir No, ShallowRenderer v16 will behave the same way.
https://github.com/facebook/react/pull/9426/files#diff-a8123b15946bc2724f8ff58d87aa3bbbR183

I think it would be nice if ShallowRenderer supported this in a future major release.

nxmohamad:
thank you @koba04 for this PR! just read through the whole thread simply because I was curious of the culmination of this feature

koba04:
@nxmohamad 
Thanks!
In React v16, ShallowRenderer never calls `componentDidMount` and `componentDidUpdate` so we need some works to support all lifecycle methods for React v16.

jose920405:
hi @koba04 

Thank you very much for the work done, it has been very helpful.

With this last comment you left, I finally make sense of a problem I'm going through https://github.com/airbnb/enzyme/issues/1279.

My Info:
```
"name": "enzyme",
"version": "3.1.0",
```

-----------------

```
"react": "16.0.0",
"react-native": "0.47.2"
```

`componentDidUpdate` not works for me.

I am testing and apparently this is a problem related to redux interactions. When I change my redux `dispatch` by a simple `setState`, `componentDidUpdate` is executed correctly.


In case this is not yet available for `React v16`. I wanted to ask if you or someone is currently working on this?

koba04:
Hi, @jose920405 

> I am testing and apparently this is a problem related to redux interactions. When I change my redux dispatch by a simple setState, componentDidUpdate is executed correctly.

According to [this comment](https://github.com/airbnb/enzyme/issues/34#issuecomment-338686348), are you using `lifecycleExperimental` flag, aren't you? What about `componentWillUpdate`?
If so, that is an unexpected behavior. Can you create a reproduce case for this?

Currently, enzyme has a bug that doesn't call componentDidUpdate for setState if you don't enable lifecycleExperimental. This is a PR to fix this https://github.com/airbnb/enzyme/pull/1261.

jose920405:
Yeah, i'm using `lifecycleExperimental`. Although I understand, for version `3.1.0` is already default to` true`. Anyway I added it to be sure.

`componentWillUpdate` is not fired either

My code:

My **LoginScreen Component**

```
// Reducer
import UserReducer from '../Redux/Reducers/UserReducer'

class LoginScreen extends Component {

  constructor(props) {
    super(props)
  }

  componentDidMount() {
    console.log('componentDidMount ===>')
  }

  componentWillUpdate() {
    console.log('componentWillUpdate ====> ')
  }

  componentDidUpdate() {
    console.log('componentDidUpdate ====> ')
  }

  render() {
    console.log('Inside component Render')
    return (
      <View style={styles.mainContainer}>
        <View style={styles.formContainer}>
          <TextInputForm
            eachFormCustomTop={styles.eachFormCustomTop}
            icon={{
              source: Images.iconLogin
            }}
            textInput={{
              style: styles.textInputs,
              placeholder: 'Email',
              autoCapitalize: 'none',
              keyboardType: 'email-address',
              value: this.props.user.email,
              onChangeText: (text) => {
                console.log('onChangeText email text ==> ', text); // this log fires.
                this.props.email_typing(text) // `componentDidUpdate` doesn't wors

                // this.setState({ // If i use `setState` works
                //   changeTest: ''
                // })

              },
            }}
          />
        </View>
      </View>
    )
  }
}

const mapStateToProps = (state) => {
  return {
    user: state.user
  }
}

const mapDispatchToProps = (dispatch) => {
  return {
    email_typing: (text) => dispatch(UserReducer.email_typing(text)),
  }
}

export default connect(mapStateToProps, mapDispatchToProps)(LoginScreen)

```

My **Test File**:

```
import React from 'react'

import { createStore, combineReducers } from 'redux';

import Enzyme, { shallow, mount } from 'enzyme';

Enzyme.configure({ adapter: new Adapter() });

import LoginScreen from '../../App/Containers/LoginScreen'

// Components
import TextInputForm from '../../App/Components/TextInputForm'

const store = createStore(
  combineReducers(
    {
      user: require('../../App/Redux/Reducers/UserReducer').reducer,
    }
  )
);

function shallowComp() {
  return shallow(
    <LoginScreen />,
    { context: { store }, lifecycleExperimental: true },
  );
}

function updateComp() {
  render = wrapper.update().dive();
}

let wrapper = shallowComp();
let render = wrapper.dive();

describe('Login Screen Test', () => {

  test('>>>>> Render Correctly', () => {
    expect(wrapper).toHaveLength(1)
  })

  test('>>>>> onChangeText in Email', () => {
    const textInputs = render.find(TextInputForm);
    const emailTextInput = textInputs.first().getElement().props.textInput;
    
    emailTextInput.onChangeText('myEmail@email.com') // Here Execute `onChangeText` inside render of LoginScreen comp

    updateComp()

    const textInputsChanged = render.find(TextInputForm);
    let emailTextInputChanged = textInputsChanged.first().getElement().props.textInput;
    expect(emailTextInput.value).not.toEqual(emailTextInputChanged.value)
  })
})

```

`this.props.email_typing(text)` inside `onChangeText`: 

It works as expected, runs the change on my reducer, and causes the render to run again. **componentDidUpdate** DOES NOT FIRE.

jose920405:
@koba04 

When you say.

`if you call setState of React Component instance directly, componentDidUpdate is never called`.

This is something that is not clear to me, since it works very well for me.

If we go to my example inside `onChangeText` and replace `this.props.email_typing(text)` to:

```
this.setState({ // If i use `setState` works
   changeTest: ''anyval"
})
```

results in the `componentDidUpdate` entering correctly.

**My problem is directly with redux actions and not with `setState`.
Although I suppose at the end of the day, a redux action must run a `setState` at some point**

koba04:
@jose920405 Thanks, That's a weird behavior. Could you create a repository or gist to be able to confirm this?

jose920405:
hi @koba04 

apologies for the delay, I was quite busy in other things these days :)

This is the repo with the example.

https://github.com/jose920405/example-react-native-redux-with-enzyme

jose920405:
hi @koba04 

apologies for the delay, I was quite busy in other things these days 🙂 

This is the repo with the example.

https://github.com/jose920405/example-react-native-redux-with-enzyme

unfortunately I have not been able to replicate what I say to you if the `componentDidUpdate` works for me when I use the` setState` 🙁 .

But I'm still testing it on my original project and there if I still use the `componentDidUpdate` 😂 , there's something I should be doing that I may not remember.

This is my proving:

captura de pantalla 2017-11-02 a la s 3 55 46 p m

____

For now it is happening to me what it says that the `componentDidUpdate` does not run.

koba04:
@jose920405 Thanks for creating the repo!
Currently, I have no time for this so no I'm not sure what I should do for this yet.
But I'll keep working on this with the repo. Thanks!

Issue #277, opened 2016-03-21
.simulate() should provide a basic event object methods by default


tarjei:
Hi,
If I do simulate('click') then it would be great if the call to the handler included a bare bones event object containing the normal preventDefault() methods instead of having to define those every time.

gmcleod-buzz:
I second this feature, would be amazing!

Cleod9:
I second this feature, would be amazing!

lelandrichardson:
@tarjei I agree this would be a great addition. I would encourage you to open a PR if this is something that you're really wanting!

I'd like to improve the event simulation for enzyme, but have not yet really gotten around to it in earnest. Any help would be appreciated!

PR #329, opened 2016-04-18
Add default events object to shallow wrapper simulate

danielhusar:
Hey,

I had to pass preventDefault mock lot of times to simulate event in shallow rendering, so I thought it would be nice to have it by default there.

Fixes: https://github.com/airbnb/enzyme/issues/277

### DIFF
--- a/src/ShallowWrapper.js
+++ b/src/ShallowWrapper.js
@@ -350,8 +350,11 @@ export default class ShallowWrapper {
     const handler = this.prop(propFromEvent(event));
     if (handler) {
       withSetStateAllowed(() => {
-        // TODO(lmr): create/use synthetic events
         // TODO(lmr): emulate React's event propagation
+        if (!args.length && window) {
+          args.push(new window.Event(event));
+        }
+
         handler(...args);
         this.root.update();
       });
--- a/test/ShallowWrapper-spec.js
+++ b/test/ShallowWrapper-spec.js
@@ -862,7 +862,23 @@ describe('shallow', () => {
       expect(wrapper.find('.clicks-0').length).to.equal(1);
       wrapper.simulate('click');
       expect(wrapper.find('.clicks-1').length).to.equal(1);
+    });
 
+    it('should pass in default event data', () => {
+      const spy = sinon.spy();
+      class Foo extends React.Component {
+        render() {
+          return (
+            <a onClick={spy}>foo</a>
+          );
+        }
+      }
+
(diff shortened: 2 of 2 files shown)


danielhusar:
It failed on old node's because of jsdom not working there :(

src/ShallowWrapper.js
```diff
@@ -350,8 +350,11 @@ export default class ShallowWrapper {
     const handler = this.prop(propFromEvent(event));
     if (handler) {
       withSetStateAllowed(() => {
-        // TODO(lmr): create/use synthetic events
         // TODO(lmr): emulate React's event propagation
+        if (!args.length && window) {
+          args.push(new window.Event(event));
```
ljharb: `window` is not available when shallow-wrapping.

danielhusar: Don't most users have it available when running tests ?
I know its not requirement for shallow rendering, but are there many people that use only shallow rendering ?

Optionally I was thinking to add jsdom there but introducing this as dependency is probably not good idea.

Maybe just hardcode some event object with most used properties ?

ljharb: No. Many users do not use `mount` at all and do not have the DOM available during tests.

`shallow` must not require or assume a browser environment.

danielhusar: 👍 So what would be better approach ? Check for `global.window` and add it only for users that have it exposed ? (Leads to inconsistency) 
Or create manually event object ? (Don't think so we can mock properly all properties)

blainekasten:
I feel like something I would rather see is a mocking event object in enzyme that I can pass to simulate.

Something like this:

```js
wrapper.simulate('click', enzyme.mockEvent);
```

And that is effectively an object that represents the shape of a typical event.

src/ShallowWrapper.js
ljharb: Manually create an event object. The behavior of `shallow` should not differ based on the presence or absence of browser globals.

lelandrichardson:
We definitely cannot rely on browser APIs for shallow.

I would love to see a well built event mocking system built for enzyme that would remove some of the boilerplate, but I hesitate to add something that only implements it half-way - which we would then have to have a potential additional breaking change for in the future.

lelandrichardson:
I'm going to close this in favor of #368 which I think is going more closely in the direction that we want, but I believe will accomplish everything you are wanting.

Thanks!

PR #335, opened 2016-04-19
Add `name` method

SimenB:
I think I'm doing something weird here, for the `shallow` one. To be able to get the name, I have to wrap the component I want in another component. Not a problem for `mount`. But this is how `type` works as well, so I don't know if I'm doing something, or if it's by design.
The weird thing is the docs for `type` says

```jsx
expect(wrapper.type()).to.equal(Foo);
```

Which is false, at least for a component like

```jsx
  render() {
    return <div />;
  }
}

expect(wrapper.type()).to.equal(Foo); // Fails, it's 'div'
```

It works correctly using `mount`, though.

```jsx
  render() {
    return <div />;
  }
}

expect(wrapper.type()).to.equal(Foo); // Succeeds
```

So I don't know if it's a bug, or if the docs for `shallow` is wrong.

But there is a test for it that checks the current behavior.
`shallow` is null: https://github.com/airbnb/enzyme/blob/1b9d317c57ba03414ae43109fa36bf4abf6e3445/test/ShallowWrapper-spec.js#L2213
`mount` is not: https://github.com/airbnb/enzyme/blob/1b9d317c57ba03414ae43109fa36bf4abf6e3445/test/ReactWrapper-spec.js#L1881

If it is correct, maybe the docs need to be updated?


test/ShallowWrapper-spec.js
```diff
@@ -2244,4 +2244,45 @@ describe('shallow', () => {
...
   });
 
+  describe('.name()', () => {
+    describe('node with displayName', () => {
+      it('should return the displayName of the node', () => {
+        class Foo extends React.Component {
+          render() { return <div />; }
+        }
+
+        class Wrapper extends React.Component {
```
SimenB: this is weird to me

test/ShallowWrapper-spec.js
```diff
@@ -2244,4 +2244,45 @@ describe('shallow', () => {
...
+      });
+    });
+
+    describe('node without displayName', () => {
+      it('should return the name of the node', () => {
+        class Foo extends React.Component {
+          render() { return <div />; }
+        }
+
+        class Wrapper extends React.Component {
```
SimenB: also this

ljharb:
I'll let @lelandrichardson comment on the shallow weirdness you mentioned.

Could you please add `React.createClass` and SFC tests for both shallow and mount?

SimenB:
Tests added for `displayName` using SFC and `React.createClass`

test/ShallowWrapper-spec.js
```diff
@@ -2244,4 +2244,72 @@ describe('shallow', () => {
     });
   });
 
+  describe('.name()', () => {
```
SimenB: The wrapper components here are what I'm talking about

test/ReactWrapper-spec.js
```diff
@@ -1896,4 +1896,62 @@ describeWithDOM('mount', () => {
...
+
+        Foo.displayName = 'CustomWrapper';
+
+        const wrapper = mount(<Foo />);
+        expect(wrapper.name()).to.equal('CustomWrapper');
+      });
+
+      describeIf(!REACT013, 'stateless function components', () => {
+        it('should return the name of the node', () => {
+          const Foo = () => <div />;
```
ljharb: instead of an arrow (since arrow functions don't have names) could this be a normal function SFC, with a differing name?

SimenB: Like

```jsx
function SFC() {
  return <div />;
}
```
?

ljharb: yup!

SimenB: done

SimenB:
I added the `createClass` and SFC tests only for `displayName`. Want me to add them for `name` as well?

ljharb:
It would probably be best to include them, to prevent future regressions.

SimenB:
Aight, added.

SimenB:
@ljharb ping :smile:

ljharb:
@SimenB i've left it for @lelandrichardson to take a final look at - although now it needs a fresh rebase ;-)

SimenB:
Ah, didn't notice. Will rebase!

lelandrichardson:
@SimenB I'm good with this addition. Looks good.

Regarding the behavior you're mentioning above, this is expected. when you run `shallow(<Foo />)`, you are *rendering* `Foo`, and then wrapping around the result... so the wrapper by default is looking at the root node (which in your example is a div).

Sorry this took me so long, but you think you can rebase this now? Just merged a big PR and would like things to be sequential. Thanks.

SimenB:
Rebased

lelandrichardson:
doh. Merged something else in and now there are conflicts :(

@SimenB yes, those docs look very misleading! I'll go in and fix. Thanks for pointing it out.

SimenB:
Rebased against latest master

SimenB:
Great, thanks!

Any plans on a new release soon? A lot of nice features has been added since 2.2 😄

jerrysu:
The examples here are also wrong: https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/props.md#example

PR #339, opened 2016-04-20
Add an `isEmptyRender()` method to both `ShallowWrapper` and `ReactWrapper`

milesj:
# Why

Added an `isNull()` method to both `ShallowWrapper` and `ReactWrapper`,
making it easier in tests to detect when a component renders nothing;
either a null or undefined value.

I chose the name `isNull()` since `isEmpty()` was used for other situations.
It's not as descriptive as it could be, so open to suggestion.

# Tested

Unit tests!


src/ReactWrapper.js
```diff
@@ -288,6 +288,15 @@ export default class ReactWrapper {
   }
 
   /**
+   * Returns true if the component rendered nothing, i.e., null or undefined.
+   *
+   * @returns {boolean}
+   */
+  isNull() {
```
sjdemartini: I feel like this name too specifically implies that the component did `return null` in its `render` method, even though `return undefined` (either explicitly or implicitly) should also be fine.

src/ReactWrapper.js
```diff
@@ -288,6 +288,15 @@ export default class ReactWrapper {
...
+    return this.single(n => findDOMNode(n) === null);
```
ljharb: `return this.type() === null`?

src/ShallowWrapper.js
```diff
@@ -254,6 +254,15 @@ export default class ShallowWrapper {
   }
 
   /**
+   * Returns true if the component rendered nothing, i.e., null or undefined.
+   *
+   * @returns {boolean}
+   */
+  isNull() {
+    return (this.node === null);
```
ljharb: please remove the unnecessary parens.

if this is the canonical location for "is null", we should call into `isNull` in a number of other places, like `.type()` for example.

src/ReactWrapper.js
milesj: That doesn't actually work. It also seems like SFC causes issues too. Trying to figure that out.

milesj:
Seems like a lot changed internally during the React 15 upgrade. I've not been able to find an implementation that works across all versions. Using `this.type()` doesn't work for all use cases, nor does `findDOMNode()`, so I had to use a conditional.

lelandrichardson:
I think we need to duplicate all of these tests for `undefined` and `false`, which are also things that are "rendering nothing" that we would probably want this function to behave identically for. Looking at the code, I don't think this will pass right now.

The other question for me is the name.  `isNull()` is pretty directly indicating that we are testing for `null`, which might not always be the case (other things can return `ReactNodeTypes.EMPTY`).

That said, I'm not really sure what the right name is. Maybe it should be `isEmpty()` or `isNothing()`?

milesj:
Perhaps something like `hasRenderedEmpty()` or `hasRenderedNothing()`?

milesj:
Ok, I changed it to `isEmptyRender()`, to be a bit more explicit. I also added `false` based tests and tried to add `undefined`, but supposedly you can't, as React errors with "Invariant Violation: Unexpected node: undefined" or "Error: Invariant Violation: Foo.render(): A valid ReactComponent must be returned. You may have returned undefined, an array or some other invalid object.".

lelandrichardson:
Hey @milesj , you're right. My bad - I guess undefined is not valid here.

We also want to ensure that this returns false in the proper cases though.

`false` => `true`
`null` => `true`
`true` => `false`
`0` => `false`
`<div />` => `false`
`<Foo />` => `false`


At the moment, `isEmptyRender() { return true; }` would pass all of these tests :)

I think at that point this will be ready to merge. Not super psyched about the name, but can't think of anything better at the moment.... so at least it's explicit.

milesj:
I was a bit surprised about the undefined thing as well. I could of sworn it allowed it at some point.

I'll look into adding some truthy tests.

ljharb:
Don't forget about `NaN` and the empty string, just to make sure the other falsy values are covered (both should return `false` i believe)

Aweary:
@milesj any updates on this? 😄

milesj:
Oops, thanks for reminding me. Been pretty swamped at work. Will try to get it this week.

Aweary:
No worries, don't feel pressured/rushed! Thanks for the contribution, looking forward to seeing how this turns out 👍

milesj:
Added a ton of return value checks.

milesj:
@lelandrichardson I added tests that check null/false, as well as React components/elements. I originally had it testing every possible falsey value, but then it just barfed all over the output, as seen by the screenshot. I don't really see the benefit of testing all the falsey values as React doesn't allow anything besides null/false.

screen shot 2016-05-19 at 12 37 07 pm

Aweary:
@milesj I think we should only be testing valid  "empty" return values 👍

ljharb:
What about `NaN`, and the empty string? Does React complain about those too?

milesj:
I tested a ton of values, all of which are: null, false, undefined, NaN, array, empty array, object, empty object, zero, string, empty string, React element/component, DOM element, etc.

All of them threw those warnings excluding null/false/React.

milesj:
Screenshot I took.
screen shot 2016-05-19 at 2 22 17 pm

Aweary:
LGTM

milesj:
@lelandrichardson Anything else you'd like done on this? Besides hitting the "Update Branch" button :P

Aweary:
@milesj can you squash this to get read of all those merge commits? I think we can merge this after that 👍

milesj:
Wouldn't Github's new merge squash feature suffice here?

I've tried rebasing the history a bit, but it's turned out troublesome since the commits happened at different points in history. I'll mess around some more.

Aweary:
Does Github do that now? I didn't realize. I'll merge.

ljharb:
(merge-squash is fine when the PR author has issues with rebasing, but in general we do strongly prefer rebased and non-squash-merged commits. github's squash-merge breaks the history between the PR and master)

milesj:
Yeah, GitHub's automatic "Update Branch" is useful, but annoying history wise. Will probably do it manually going forward.

PR #340, opened 2016-04-21
Update doc regarding webpack and React 15

tylercollier:
I've updated doc regarding webpack and React 15, which came from the [advice here](https://github.com/airbnb/enzyme/issues/47#issuecomment-207498885).

What I had to do to get things working with webpack and React 0.14 was different than in the guide, and I wrote about it [here](https://github.com/airbnb/enzyme/issues/47#issuecomment-213032872). I did not include it in this PR due to not really understanding what's going on and if my solution is a universal one, but I think a better guide will make sure people don't abandon using enzyme because they feel it's too hard to set up.

### DIFF
--- a/docs/guides/webpack.md
+++ b/docs/guides/webpack.md
@@ -46,6 +46,20 @@ react-dom/server
 react-addons-test-utils
 ```
 
+## React 15 Compatability
+
+If you are using React 15, your config should include these externals:
+
+```js
+/* webpack.config.js */
+// ...
+externals: {
+  'react/addons': true,
+  'react/lib/ExecutionEnvironment': true,
+  'react/lib/ReactContext': true
+}
+// ...
+```
 
 ## Example Projects
 


lelandrichardson:
LGTM. We added react 15 support, but didn't really update the documentation. Thanks for the help!

tylercollier:
Can you update the doc for React 0.14 too?

The current guide says:

> If you are using React 0.14, the instructions above will be the same but with a different list of externals:

In its list that follows, it isn't as helpful as what's shown for using with React 0.13 (which is actual javascript). I recommend writing actual javascript because with just a list, it's unclear what the *values* should be (`'window'` or `true`).

So something like this:

```javascript
/* webpack.config.js */
// ...
externals: {
  'cheerio': 'window',
  'react-dom': ???
  'react/lib/ExecutionEnvironment': true,
  'react/lib/ReactContext': true
}
// ...
```

Note the `???`. For me, to get things working, I actually removed that line and replaced it with `'react-deom/server': 'window'`. I'm not sure how 'correct' that is vs just my environment.

PR #342, opened 2016-04-22
ShallowRenderer.simulate supports batched updates

koba04:
This PR is for `ShallowWrapper.simulate()` supports batched updates using `ReactUpdates.batchedUpdates`.

`React.addons.TestUtils.Simulate.{eventName}` is supporting batched updates.
ref. https://jsfiddle.net/koba04/6Lchbest/

`ReactUpdates.batchedUpdates` is exported as `ReactDOM. unstable_batchedUpdates` from React v0.14. In React v0.13, it is exported as  `React.addons.batchedUpdates`.

It might be better I use the exported interface.
But I guess it makes your configurations complex than using `ReactUpdates.batchedUpdates` directly.

FYI: This batched updates feature is required #318.

Thanks.

### DIFF
--- a/src/react-compat.js
+++ b/src/react-compat.js
@@ -10,9 +10,9 @@ let findDOMNode;
 let childrenToArray;
 let renderWithOptions;
 let unmountComponentAtNode;
+let batchedUpdates;
 
 const React = require('react');
-const batchedUpdates = require('react/lib/ReactUpdates').batchedUpdates;
 
 if (REACT013) {
   renderToStaticMarkup = React.renderToStaticMarkup;
@@ -21,6 +21,7 @@ if (REACT013) {
   unmountComponentAtNode = React.unmountComponentAtNode;
   /* eslint-enable react/no-deprecated */
   TestUtils = require('react/addons').addons.TestUtils;
+  batchedUpdates = require('react/addons').addons.batchedUpdates;
   const ReactContext = require('react/lib/ReactContext');
 
   // Shallow rendering in 0.13 did not properly support context. This function provides a shim
@@ -77,6 +78,7 @@ if (REACT013) {
   renderToStaticMarkup = require('react-dom/server').renderToStaticMarkup;
   findDOMNode = ReactDOM.findDOMNode;
   unmountComponentAtNode = ReactDOM.unmountComponentAtNode;
+  batchedUpdates = ReactDOM.unstable_batchedUpdates;
   // We require the testutils, but they don't come with 0.14 out of the box, so we
   // require them here through this node module. The bummer is that we are not able
   // to list this as a dependency in package.json and have 0.13 work properly.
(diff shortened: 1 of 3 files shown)


lelandrichardson:
Interesting. So you are planning on using this to ensure that the batching behavior of setState is accurately depicted in #318?

koba04:
@lelandrichardson 

> So you are planning on using this to ensure that the batching behavior of setState is accurately depicted in #318?

Yes, it would be nice if I could ensure that.
But I'm not sure that it is possible because I'm trying it yet.

This is relevant to #318 but independent.

src/react-compat.js
```diff
@@ -134,6 +134,8 @@ if (REACT013) {
   };
 }
 
+const batchedUpdates = require('react/lib/ReactUpdates').batchedUpdates;
```
blainekasten: This might become an issue due to https://github.com/facebook/react/issues/6460. depending on which way they go. Hopefully this might help sway them away.

Aweary: I think we'll likely have to be reactive when it comes to https://github.com/facebook/react/issues/6460. We should definitely try to work with the React team if they want to support libraries like enzyme in some capacity.

Also @koba04 can you put this `require` up at the top with the main `const React = require('react')` statement, since this is not a version dependant import at all.

koba04: @Aweary I've updated it!

Aweary:
Since there is a public API for `batchedUpdates` maybe we should just wrap the `batchedUpdates` assignment call in the `REACT013` check and use the appropriate access points mentioned in the OP. It's more complicated but at least we can avoid reacting into `React/lib` for now.

What do you think @blainekasten @lelandrichardson @koba04?

koba04:
@Aweary I've updated this PR to use public APIs for batchedUpdates! It's not complicated for me 😄

Aweary:
LGTM

nfcampos:
LGTM

nfcampos:
I'll have to find out whether react batches updates for all handlers triggered by an event for #368, if anyone knows let me know

Aweary:
@nfcampos [a cursory glance suggest that they do](https://github.com/facebook/react/blob/62f64a84a43b9d8288df5136b816d76dcc7e3321/src/renderers/dom/client/ReactEventListener.js#L170) but we'll have to investigate that further to verify.

Aweary:
@ljharb @lelandrichardson it doesn't seem like the LGTM check seems to be working?

lelandrichardson:
LGTM

lelandrichardson:
@Aweary it looks like there's still some configuration we need to do with it to add you guys. i'll see if @ljharb can finish that today. thanks!

ljharb:
@Aweary turns out that the issue is https://github.com/lgtmco/lgtm/issues/19 - i'll update our MAINTAINERS file in the meantime.

koba04:
rebased

Aweary:
@koba04 can you rebase please? I think we can merge this afterwards.

koba04:
@Aweary rebased!

Aweary:
Thanks @koba04!

koba04:
Thanks for your help 🎉

PR #344, opened 2016-04-23
Typofix in props.md

volrath:
Fixes `.prop` method signature in "Related Methods"

### DIFF
--- a/docs/api/ShallowWrapper/props.md
+++ b/docs/api/ShallowWrapper/props.md
@@ -16,6 +16,6 @@ expect(wrapper.props().foo).to.equal(10);
 
 #### Related Methods
 
-- [`.prop() => Object`](prop.md)
+- [`.prop(key) => Any`](prop.md)
 - [`.state([key]) => Any`](state.md)
 - [`.context([key]) => Any`](context.md)


docs/api/ShallowWrapper/props.md
```diff
@@ -16,6 +16,6 @@ expect(wrapper.props().foo).to.equal(10);
 
 #### Related Methods
 
-- [`.prop() => Object`](prop.md)
+- [`.prop(key) => Any`](prop.md)
```
ljharb: both are actually valid, and the argumentless form is more important anyways. Perhaps we should list both here?

volrath: I tried `wrapper.prop()` and it returns `undefined`, and it actually [makes sense](https://github.com/airbnb/enzyme/blob/master/src/ShallowWrapper.js#L484) that it does. It's different from `wrapper.props()` which does return an `Object`. Plus the signature in [`prop.md`](https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/prop.md) is `.prop(key) => Any`. I'm not sure if I'm missing something here.

ljharb: aha, prop vs props :-) this is a good fix, thanks

volrath: yw :)

PR #354, opened 2016-04-29
Clearer doc on how to use .simulate()

gCardinal:
I actually spent a good while on this. I'm using MaterialUI and those components don't have `onClick` props, they use `onTouchTap` and all sorts of custom events of the sort to handle touch and clicks.

It was not clear that .simulate() actually called the prop itself and did not cause a "real" event to be sent to the component, so I was trying to `.simulate('click')`. With no success obviously.

I'm not sure if there's a specific PR policy for this repo, but I figured I'd try my hand at clarifying `.simulate()`'s documentation.


docs/api/ShallowWrapper/simulate.md
```diff
@@ -53,3 +53,8 @@ expect(wrapper.find('.clicks-1').length).to.equal(1);
 - Currently, event simulation for the shallow renderer does not propagate as one would normally
 expect in a real environment. As a result, one must call `.simulate()` on the actual node that has
 the event handler set.
+- Even though the name would imply this simulates an actual event, `.simulate()` will in fact 
+target the component's prop based on the event you give it. For example, `.simulate('click')` will
+actually get the `onClick` prop and call it. This is important to keep in mind when using
+libraries such as MaterialUI where clicks are set on a `onTouchTap` prop. You will need to call
```
Aweary: Maybe generalize this so it doesn't mention a specific library? Just for those who don't know what it is. If not, it should be Material-UI (with the dash), per their website.

joncursi:
I like the idea of this PR, but I think the example is wrong. React doesn't seem to support `onTouchTap` simulation:

```
TypeError: ReactWrapper::simulate() event 'touchTap' does not exist
    at ReactWrapper.<anonymous> (packages/modules.js?hash=b053b68f533e45d134e516cff9d658226e7fa66b:25409:19)
    at ReactWrapper.single (packages/modules.js?hash=b053b68f533e45d134e516cff9d658226e7fa66b:25975:19)
    at ReactWrapper.simulate (packages/modules.js?hash=b053b68f533e45d134e516cff9d658226e7fa66b:25405:14)
    at Context.<anonymous> (app/app.js?hash=48ae892390da5ada9154e1bedf99219291ef4069:875:49)
    at callFn (packages/practicalmeteor_mocha.js?hash=489e07e2e98906b5ed07059bc93d8f53fe79c00b:3227:21)
    at Test.Runnable.run (packages/practicalmeteor_mocha.js?hash=489e07e2e98906b5ed07059bc93d8f53fe79c00b:3220:7)
    at Runner.runTest (packages/practicalmeteor_mocha.js?hash=489e07e2e98906b5ed07059bc93d8f53fe79c00b:3688:10)
    at packages/practicalmeteor_mocha.js?hash=489e07e2e98906b5ed07059bc93d8f53fe79c00b:3799:12
    at next (packages/practicalmeteor_mocha.js?hash=489e07e2e98906b5ed07059bc93d8f53fe79c00b:3608:14)
    at packages/practicalmeteor_mocha.js?hash=489e07e2e98906b5ed07059bc93d8f53fe79c00b:3618:7
```

https://facebook.github.io/react/docs/events.html#touch-events

blainekasten:
@joncursi this PR is targeting `ShallowWrapper` docs while your error is within the `ReactWrapper` (mount).

docs/api/ShallowWrapper/simulate.md
blainekasten: I agree. We should either just trim out the enter last part, or generalize it. I think it's helpful enough to just have it say the shallow simulate's simply call the `onEvent` function.

joncursi:
I'm just suggesting a different event be used in the example rather than `touchTap` which has known compatibility issues inside of React. Might make more sense that way.

gCardinal:
Like @blainekasten suggested, I removed the part where I mention Material-UI. I hope it will still be clear enough.

Aweary:
I think this may need some updates for when https://github.com/airbnb/enzyme/pull/368 gets merged.

blainekasten:
@Aweary i'm not following that PR very closely. But does it change it so it's not calling the prop anymore? If not, I don't think it hurts to merge doc PR's. They can always be updated later easily.

nfcampos:
@blainekasten it's still calling the prop like before but now it simulates event propagation and calls that prop with an event-like object, so the result of calling `simulate` is more in line with the result of dispatching an actual event.

Aweary:
It may introduce [deprecations](https://github.com/airbnb/enzyme/pull/368#issuecomment-218554950) related to calling custom events, which would be something we'd also want to document.

blainekasten:
If it's still calling the prop. Then his doc suggestions are correct, and might need updating later. I say this is good to merged as is and we can update later. Then @gCardinal can get on the contributor list at least :)

LGTM

Aweary:
Yeah you're right, they are still technically correct (just potentially incomplete pending that PR) so I'm good with merging and updating later 😄

gCardinal:
Well, I guess it's possible to future-proof how I wrote it. What about:

> Be aware that `.simulate()` uses the component's props to simulate the event, meaning `.simulate('click')` will call whatever function is bound to your component's `onClick` prop and `.simulate('touch')` would call the `onTouch` prop.

I think it reflects well the current behavior and will still apply when that PR gets merged. What do you think?

Aweary:
@gCardinal I think what you have now is fine, I don't think we've decided on what exactly the future behavior is going to be 🙂

Aweary:
@gCardinal can you update your PR with the master branch? I think we can merge after that.

gCardinal:
I had never use Github's "auto fork feature-thingy" when simply editing `.MD` files before, so I'm not sure I merged properly, but it seems alright. Let me know if there's a problem with it.

blainekasten:
Looks good! Thanks @gCardinal

Issue #273, opened 2016-03-19
Option to have a prop ignored on contains


DanielMSchmidt:
Hi there,

First of all thanks for this great library! I am currently testing a React Native application and I found it hard to do so, as `contains` can't be differed to be more allowing.

My use case is that I have a component render a `<Text`> element which is set via prop, much like in your [example test](https://github.com/lelandrichardson/enzyme-example-react-native/blob/master/test/MyComponent.js). Unfortunately my <Text> element needs a style prop, which I don't intend to share between my tests and application. Would it be possible to add a whitelist to the compare functions as additional argument? I would be glad to help and do a PR if this feature is chosen to be useful for anyone else.

Have a nice day!

gm758:
I also found this a bit counter intuitive my first time using enzyme. I currently use the approach below when I want to test the contents of an element.

```
  it('should render text inside the Text component', () => {
    const text = 'Test Text';
    const wrapper = shallow(
      <Button text={text} />
    );

    expect(wrapper.find(Text).prop('children')).to.equal(text);
  });
```

smacker:
same idea: https://github.com/airbnb/enzyme/issues/204#issuecomment-185607556 :)

Aweary:
The new API introduced in https://github.com/airbnb/enzyme/pull/362 should cover this use case, so I'm going to close this as it seems resolved. Let me know if you have any questions!

Issue #204, opened 2016-02-18
How to use contains with function props?


smacker:
`contains` method makes very simple testing complex component and don't care about markup. Something like this works great:

```javascript
const MyComponent = () => (<div><Input /></div>)

it('should render input', () => {
    const wrapper = shallow(<MyComponent />);
    expect(wrapper.contains(<Input />)).to.equal(true);
});
```

But if I have function props, it doesn't work anymore:

```javascript
class MyComponent extends React.Component {
    _onChange() {}

    render() {
        return <div><Input onChange={this._onChange.bind(this)} /></div>;
    }
}

it('should render input', () => {
    const wrapper = shallow(<MyComponent />);
    expect(wrapper.contains(<Input onChange={what_should_I_put_here?} />)).to.equal(true);
});
```

In some cases it is possible to make binding in constructor (then I can get function from instance and pass as prop), but not always.

Do you see any possible solution for this problem? (for example https://github.com/algolia/expect-jsx just skip all function props when compere jsx)

ljharb:
@smacker generally you'd want to do `this.onChange = this.onChange.bind(this)` in the constructor anyways, to avoid creating a function in the render path.

If you do that, then `wrapper().instance().onChange` would be the function.

smacker:
I mentioned that it's possible to move binding in constructor in some cases, but not always.

Consider this examples:
```javascript
items.map((item) => <Input onChange={this._onInputChange.bind(this, item.type))} />

items.map((item) => <Input onChange={this._makeOnChangeCallback(item.name)} />
```

ljharb:
Ah, true you did. In that case, you created a new function and didn't store it anywhere else - it doesn't seem possible to locate things based on that function. You'd probably want to spy on the `onInputChange` etc prior to calling `render`, and then assert that the prop function calls through to your original.

smacker:
Could you please give a example with `onChange={this._onChange.bind(this, item.type)}`?
I ended up with a very ugly code, that overrides bind method :(

My point is, that in some cases I want to find component, but skip some props, i'll tests a behavior in other tests like:

```javascript
it('change of input should set name state', () => {
  const newValue = 'newValue';
  wrapper.find(<Input />).props('onChange')(newValue);
  expect(wrapper.getState('name')).toEqual(newValue);
})
```
In this case I don't care about private callbacks that were passed to component, I test only behavior.

ljharb:
So, in this case, I'd do something like the following:
```
[...]

PR #362, opened 2016-05-03
Add loose equals and loose contains functions on ShallowWrapper

mathieuancelin:
This PR add two functions on the `ShallowWrapper` to perform loose equality and loose contains check on a tree. The functions are named `.rightContains()` and `.rightEquals()` because the equality is based on the expected element (the right one) instead of the wrapped element (the left one). The function names are not very good and I'll be happy to change it for something more relevant and meaningful.

I added these functions because it is really really handy to write clean tests that check just the necessary parts of the tree for the test, i.e. "I want to check if that particular node is present in my tree, but I don't care about its style or onClick handler".

It works by checking if at least, all the props (including children) of the expected element are present on the wrapper element and equal to each other. It is based on the `Utils.nodeEqual` function with a little tweak on the props length comparison.

To use it, you just have to write something like : 

```javascript
const MyComponent = React.createClass({
  handleClick() { ...  },
  render() {
    return (
      <div>
        <div onClick={this.handleClick} className="foo bar">Hello</div>
      </div>
    );
  }
});

const wrapper = shallow(<MyComponent />);

// here we don't need to check if onClick is present on the node, so we just skip it
// same for the className
expect(wrapper.rightEquals(
  <div>
    <div>Hello</div>
  </div>
)).to.equal(true);

// here we don't need to check if onClick is present on the node, so we just skip it
// same for the className
expect(wrapper.rightContains(
  <div>Hello</div>
)).to.equal(true);

// here we don't need to check if onClick is present on the node, so we just skip it
expect(wrapper.rightContains(
  <div className="foo bar">Hello</div>
)).to.equal(true);
```

As the contributors guide suggest, I added some documentation and some tests that should pass with all versions of react.

These kind of feature are suggested on some issues like https://github.com/airbnb/enzyme/issues/204 or https://github.com/airbnb/enzyme/issues/273


Aweary:
I like this, with the exception of the API name. Maybe something like `looseContains`/`looseEquals` would be more accurate?

docs/api/ShallowWrapper/rightContains.md
```diff
@@ -0,0 +1,64 @@
...
+#### Arguments
+
+1. `nodeOrNodes` (`ReactElement|Array<ReactElement>`): The node or array of nodes whose presence you are detecting in the current instance's
+render tree.
+
+
+
+#### Returns
+
+`Boolean`: whether or not the current wrapper has a node anywhere in it's render tree that looks
```
Aweary: should be "its" not "it's"

docs/api/ShallowWrapper/rightEquals.md
```diff
@@ -0,0 +1,46 @@
...
+#### Arguments
+
+1. `node` (`ReactElement`): The node whose presence you are detecting in the current instance's
+render tree.
+
+
+
+#### Returns
+
+`Boolean`: whether or not the current wrapper has a node anywhere in it's render tree that looks
```
Aweary: same thing here, "its" not "it's"

ljharb:
I definitely am not a fan of the names.

Does this comparison ignore all props, or just `style` and `onClick`? I think we'd need to very explicitly define which things are compared and which are ignored, and ideally the names of the functions took this into account.

Aweary:
@ljharb I believe it ignores all props that aren't present on the expression passed into rightContains`/`rightEquals`. So something like `rightContains('<div><div/></div>')` would match any structure that contains a nested `div`, is that right @mathieuancelin?

ljharb:
so something like `matchesElement`, and `containsMatchingElement`?

mathieuancelin:
@Aweary @ljharb I'm not a fan of the name either, but first I wanted to write something that works, see if you guys are interested in such feature and then discuss the name, I know it can be tricky to agree on public API names ;-)

docs/api/ShallowWrapper/rightContains.md
```diff
@@ -0,0 +1,64 @@
+# `.rightContains(nodeOrNodes) => Boolean`
+
+Returns whether or not the current wrapper has a node anywhere in it's render tree that looks like
+the one passed in. Equality is based on the expected element (`nodeOrNodes`) and not on the wrapper. It will determine if an element in the wrapper `looks like` the expected element by checking if all props of the expected element are present on the wrappers element and equals to each other.
```
ljharb: "looks like" here should be non curly quotes, not backticks, since it's not code

mathieuancelin:
@ljharb yes, @Aweary is right, these function just check props that are present on the elements passed into `rightContains`/`rightEquals` so you can check `style` or `onClick` if you want to, or you can just ignore them because it's not relevant for you test

docs/api/ShallowWrapper/rightContains.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+#### Arguments
+
+1. `nodeOrNodes` (`ReactElement|Array<ReactElement>`): The node or array of nodes whose presence you are detecting in the current instance's
```
ljharb: can we be very very explicit about whether this has `some` or `every` semantics?

I almost would prefer three separate methods: `containsMatchingElement`, `containsAllMatchingElements`, and `containsAnyMatchingElements` - taking a node, an array, and an array, respectively - rather than risking any confusion about what conditions cause `true` or `false` to be returned. Thoughts?

src/Utils.js
```diff
@@ -10,6 +10,10 @@ import {
   REACT15,
 } from './version';
 
+function equality(a, b) {
```
ljharb: this doesn't work properly for `NaN` or `-0` - we should pull in https://www.npmjs.com/package/object-is for this.

mathieuancelin: Ok, I will. 

Actually it's just a rewrite of the actual code (located [here](https://github.com/airbnb/enzyme/blob/master/src/Utils.js#L98))

ljharb: sure, comparing lengths will always work with `===` :-) but a generic function named "equality" should probably be actually generic.

docs/api/ShallowWrapper/rightContains.md
Aweary: If you're going for emphasis, try italics 😄

docs/api/ShallowWrapper/rightContains.md
mathieuancelin: It actually just work like `wrapper.contains(nodeOrNodes)` except the "source of truth" is inverted. It's not A.contains(B) anymore but B.contains(A) instead.

src/Utils.js
mathieuancelin: Ok, no problem for me :-)

docs/api/ShallowWrapper/rightContains.md
ljharb: I don't think that's accurate. `shallow(<A a><B b /></A>).containsMatchingElement(<B />)` isn't the same thing as `shallow(<B />).contains(<A a><B b /></A>)`.

I see this as simply "contains", but using the element-to-be-found's prop list rather than the element-being-searched.

mathieuancelin: > I see this as simply "contains", but using the element-to-be-found's prop list rather than the element-being-searched.

Yes, that's what I was trying to explain :-)

mathieuancelin:
Hey @ljharb and @Aweary, 

I've just pushed some new code with the modifications that you suggested.
Let me know if something's wrong.

ljharb:
Nothing immediately jumps out at me. Is this perhaps something we also want exposed on `mount`? Both mount and shallow's wrappers have `contains`.

mathieuancelin:
Yeah, I thought exactly the same thing. I will take a look at `mount` today

mathieuancelin:
@ljharb I just added the contains* function to the `ReactWrapper` with some tests and the documentation

docs/README.md
```diff
@@ -39,6 +42,7 @@
     * [key()](/docs/api/ShallowWrapper/key.md)
     * [last()](/docs/api/ShallowWrapper/last.md)
     * [map(fn)](/docs/api/ShallowWrapper/map.md)
+    * [matchesElement(node)](/docs/api/ShallowWrapper/matchesElement.md)
```
ljharb: we still need `matchesElement` on `ReactWrapper`, yes?

mathieuancelin: Oooops, done !

mathieuancelin:
@ljharb it should be good now :-)

Aweary:
@lelandrichardson @ljharb do you think we can get this merged soon?

ljharb:
@Aweary it needs to be rebased on latest master :-)

Aweary:
Yeah, I meant more in terms of whether you guys will be accepting it. If that's the case then @mathieuancelin it'd be great if you could rebase 😄

mathieuancelin:
@Aweary it's rebased, just waiting on CI :-)

Aweary:
Maybe it's just me, but it says it's still out-of-date with master 😬

mathieuancelin:
@Aweary it seems good now, but travis seems a bit overloaded at the moment

Aweary:
I believe @ljharb is going to want that last merge commit (3394025) to be squashed as well

mathieuancelin:
@Aweary It should be okay now. I don't even where does this merge come from. And that's why I should never try to update a PR before going to sleep :-)

lelandrichardson:
I had considered a similar API before, but never ended up writing it. I think this is quite useful in practice, though it always feels a little messy to me in my head.  I think it's a reasonable addition though.

What do we think about the names `fuzzyContains` and `fuzzyEquals`? cc @ljharb

ljharb:
I like the methods but I don't like the name "fuzzy" - the current names were my suggestion and atm I still find them to be the clearest.

lelandrichardson:
The problem I have with the current names is that they don't seem to indicate to me that they are not "strict".  using the "fuzzy" prefix makes it obvious that this is not strict equality.

ljharb:
To me, "matching" is a non-strict thing - since in every test framework, "to match" generally means "is kind of the same as". Another common word to describe non-strict-equality is "loose equality". "Fuzzy" might imply that `{ foo: 3 }` and `{ Foo: 3 }` are the same - because you don't know to what depth "fuzzy" refers to.

lelandrichardson:
Considering test libraries use the `match` nomenclature, i'll go ahead and agree with the current naming :)

OK. Let's do this! We've got some fresh conflicts. Mind rebasing and fixing? Thanks!

mathieuancelin:
@lelandrichardson it should be good now :-)

mathieuancelin:
Thanks @lelandrichardson :-)

PR #368, opened 2016-05-06
[RFC] shallow wrapper .simulate calls event handlers with SyntheticEvent

nfcampos:
What do you guys think about this? I just added an empty React SyntheticEvent as the first argument that `.simulate` calls event handlers with.
This provides the regular event methods `stopPropagation` an friends by default but still allows the user to replace them eg. with spies, exactly the same way they do now.
The only failing tests are the assertions about which arguments an event handler is called with (obviously ;)


nfcampos:
This is not a complete solution to all the shallow wrapper's event shortcomings (still no propagation) but to me it looks like a change that should be compatible with whatever ends up being done for propagation, since all it does is call event handlers with an object with an event-like interface

nfcampos:
I've added a POC for event propagation :) let me know what you think

src/ShallowWrapper.js
```diff
@@ -346,16 +347,22 @@ export default class ShallowWrapper {
...
-      withSetStateAllowed(() => {
-        // TODO(lmr): create/use synthetic events
-        // TODO(lmr): emulate React's event propagation
-        handler(...args);
-        this.root.update();
-      });
-    }
+  simulate(event, mock, ...args) {
+    const eventProp = propFromEvent(event);
+    const e = Object.assign(new SyntheticEvent(undefined, undefined, { type: event }), mock);
```
nfcampos: if the mock replaces the `stopPropagation` method event propagation stops working as intended so this `Object.assign` is probably not a great idea

src/ShallowWrapper.js
```diff
@@ -26,6 +26,7 @@ import {
   createShallowRenderer,
   renderToStaticMarkup,
 } from './react-compat';
+import SyntheticEvent from 'react/lib/SyntheticEvent';
```
Aweary: This should likely be re-exported from `react-compat`, not directly imported from `react/lib`.

nfcampos: yeah i realised that afterwards, ill change it if this goes forward

src/ShallowWrapper.js
```diff
@@ -346,16 +347,22 @@ export default class ShallowWrapper {
...
-        // TODO(lmr): emulate React's event propagation
-        handler(...args);
-        this.root.update();
-      });
-    }
+  simulate(event, mock, ...args) {
+    const eventProp = propFromEvent(event);
+    const e = Object.assign(new SyntheticEvent(undefined, undefined, { type: event }), mock);
+    withSetStateAllowed(() => {
+      let current = this;
```
Aweary: Why is this necessary? `do` blocks do not create a new scope.

nfcampos: thats the variable that keeps getting updated as i traverse the tree from the event's target up until the root, i should probbaly make this a for loop to make it clearer

lelandrichardson:
@nfcampos thanks for this contribution!  I think that this is on the right track.  I have a couple of things though:

1. As Aweary mentioned, let's pull this inside of `react-compat`
2. If user provides explicit `null` as first argument to simulate, no event object ends up getting passed into the event handler.

There are a couple of questions this raises:

1. Is this a breaking change? this will likely make a lot of code using `simulate` follow new code paths that could cause breakage.
2. There is now no way to simulate a custom event (ie, `onCustomThingHappens`) where the data being passed into that is not supposed to be an event object, but rather some data object that is domain specific. How do we preserve the ability to simulate events and pass in arbitrary data structures instead of synthetic events?

Unfortunately my first `.simulate()` API wasn't forward-thinking enough to forecast these backwards compatibility issues.

@ljharb do you have any thoughts?

My thinking is that this will have to be a breaking change, and that perhaps we need to provide an additional way for users to specify *not* using the synthetic event.

nfcampos:
@lelandrichardson I've pulled the import of SyntheticEvent into `react-compat`.

I think there's no way around this being a breaking change, especially because of the event propagation this introduces, it will simply do stuff in your tests that it didn't do before. 

A couple questions:
- does it make sense for the `.simulate` method to support custom events? We at least need to make it clear to the user that even custom events will follow the propagation logic
- In the spirit of making the shallow wrapper have an api as similar to the full rendering wrapper as possible, I don't think it makes sense for the `.simulate` method to:
  1. support custom events (the full rendering wrapper doesn't https://github.com/airbnb/enzyme/blob/master/src/ReactWrapper.js#L384)
  2. support arbitrary event data (the full rendering wrapper doesn't https://github.com/facebook/react/blob/master/src/test/ReactTestUtils.js#L522 the passed in arg is Object.assign'ed into the SyntheticEvent)
  3. support more arguments beyond the first (the full rendering wrapper doesn't https://github.com/facebook/react/blob/master/src/test/ReactTestUtils.js#L493 , this is in fact a mistake in  our wrapper around the TestUtils simulate, we send more arguments which are ignored, https://github.com/airbnb/enzyme/blob/master/src/ReactWrapper.js#L387)
- like the full rendering `.simulate` the shallow simulate should be guarded with `this.single` to prevent you simulating an event on more than one node at a time (you can always do that explicitly with `forEach`, but we shouldn't magically 'dispatch' several events on a single call to `.simulate`, do you agree?

nfcampos:
another thing, I think there should be a command in `package.json`  to run tests without jsdom, to make sure `shallow` doesn't depend on `window` being available. Ideally each test file that needs jsdom would import it but since mocha runs all test files in the same process it would end up being available in all the tests...

src/react-compat.js
```diff
@@ -134,6 +134,8 @@ if (REACT013) {
   };
 }
 
+import SyntheticEvent from 'react/lib/SyntheticEvent';
```
lelandrichardson: nit: please move this to the top of the file. `import` statements get executed before anything else when transpiled by babel anyway (and in the spec), so putting this here can be misleading.

test/ShallowWrapper-spec.js
```diff
@@ -860,6 +860,47 @@ describe('shallow', () => {
...
+    it('should respect event.stopPropagation()', () => {
+      const innerOnClick = sinon.spy(e => {
+        e.stopPropagation();
+      });
+      const outerOnClick = sinon.spy();
+      class Foo extends React.Component {
+        render() {
+          return (
+            <div onClick={outerOnClick}>
+              <a onClick={innerOnClick}>foo</a>
```
lelandrichardson: some good additional test cases for propogation might be:

1. ensure that it can skip over parent nodes that don't have `onClick` defined, etc.
2. ensure that it propogates beyond n=2 (ie, parent + grandparent).

nfcampos: I've added those to the list of missing tests above

lelandrichardson:
@nfcampos all good points.  I think you are correct, the only way out of this is to create another method that handles non-standard event simulation.

To reiterate your points, I think we should move forward with the following guides:

1. `.simulate(event, mock)` should have semantically equivalent behavior for both `shallow` and `mount` (to the extent possible, consider shallow is only a shallow tree).
2. An additional method (how about `invoke(name, ...args)`?) will be provided that will allow for custom events, and *not* handle propogation

Further questions i have:

1. what about the capture phase? do we want to simulate `onClickCapture`?
2. If i have a `<Button />` component with an `onClick` handler, are we still allowing for `button.simulate('click')`?

Note: I also agree that this should be wrapped inside of `this.single` 👍

src/react-compat.js
nfcampos: I agree. I've moved it to the top

nfcampos:
> If i have a <Button /> component with an onClick handler, are we still allowing for button.simulate('click')?
@lelandrichardson is this https://github.com/airbnb/enzyme/pull/368/commits/b4014e002314a1179b326e14d8c4bc5ba4e4fff1 what you mean? If yes, yeah that still works.

nfcampos:
> what about the capture phase? do we want to simulate onClickCapture?

If you think it's worth it, I can add that, it's doable, do you think it's worth it?

lelandrichardson:
> If you think it's worth it, I can add that, it's doable, do you think it's worth it?

I think that if we decide we need it later that it will need to be another breaking change, so I think it's worth it for us to get it right this time.

nfcampos:
Yeah I tend to agree. Then there is one other thing we should think about now: events that have exotic behaviour (I'm not sure what to call it) eg. calling `click` on a `<button type='submit' />` will call `submit` on the parent `form`. Should we also support this? Are there other behaviours like this? I'm only aware of this one

lelandrichardson:
^ I Think @ljharb will have some thoughts on this.

The main thing we are going to face is that it is extremely hard to emulate this behavior reliably on a shallow wrapper because by design we are looking at an incomplete picture of the DOM.

Thinking about this further, we are realistically going to have the same problem with click/capture event phases because we have no way of knowing if a parent component ends up preventing the capture phase or not.

nfcampos:
Hmm, re: capturing I think it's doable, think of eg. this tree
```js
<div onClick={e => e.stopPropagation()}>
  <a onClick={innerOnClick}>foo</a>
</div>
```
If `click` on `a` is simulated in capture mode we first traverse the whole tree from the target, `a`, up until the root, in this case only `a` then `div`, then call the handlers on each (if they exist) in the opposite order, starting with the `div`. Since the `div` stops propagation then we bail out of further calls and never call `innerOnClick`.
Am I missing something?

Obviously we only know whether a parent node that you actually rendered stops propagation or not, when you end up using the component you're testing in your app it might be inside a component that stops propagation. But that's the whole point of unit testing components individually, no?

lelandrichardson:
Yes the capturing will work so long as you're not worries about effects from outside your component. This is reasonable, but is definitely not the whole picture. This is an instance where there is a sort of implicit public contract of the component that we are not able to test. In reality we could provide an API to capture events at the root node of the wrapper, and then the tester could mock out the "response" of the parent component, which could be preventing the capture phase.

These are definitely corner cases, but perhaps something we should consider nonetheless.

nfcampos:
So this is in a way similar to context in which a parent component changes the behaviour of its children implicitly (ie. through something other than props), right?
Any other instance of that sort of situation?
I'm thinking the clearest way to test stuff like this would be to able to specify a root component to render the component you're testing inside of. Such a component could provide context, or register capture event handlers or do whatever else might change the behaviour of the component you're trying to test.

lelandrichardson:
Yeah, although props and context are somewhat similar (props are just less explicit) in that they are top => down.  Event bubbling is in the opposite direction and seems trickier to formulate an API that makes sense around it.

nfcampos:
Bubbling is not that hard to test, the only way an outside component can mess with a component's behaviour through event bubbling is if that component accepts `children` and the children stop event propagation. But you're already free to test components with whatever children you may like, it's just another prop.
Capturing is the difficult one, because it's implicit, it all happens outside the component you're testing. I'm not sure what the best thing for that is. Allowing a custom root component to render your component in allows you to test whatever you want but makes tests more verbose, you have to create a component that does whatever it is you're testing.

lelandrichardson:
The API I was thinking would be something like:

```
const click = spy();
const capture = spy();
const Foo = () => (<div><a onClick={click} onClickCapture={capture} /></div>);
// register a listener for events bubbling up, (or down in the case of the capture phase)
wrapper.on('click', e => { e.stopPropogation(); });
wrapper.find('a').simulate('click');
expect(click).to.have.callCount(1);
expect(capture).to.have.callCount(0);
```

`.on()` might be a bit too general of a name for this API, but maybe something like `onEventBubble('click', handler)` or something.

nfcampos:
@lelandrichardson actually in that example I think neither `click` nor `capture` would get called, stopping propagation on the capture phase skips the bubbling phase as a whole, since it happens before. Correct if I'm wrong, but this `.on` method would be useful only for registering capture handlers, right? seeing as it registers a root handler for the event, a root bubble phase handler would be useless, it would fire after all other event handlers...

lelandrichardson:
@nfcampos hmm. Maybe i'm misunderstanding the bubbling behavior. My impression was that the event bubbled all the way to the root (the real root, ie `document`), and then would bubble back down in reverse order with the `Capture` phase. I was under the impression that if `stopPropogation` was called at any time, it would stop the process.

nfcampos:
I believe that's exactly what happens but the capture phase happens before the bubble phase so, with following tree, `div` > `span` > `a`, the order of calling the event handlers is: div.onClickCapture. span.onClickCapture, a.onClickCapture, a.onClick,span.onClick, div.onClick, stooping at any point if stopPropagation is called, as you said

lelandrichardson:
@nfcampos whoops. that was my mistake. I flipped the order of capture versus not. whoops :P

nfcampos:
@lelandrichardson i've just added the capture phase :) still missing a bunch of tests though

nfcampos:
(the build fails in node 0.10 because there's no `Symbol`. I guess I should replace the for-of loop with a regular for loop, right?)

ljharb:
@nfcampos even better, no loop at all - please use `Object.keys(handlers).forEach` here, or similar.

nfcampos:
but how do I break out of a `forEach`? (when the propagation is stopped)

ljharb:
@nfcampos ah, in that case, `Object.keys(handlers).some(…)` and `return true` when you want to stop the iteration

nfcampos:
done

test/parity-spec.js
```diff
@@ -0,0 +1,55 @@
+import { describeWithDOM } from './_helpers';
```
nfcampos: what do you guys think about a test like this? this effectively tests that the API of methods on `shallow` is the same as the API of the same methods on `mount`, which seems like an important thing to test explicitly

Aweary: I think this is an important thing to test for any API method that should have feature parity between `mount` and `shallow` 👍

nfcampos:
should I add this check to `shallow::simulate` as it exists in `mount::simulate`?
```
const eventFn = Simulate[mappedEvent];
      if (!eventFn) {
        throw new TypeError(`ReactWrapper::simulate() event '${event}' does not exist`);
      }
```
ie. should one of the `simulate` methods allow you to simulate events that the other one doesn't?

nfcampos:
I've written all the tests (it's missing the tests for functional components) I wanted to write for this, can someone have a look around and see if there is anything else I should be testing?

Aweary:
@nfcampos I think we'd want this before doing that: 
>An additional method (how about invoke(name, ...args)?) will be provided that will allow for custom events, and not handle propogation

Maybe we could implement something like `invoke` and then add a deprecating warning instead if its a custom event, like:

```js
if (!eventFn) {
  console.warn(
    'ShallowWrapper::simulate() event '${event}' doesnt exist. Use invoke() if youre trying to simulate a custom event.
);
return invoke(args)
}

**Comment by nfcampos** [comment]
@Aweary yeah makes sense to do that first

**Comment by nfcampos** [comment]
@lelandrichardson should I add the `invoke` method in this PR as well? other than that I think this is ready.

**Comment by Aweary** [comment]
@nfcampos I think that since this PR is what makes `invoke` necessary that it's fine to add here (IMO)

**Comment by nfcampos** [comment]
I've added the `invoke` method to `ShallowWrapper`. This should now be complete, I believe.

**Comment by Aweary** [comment]
@nfcampos should we also add the check for the `eventFn` and call `invoke` with a warning--like we talked about before:

```js
if (!eventFn) {
  console.warn(
    `ShallowWrapper::simulate() event '${event}' doesnt exist. Use invoke() if
     youre trying to simulate a custom event.`
);
return invoke(args)
}
```

Maybe I missed it, I didn't see that there

nfcampos:
@Aweary I haven't added that. I'm reconsidering whether that warning should really exist. Does it make sense?

Aweary:
@nfcampos @ljharb any blockers here?

ljharb:
I'll take another look shortly; either way let's hold off merging in this breaking change (and the others queued up) until we've merged in all the non-majors and cut a release.

docs/api/ShallowWrapper/invoke.md
```diff
@@ -0,0 +1,51 @@
...
+class Foo extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { count: 0 };
+  }
+  render() {
+    const { count } = this.state;
+    return (
+      <div>
+        <div className={`clicks-${count}`}>
```
ljharb: not actually a blocker ofc, but i'd prefer to use a data attribute here, so that the example follows best practices :-)

nfcampos: You're right, I've changed it

blainekasten:
> I'll take another look shortly; either way let's hold off merging in this breaking change (and the others queued up) until we've merged in all the non-majors and cut a release.

Non-majors keep coming in. Maybe we make a project of what to batch in the next major release so we can actually act on this.

nfcampos:
@ljharb I was trying to rebase this and it seems with react 15.5 SyntheticEvent is no longer importable from react, so I'm not sure where to go from here.

ljharb:
@nfcampos they likely moved it to another package? if not, then we should file an issue on react asking for it to be exported.

nfcampos:
@ljharb thinking about this some more, I don't think we want to start depending on SyntheticEvent, seeing as that is React-specific and we're going in the opposite direction in v3. What do you suggest we do with this, drop it or somehow not use SyntheticEvent?

ljharb:
I've come to believe that `.simulate` is an inherently problematic API, and shouldn't be used at all - so I'm not sure it's worth enhancing it.

nfcampos:
Why problematic?

ljharb:
It conveys a false promise - it doesn't actually simulate anything. There's not necessarily an event, there's not necessarily bubbling, there's not any chain of events (like if you simulate a change, you don't get all the interim key events).

Whereas if you use `.prop('foo')()`, you explicitly invoke a function, and you know what arguments you're passing it.

PR #371, opened 2016-05-08
Test on node v6

SimenB:
Io.js is not maintained anymore, and v6 is the latest

(example of io.js not maintained: https://nodejs.org/en/blog/vulnerability/openssl-may-2016/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 node_js:
+  - "6"
   - "5"
   - "4"
-  - "iojs"
   - "0.12"
   - "0.10"
 before_install:
@@ -26,6 +26,16 @@ sudo: false
 matrix:
   fast_finish: true
   exclude:
+  - node_js: "5"
+    env: EXAMPLE=mocha
+  - node_js: "5"
+    env: EXAMPLE=karma
+  - node_js: "5"
+    env: EXAMPLE=react-native
+  - node_js: "5"
+    env: EXAMPLE=karma-webpack
+  - node_js: "5"
+    env: EXAMPLE=jest
   - node_js: "4"
     env: EXAMPLE=mocha
   - node_js: "4"


lelandrichardson:
cc @ljharb 
this seems fine to me. If you're good with it go ahead and merge.

.travis.yml
```diff
@@ -26,6 +26,16 @@ sudo: false
 matrix:
   fast_finish: true
   exclude:
+  - node_js: "5"
```
ljharb: if we're going to add these, we should also add them for node 6, and also remove the iojs ones.

ljharb:
fwiw, "lack of official support" is not a good reason imo to stop supporting a platform. 0.10 should remain tested long after its official maintenance period has ended in October.

However, since this is a testing lib, it's not very critical that it work in io.js, so I'm OK removing it from the travis config.

.travis.yml
lelandrichardson: @ljharb to be clear, these are *exclusions*, so this is making node 6 the only platform in the matrix that has the example projects run.

ljharb: oops, my bad - yeah this makes sense then

nfcampos:
btw, never remove 0.10 (or 0.12 I'm not sure which) since the fact it doesn't support jsdom is what makes the travis build as a whole fail when the shallow wrapper erroneously ends up depending on the presence of `window`, or better yet we should be testing that explicitly ;)

lelandrichardson:
@nfcampos For each node version, we run the test suite both with and without jsdom, so running without shallow relying on `window` or `document` is reliably tested on every node version, just FYI.

nfcampos:
Are you sure? see this travis build, I purposefully undid a previous commit that fixed react 13 trying to access `window` when creating a SyntheticEvent, the tests will pass (it hans't finished yet) everywhere other than 0.10 where jsdom doesn't run https://travis-ci.org/airbnb/enzyme/builds/128727535

lelandrichardson:
@nfcampos looks like you might be right...  I'll double check to make sure. If so, i'll adjust travis config to run an npm script that does both, because we should definitely be testing for that.

Thanks for surfacing!

PR #373, opened 2016-05-09
added error checking for state, setState, context, instance methods on ShallowWrapper

nfcampos:
- `.instance` was missing the `root` check even though it was mentioned in the description above.
- `state` and `setState` don't make sense for functional components so they now throw if called on a functional component (error message like the `root` one)
- also added the existing error checking of `setContext` to `context`


lelandrichardson:
Awesome!

PR #381, opened 2016-05-11
Fix mount::simulate() signature

nfcampos:
The signature was wrong because the underlying function from React TestUtils takes only 2 arguments.
There was no test to fix because we were testing only the first argument.

### DIFF
--- a/src/ReactWrapper.js
+++ b/src/ReactWrapper.js
@@ -468,7 +468,7 @@ export default class ReactWrapper {
    * testing events should be met with some skepticism.
    *
    * @param {String} event
-   * @param {Array} args
+   * @param {Object} mock (optional)
    * @returns {ReactWrapper}
    */
   simulate(event, mock = {}) {
--- a/src/ReactWrapper.js
+++ b/src/ReactWrapper.js
@@ -471,7 +471,7 @@ export default class ReactWrapper {
    * @param {Array} args
    * @returns {ReactWrapper}
    */
-  simulate(event, ...args) {
+  simulate(event, mock) {
     this.single(n => {
       const mappedEvent = mapNativeEventNames(event);
       const eventFn = Simulate[mappedEvent];
@@ -479,7 +479,7 @@ export default class ReactWrapper {
         throw new TypeError(`ReactWrapper::simulate() event '${event}' does not exist`);
       }
 
-      eventFn(findDOMNode(n), ...args);
+      eventFn(findDOMNode(n), mock);
     });
     return this;
   }
--- a/src/ReactWrapper.js
+++ b/src/ReactWrapper.js
@@ -471,7 +471,7 @@ export default class ReactWrapper {
    * @param {Array} args
    * @returns {ReactWrapper}
    */
-  simulate(event, mock) {
+  simulate(event, mock = {}) {
     this.single(n => {
       const mappedEvent = mapNativeEventNames(event);
       const eventFn = Simulate[mappedEvent];
--- a/docs/api/ReactWrapper/simulate.md
+++ b/docs/api/ReactWrapper/simulate.md
@@ -1,4 +1,4 @@
-# `.simulate(event[, ...args]) => Self`
+# `.simulate(event[, mock]) => Self`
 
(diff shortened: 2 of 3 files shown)


src/ReactWrapper.js
```diff
@@ -471,15 +471,15 @@ export default class ReactWrapper {
    * @param {Array} args
    * @returns {ReactWrapper}
    */
-  simulate(event, ...args) {
+  simulate(event, mock) {
```
ljharb: since the mock is optional, it should probably explicitly default to a value - that way `simulate`'s `length` won't change.

nfcampos: to default to a value let's have it default it to `{}`?

nfcampos: this is its only use https://github.com/facebook/react/blob/master/src/test/ReactTestUtils.js#L583 (in react v15) so I guess `undefined`, `null` or `{}` are all harmless, which one do you like best?

ljharb: probably `{}`, since it's eventually `assign`ed?

nfcampos: done

Aweary:
LGTM

Aweary:
@ljharb I'd merge this but it seems I don't have write access 😮

ljharb:
@Aweary hm, you should definitely have write access, but you won't be able to merge it until 1) it's been LGTM'd after the last commit (LGTM, fwiw) and 2) the travis tests have passed.

nfcampos:
@ljharb same here, I can't merge either #354 or #342 even though both checks have passed

ljharb:
@nfcampos both are out of date with master, so the user has to rebase before they can be merged

nfcampos:
hmm, but that panel at the bottom for me looks like this 
screenshot 2016-05-12 08 47 55
which is exactly what the panel for this PR looks like for me, and this one is up to date with master

ljharb:
@nfcampos hm, that's super weird. I'll play with the protected branch settings. Are you capable of merging this one?

ljharb:
@nfcampos ah wait, nvm, i think i fixed the setting. i'll clean up our chatter on this PR now :-)

src/ReactWrapper.js
```diff
@@ -468,18 +468,18 @@ export default class ReactWrapper {
...
+  simulate(event, mock = {}) {
     this.single(n => {
       const mappedEvent = mapNativeEventNames(event);
       const eventFn = Simulate[mappedEvent];
       if (!eventFn) {
         throw new TypeError(`ReactWrapper::simulate() event '${event}' does not exist`);
       }
 
-      eventFn(findDOMNode(n), ...args);
+      eventFn(findDOMNode(n), mock);
```
blainekasten: I might be overthinking this. But is there a chance this broke events that were dependent on second and third React arguments?

React usually sends a couple arguments to a function:
```handler(syntheticEvent, ...args, reactId, nativeEvent)```

So if anyone was depending on the `reactId` argument or the last `nativeEvent` argument we just broke their tests. I think a simple fix would be adding the `...args` back in after the mock so people can specify those events if they need to.

Comment (Aweary) [inline src/ReactWrapper.js]: @blainekasten `makeSimulator` returns a function that takes two arguments

```js
function makeSimulator(eventType) {
  return function(domComponentOrNode, eventData) {
  ...
```

[src/test/ReactTestUtils.js#L497](https://github.com/facebook/react/blob/ba9b985406ff56c2dea192b119328b6096895097/src/test/ReactTestUtils.js#L497)

I'll have to check previous releases of React to see if it's any different

nfcampos: @blainekasten I think if you use `mount::simulate()` on the currently released version you'll find that any argument after the first never finds its way to the event handlers, that's just the way the `Simulate` method from react TestUtils is written

blainekasten: If that's how React TestUtils does it, it's probably fine. It'd definitely be an edge case where a user does need those second/third arguments. So we can wait until somebody brings up an issue needing it.

Issue #314, opened 2016-04-11
Id selector not working when combined with a tag selector


rouanw:
I've been trying to get the following unit test code to work:

```js
const component = React.createElement(MyComponent, { });
const wrapper = shallow(component);
expect(wrapper.find('input#myid').length).to.equal(1);
```

But the `find` method is not finding anything. It is able to find the input box just fine with `input` as the selector and with `#myid` as the selector, but not when I specify both.

I'm using Enzyme v2.2.0.

blainekasten:
FWIW, using id's is an anti-pattern in react since it breaks re-usability with w3c DOM validation.
That said, this does appear to a be a bug.

cc @lelandrichardson

rouanw:
Thanks @blainekasten.

havenchyk:
The fix is merged two months ago, can it be closed?

Aweary:
@havenchyk yupp, thanks!

rixth:
@Aweary this caused a regression. It broke attribute selectors with values containing `#, ex `a[href='/page#anchor']`.

PR #387, opened 2016-05-12
Id selector not working when combined with a tag selector (#314)

nfpiche:
For issue #314


Aweary:
Seems reasonable enough, LGTM

PR #392, opened 2016-05-13
add event to simulator

eUstundag:
I noticed a few places 'onBlur' is utilized in our components, figured this would eventually help someone!

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -241,6 +241,7 @@ export function coercePropValue(propName, propValue) {
 
 export function mapNativeEventNames(event) {
   const nativeToReactEventMap = {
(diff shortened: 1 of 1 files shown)


nfcampos:
Thanks for the PR @eUstundag! I don't think this is necessary though, the purpose of this map is to map lowercase event names to their camel case equivalent, for `blur` they're the same. If you look at the return statement at the end of this function you'll see that events not found in the map are just returned as is https://github.com/airbnb/enzyme/blob/master/src/Utils.js#L286

eUstundag:
my mistake! sorry about that.

nfcampos:
No problem, thanks anyway!

Issue #261, opened 2016-03-15
Beef up tests for Stateless Functional Components


lelandrichardson:
Enzyme was originally written to be compatible with React 0.13, before Stateless Functional components were a thing. It really seems like we've been getting a lot of PRs and Issues highlighting the incompatibility of SFCs with enzyme.

I love SFCs - so i'm hoping we can change this. i think the first step might be to increase the number of tests we have that use SFCs.

If anyone is interested in collaborating on enzyme, this is some relatively straight-forward work that I'd be hugely grateful for!

Help us make enzyme more stable!

gm758:
Recently started using enzyme for my first react native app. Looking forward to taking a look at this issue.

Andarist:
I think I could help a little bit, especially that I am using SFC all over the place and it would be nice to have a possibility to test them correctly ;)

Although I would need some guidance on what I can work on and maybe what would be the best approach to solving particular task.

tf:
Related #262

lelandrichardson:
One approach that I've considered, though it isn't very DRY, is to literally just go through each test for mount/shallow and duplicate it if it's using a component and use an SFC instead.

Might be a better approach than that though...

ljharb:
I find that sometimes it's nicer when tests do repeat themselves - that's probably the simplest approach here.

lencioni:
> I find that sometimes it's nicer when tests do repeat themselves

+1 to this. Tests should prioritize ease of understanding any given example to make it easier on folks who run into failing specs and need to understand what is going on.

gm758:
I have finished adding tests for shallow, but have encountered issues writing tests for 'mount' (one of these issues is mentioned in #45). From what I can tell, stateless function components do not play well with findDOMNode. This means that calling .text() on the wrapper of an SFC will not work. Additionally, trying to get an SFC with .find() does not work either.

I would appreciate some guidance on how to proceed. Should I write failing tests that attempt to use these functionalities (so that they can be fleshed out in future releases), or should I use a work arounds like in the example below?

`
const InnerComponent = () => (
  <div className="foo">foobar</div>
);
const OuterComponent = () => (
  <div className="bar"><InnerComponent /></div>
)
const innerWrapper = mount(<InnerComponent />);
const outerWrapper = mount(<OuterComponent />);
`

Will not work as expected:
`
innerWrapper.text();
outerWrapper.find(InnerComponent);
`

Will work:
`
innerWrapper.find('.foo').text()
[...]

PR #394, opened 2016-05-15
SFC tests for mount + more

gm758:
This PR should complete the task of duplicating all relevant class-based component tests to also include stateless function components.

Happy to squash commits if everything looks good.

Related to issue #261


package.json
```diff
@@ -17,7 +17,7 @@
     "test:single": "mocha --require withDom.js --compilers js:babel-core/register --watch --reporter dot",
     "test:watch": "mocha --require withDom.js --compilers js:babel-core/register --recursive test/*.js --watch --reporter dot",
     "test:env": "sh ./example-test.sh",
-    "test:all": "npm run react:13 && npm test && npm run react:14 && npm test && npm run react:15 && npm test",
+    "test:all": "npm run lint && npm run react:13 && npm run test:only && npm run react:14 && npm run test:only && npm run react:15 && npm run test:only",
```
ljharb: for a nicer hook, let's run `npm run lint` in "pretest" instead of `test`, and then here, we can `npm run pretest` everywhere we currently run `npm run lint` (that's not "test")?

ljharb:
Pending my one comment, this LGTM. Thanks for doing this work!

package.json
gm758: Just to clarify, we'd have something like:
```
"pretest": "npm run lint",
"test": "npm run test:only",
"check": "npm run test:all", 
"test:all": "npm run pretest && npm run react:13 && npm run test:only && npm run react:14 && npm run test:only && npm run react:15 && npm run test:only"
```

I'm not clear on what purpose `check` serves. Before my changes, it was just running lint a fourth time on top of the three times `test:all` ran lint via 3 calls to `test`.

ljharb: yeah, you're right - perhaps reverting your changes, and instead only changing the `npm test` commands on this line to be `npm run test:only`, would be best?

nfcampos:
This looks great @gm758 👍

gm758:
@ljharb How's that look? `pretest` gets rid of the need for `npm run lint` in `test`, `check` already runs the linter, so no need to run it in `test:all`

ljharb:
@gm758 looks great, thanks! It'll just need a fresh rebase on top of master :-)

gm758:
Should be good to go

Aweary:
Thanks for the great work @gm758!

PR #399, opened 2016-05-18
Update Jest docs.

cpojer:
The moduleFileExtensions option is to list the extensions Jest looks at. If you overwrite it to `json` it will not load `js` files and will not work. See https://github.com/facebook/jest/issues/1014#issuecomment-219900877

Also updated the paths to work better (because they are paths, not module names).


lelandrichardson:
LGTM

blainekasten:
Thanks @cpojer!

PR #412, opened 2016-05-25
Support additional characters in attribute selectors

jordwest:
This change allows `.find('[data-attr1]')` and `.find('[data-attr_foo]')` to match as expected.

According to the HTML5 data-* spec, digits are allowed in data-* attributes.

Additionally, this prevents matching of invalid attributes, such as `:foo`, `-foo`, or `123-foo`,

### DIFF
--- a/test/ReactWrapper-spec.js
+++ b/test/ReactWrapper-spec.js
@@ -369,12 +369,13 @@ describeWithDOM('mount', () => {
     it('should support data prop selectors', () => {
       const wrapper = mount(
         <div>
-          <span data-foo="bar" />
+          <span data-foo="bar" data-foo123="bar" />
         </div>
       );
 
       expect(wrapper.find('[data-foo="bar"]')).to.have.length(1);
       expect(wrapper.find('[data-foo]')).to.have.length(1);
+      expect(wrapper.find('[data-foo123]')).to.have.length(1);
     });
 
     it('should find components with multiple matching props', () => {
--- a/test/ReactWrapper-spec.js
+++ b/test/ReactWrapper-spec.js
@@ -387,6 +387,7 @@ describeWithDOM('mount', () => {
           <span data-foo="bar" />
           <span data-foo-123="bar2" />
           <span data-123-foo="bar3" />
+          <span data-foo_bar="bar4" />
         </div>
       );
 
@@ -398,6 +399,9 @@ describeWithDOM('mount', () => {
 
       expect(wrapper.find('[data-123-foo]')).to.have.length(1);
       expect(wrapper.find('[data-123-foo="bar3"]')).to.have.length(1);
+
+      expect(wrapper.find('[data-foo_bar]')).to.have.length(1);
+      expect(wrapper.find('[data-foo_bar="bar4"]')).to.have.length(1);
     });
 
     it('should find components with multiple matching props', () => {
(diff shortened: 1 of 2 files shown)


test/ReactWrapper-spec.js
```diff
@@ -369,12 +369,13 @@ describeWithDOM('mount', () => {
...
       const wrapper = mount(
         <div>
-          <span data-foo="bar" />
+          <span data-foo="bar" data-foo123="bar" />
         </div>
       );
 
       expect(wrapper.find('[data-foo="bar"]')).to.have.length(1);
       expect(wrapper.find('[data-foo]')).to.have.length(1);
+      expect(wrapper.find('[data-foo123]')).to.have.length(1);
```
ljharb: can you add another test for attributes that start with a digit? I'm pretty sure `123-abc` isn't a valid attribute (but either way we should have a test for it)

jordwest: Sure thing, is that `[data-123-abc]`? Or just `[123-abc]`?

ljharb: `data-123-abc` should be valid, i think `123-abc` should be invalid.

jordwest:
I've added tests to ensure that `data-123-foo` is valid, and that the following are invalid:
 - `123-foo`
 - `-foo`
 - `:foo`

The last two were failing in master, but a change to the regexp now ensures the first character of an attribute is always an alphabet character.

test/ReactWrapper-spec.js
```diff
@@ -366,15 +366,34 @@ describeWithDOM('mount', () => {
...
           <span data-foo="bar" />
+          <span data-foo-123="bar2" />
+          <span data-123-foo="bar3" />
         </div>
       );
 
       expect(wrapper.find('[data-foo="bar"]')).to.have.length(1);
       expect(wrapper.find('[data-foo]')).to.have.length(1);
+      expect(wrapper.find('[data-foo-123]')).to.have.length(1);
+      expect(wrapper.find('[data-123-foo]')).to.have.length(1);
```
blainekasten: can you add tests to ensure the value matching still works out correct?

Things like:

```
find('[data-foo-123=3'])
find('[data-foo-123="3"'])
```

test/ReactWrapper-spec.js
```diff
@@ -366,15 +366,34 @@ describeWithDOM('mount', () => {
...
 
+    it('should not find components with invalid attributes', () => {
+      // Invalid attributes aren't valid JSX, so manual instantiation is necessary
+      const wrapper = mount(
+        React.createElement('div', null, React.createElement('span', {
+          '123-foo': 'bar',
+          '-foo': 'bar',
+          ':foo': 'bar',
+        }))
+      );
```
Aweary: Can you just use JSX here to be consistent with the other tests? Thanks!

Aweary:
@jordwest can you add support for underscores as well? React will render this `<div/>` with the corrent attribute just fine:

```js
<div data-foo_bar='true' />
```
but this query currently fails: 

```js
expect(wrapper.find('[data-foo_bar="true"]')).to.have.length(1);
```

jordwest:
@blainekasten:
> can you add tests to ensure the value matching still works out correct?

Added in 31d0ec0

@Aweary:
> can you add support for underscores as well?

Added in 4c60f7b

Aweary:
Thanks @jordwest! LGTM

PR #420, opened 2016-05-27
tests for shallow#props() and prop()

nfcampos:
- added test ensuring that props() returns props of the root rendered node, not props passed to composite root component
- added tests for untested prop() method

### DIFF
--- a/test/ShallowWrapper-spec.js
+++ b/test/ShallowWrapper-spec.js
@@ -1381,6 +1381,95 @@ describe('shallow', () => {
       expect(wrapper.find('.baz').props().onClick).to.equal(fn);
       expect(wrapper.find('.foo').props().id).to.equal('fooId');
     });
+
+    it('should return props of root rendered node', () => {
+      class Foo extends React.Component {
+        render() {
+          return (
+            <div className={this.props.bar} id={this.props.foo} />
+          );
+        }
+      }
+
+      const wrapper = shallow(<Foo foo="hi" bar="bye" />);
+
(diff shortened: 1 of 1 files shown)


nfcampos:
I wanted to do the same for `mount` but apparently the following
```js
        render() {
          return (
            <div className={this.props.bar} id={this.props.foo} />
          );
        }
      }

      const wrapper = mount(<Foo foo="hi" bar="bye" />);

      expect(wrapper.props()).to.eql({ className: 'bye', id: 'hi' });
```
is not true for `mount`. `wrapper.props()` is actually the props passed in (`{foo: 'hi', bar: 'bye'}`)
Is this on purpose or is this a bug?

Aweary:
@nfcampos I believe that's on purpose. `shallow` doesn't render the actual top level component, looking back at [this comment from @lelandrichardson](https://github.com/airbnb/enzyme/pull/377#issuecomment-218517694) but mount does.

Aweary:
I'm also curious to see if these changes https://github.com/airbnb/enzyme/pull/385 pass these new tests too.

nfcampos:
yeah probably it's on purpose, but it is surprising, anyway it should be tested explicitly either way so I'll open another PR with these same tests for `mount` (changing these assertions to match the behaviour of mount)

nfcampos:
this one is ready to merge I believe

nfcampos:
rebased

ljharb:
LGTM

Issue #375, opened 2016-05-10
Implement SVG support


EvNaverniouk:
When using `.find()` on components with an `<svg/>` tag, enzyme explodes with:

```
TypeError: classes.replace is not a function
    at instHasClassName (webpack:///./~/enzyme/build/MountedTraversal.js?:69:21)
    at eval (webpack:///./~/enzyme/build/MountedTraversal.js?:224:20)
    at findAllInRenderedTreeInternal (webpack:///./~/enzyme/build/MountedTraversal.js?:282:13)
    at findAllInRenderedTreeInternal (webpack:///./~/enzyme/build/MountedTraversal.js?:294:24)
    at findAllInRenderedTreeInternal (webpack:///./~/enzyme/build/MountedTraversal.js?:279:12)
    at treeFilter (webpack:///./~/enzyme/build/MountedTraversal.js?:307:10)
    at ReactWrapper.eval (webpack:///./~/enzyme/build/ReactWrapper.js?:62:12)
    at eval (webpack:///./~/enzyme/build/ReactWrapper.js?:1166:21)
    at Array.map (native)
    at ReactWrapper.flatMap (webpack:///./~/enzyme/build/ReactWrapper.js?:1165:32)
```

Here is a very short and simple test case to reproduce the issue:

```
import {mount} from 'enzyme';
const component = mount(<svg />);
component.find('.test');
```

The problem is that SVG elements do not report their className as a string, but as an object instead.

Aweary:
@EvNaverniouk what version of React are you using?

EvNaverniouk:
I'm on React 15.0.2, Enzyme 2.3.0, JSDom 9.0.0. I'm also using the `css-modules-require-hook` module, but disabling it doesn't appear to fix the issue either.

Aweary:
Can you share the exact code that is failing for you?

Aweary:
@EvNaverniouk are you running this test in a real browser environment at all? Can I see your `jsdom` setup?

Using JSDom `9.0.0` I see the following behavior:

```js
var svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
console.log(svg.className)
// ''
```

Which is incorrect, it should return an object with the shape `{baseVal: "", animVal: ""}` so it looks like the reason it's passing for me is that JSDom isn't respecting `createElementNS`, so I'm curious whats happening on your end.

EvNaverniouk:
[...]

PR #421, opened 2016-05-27
running tests with karma

nfcampos:
where this is at:
- all tests are running
- ~~2 are failing~~ fixed now (were failing because the tests were assuming the body was empty)
- ~~only running in chrome for now~~ chrome and firefox

## todo
- [x] add source maps
- [x] see why 2 tests are failing
- [x] run in firefox
- [x] set up for travis
- [x] run with react 0.13
- [x] run with react 0.14
- [x] run with react 15


Aweary:
Awesome start @nfcampos this is going to be great. Should let us support + test SVGs for #375

nfcampos:
this is ready to merge now I believe

.travis.yml
```diff
@@ -8,15 +8,22 @@ node_js:
 before_install:
   - 'if [ "${TRAVIS_NODE_VERSION}" != "0.9" ]; then case "$(npm --version)" in 1.*) npm install -g npm@1.4.28 ;; 2.*) npm install -g npm@2 ;; esac ; fi'
   - 'if [ "${TRAVIS_NODE_VERSION}" != "0.6" ] && [ "${TRAVIS_NODE_VERSION}" != "0.9" ]; then npm install -g npm; fi'
-before_script: "sh install-relevant-react.sh"
+before_script:
+  - "sh install-relevant-react.sh"
+  - if [ $KARMA == 1 ]; then export CHROME_BIN=chromium-browser; fi
+  - if [ $KARMA == 1 ]; then export DISPLAY=:99.0; fi
```
ljharb: can all three of these commands be combined into one line, if they're related?

nfcampos: good point, I've combined them

nfcampos:
3 tests fail when ran on Firefox 45.0 and react 15.1.0 with karma because of this firefox bug https://github.com/facebook/react/pull/6896 it'll be fixed in a future version of react (or firefox) so I don't think there's anything for us to here other than be aware of it.
screen shot 2016-06-06 at 18 12 19

nfcampos:
oops, accidentally closed this, will reopen tomorrow

Aweary:
@nfcampos what's the status on this?

nfcampos:
I'll rebase and after that it should be ready to go

nfcampos:
@Aweary do you want to take a look at this before it gets out of date again? :)

nfcampos:
I've rebased this on the current master, does someone want to review this?

.travis.yml
```diff
@@ -26,6 +31,30 @@ sudo: false
 matrix:
   fast_finish: true
   exclude:
+  - node_js: "6"
```
ljharb: this might get much simpler after #564 is merged

package.json
```diff
@@ -62,6 +63,7 @@
     "babel-cli": "^6.11.4",
     "babel-core": "^6.13.2",
     "babel-eslint": "^6.1.2",
+    "babel-loader": "~6.2.5",
```
ljharb: why the tilde instead of a carat?

src/version.js
```diff
@@ -1,6 +1,12 @@
...
 
-export const VERSION = React.version;
-export const REACT013 = VERSION.slice(0, 4) === '0.13';
-export const REACT014 = VERSION.slice(0, 4) === '0.14';
-export const REACT15 = VERSION.slice(0, 3) === '15.';
+var React = require('react');
+
+var VERSION = React.version;
+
+module.exports = {
```
ljharb: why is this file changing from ES6 exports to node exports?

karma.conf.js
```diff
@@ -0,0 +1,91 @@
+/* eslint-disable no-var,prefer-arrow-callback */
+
+var IgnorePlugin = require('webpack').IgnorePlugin;
+var REACT013 = require('./src/version').REACT013;
```
ljharb: this should work without the change in src/version

package.json
nfcampos: my bad, i'll change that

src/version.js
nfcampos: because I need to import it in the karma conf file and I don't want to transpile that
makes sense?

.travis.yml
nfcampos: yep, i should change this once that's merged

src/version.js
ljharb: I'd prefer `npm run build` before the karma tests, so that the tests are always running on the build output.

nfcampos: hmm ok, makes sense. so i guess the `test:karma` command in package.json should run `build` before? or we should do that manually when testing with karma?

nfcampos: hmm, actually that's pretty inconvenient,  because the test files themselves import the source files from `src`, not from `build`, what do you suggest for that?

ljharb: hm, that's a good point. since the mocha tests use babel-register, let's just use that here too.

nfcampos: done

.travis.yml
nfcampos: I've changed this

(weirdly enough one of the two allowed failures on travis works after this, the karma example)

ljharb: should we remove it from "allowed failures" then?

nfcampos: I guess, even though I'm not entirely sure why it now works

nfcampos: I've removed

karma.conf.js
```diff
@@ -0,0 +1,92 @@
...
+      'test/_*.jsx',
+    ],
+
+    browsers: [
+      process.env.TRAVIS ? 'Chrome_travis' : 'Chrome',
+      'Firefox',
+    ],
+
+    preprocessors: {
+      'test/*.jsx': ['webpack', 'sourcemap'],
```
ljharb: `.js` might need preprocessing too - we use babel for every file, and `.jsx` is only for files that contain JSX

test/ReactWrapper-spec.jsx
```diff
@@ -929,8 +929,9 @@ describeWithDOM('mount', () => {
       }
 
       const similarException = ((() => {
+        const user = {};
         try {
-          return undefined.givenName;
+          return user.name.givenName;
```
ljharb: the difference here seems to be, "user is undefined" vs "user.name is undefined" - can you elaborate on why one is more correct than the other?

nfcampos: Firefox's error message for the error in the component in this test is different from chrome's/node's, if I remember correctly it's something like "cannot read property givenName of user.name", while in chrome is "cannot read property givenName of undefined". that's why I made this change, to ensure the property/variable names are the same in `similarException` and in the component

karma.conf.js
nfcampos: good point, will change

nfcampos: I've changed this

nfcampos:
does anyone else want to review this?

nfcampos:
@Aweary @ljharb do we want to merge this before it gets out of date again ;) ?

ljharb:
LGTM

PR #422, opened 2016-05-27
tests for mount#props() and prop()

nfcampos:
- added test ensuring that props() returns props of the root node
- added tests for untested prop() method

~~@lelandrichardson can you confirm that this behaviour of returning the props passed to the root node instead of the props of rendered root node is indeed the desired behaviour?~~ nevermind, I see why it is like this

### DIFF
--- a/test/ReactWrapper-spec.js
+++ b/test/ReactWrapper-spec.js
@@ -1533,7 +1533,7 @@ describeWithDOM('mount', () => {
       expect(wrapper.find('.foo').props().id).to.equal('fooId');
     });
 
-    it('should return props of root rendered node', () => {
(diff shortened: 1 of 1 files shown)


blainekasten:
LGTM

ljharb:
LGTM pending a rebase to get rid of the merge commit

nfcampos:
i'll rebase now

blainekasten:
Don't you just squash commits? Does the rebase really matter then?

nfcampos:
i've rebased

ljharb:
@blainekasten ideally we never squash or have merge commits. The best way to do it is to have a fast-forwardable rebase that can simply be merged (whether ff or non-ff).

PR #423, opened 2016-05-28
Add tap method to sidebar

pbernasconi:


nfcampos:
Hi @pbernasconi , thanks for doing this! Do you mind adding it here as well? https://github.com/airbnb/enzyme/blob/master/docs/README.md

pbernasconi:
@nfcampos ⚡ updated and squashed.

docs/README.md
```diff
@@ -60,6 +60,7 @@
     * [some(selector)](/docs/api/ShallowWrapper/some.md)
     * [someWhere(predicate)](/docs/api/ShallowWrapper/someWhere.md)
     * [state([key])](/docs/api/ShallowWrapper/state.md)
+    * [tap()](/docs/api/ShallowWrapper/tap.md)
```
nfcampos: can you add the argument to the signature like in the other two?

nfcampos: oh and let's call the argument by the same name it is called in the doc for `tap` to be consistent with the other methods, so `intercepter` instead of `fn`

nfcampos:
Thanks! LGTM

docs/api/mount.md
```diff
@@ -192,6 +192,9 @@ Reduces the current array of nodes to a value
 #### [`.reduceRight(fn[, initialValue]) => Any`](/docs/api/ReactWrapper/reduceRight.md)
 Reduces the current array of nodes to a value, from right to left.
 
+#### [`.tap(intercepter) => ReactWrapper`](ReactWrapper/tap.md)
```
nfcampos: @pbernasconi sorry to have realised this only now but the signature should be `.tap(intercepter) => Self` to match the api docs for `tap`

pbernasconi: ⚡ done

nfcampos:
Thanks again @pbernasconi! LGTM

PR #424, opened 2016-05-28
cache node_modules for travis builds

nfcampos:
let's see if the builds are faster

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,9 @@ env:
   - EXAMPLE=karma-webpack
   - EXAMPLE=jest
 sudo: false
+cache:
+  directories:
+    - node_modules
 matrix:
   fast_finish: true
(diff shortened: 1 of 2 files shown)


ljharb:
It will be interesting to compare the builds, but I'm very hesitant to do this - correct is always better than fast, and we install multiple versions of react depending on the build, so I doubt it will yield us much improvement.

nfcampos:
build with empty cache: https://travis-ci.org/airbnb/enzyme/builds/133583796
build from existing cache: https://travis-ci.org/airbnb/enzyme/builds/133587742
it appears to like 10-20% faster with the cache so I don't think it's enough to be worth it.
(and apparently the cache broke something for the node 0.10 builds...)

ljharb:
Agreed.

PR #425, opened 2016-05-28
fix travis test command to actually run example tests

nfcampos:
example tests were not being run, they were just printing `Test skipped`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,15 +36,15 @@ matrix:
     env: EXAMPLE=karma-webpack
   - node_js: "6"
     env: EXAMPLE=jest
-  - node_js: "4"
+  - node_js: "5"
     env: EXAMPLE=mocha
-  - node_js: "4"
+  - node_js: "5"
     env: EXAMPLE=karma
-  - node_js: "4"
+  - node_js: "5"
     env: EXAMPLE=react-native
-  - node_js: "4"
+  - node_js: "5"
     env: EXAMPLE=karma-webpack
-  - node_js: "4"
+  - node_js: "5"
     env: EXAMPLE=jest
   - node_js: "0.12"
     env: EXAMPLE=mocha
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,15 +26,15 @@ sudo: false
 matrix:
   fast_finish: true
   exclude:
-  - node_js: "5"
+  - node_js: "6"
     env: EXAMPLE=mocha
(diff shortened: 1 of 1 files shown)


nfcampos:
the build is failing because two examples are failing:
- [react-native](https://travis-ci.org/airbnb/enzyme/jobs/133596121)
- [karma-webpack](https://travis-ci.org/airbnb/enzyme/jobs/133596122)

I'll take a look at the examples soon-ish to figure out why they're failing if no one beats me to it

.travis.yml
```diff
@@ -10,7 +10,7 @@ before_install:
   - 'if [ "${TRAVIS_NODE_VERSION}" != "0.6" ] && [ "${TRAVIS_NODE_VERSION}" != "0.9" ]; then npm install -g npm; fi'
 before_script: "sh install-relevant-react.sh"
 script:
-  - 'if [ -z "$REACT" ] && [ "${TRAVIS_NODE_VERSION}" = "4" ]; then npm run test:env -- "${EXAMPLE}" ; elif [ -z "$REACT" ]; then echo "Test Skipped" ; elif [ "${TRAVIS_NODE_VERSION}" = "4" ]; then npm run lint && npm run travis ; elif [ "${TRAVIS_NODE_VERSION}" = "0.12" ]; then npm run travis ; else npm test ; fi'
```
ljharb: the intention is to skip them in all envs except one

nfcampos: I know, but the `exclude` entries in `matrix` do that already

nfcampos: which is why the example tests were not running, `matrix.exclude` were restricting them to node 6, this was restricting them to node 4, thus they never actually ran

ljharb: ah, then when someone updated `exclude` to 6, they must have forgotten to update this. Good catch!

nfcampos:
both failing examples run on my machine without failing, does anyone have any idea why they don't work in travis, do you guys remember if when these examples were set up initially something similar happenned?

ljharb:
It looks like `ws` is having trouble compiling. If we run them in node 4 or 6 instead of node 5, would that help?

nfcampos:
I've changed from running examples in node 6 to running them in node 5, let's see what happens

nfcampos:
apparently it still fails with node 5, I think the problem is further below, those build errors happen in the builds to test enzyme itself, not the examples, and those pass

PR #427, opened 2016-05-30
bug fix: getSeletors()

rifeljm:
selectors shouldn't be split by space, because selector itself can contain a space
Example: .find('div[data-foo="foo bar"]')

### DIFF
--- a/test/ReactWrapper-spec.js
+++ b/test/ReactWrapper-spec.js
@@ -549,6 +549,15 @@ describeWithDOM('mount', () => {
       expect(() => wrapper.find(null)).to.throw(Error);
     });
 
+    it('Should query attributes with spaces in their values', () => {
+      const wrapper = mount(
+        <div>
+          <h1 data-foo="foo bar">Hello</h1>
+        </div>
+      );
+      expect(wrapper.find('[data-foo="foo bar"]')).to.have.length(1);
+    });
+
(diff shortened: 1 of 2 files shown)


test/ReactWrapper-spec.js
```diff
@@ -549,6 +549,15 @@ describeWithDOM('mount', () => {
       expect(() => wrapper.find(null)).to.throw(Error);
     });
 
+    it('Should query attributes with spaces in their values', () => {
+      const wrapper = mount(
+        <div>
+          <h1 data-foo="foo bar">Hello</h1>
+        </div>
+      );
+      expect(wrapper.find('[data-foo="foo bar"]')).to.have.length(1);
```
Aweary: Can you add another test to verify that attribute queries still work without the attribute value? 

```js
wrapper.find("[data-foo]")
```

Aweary:
@rifeljm can we make sure this works with/covers the following cases as well:

* Attribute values with multiple words (`'[data-foo="bar baz quz]'`)
* Attribute values with multiple spaces between the words (`'data-foo="bar    baz     quz'`)

blainekasten:
LGTM

Aweary:
@rifeljm can you the branch with master? After that I think we can merge 👍

blainekasten:
LGTM

blainekasten:
Thanks @rifeljm

PR #432, opened 2016-06-01
Loose contains and default props

jkimbo:
Came across this issue today where the `.containsMatchingElement` function gets tripped up with default props. If a custom component has a default prop that is overridden when it's rendered you have to add that override to the component your matching against for `.containsMatchingElement` to work. I've added a minimal failing test for both the shallow and full dom render wrappers to illustrate it.

In my mind the `.containsMatchingElement` (and it's siblings) should not care that a default prop is different between the rendered component and the matching one since that is not what I'm testing but I can see why that is the case since at the point of checking React doesn't differentiate. Any thoughts on this?

Thanks!

### DIFF
--- a/test/ReactWrapper-spec.js
+++ b/test/ReactWrapper-spec.js
@@ -2616,6 +2616,26 @@ describeWithDOM('mount', () => {
         <div onClick={spy2}>Au revoir le monde</div>
       )).to.equal(false);
     });
+    it('should match custom component that looks like the rendered one', () => {
+      class TestComponent extends React.Component {
+        render() {
+          return (<div>Hello World</div>);
+        }
+      }
+
+      TestComponent.defaultProps = {
+        foo: 'bar',
+      };
+
+      const wrapper = mount(
+        <div>
+          <TestComponent foo="buzz" />
+        </div>
+      );
+      expect(wrapper.containsMatchingElement(
+        <TestComponent />
+      )).to.equal(true);
+    });
   });
   describe('.containsAllMatchingElements(nodes)', () => {
     it('should match on an array of nodes that all looks like one of rendered nodes', () => {
--- a/test/ShallowWrapper-spec.js
+++ b/test/ShallowWrapper-spec.js
@@ -2559,6 +2559,26 @@ describe('shallow', () => {
         <div onClick={spy2}>Au revoir le monde</div>
       )).to.equal(false);
     });
+    it('should match custom component that looks like the rendered one', () => {
+      class TestComponent extends React.Component {
+        render() {
+          return (<div>Hello World</div>);
+        }
+      }
+
+      TestComponent.defaultProps = {
+        foo: 'bar',
+      };
+
+      const wrapper = shallow(
+        <div>
+          <TestComponent foo="buzz" />
+        </div>
+      );
+      expect(wrapper.containsMatchingElement(
+        <TestComponent />
+      )).to.equal(true);
+    });
   });
   describe('.containsAllMatchingElements(nodes)', () => {
     it('should match on an array of nodes that all looks like one of rendered nodes', () => {


blainekasten:
I think this makes sense. From a testing perspective, It would be way to hard to have to check a components defaultProps to see how it will behave. And then adding a defaultProp breaks tests? That seems like a bad idea.

I think these tests are correct and we should get them passing.

ljharb:
I disagree. Adding a defaultProp that changes the way the component behaves in production SHOULD break tests. You should be testing with the actual prop values that React will use in production, not the props you chose to pass in. I'm :x: on this one.

jkimbo:
@ljharb I get your point but for me it's all about what you're actually asserting in your tests. As shown in the tests in this pr I don't actually care what the value of the prop `foo` is in my assertion. The change from the default value for the prop might well change the behaviour of the component however that is not what my test is checking for.

A different way of putting it is if instead of using a default prop I have `const foo = this.props.foo || 'bar'` in the render method then these tests would not fail (correctly) since it's only checking that the props I include in my assertion are the same.

nfcampos:
I agree with @ljharb on this one, default props are part of the what the component is so in your test case
```js
      class TestComponent extends React.Component {
        render() {
          return (<div>Hello World</div>);
        }
      }

      TestComponent.defaultProps = {
        foo: 'bar',
      };

      const wrapper = mount(
        <div>
          <TestComponent foo="buzz" />
        </div>
      );
      expect(wrapper.containsMatchingElement(
        <TestComponent />
      )).to.equal(true);
```
there is in fact no `TestComponent` element without props in the rendered tree because all TestComponent elements get `foo='bar`via defaultProps

jkimbo:
@nfcampos agree that the test case should fail if using the `.contains` query but the `.containsMatchingElement` finds elements that **look like** the expected element and in my opinion there is a `TestComponent` element that looks like the one in the assertion.

nfcampos:
@jkimbo I wasn't as clear as I wanted to be, if `TestComponent` is
```js
      class TestComponent extends React.Component {
        render() {
          return (<div>Hello World</div>);
        }
      }

      TestComponent.defaultProps = {
        foo: 'bar',
      };
```
then `<TestComponent />` is virtually indistinguishable from `<TestComponent foo='bar />` (`defaultProps` are applied on the call to `React.createElement`).

While I understand your point about wanting to use `wrapper.containsMatchingElement(<TestComponent />)` to test whether **there is any `TestComponent` element present irrespective of its props**, React applies `defaultProps` as soon as that JSX expression is evaluated.

As such you're not checking whether **there is any `TestComponent` element present irrespective of its props** but are instead testing whether **there is any `TestComponent` element present whose `defaultProps` were not overridden**.

I don't see how we could change the behaviour of `containsMatchingElement` to allow your use case without breaking with the semantics of what a JSX expression of a React element means, but I'm happy to be proven wrong.

jkimbo:
@nfcampos apologies for the late reply to this, I've been on holiday for the past couple of weeks.

So, while I understand your argument around correctness, I would argue that as a test writer my assumption would be that the props that I am not asserting against would not affect my tests. I am not trying to say that `containsMatchingElement` should match an element **irrespective of its props** but that it should match an element **with the props that I am checking for** and no more. The docs on `containsMatchingElement` specifically say:

> ...element in the wraper **looks like** the expected element...

which to me indicates that it is using a heuristic to determine if an element **looks like** what is being rendered. I would argue that that heuristic should ignore default props when matching.

Here is a more complete example which I hope illustrates my point a bit better.

The following should pass:

```js
      class TestComponent extends React.Component {
        render() {
          return (<div>Hello World</div>);
        }
      }
      TestComponent.defaultProps = {
        fizz: true,
      };

      const wrapper = shallow(
        <div>
          <TestComponent foo="buzz" className="bar" />
        </div>
      );
      expect(wrapper.containsMatchingElement(
        <TestComponent className="bar" />
      )).to.equal(true);
```

However if I set the default prop of `foo` on `TestComponent` to something other than `buzz`:

```js
      TestComponent.defaultProps = {
        fizz: true,
        foo: 'bar',
      };
```

then the test will fail. In my eyes that's not the expected because `<TestComponent className="bar" />` **looks like** `<TestComponent foo="buzz" className="bar" />`

Aweary:
I get what you're saying, but as @ljharb said `defaultProps` affect the component's behavior and shouldn't be ignored. When you do

```js
   TestComponent.defaultProps = {
        fizz: true,
        foo: 'bar',
      };
```
`<TestComponent className="bar" />` is essentially identical to `<TestComponent foo='bar' fizz={true} className="bar" />`. I feel like we'd make it easier for people to write tests that *should* fail but don't, since we're diverging from how React actually renders components with `props` and `defaultProps`.

If anything we should mention this in the [Common Gotchas](http://airbnb.io/enzyme/docs/api/ShallowWrapper/containsMatchingElement.html#common-gotchas) section in the docs for `containsMatchingElement` and possibly clarify what "looks like" means specifically.

I'm going to close this since it seems that the majority of contributors aren't in favor of this, but feel free to continue the discussion and we can re-open if we decide it's a change we want to pursue!

PR #433, opened 2016-06-02
Fix example in reactwrapper context doc

romainberger:
Hi,

Just a small fix in a example of the doc. Not a big change but I spent 5 minutes trying to figure out why my code wasn't working so I figured someone else might have the same issue :)


Aweary:
Thanks @romainberger! LGTM

PR #449, opened 2016-06-09
Fix syntax highlighting

rpunkfu:


README.md
```diff
@@ -67,7 +67,7 @@ Basic Usage
 
 ## [Shallow Rendering](/docs/api/shallow.md)
 
-```jsx
+```javascript
```
ljharb: this will break syntax highlighting for the jsx examples. What is this intended to "fix"?

rpunkfu: Because right now syntax highlighting is broken with `jsx` instead of `javascript`.

That's how it looked before fix:


and after:

Aweary: It seems like Github doesn't support JSX syntax highlighting very well. If you look at https://github.com/airbnb/enzyme/blob/master/src/ReactWrapperComponent.jsx you'll see there's very little highlighting going on

ljharb:
in that case, LGTM

Issue #452, opened 2016-06-13
hasClassName throws Error if component has property className which is not a string


trmjoa:
Hi,

We use a utility like JedWatsons classnames(https://github.com/JedWatson/classnames) for setting classnames on our components. Syntax is like this:

```
<MyComponent className={ { classA : true, classB : false } } />
```

This makes the internal hasClassName and all functions in the api blow up with the exception `TypeError: classes.replace is not a function` since the property `className` is a object.

In my view the hasClassName function should not throw a exception if it encounters a custom component with a className attribute that is not a string. From my usage of enzyme I have the assumption that you are generally looking for native components when the hasClassName function is used. If that assumption is considered true I am tempted to suggest the following patch:
- If node.type is a non-native component(function) and className is not a string return false. 
- If node.type is a non-native component and className is a string, run the check as today (for backwards compatibility)

It could probably also be possible to take the defined proptype into consideration.

/Joakim

Aweary:
This is documented in https://github.com/airbnb/enzyme/issues/375, you can watch that issue for updates! Thanks!

Aweary:
Also, classnames doesn't return an object. It takes an object and returns a string based on the key's value.
 
```js
import classNames from 'classnames'

const className = classNames({
  foo: true,
  bar: true,
  baz: false
});
// returns "foo bar"

**Comment by bannaN** [comment]
Hi,

Thanks for your input. Might be that I misunderstand the proposed fix in the referenced issue, but after I looked through the PR that is connected with the ticket you referenced, it seems to me that it will not solve the issue. 

They are discussing a SVG specific solution(using `baseVal`) which does not solve the general naming collision where a custom created component has a `className` attribute. You are right that the classnames utility returns a string, and how we use it, is something like this:

```
import classNames from 'classnames';
class MyComponent extends React.Component{
  render() {
    <div className={ classNames( this.props.className ) }> { this.props.children} </div>
  }
}

ReactDOM.render(
  React.createElement(
    MyComponent, 
    { className : { classA : true, classB : shouldIAppendClassB(), classC : false }}
  )
);
```
Wouldn't it be more appropriate in the `hasClassName` method to assume the check is not applicable if the component has no `className` attribute that should be a string(defined by propTypes on custom components)? The problem comes in the find methods, because they iterate all components, also non-native.

Also, in the issue you linked, it seems to me that the proposed fix only affects the full rendering functionality, not the shallow part.

Best regards
Joakim

[...]

PR #453, opened 2016-06-13
hasClassName throws Error if component has property className which is not a string

trmjoa:
A proposal to fix https://github.com/airbnb/enzyme/issues/452

I now see that there are PR's pending for related issues (https://github.com/airbnb/enzyme/pull/448). I'm not sure if the proposed fixes/code changes are overlapping.

If you feel its a viable approach, but want changes, please advise me on how to proceed and I will give it a try.

### DIFF
--- a/src/ShallowTraversal.js
+++ b/src/ShallowTraversal.js
@@ -27,6 +27,9 @@ export function childrenOfNode(node) {
 
 export function hasClassName(node, className) {
   let classes = propsOfNode(node).className || '';
+  if (node && typeof node.type === 'function' && typeof classes !== 'string') {
+    return false;
+  }
   classes = classes.replace(/\s/g, ' ');
   return ` ${classes} `.indexOf(` ${className} `) > -1;
 }
--- a/test/ShallowTraversal-spec.js
+++ b/test/ShallowTraversal-spec.js
@@ -60,46 +60,21 @@ describe('ShallowTraversal', () => {
     });
 
     it('should not throw with custom components', () => {
-
-      // Checks that a Error is not thrown if the component has a object as a className value
-      {
-        class CustomComp extends React.Component {
-          render() {
-            const classes = Object.keys(this.props.className)
-              .filter(k => this.props.className[k]).join(' ');
-            return React.createElement(
-              'div',
-              { className: classes },
-              this.props.children
-            );
-          }
+      class CustomComp extends React.Component {
+        render() {
+          const classes = Object.keys(this.props.className)
+            .filter(k => this.props.className[k]).join(' ');
+          return (
+            <div className={ classes }>{ this.props.children}</div>
+          );
         }
-        CustomComp.defaultProps = { className: {} };
-
-        const node = (<CustomComp />);
-        expect(() => hasClassName(node, 'something')).to.not.throw(Error);
-
-        expect(hasClassName(node, 'something')).to.equal(false);
       }
+      CustomComp.defaultProps = { className: {} };
 
-      // Checks that if the component has a string as a className value
-      // it will still work (for backwards compatibility)
-      {
-        class CustomComp extends React.Component {
-          render() {
-            return React.createElement(
-              'div',
-              { className: this.props.className },
-              this.props.children
-            );
-          }
-        }
-        const node = (<CustomComp className="myClass" />);
+      const node = (<CustomComp />);
+      expect(() => hasClassName(node, 'something')).to.not.throw(Error);
 
-        expect(hasClassName(node, 'myClass')).to.equal(true);
-      }
-
-
+      expect(hasClassName(node, 'something')).to.equal(false);
     });
 
   });
--- a/test/ShallowTraversal-spec.js
+++ b/test/ShallowTraversal-spec.js
@@ -59,6 +59,49 @@ describe('ShallowTraversal', () => {
       expect(hasClassName(node, 'foo-bar')).to.equal(true);
     });
 
+    it('should not throw with custom components', () => {
+
+      // Checks that a Error is not thrown if the component has a object as a className value
+      {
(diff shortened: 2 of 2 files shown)


Aweary:
Thanks for the PR @bannaN! I think we will be going with #448 since it was opened first, but we appreciate the effort. We also don't want to simply return `false` when the `className` is an object, as that's essentially a silent failure, which is what we had before 2.3.0.

src/ShallowTraversal.js
```diff
@@ -27,6 +27,9 @@ export function childrenOfNode(node) {
 
 export function hasClassName(node, className) {
   let classes = propsOfNode(node).className || '';
+  if (node && typeof node.type === 'function' && typeof classes !== 'string') {
```
Aweary: Can we just do the `typeof` check instead of all these checks on `node`? No matter what, if `classes` is not a string we should exit early since the next operation is a string operation.

test/ShallowTraversal-spec.js
```diff
@@ -59,6 +59,49 @@ describe('ShallowTraversal', () => {
       expect(hasClassName(node, 'foo-bar')).to.equal(true);
     });
 
+    it('should not throw with custom components', () => {
+
+      // Checks that a Error is not thrown if the component has a object as a className value
+      {
```
Aweary: Please remove these comments and extra brackets around the tests, they have no purpose.

test/ShallowTraversal-spec.js
```diff
@@ -59,6 +59,49 @@ describe('ShallowTraversal', () => {
...
+        class CustomComp extends React.Component {
+          render() {
+            const classes = Object.keys(this.props.className)
+              .filter(k => this.props.className[k]).join(' ');
+            return React.createElement(
```
Aweary: Use JSX instead, to be consistent with our current tests

test/ShallowTraversal-spec.js
```diff
@@ -59,6 +59,49 @@ describe('ShallowTraversal', () => {
...
+              'div',
+              { className: this.props.className },
+              this.props.children
+            );
+          }
+        }
+        const node = (<CustomComp className="myClass" />);
+
+        expect(hasClassName(node, 'myClass')).to.equal(true);
+      }
```
Aweary: You don't need to add a backwards compatibility test, the current tests should cover this.

src/ShallowTraversal.js
bannaN: I don't have the overview of the whole codebase, but I would consider it good if the method throws if its a native component and `className` is not a string. That should not be a  silent error?
So;
`node && typeof node.type === 'function'` if that is true the node currently being processed is a custom component. If its not a custom component e.g a regular div, span etc, it will continue and have the same behavior as today, that is throw a error if you have passed something else then a string to `className`. 

That's my take on it. What do you think? Your decision, of course :)

test/ShallowTraversal-spec.js
bannaN: Removed.

test/ShallowTraversal-spec.js
bannaN: Done

test/ShallowTraversal-spec.js
bannaN: Removed

bannaN:
Updated the PR according to your comments. Didn't update the code inside `hasClassName` itself as I just wanted to clarify the behavior first. Whats your policy on patches. Do you want me to rebase and squash the commits into one?

Aweary:
@bannaN do you see this conflicting with https://github.com/airbnb/enzyme/pull/448 in any way?

bannaN:
Well... We're not using the mounted version of enzyme that much yet. But from looking at the code in MountedTraversal.js and the patch in #448 they both address issues around the class checking. The #448 patch just adds support for checking the `classList` property first if it exists, and if it doesn't and `className` is a object it uses the `baseVal` property if it exists.

The problem described in this issue (#453) doesn't exist in the mounted part of the code because in MountedTraversal.js a check for `!isDOMComponent(inst)` is executed. It will return false for custom components, I also do this in my patch so the behavior will be consistent with the version used in the mounted part.

The only "conflict" I see is that the patch in #448 should maybe be applied to the ShallowTraversal.js also, or at least ported. So that the behavior is consistent. Then both methods that check for classnames has the following behavior:

1. Check that this is a node representing a native html element(div,span, p etc); Otherwise return false
2. Check the `classList` property, if present check if the classname is in it and return the result.
3. If the instance doesn't have a `classList` property, check the `className` property. If it exists and is a object assign the objects `baseVal` property to the `classes` variable(this is to suppoert SVG elements).
4. Do the check as it is done today and return the result

trmjoa:
@Aweary Are you interested in getting this in, or should I close the issue?

lencioni:
@Aweary what do you think we should do with this PR?

Aweary:
I don't think we should throw if it's not a string, using an object with an overloaded `toString` method is perfectly valid. https://jsfiddle.net/tLs53LnL/

We should probably just coerce `className` to a string before calling `replace`.

ljharb:
This looks worthwhile to get in in some form - either way it needs to be rebased, now that #448 is already merged.

Aweary:
@ljharb looks like we posted at the same time, ping just in case you missed my comment 🙂

ljharb:
Coercing to string (with `String()` only) seems reasonable also.

Aweary:
@trmjoa do you want to update your PR to just coerce `className` to a string using `String()`?

trmjoa:
I will rebase and update the PR over the weekend.

ljharb:
The linked issue was closed with 3070850ca889be2ca9d396c0433cc9c536341b48

PR #454, opened 2016-06-13
Fix broken link to Lab setup instructions.

ludovicofischer:
On the documentation home page, the link to the Lab instructions should lead you to a web page, instead you get the option to download a markdown file.

This ensures the link points to the generated HTML by adding the Lab page to the table of contents. The build gives many similar errors  (`x contains an hyperlink to resource outside spine`), so if you think this fix is correct I could also go over the rest.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -9,6 +9,7 @@
   * [Karma](/docs/guides/karma.md)
   * [Mocha](/docs/guides/mocha.md)
   * [React Native](/docs/guides/react-native.md)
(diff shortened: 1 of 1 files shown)


Aweary:
I haven't touched the Gitbook setup, but this seems fine to me. I believe it might require incrementing the version in `book.json` as well (ping @ljharb)

ludovicofischer:
It’s time for bed in my side of the world, so I’ll be back in around 8 hours.

lelandrichardson:
LGTM

Issue #456, opened 2016-06-14
is ":not" selector not supported?


arch-mage:
What I understand is that ":not" is a css selector to find an element without specified attribute.
So, with this component:

``` javascript
const Table = () => (
  <table>
    <thead>
      <tr>
        <th></th>
        <th></th>
      <tr>
    </thead>
    <tbody>
      <tr>
        <th scope="row"></th>
        <td></td>
      <tr>
    </tbody>
  </table>
);
```

But why this is happening?

``` javascript
wrapper.find('th[scope="row"]').length // == 1
wrapper.find('th:not([scope="row"])').length // == 0 ?
```

koba04:
You can see enzyme is supporting CSS selectors in the following.

https://github.com/airbnb/enzyme/blob/master/docs/api/selector.md#1-a-valid-css-selector

blainekasten:
We'd gladly accept a PR if you were wanting to add support for the `:not` selector!

AugustinLF:
I'm giving it a try, I'll ley you know soon of my progress!

ljharb:
This was added in #460.

akegan:
Looks like the PR referenced above ended up getting closed rather than merged. Does anyone know whether or not it got implemented, and if it did, which version it was released in?

ljharb:
@akegan it ended up moving to #1086, and should be in all versions of enzyme 3.

Issue #439, opened 2016-06-06
attribute selector returning null when selector contains a dot or space


andrewmclagan:
example:

`(wrapper.find('[name="user.first_name"]')` returns null.

`(wrapper.find('[placeholder="First name"]')` returns null.

As far as I'm aware dot syntax is totally acceptable and valid?

ljharb:
Can you provide how you created `wrapper`?

andrewmclagan:
````
const defaultProps = {
  editProfile: () => {},
  submitting: false,
  handleSubmit: () => {},
  invalid: false,
  initialValues: {},
};

test('it renders the correct form elements', t => {
  const wrapper = shallow(<ProfileForm {...defaultProps} />);
  t.is(wrapper.find('[name="user.first_name"]').length, 1);
});
````

something like that

**Comment by ljharb** [comment]
And what does `ProfileForm` render? (also what test framework is that?)

**Comment by andrewmclagan** [comment]
The test runner is AVA https://github.com/avajs/ava

I overcame the issue by using object notation `{ name: 'user.name' }` just thought i'd log it.

ProfileForm
````
  render() {

    const { invalid, submitting } = this.props;

    return (
      <div>
        <form className="signup-form" onSubmit={this.props.handleSubmit(this.submit)}>

          <div className="row">
            <FormField
              name="user.first_name"
              component="input"
              type="text"
              placeholder="First name"
              wrapClassName="col-md-6 form-group"
            />
            
           // ...
````

**Comment by Aweary** [comment]
@andrewmclagan for reference, we are already aware that a space in an attribute selector causes it to fail, see https://github.com/airbnb/enzyme/issues/416

**Comment by Aweary** [comment]
As for dots in attribute values, I was able to reproduce the issue. You can see the failing test case here: 
[Aweary/enzyme-test-repo/blob/issue-439/test.js
](https://github.com/Aweary/enzyme-test-repo/blob/issue-439/test.js)

```js
class TestComponent extends React.Component {
  render() {
    return (
      <div data-foo='foo.bar'>
        <h1>Hello, {this.props.name}</h1>
      </div>
    );
  }
}


describe('Issue #439', () => {
  it('should find attributes with dots in the value', () => {
    const wrapper = mount(<TestComponent/>);
    expect(wrapper.find('[data-foo="foo.bar"]')).to.have.length(1);
  });
```

Aweary:
So the issue is that the `"."` is flagging the selector as a compound selector

```js
[...]

Issue #413, opened 2016-05-25
Trouble with compound selector


fresk-nc:
I use [css-modules](https://github.com/webpack/css-loader#css-modules)  and have trouble with this selector:

```
.PlaybackBar__minimize___Psnl-.PlaybackBar__button___mzBxL
```

```
const wrapper = shallow(<div><button className="PlaybackBar__minimize___Psnl- PlaybackBar__button___mzBxL"></button></div>);
expect(wrapper.find('.PlaybackBar__minimize___Psnl-.PlaybackBar__button___mzBxL')).to.have.length(1);
```

Problem here https://github.com/nfpiche/enzyme/blob/master/src/Utils.js#L179.
I suggest to supplement the regex.

Aweary:
Hey @fresk-nc, thanks for the issue! I was able to reproduce a simplified case. You can see it here: [Aweary/enzyme-test-repo/blob/issue-413/test.js](https://github.com/Aweary/enzyme-test-repo/blob/issue-413/test.js). The issue is the dash after the first class name. You can see in the test case that by adding a trailing dash: 

```js
const firstClass = 'foo'
const secondClass = 'bar'

const transformedFirstClass = firstClass + '-'
const transformedSecondClass = secondClass
```

It causes the query to fail. I did verify that a trailing dash works fine in an actual DOM environment:

```js
div = document.createElement('div')
div.classList.add('foo-')
div.classList.add('bar')
document.body.appendChild(div)
document.querySelector('.foo-.bar')
// > <div class=​"foo- bar">​</div>​
```

So we should definitely fix this so it matches `querySelector`'s behavior. We'll look into this as soon as we can 👍

Aweary:
I'm going to use this issue to document any other valid compound selectors that aren't being identified correctly:

* `'[foo="true"].bar'`

jacobp100:
Also broken,

```js
test.only('weird-but-not-that-weird input names', t => {
  const $ = mount(<input type="checkbox" name="hours[]" value="0" />);
  t.is($.find('[name="hours[]"]').length, 1);
});
```

wmira:
Compound selector is unable to handle if className starts with _ or - it seems. I just hit this one..

```javascript
it.only('Selector Bug', () => {
        const wrapper = shallow(
            <div className='abc'>
                <div className='_startsWithUnderscore'/>
                <div className='nounderscore'/>
            </div>           
        );
[...]

PR #458, opened 2016-06-16
Use postcss-selector-parser for selector handling

jacobp100:


jacobp100:
This broke one test: *should not find components with invalid attributes* (i.e. `[-foo]` won't match anything).

I'm not sure what's best here. If we want to keep this behaviour, it'll have to be an extra check on attribute selectors. I feel like selecting invalid JSX attributes should throw.

Aweary:
Very nice, I was [just mentioning](https://github.com/airbnb/enzyme/issues/439#issuecomment-225700390) doing something like this here. Do you know if this would also resolve https://github.com/airbnb/enzyme/issues/456?

ping @ljharb what do you think?

src/ComplexSelector.js
```diff
@@ -7,31 +7,31 @@ export default class ComplexSelector {
...
 
-      list.push(sel);
       return list;
     }, []);
   }
 
   handleSelectors(selectors, wrapper) {
     const recurseSelector = (offset, fn, pre) => {
-      const predicate = pre || this.buildPredicate(selectors[offset]);
+      const predicate = pre || this.buildPredicate(selectors[offset].toString());
```
ljharb: `toString` is not guaranteed to return a String - please wrap in `String()` instead, throughout.

src/Utils.js
```diff
@@ -161,13 +162,14 @@ export function withSetStateAllowed(fn) {
   }
 }
 
-export function splitSelector(selector) {
-  return selector.split(/(?=\.|\[.*\])|(?=#|\[#.*\])/);
+export function getAst(selector) {
```
ljharb: This name is pretty broad for just dealing with CSS selectors - maybe `getSelectorAST`?

src/Utils.js
```diff
@@ -161,13 +162,14 @@ export function withSetStateAllowed(fn) {
...
+  let value;
+  parser(ast => { value = ast; }).process(selector);
+  return value;
 }
 
-export function isSimpleSelector(selector) {
-  // any of these characters pretty much guarantee it's a complex selector
-  return !/[~\s:>]/.test(selector);
+export function isValidPropName(propName) {
+  return /^[$A-Z_][0-9A-Z_$-:]*$/i.test(propName);
```
ljharb: Rather than using `/i`, could we just specify `a-zA-Z`? I think it's equally readable but would be more efficient.

ljharb:
Made a few comments - overall I think this looks awesome! @jacobp100, could you add more shallow and mount tests that cover a number of expected-to-work selectors, so that things don't break without us knowing in the future?

Aweary:
@jacobp100 I mentioned a couple of valid selectors that are broken in [this comment](https://github.com/airbnb/enzyme/issues/413#issuecomment-221636721). If we could test those as well that'd be great.

src/ShallowTraversal.js
```diff
@@ -108,34 +107,34 @@ export function nodeMatchesObjectProps(node, props) {
   return isSubset(propsOfNode(node), props);
 }
 
+function buildSelectorPredicate(selector) {
+  const nodes = getAst(selector).nodes[0];
+  return AND(nodes.map(node => {
+    switch (node.type) {
+      case 'class':
```
blainekasten: It's a bit nit, but the old way of using object constants seemed safer/nicer than just strings. Can you update this to use constants?

blainekasten:
This is awesome and exciting. Thanks for doing this work @jacobp100

src/ShallowTraversal.js
jacobp100: The strings come from `postcss-selector-parser`, and aren't exported. But I could definitely put a map in util if you think that's best!

blainekasten: Yeah. Just put it in a map so there is a central place of seeing the node types supported

ljharb: It would also be great to file an issue or send a PR upstream to get `postcss-selector-parser` to export the strings.

ljharb:
This LGTM! @lelandrichardson, any final thoughts before we merge?

Aweary:
For reference, I've opened a PR upstream to export the selector types: https://github.com/postcss/postcss-selector-parser/pull/70

lelandrichardson:
Really liking this direction.... Thanks for this work @jacobp100 . I have a couple of questions though:

1. I gather we are ignoring combined selectors (ie, `.foo, .bar`) in some situations (or is it all situations?). In this case, we should probably throw a warning or something to the developer explaining that only the first one is being used, or else it might be misleading.

2. are we no longer warning at all for unsupported CSS selectors? It seems there are still a number left.

3. this gets an AST for the CSS selectors (which is great!) but leaves traversal up to us. For descendant selectors etc. this can be quite hairy. I don't seem to see a lot of tests covering this behavior? It seems important to me if we are adding descendant (and other nonlocal selectors like sibling, direct descendant, etc.) then we will really want to make sure we get this right. 

Sorry if I'm missing something... but I feel like this needs to be merged in with a little bit more care. I love the direction and it seems like this will be much more robust than attempting the parsing ourselves, but let's make sure we understand the change we are making, and ensure that any functionality that we are adding ends up working as close to the CSS spec as it can be, or if it differs, documenting the differences.

Does that seem reasonable?

jacobp100:
Agreed. I think it would make sense to throw for multiple selectors.

I think I throw for all invalid selectors. Can you see a case where I haven't?

I'm actually looking into traversals in another branch. It's a bigger undertaking though.

Aweary:
One concern I have is bringing in a dependency that isn't being actively maintained. Looking at `postcss-selector-parser` it doesn't seem like they are actively maintaining it at the moment. The last commit was almost a month ago and there are PRs that haven't been addressed for about the same amount of time.

If we make this a dependency and run into upstream issues we may be facing an uphill battle getting those fixed. They could just be busy right now, but it's worth considering.

ljharb:
If it's abandoned that's one thing, but we can always fork it and publish our own in that case - I'm not too concerned about anything that's in a git repo.

lelandrichardson:
An alternative that's been brought up before is this: https://github.com/jquense/bill which is used in https://github.com/jquense/teaspoon

cc @jquense to see if he has anything to add on this thread!

jquense:
I've suggested bill before as a possibility. i'd be happy to have it consumed by more testing libraries, I'd also be happy to have more collaborators which may be a better option to forking something :).  bill solid at this point and supports back to react13

Aweary:
@jquense coming back to this, it looks like `bill` would take care of traversal for us right? I would be open to trying to implement this with `bill` instead so we don't have to include a bunch of AST traversal logic here too.

jquense:
@Aweary ya `bill`, exposes a bunch of traversal utilities, and abstracts out the differences between element and instance trees so that you can use the same traversal logic (where that makes sense) for both. High level it wraps matched trees in `Node` objects that expose DOM Element-esque traversal properties `prevSibling`, `children`, `parentNode`, etc; [it makes implementing a wide range of selectors nice and easy](https://github.com/jquense/bill/blob/master/src/selectors.js)

feel free to ping me where ever if you have any questions/concerns 👍

aweary:
@jacobp100 is there any chance you still want to work on this?

jacobp100:
I think this pr should be complete. But let me know if you have any bugs!

ljharb:
@jacobp100 at the least it'd need a fresh rebase - do you mind doing that?

aweary:
@jacobp100 would you mind rebasing and resolving the merge conflicts? If not, no worries. I can try to take this over if you don't have the time. I'd really like to get this merged soon.

jquense:
@aweary you are everywhere man

aweary:
@jquense so many interesting projects, so little time 😄

jquense:
@aweary I know! any shot of y'all incorporating https://github.com/jquense/bill ? id love to not maintain my other testing library on top of it and just use enzyme :P

aweary:
@jacobp100 At this point Enzyme has been entirely rewritten and this PR has fallen really far out of sync. I'm going to close this in favor of https://github.com/airbnb/enzyme/pull/1086 which uses a custom selector parser that supports the variant selector grammar we expect and utilizes the new RST spec we've defined.

Sorry this never got merged, we really appreciate all the hard work you did on this!

PR #459, opened 2016-06-16
Fix falsy attributes being ignored

jacobp100:


Aweary:
LGTM

test/ReactWrapper-spec.js
```diff
@@ -429,15 +429,24 @@ describeWithDOM('mount', () => {
...
+      expect(wrapper.find('[value=1]')).to.have.length(1);
+      expect(wrapper.find('[value=2]')).to.have.length(0);
+
+      expect(wrapper.find('[value=false]')).to.have.length(1);
+      expect(wrapper.find('[value=true]')).to.have.length(1);
+
+      expect(wrapper.find('[value="0"]')).to.have.length(0);
+      expect(wrapper.find('[value="1"]')).to.have.length(0);
+      expect(wrapper.find('[value="false"]')).to.have.length(0);
+      expect(wrapper.find('[value="true"]')).to.have.length(0);
```
ljharb: Can we add tests for `wrapper.find('[value]')`, and also include `<a value={undefined} />` and `<a value={null} />`? (in both shallow and mount)

Aweary: @ljharb we do already have tests for value-free attribute queries [here](https://github.com/jacobp100/enzyme/blob/35bb46f608702f0ae0fe0f0dc63d5b3d9d06a5a4/test/ReactWrapper-spec.js#L568)

ljharb: Gtk, but I think it would still be useful in this test case, among a number of elements with falsy-valued props

jacobp100: There's tests for null and undefined in `should support object property selectors`.

ljharb: I specifically want to check for string selectors, since that's what's affected by this PR.

Aweary:
@jacobp100 can you add `<a value={undefined} />` and `<a value={null} />` to the shallow and mount test and resolve the merge conflicts? I'd like to merge this afterwards 👍

lencioni:
@jacobp100 when you have a moment, would you also mind rebasing this onto latest master in addition to adding the tests requested by https://github.com/airbnb/enzyme/pull/459#issuecomment-239576011?

ljharb:
This seems like it will be superseded by #662.

PR #460, opened 2016-06-16
add ":not" to negate css prop selector

arch-mage:
this change based on [this issue](https://github.com/airbnb/enzyme/issues/456).

P.S: I've never done any pull request to any project before. This is my first time, please tell me if I did something wrong or I miss anything important.

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -182,7 +182,6 @@ export function selectorError(selector) {
   );
 }
 
-// export const isCompoundSelector = /([a-z]\.[a-z]|[a-z]\[.*\]|[a-z]#[a-z])/i;
 export const isCompoundSelector = /([a-z]\.[a-z]|[a-z]\[.*\]|[a-z]:not\(\[.*\]\)|[a-z]#[a-z])/;
 
 const isPropSelector = /^\[.*\]$/;


Aweary:
@Arch-Mage great work on your first PR, we really appreciate it! This is going to be affected by https://github.com/airbnb/enzyme/pull/458, which totally reimplements how we parse selectors, so I'm not sure if we want to merge this if it's not going to be compatible.

ljharb:
Personally I like the idea of merging this first, and then making sure #458 still passes the new tests after a rebase :-)

Aweary:
That's fine with me :) @Arch-Mage can we add some more tests to verify that this works with other selectors like classes and ids?

Arch-Mage:
I think it's better to wait for #458 first. Because, when I wrote this, i only think about negating attribute. So it won't work on `:not(#foo)` or `:not(.bar)`. It's kind a little hacking I think which obviously will break. I will see then if I can do adding `:not` selector

ljharb:
@aweary is this still needed in v3?

aweary:
@ljharb it's still needed, but this PR is now totally outdated. I'll be sending a PR for this today 😃  @arch-mage thanks for your work on this, sorry we didn't get it merged!

PR #462, opened 2016-06-16
Increase test legibility, added click() and val() shorthands on top of simulate()

bettiolo:
Added:
- `wrapper.click();` as a shorthand of `wrapper.simulate('click');`
- `wrapper.val('new value');` as a shorthand of `wrapper.simulate('change', {target: {value: 'new value'}};`


blainekasten:
Thanks for your contribution! I'm not personally a fan of these api's, though I could easily be out-ruled. Here's my rationale:

> wrapper.click(); as a shorthand of wrapper.simulate('click');

I would feel obligated to add a shorthand for every method, at which point the `simulate` method is pointless. I feel like the `simulate` is supportive enough.

> wrapper.val('new value'); as a shorthand of wrapper.simulate('change', {target: {value: 'new value'}};

I really don't like this one. for a couple reasons. `val` does not indicate that you are simluating a change event with a value. This is pigeon-holing to simulations that only need a target. But if your function called `e.stopPropagation`, now this method doesn't work for it.

blainekasten:
Do keep in mind, you could always do these sort of things in userland by extending the prototypes.

```js
// before tests are executed..
MountWrapper.prototype.click = function() { 
  this.simulate('click', arguments);
}
```

bettiolo:
`val()` has been chosen to map jquery/cheerio methods: http://api.jquery.com/val/#val2

blainekasten:
That is true. But they are slightly different uses. Theirs is directly setting the value on a node. Yours is simulating an event which could be handled in unknown ways. But the name of it was the lesser point. The main point was this:

> This is pigeon-holing to simulations that only need a target. But if your function called e.stopPropagation, now this method doesn't work for it.

I'm curious to know if other maintainers disagree with me. We'll sit on this and get some other feedback.

Aweary:
I feel essentially the same @blainekasten, I don't think we should start adding shorthand methods for simulated events and on top of that `.val()` would be potentially confusing IMO.

nfcampos:
- re: `.click()`, 👎 from me, I don't see a big point in adding a shortcut for something that is as simple as `.simulate('click')`.
- re: `.val()`, I do see some value in a method to set the value of a node, but as @Aweary and @blainekasten I don't like that such a method simulates a `change` event, especially for people familiar with jquery it can be surprising to find here a method that appears to do the same thing but simulates an event...

blainekasten:
Thanks for the PR. Sorry this one isn't going to land as it doesn't serve the best interest of this project. We hope to see more PRs from you though! Thanks again!

PR #464, opened 2016-06-21
docs: add systemjs guide

tamird:
Related to #302, https://github.com/jspm/jspm-cli/issues/1875.

### DIFF
--- a/docs/guides/systemjs.md
+++ b/docs/guides/systemjs.md
@@ -1,0 +1,73 @@
+# Using Enzyme with SystemJS
+
+If you are using a test runner that runs code in a browser-based environment,
+you may be using [SystemJS]() in order to bundle your React code.
+
+SystemJS uses static analysis to create a dependency graph at build-time of
+your source code to build a bundle. Enzyme has a hand full of conditional
+`require()` calls in it in order to remain compatible with React 0.13 and
+React 0.14.
+
+Unfortunately, these conditional requires mean there is a bit of extra setup
+with bundlers like SystemJS.
+
+In your SystemJS configuration, you simply need to make sure that the
+following files are labeled as "external", which means they will be ignored:
+
+```
+react/addons
+react/lib/ReactContext
+react/lib/ExecutionEnvironment
+```
+
+Here is an example piece of configuration code marking these as external:
+
+```json
+/* package.json */
+{
+  "jspm": {
+    "overrides": {
+      "npm:enzyme@2.3.0": {
+        "map": {
+          "react/addons": "@empty",
+          "react/lib/ExecutionEnvironment": "@empty",
+          "react/lib/ReactContext": "@empty"
+        }
+      }
+    }
+  }
+}
+```
+
+
+## React 0.14 Compatibility
+
+If you are using React 0.14, the instructions above will be the same but with a different list of
+externals:
+
+```
+react-dom
+react-dom/server
+react-addons-test-utils
+```
+
+## React 15 Compatability
+
+If you are using React 15, your config should include these externals:
+
+```json
+/* package.json */
+{
+  "jspm": {
+    "overrides": {
+      "npm:enzyme@2.3.0": {
+        "map": {
+          "react/addons": "@empty",
+          "react/lib/ExecutionEnvironment": "@empty",
+          "react/lib/ReactContext": "@empty"
+        }
+      }
+    }
+  }
+}
+```
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ testing your React components, you can consider using [jasmine-enzyme](https://g
 
 [Using Enzyme with Browserify](/docs/guides/browserify.md)
 
+[Using Enzyme with SystemJS](/docs/guides/systemjs.md)
+
 [Using Enzyme with WebPack](/docs/guides/webpack.md)
 
 [Using Enzyme with JSDOM](/docs/guides/jsdom.md)


docs/guides/systemjs-webpack.md
```diff
@@ -0,0 +1,93 @@
...
+### SystemJS
+
+```json
+/* package.json */
+{
+  "jspm": {
+    "overrides": {
+      "npm:enzyme@2.3.0": {
+        "map": {
+          "cheerio": "@empty",
```
tamird: Not super sure about this line. Perhaps @guybedford knows the correct mapping to Webpack's `'window'`?

docs/guides/systemjs-webpack.md
```diff
@@ -0,0 +1,93 @@
...
+  'react/addons': true,
+  'react/lib/ExecutionEnvironment': true,
+  'react/lib/ReactContext': true
+}
+// ...
+```
+
+## React 15 Compatability
+
+If you are using React 15, your configuration might include the following:
```
Aweary: "might" or "should"? If its required we should make this more explicit

Aweary:
As far as I know, SystemJS and jspm are distinct projects, with jspm being implemented on top of SystemJS. I don't think we should be including instructions for SystemJS, webpack, and jspm in one doc.

docs/guides/systemjs-webpack.md
tamird: updated

tamird:
SystemJS and JSPM are tightly coupled, but one is not built on the other.

The reason to put these instructions in the same doc is that they are largely the same, and the reason for the workaround is also the same. What's the upshot to splitting them up? Updates to one are likely to require updating the other.

Aweary:
SystemJS and jspm may be coupled, but why are we including it in the webpack documentation? If anything jspm specific instructions should be in a separate doc.

tamird:
Note that I've renamed the file, so it's not "the webpack documentation".

Aweary:
> Note that I've renamed the file, so it's not "the webpack documentation".
We have individual doc pages for bundlers, we should keep that same setup. my point is that I don't want to lump together webpack and jspm.

tamird:
Oh, I didn't realize e.g. browserify was separate. Updated.

tamird:
@Aweary ping

Aweary:
I don't know enough about SystemJS to know if this is accurate, unfortunately

@ljharb @lelandrichardson does this look OK to you?

tamird:
We're using this configuration in CockroachDB: https://github.com/cockroachdb/cockroach/blob/master/ui/package.json#L51:L58

Aweary:
> 
We're using this configuration in CockroachDB: https://github.com/cockroachdb/cockroach/blob/master/ui/package.json#L51:L58

👍  I just want one other contributor to take a look at it before we make any changes to our docs.

docs/guides/systemjs.md
```diff
@@ -0,0 +1,73 @@
...
+        "map": {
+          "react/addons": "@empty",
+          "react/lib/ExecutionEnvironment": "@empty",
+          "react/lib/ReactContext": "@empty"
+        }
+      }
+    }
+  }
+}
+```
```
guybedford: Rather than documenting this for all users to do manually, it may just be better to check this override in to the jspm registry or package.json file?

tamird: the override depends on the react version in use.

tamird:
they're conditional on the version of another dependency - is there a way to express that?

edrex:
The override provided for react 0.15.x isn't working for me: 
```
24 06 2016 15:16:34.201:WARN [web-server]: 404: /base/jspm_packages/npm/react@15.1.0/addons.js
24 06 2016 15:16:34.216:WARN [web-server]: 404: /base/jspm_packages/npm/react@15.1.0/lib/ReactContext.js
24 06 2016 15:16:34.226:WARN [web-server]: 404: /base/jspm_packages/npm/react@15.1.0/lib/ExecutionEnvironment.js
Chrome 51.0.2704 (Mac OS X 10.11.4) ERROR
  Error: (SystemJS) Error: XHR error (404 Not Found) loading http://localhost:9876/jspm_packages/npm/react@15.1.0/addons.js
  	Error loading http://localhost:9876/jspm_packages/npm/react@15.1.0/addons.js as "react/addons" from http://localhost:9876/jspm_packages/npm/enzyme@2.3.0/build/react-compat.js
```
(using karma-jspm)

guybedford:
@tamird yes we can do a conditional map configs, but it may be a little complex for this use case, so worth encouraging the manual process at least for now.

Aweary:
@guybedford thanks for jumping in here, I'll defer to your judgement whether these docs look good for jspm/SystemJS. If there's anything enzyme specific you're not sure of I'm happy to help, but otherwise if this looks good I'm sure we can get this merged.

docs/guides/systemjs.md
```diff
@@ -0,0 +1,73 @@
...
+React 0.14.
+
+Unfortunately, these conditional requires mean there is a bit of extra setup
+with bundlers like SystemJS.
+
+In your SystemJS configuration, you simply need to make sure that the
+following files are labeled as "external", which means they will be ignored:
+
+```
+react/addons
```
nfcampos: since down below you specify the externals for react 0.14 and 15 specifically it's unclear for which version of react this list up here is meant for

docs/guides/systemjs.md
```diff
@@ -0,0 +1,73 @@
...
+react-dom/server
+react-addons-test-utils
+```
+
+## React 15 Compatability
+
+If you are using React 15, your config should include these externals:
+
+```json
+/* package.json */
```
nfcampos: could we be consistent here and for all supported react versions show either a list of externals or the config entry for `package.json`? I don't have a preference either way (maybe one of the other maintainers), but I do think they should all be either one or the other.

nfcampos:
@Aweary I've taken a look and it looks alright to me (other than the 2 comments I left), however, i've no experience with systemjs so I can't be much help there

guybedford:
@Aweary I'm not that familiar with the exact mechanics for these configs, but it seems like the right sort of approach to me here although it would definitely be worth looking at automating this configuration in future. Thanks so much for being open to including instructions for this, and I'll aim to keep an eye on this use case as well.

docs/guides/systemjs.md
tamird: I agree that'd be good, but for the time being this follows the example of the other guides. They should all be updated at the same time (in fact, they probably should be merged since they overlap so greatly); I think that can be done in a follow-up PR.

Aweary:
@tamird if you can address @nfcampos's comment(s) and if @guybedford thinks this looks acceptable I think we can merge this after you merge the latest changes from master.

lelandrichardson:
This LGTM.

@tamird one last thing: You should add this file to : https://github.com/airbnb/enzyme/blob/master/docs/README.md so that it shows up in the docs site!

tamird:
Done.

On Fri, Jul 8, 2016 at 12:22 AM, Leland Richardson <notifications@github.com

Aweary:
Thanks @tamird 🎉

PR #469, opened 2016-06-23
Fix syntax in enzyme changelog

JHilker:
Super minor PR, simply noticed it while I was reading the changelog.


Aweary:
LGTM thanks!

PR #476, opened 2016-06-28
fixes mount().debug() output with mixed children

### LINKED ISSUE #475: Does not render static + dynamic children
I am doing full DOM rendering, when a component contains static OR dynamic children it works. However, when a component contains static AND dynamic children, it does not:

```javascript
let compA = mount(<div>hello</div>);
console.log('A', compA.debug());

let compB = mount(<div>{'world'}</div>);
console.log('B', compB.debug());

let compC = mount(<div>hello {'world'}</div>);
console.log('C', compC.debug());
```

__output:__
```
A <div>
  hello
</div>
B <div>
  world
</div>
C <div />
```

Notice how C is being redered as `<div />`. Expected: `<div>hello world</div>`.

Not sure if it is relevant but I am using jsdom.

nfcampos:
closes #475 

could one of the other maintainers look at this to make sure this change actually makes sense?

### DIFF
--- a/src/Debug.js
+++ b/src/Debug.js
@@ -81,6 +81,10 @@ export function debugInst(inst, indentLength = 2) {
   if (typeof inst === 'string' || typeof inst === 'number') return escape(inst);
   if (!inst) return '';
 
+  if (inst._stringText) {
+    return inst._stringText;
+  }
+
   if (!inst.getPublicInstance) {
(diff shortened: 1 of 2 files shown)


ljharb:
LGTM. should we make sure we have a similar test for shallow debug?

nfcampos:
@ljharb good point, I've added it

PR #481, opened 2016-06-30
add tests for SFCs that return null in react 15

gm758:
React 15 added support for SFCs that return null. I added tests for this scenario in mount and shallow to complement preexisting tests for class components that return null


ljharb:
Sweet, thanks! LGTM

Issue #480, opened 2016-06-30
Clean up tests in prep for react@15.2.0


blainekasten:
Based on this [twitter thread from dan abramov](https://twitter.com/soprano/status/748546919895441408) React 15.2.0 will be adding a warning for props passed through to DOM nodes that aren't DOM attributes. 

I don't _think_ this will have any implications on the API of this library. It only would if someone did something like this:

```js
mount(<div foo />);
```

because we pass the props through, but at that point, that's their fault and issue to fix. 

But internally, we do that a lot with our tests (see screenshot below). So effectively we need to update our tests to not put arbitrary props on DOM nodes. 

This can be tested currently by installing `react@15.2.0-rc.1`

![image](https://cloud.githubusercontent.com/assets/2916023/16496194/6c818ac8-3eb6-11e6-9d04-9ec4cc190b14.png)

ljharb:
There are a number of boolean HTML attributes that should be totally fine - `<input checked />` for example - so hopefully React will maintain a whitelist of these, and not warn for them.

Similarly, any actual HTML attribute without a value, like `<div foo>`, is sugar for `<div foo="foo">`, so that warning would be a very bad idea.

If I'm misunderstanding, and it's solely a warning for actually-invalid HTML attribute names, then I'm on board :-)

blainekasten:
You nailed it. We just have things like this in our tests `<div amount={3} />` which react will warn for. But they do whitelist legitimate attributes.

ljharb:
We should just change those to `data-amount`, `data-foo`, etc.

PR #487, opened 2016-07-03
Remove arbitrary props from DOM elements

AugustinLF:
This is a work in progress, since I'm stuck with a couple of case. Perhaps someone could help me with that? I skipped the two failing tests, I'll comment to explain my difficulties.

### DIFF
--- a/test/ReactWrapper-spec.jsx
+++ b/test/ReactWrapper-spec.jsx
@@ -282,6 +282,15 @@ describeWithDOM('mount', () => {
   });
 
   describe('.find(selector)', () => {
+    let sandbox;
+
+    beforeEach(() => {
+      sandbox = sinon.sandbox.create();
+    });
+
+    afterEach(() => {
+      sandbox.restore();
+    });
 
     it('should find an element based on a class name', () => {
       const wrapper = mount(
@@ -384,7 +393,9 @@ describeWithDOM('mount', () => {
 
     });
 
-    it.skip('should not find components with invalid attributes', () => {
+    it('should not find components with invalid attributes', () => {
+      // React 15.2 warns when setting a non valid prop to an DOM element
+      sandbox.stub(console, 'error');
       // Invalid attributes aren't valid JSX, so manual instantiation is necessary
       const wrapper = mount(
         React.createElement('div', null, React.createElement('span', {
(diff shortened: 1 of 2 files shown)


test/ReactWrapper-spec.js
```diff
@@ -372,7 +372,7 @@ describeWithDOM('mount', () => {
 
     });
 
-    it('should not find components with invalid attributes', () => {
+    it.skip('should not find components with invalid attributes', () => {
```
AugustinLF: The point of this test is to test on purpose wrong attributes that should be ignored by enzyme. I'm not sure there's anything to do. Since I don't like warnings when running tests, I would do:
```
const _warn = console.warn;
console.warn = () => {};
// tests
console.warn = _warn;
```

test/ReactWrapper-spec.js
```diff
@@ -502,16 +502,21 @@ describeWithDOM('mount', () => {
...
-          <button data-test="ref" prop={undefined} />
-          <span data-test="ref" prop={null} />
-          <div data-test="ref" prop={123} />
-          <input data-test="ref" prop={false} />
-          <a data-test="ref" prop />
+          <Foo prop={undefined} />
+          <Foo prop={null} />
+          <Foo prop={123} />
+          <Foo prop={false} />
+          <Foo prop />
```
AugustinLF: Here, all the associated find tests `find({ prop: undefined })` fail when using a React Component instead of a simple element. I'm quite surprised of this behaviour, for me it should behave in the same way, perhaps there is a enzyme behaviour that I'm unaware of?

Aweary: That will likely be resolved by https://github.com/airbnb/enzyme/pull/377. Why did change this from using DOM nodes to a component, though? I think we should keep this test as-is.

ljharb: Or better, keep the DOM tests and still add the Foo tests.

test/ReactWrapper-spec.js
Aweary: We could probably just use `sinon.spy` to mock `console.warn` in the test itself, and revert the mock (does sinon do that by itself? I'm not sure...) in the `beforeEach` block.

test/ReactWrapper-spec.js
Aweary: Since it fails now with components, I can just those tests to #377.

test/ReactWrapper-spec.js
ljharb: sinon-sandbox will do that by itself - if we're not using that yet (and restoring it afterEach test), we should

AugustinLF:
I updated the PR to fix the omitted tests, and rewrite some using `data-amount`, `data-foo`, to follow @ljharb. If everything's fine, I'll squash the commits, and rebase the whole branch!

Aweary:
LGTM @AugustinLF can you resolve the merge conflicts here? I think we can merge afterwards.

blainekasten:
closes #480

AugustinLF:
Done! Should I squash the commits?

test/ReactWrapper-spec.jsx
```diff
@@ -385,6 +394,8 @@ describeWithDOM('mount', () => {
     });
 
     it('should not find components with invalid attributes', () => {
+      // React 15.2 warns when setting a non valid prop to an DOM element
+      sandbox.stub(console, 'error');
```
ljharb: why is this necessary? the test will fail if a non-valid prop is set, won't it?

AugustinLF: The test will pass, it will simply warn in the console.
image

ljharb: If I haven't already, I'll soon be making console warnings and errors fail tests - so if a test would be issuing a warning in a version of React, we shouldn't be running that test in that version of React (since React covers it for us by issuing the warning).

Can we restrict this test to only React < 15, and remove the stub?

AugustinLF: Done! I changed so that the test is only run when react 0.13 or 0.14.

package.json
```diff
@@ -84,6 +84,6 @@
     "sinon": "^1.17.5"
   },
   "peerDependencies": {
-    "react": "0.13.x || 0.14.x || ^15.0.0-0"
+    "react": "0.13.x || 0.14.x || ^15.x"
```
ljharb: Is this change necessary? We want to include the v15 prereleases.

AugustinLF: No it's not anymore. I did it because I wrote that line before the release of 15.2. I was working with 15.2.0-rc.1
 or 2, which doesn't match ^15.0.0-0). I added a commit to fix that.

ljharb: ah, good call. Let's make this `0.13.x || 0.14.x || ^15.0.0-0 || 15.x` then :-)

AugustinLF: Let's hope it's fine now!

ljharb:
@AugustinLF i think you can simply remove the last two commits now :-)

AugustinLF:
@ljharb Yep, it's done!

AugustinLF:
I rebased the branch to fix the conflicts :)

Aweary:
Thanks @AugustinLF!

PR #489, opened 2016-07-07
Update react-native.md

esauter5:
Specify that the docs are talking about React Native 0.18 rather than React 0.18 to avoid confusion.

### DIFF
--- a/docs/guides/react-native.md
+++ b/docs/guides/react-native.md
@@ -1,6 +1,6 @@
 # Using Enzyme to Test Components in React Native
 
-As of React Native 0.18, React Native uses React as a dependency rather than a forked version of the library,
+As of v0.18, React Native uses React as a dependency rather than a forked version of the library,
 which means it is now possible to use Enzyme's `shallow` with React Native components.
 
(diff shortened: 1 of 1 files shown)


docs/guides/react-native.md
```diff
@@ -1,6 +1,6 @@
 # Using Enzyme to Test Components in React Native
 
-As of React 0.18, React Native uses React as a dependency rather than a forked version of the library,
+As of React Native 0.18, React Native uses React as a dependency rather than a forked version of the library,
```
Aweary: Maybe instead we can say "As of version 0.18, React Native" or just "As of  0.18, React Native"? It sounds weird having "React Native" repeat within one word.

esauter5: I thought that sounded awkward as well, but since I was trying to be explicit about what we were talking about, I didn't want to completely remove it.

However, removing it completely still accomplishes the task of making sure it doesnt just say "React" so I agree with this.

Aweary:
LGTM thanks!

PR #490, opened 2016-07-08
ensure ShallowWrapper render output can't get stale

### LINKED ISSUE #360: Removing or documenting the need for Shallow Renderer's .update()
### The problem

When testing interactions with shallow rendered components, what interactions do and don't need `.update()` after them can feel inconsistent. If the component's rendered output doesn't match what you think it should after performing some action, you can end up littering tests with `.update()` hoping to get lucky.

It wasn't until I peeked at the source and saw the `.update()` call in `.simulate()` that everything made sense. Even knowing that unless I'm using a few "blessed" methods I'll need to call `.update()` myself it's still going to trip me up at times.

### The solution
Is it feasible to wrap rendering such that whenever a component is rendered `.update()` is called afterward automagically? Via wrapping .prototype.render, I suppose, though hopefully there's a more clever solution or hook somewhere.

Or at least, an update to `.update()`'s docs to be clear about the cases where it's called for you could help greatly. From what I saw, this is just `.simulate` and `.setProps/State/Context`?

jwbay:
Replace .node and .nodes with 'get' functions in ShallowWrapper to ensure render output is always fresh.  Add equivalent functions in ReactWrapper. Removes need for ShallowWrapper's .update, fixes #360

Many thanks to @guncha for the idea in #360.


src/ShallowWrapper.js
```diff
@@ -80,8 +80,16 @@ export default class ShallowWrapper {
           }
         });
       });
-      this.node = this.renderer.getRenderOutput();
-      this.nodes = [this.node];
+      Object.defineProperty(this, 'node', {
+        get() {
```
ljharb: ugh :-/ getters are incredibly slow and deoptimize all code they touch. I'd rather do a breaking change to make this a method, than use a getter.

blainekasten: I would also be down for using functions. I've always felt nervous about using `node` / `nodes` as I felt like it was reaching into enzyme internals that could be changed without notice.. Functions feel more like a public API.

Aweary: Well, `node` and `nodes` aren't technically public API are they? We don't document them as far as I can see. I agree that using a function would be an improvement, so maybe we can make that change for the next breaking release and then add `node()` and `nodes()` to the API documentation?

ljharb: Everything accessible - not just everything documented - is public API, even if it has a magical underscore in the front of the property name.

Aweary: So any change to any object property would be considered a breaking change? If someone is using `ReactWrapper.complexSelector.find` and we rename/move that, would that require a major release? Maybe we're not talking about the same thing. I usually define public API as the API that we document for the public and that we support.

ljharb: Yes, absolutely it would (when it's not fixing a bug, which admittedly can be a grey area)

If we want to be able to change things without it being a breaking change, then those things have to be *impossible to reach by consumers*, even by reflection.

Aweary: That makes sense. A breaking change doesn't necessarily have to involve an API that we document and explicitly support. Thanks @ljharb 👍 

So yeah, I think implementing `node()` and `nodes()` would be a good idea for the next major release.

guncha: I ran some benchmarks comparing changes here with the commit before it and while I'm sure there's a performance penalty, it gets lost in the noise. Both can do about 35k `wrapper.find("foo")` calls per second.

If the performance is equivalent, why break anyone's code at all?

https://github.com/guncha/enzyme/commit/633b0c5a49d47e3aae76670f333a14e5c1884d65

ljharb: Getters and setters are horrifically bad for readability and maintainability. imo they should never be used, in any project. A property should be a property, not a function.

Since this is not a bug fix, there's no urgency to add this change, and I'd prefer to convert `.node` to a function if we want more behavior than "data property".

guncha: While I agree that getters don't improve readability, this is definitely a bug. Ask anyone with passing familiarity with Enzyme about what this should do and compare it with what actually happens:

```javascript
let wrapper = Enzyme.shallow(<MyInput debounce={ 500 } />)
wrapper.simulate("keypress", {char: "b"})
wrapper.simulate("keypress", {char: "a"})
wrapper.simulate("keypress", {char: "r"})
expect(wrapper.find(".result").text()).toBe("")
jasmine.clock().tick(1000)
expect(wrapper.find(".result").text()).toBe("bar")
```

ljharb: The requirement to manually `update()` is intentional. I agree it's a UX improvement to do it automatically, but it's not a bug.

jwbay: @ljharb I can convert these to functions. Would that be acceptable? Want to make sure because it's a more involved change. 😄  I understand it would have to wait for a 3.x release.

Are there any concerns about ReactWrapper having properties for these and ShallowWrapper having functions?

ljharb: I think we should convert both to functions at the same time, for consistency.

jwbay:
This is ready for a second look. The first commit was only updated to get tests passing on older versions of React/Node.

I didn't switch everything in ReactWrapper to use the new functions because the only benefit would've been for *internal* uniformity at the cost of function calls where they weren't needed. Can switch if desired.

Outstanding questions:
* Should the new functions should be added to documentation?
* How should ShallowWrapper's .update be handled?
  * soft deprecation (current approach), possibly with a logged warning
  * outright removed since this is already a breaking change
  * repurposed to call `.instance().forceUpdate()` for non-SFCs

docs/api/shallow.md
```diff
@@ -169,9 +169,6 @@ Manually sets context of the root component.
 #### [`.instance() => ReactComponent`](ShallowWrapper/instance.md)
 Returns the instance of the root component.
 
-#### [`.update() => ShallowWrapper`](ShallowWrapper/update.md)
```
ljharb: This method still seems useful in case someone mutates state/props without calling the appropriate methods and wants to force an update.

ljharb:
fwiw I think internal uniformity is much more important than avoiding function calls, which are pretty cheap anyways.

jwbay:
Does anyone have further thoughts on this?

ljharb:
The more I think about this, the more I'm confused why there's both "getNode" and "getNodes"?

jwbay:
Don't think there's a reason besides avoiding lots of `[0]` when you're dealing with one node. Everything wrapped in .single would need it. Might be nice for .single to pass you the single node in the callback instead?

ljharb:
If it's just a convenience, perhaps `getNode` should throw when there's more than one? Also I don't think we need two internal sources of truth - `this.nodes` seems sufficient.

jwbay:
Good points, updated!

lelandrichardson:
@jwbay this is a great and very welcome change!

Although I agree w/ @ljharb  that with a `.getNode()` method, keeping track of `.node` and `.nodes` is potentially unnecessary...  however, i'm worried that if we get rid of `.node`, then we might need to consider this a breaking change. Although `.node` isn't really an advertised property of the wrapper object, I know that in practice people use it.

Thoughts?

jwbay:
@lelandrichardson Thanks! I agree this is a breaking change.

ljharb:
I'd also be fine keeping the internal .node updated for now, but removing it in the next major version bump.

jwbay:
So restore the old .update functionality? Should it still call .forceUpdate on the instance as well? It didn't before this PR.

ljharb:
@jwbay ah, if the PR can't be easily made to be semver-minor, then we may as well go full semver-major :-)

jwbay:
Oh I don't mind making the change, especially if it can get this shipped sooner :) I just wasn't sure whether adding .forceUpdate calls where there weren't any before would be a breaking change as well. Seems like it would actually. I'll take a look at what's involved for back-compat.

jwbay:
@ljharb @lelandrichardson we should be semver-minor now 👍

jwbay:
Looks like a lint error snuck into master somehow. Fixed it in this branch but also opened a separate PR just in case it's wanted: #613

lencioni:
This looks good to me. @jwbay will you please rebase onto latest master when you have a chance?

jwbay:
@lencioni done

lencioni:
LGTM

jwbay:
Resolved conflicts.

bookman25:
Is there any chance of this branch getting merged?  This would reduce a lot of boilerplate code for our team.

kentor:
any timeline on releasing this?

ljharb:
I'll cut a few releases after #670 gets in.

ChrisFindlay:
Seems to work under `jest` but not under `jest --coverage`...

PR #494, opened 2016-07-10
docs: add AVA guide

mikenikles:
This adds an example of how to use enzyme with [AVA](https://github.com/avajs/ava).

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ testing your React components, you can consider using [jasmine-enzyme](https://g
 
 [Using Enzyme with Lab](/docs/guides/lab.md)
 
+[Using Enzyme with Tape and AVA](/docs/guides/tape-ava.md)
+
 ### [Installation](/docs/installation/README.md)
 
 To get started with enzyme, you can simply install it with npm:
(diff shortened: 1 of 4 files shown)


ljharb:
We should probably add tape as well - since ava is basically a runner for tape, should those be the same document, or a different one?

mikenikles:
Given that AVA and tape are so similar, I think keeping both in the same document is ok. I can apply that to this PR.

nfcampos:
tape and ava do have some significant differences but they mostly only appear when testing async code. since most test code using enzyme will be synchronous I think they can be the same guide

ljharb:
@mikenikles would you mind rebasing this, so we can get it in?

ljharb:
@mikenikles Thanks! Any chance you'd mind rebasing locally so as to remove the merge commit?

mikenikles:
I clearly made this worse... I gave rebasing a shot based on some docs, any hint you can give me to do that properly?

mikenikles:
Looks better... I hope :)

ljharb:
Thanks!

Issue #508, opened 2016-07-21
containsMatchingElement does not work with JSX expressions


sahat:
Not sure if this is intended or not, but the following returns `false`. My expectation is that it should return `true`.

``` jsx
// component
import React from 'react';

const MyComponent = (props) => {
  const season = 2;
  const episode = 10;

  return (
   <section>
      <div className="thumbnail">Season {season}, Episode {episode}</div>
   </section>
  )
};

export default MyComponent;
```

``` jsx
// test
expect(component.containsMatchingElement(<div className="thumbnail">Season 2, Episode 10</div>)).to.equal(true);
// result: test fails
```

If I set the node's value explicitly to `Season 2, Episode 10` the test passes, but not when using JSX expression.

ljharb:
hmm - it seems like it doesn't *contain* a matching element, it *is* a matching element. What does `component.matchesElement(<div className="thumbnail">Season 2, Episode 10</div>)` output?

sahat:
@ljharb Sorry, I oversimplified the example. It is a nested child inside several divs.

ljharb:
ah k - that matters :-) in that case, i'd expect that to work. If you output `component.debug()`, what do you get?

The likely issue is that the component has `['Season ', season, ', Episode ', episode]` as a child while your test element has the full text as a single string.

eronisko:
I've come across this problem too.

Nodes `<div>mixed child</div>` and `<div>mixed {'child'}</div>` are not considered equal/matching in [`nodeEqual()`](https://github.com/airbnb/enzyme/blob/master/src/Utils.js#L76).

Is there a reason for this behaviour? I'd say that comparison of the values before and after interpolation makes sense (and it makes sense for the nodes to match)

sahat:
@ljharb After running `component.debug()` and printing results to console, I am getting the following output:

**console.log**
```js
<div className="Thumbnail__meta">
Season 
2
, Episode 
10
</div>
```

or

**console.dir**
```js
<div className="Thumbnail__meta">\nSeason \n2\n, Episode \n10\n</div>
```

ljharb:
Gotcha - seems like before comparing for matching elements, we need to replace all consecutive whitespace in text content with a single space.

sahat:
@ljharb Hey Jordan,

I think I was able to find a solution. All tests pass, including the one above.

[...]

PR #512, opened 2016-07-24
Match children before and after interpolation

eronisko:
Proposed fix for #508.

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -134,8 +134,8 @@ function internalNodeCompare(a, b, lenComp, isLoose) {
   const childCompare = isLoose ? childrenMatch : childrenEqual;
   if (leftHasChildren || rightHasChildren) {
     if (!childCompare(
-      childrenToArray(left.children),
-      childrenToArray(right.children),
+      childrenToSimplifiedArray(left.children),
+      childrenToSimplifiedArray(right.children),
       lenComp,
     )) {
       return false;
@@ -166,6 +166,27 @@ export function containsChildrenSubArray(match, node, subArray) {
 
 function arraysEqual(match, left, right) {
   return left.length === right.length && left.every((el, i) => match(el, right[i]));
+}
+
+export function childrenToSimplifiedArray(nodeChildren) {
+  const childrenArray = childrenToArray(nodeChildren);
+  const simplifiedArray = [];
+
+  for (let i = 0; i < childrenArray.length; i += 1) {
+    const child = childrenArray[i];
+    const previousChild = simplifiedArray.pop();
+
+    if (previousChild === undefined) {
+      simplifiedArray.push(child);
+    } else if (isTextualNode(child) && isTextualNode(previousChild)) {
+      simplifiedArray.push(previousChild + child);
+    } else {
+      simplifiedArray.push(previousChild);
+      simplifiedArray.push(child);
+    }
+  }
+
+  return simplifiedArray;
 }
 
 function childrenOfNode(node) {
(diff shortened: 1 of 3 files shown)


ljharb:
I think this LGTM, pending another contrib's OK, and a rebase.

ljharb:
I'm not sure yet if this is a patch or a minor.

sahat:
Probably better to be on a safe side by making it a minor version instead of a patch.

ljharb:
That's not really safer, since neither is supposed to have a breaking change :-)

sahat:
Ah, that came out wrong. I was thinking of a versioning scenario that doesn't apply here and in the case where it does cause a breaking change accidentally by slipping through existing tests. Flip a coin or use your best judgement then. ;)

src/Utils.js
```diff
@@ -125,6 +125,31 @@ function arraysEqual(match, left, right) {
   return left.length === right.length && left.every((el, i) => match(el, right[i]));
 }
 
+export function childrenToSimplifiedArray(nodeChildren) {
+  const childrenArray = childrenToArray(nodeChildren);
+  const simplifiedArray = [];
+
+  for (let i = 0; i < childrenArray.length; i++) {
+    const child = childrenArray[i];
+    const simplifiedChild = simplifiedArray.pop();
```
nfcampos: how about calling this `previousChild`?

test/Utils-spec.jsx
```diff
@@ -155,6 +156,13 @@ describe('Utils', () => {
         )).to.equal(false);
       });
 
+      it('should match children before and after interpolation', () => {
+        expect(nodeEqual(
+          <div>{2}{' children'}{<span />}</div>,
```
nfcampos: can you make this line be `<div>{2}{' children'}{<span />} abc {'hey'}</div>,` so that we test that we work correctly with spaces irrespective of whether they're defined like `{'hey '}there`  or like `{'hey'} there`?

src/Utils.js
```diff
@@ -125,6 +125,31 @@ function arraysEqual(match, left, right) {
...
+
+    if (simplifiedChild === undefined) {
```
nfcampos: can you re-write this with `else if` to get rid of `continue` and appease the linter?
eg.
```js
    if (simplifiedChild === undefined) {
      simplifiedArray.push(child);
    } else if (isTextualNode(child) && isTextualNode(simplifiedChild)) {
      simplifiedArray.push(simplifiedChild + child);
    } else {
      simplifiedArray.push(simplifiedChild);
      simplifiedArray.push(child);
    }
```

nfcampos:
@ljharb if we define majors for enzyme as `some tests that passed/failed with the previous version now will, respectively, fail/pass` then this is a major. if not it's a patch I think

nfcampos:
LGTM pending my minor comment being addressed

ljharb:
@nfcampos it's a minor if it's a new feature, but that suggests it's an intentional change which makes tests have different results, which makes it major. If tests were never supposed to have the current results and this fixes it, it's a patch. So, as you say - just very fuzzy :-/

nfcampos:
LGTM after the changes, now we just need to decide on what kind of change this is

sahat:
Has there been any consensus on whether to release this under minor, major or patch release?

ljharb:
@eronisko are you able to rebase this on top of latest master? it'd be great to get this in.

eronisko:
@ljharb done

Issue #472, opened 2016-06-27
Discussion: Framework "Recommendations"


blainekasten:
I think it'd be good to have a discussion to build a doc for our "recommendation" for testing framework wrapped components (redux). This comes up a lot. The real answer is that it's not this libraries scope to handle framework specifics. But it comes up a lot so we should have a discussion about how to handle it.

Personally I think there are a couple good options here:

* A seperate library for making library specific connections easier. e.g., `enzyme-redux`, `enzyme-coflux`, `enzyme-mobx`.
  * This would provide helper utilities to make it simple (maybe there are already libraries for this?).
* Docs on our recommendations

As far as docs go. It seems like the easiest recommendation that keeps coming up is to suggest exporting the non-framework-wrapped version of a component.

```js
export class PrivateComp { }

export const Component = frameworkWrapper(PrivateComponent, ...);
// or
export default frameworkWrapper(PrivateComponent, ...);
```

Looking to hear from the community on how they handle testing framework wrapped components. Extra curious in MobX users as it's got a lot of attention lately.

jacobrask:
Note that Redux already exposes the unconnected component as the `WrappedComponent` property on the connected component.

Aweary:
This is a problem for anyone who uses higher-order components, if anything we might want to document how to handle those in the general case, and then maybe use Redux or something as a specific example.

jacobrask:
I think there are 2 issues. The first is to test the a component wrapped in a higher-order component. In that case I think the `WrappedComponent` approach works well and can be encouraged.

The other is to test a component that has a child that depends on context, like Redux.

EmilTholin:
I don't want to talk for the entire MobX-community, but I personally think enzyme in conjunction with something lightweight like [tape](https://www.npmjs.com/package/tape) or [blue-tape](https://www.npmjs.com/package/blue-tape) is all that is needed. I don't see the need for an entirely new module:

```js
// counterFactory.js
function counterFactory() {
  let c = observable({
    value: 0
  });
  c.increment = function () {
    ++c.value;
  };
  return c;
};

// IncrementButton.js
const IncrementButton = observer(['counter'], React.createClass({
  render: function () {
    let {counter} = this.props;
    return <button onClick={counter.increment}>{counter.value}</button>;
  }
}));

// App.js
const c = counterFactory();

ReactDOM.render(
  <Provider counter={c}>
    <div>
      <IncrementButton />
    </div>
  </Provider>,
  document.getElementById('app')
[...]

PR #533, opened 2016-08-09
Adding a fluxible guide

jsmey:
Example of testing Fluxible components

### DIFF
--- a/docs/guides/fluxible.md
+++ b/docs/guides/fluxible.md
@@ -1,0 +1,134 @@
+# Using Enzyme with Fluxible
+
+Testing Fluxible higher-order components using Enzyme and Sinon.
+
+First, setup jsdom so we can use Enzyme mount.
+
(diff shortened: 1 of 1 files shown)


Aweary:
@jsmey great write up, we really appreciate the effort. I'm not sure we want to include this though; if you look at our current guides we really only provide official guidance on build systems or testing environments. If we added a guide for fluxible, we couldn't really argue against adding it for other misc. libraries (Redux, Relay, Mobx, etc).

I think if anything we'd want to add a guide that addresses the general case of testing HOCs.

blainekasten:
I agree. This seems like it would be a great blog that you could post on your own! But for the docs I think general guides are better than specifics. That was the goal of #472

Aweary:
I'm going to close this out then, I agree with @blainekasten, it'd be a great blog post!

PR #538, opened 2016-08-11
Fix Mount CSS prop selector for parity with Shallow

### LINKED ISSUE #534: Selecting React Components via CSS prop selector
I was looking at this yesterday and it appears that when using `mount` you can't use CSS prop selectors. I did some digging and found the PR (#70), and after reading through it a bit it appears that it was inertial. Instead of using a string (via CSS prop selector), the go to method was to instead use the object selector. At this point I stopped digging and assumed that the CSS prop selector was not supported in this way. However, when shallow rendering you can use a CSS prop selector with a React Component. 

``` diff
diff --git a/test/ReactWrapper-spec.jsx b/test/ReactWrapper-spec.jsx
index 503eee3..c8d7a65 100644
--- a/test/ReactWrapper-spec.jsx
+++ b/test/ReactWrapper-spec.jsx
@@ -384,6 +384,20 @@ describeWithDOM('mount', () => {

     });

+    it('should find a React Component based on a component props via css selector', () => {
+      class Foo extends React.Component {
+        render() { return <div />; }
+      }
+      const wrapper = mount(
+        <div>
+          <Foo bar="baz" />
+        </div>
+      );
+
+      expect(wrapper.find('Foo[bar="baz"]')).to.have.length(1);
+      expect(wrapper.find('[bar]')).to.have.length(1);
+    });
+
     it('should not find components with invalid attributes', () => {
       // Invalid attributes aren't valid JSX, so manual instantiation is necessary
       const wrapper = mount(
diff --git a/test/ShallowWrapper-spec.jsx b/test/ShallowWrapper-spec.jsx
index 47b570f..faf83b5 100644
--- a/test/ShallowWrapper-spec.jsx
+++ b/test/ShallowWrapper-spec.jsx
@@ -481,6 +481,21 @@ describe('shallow', () => {
       expect(wrapper.find('[data-foo]')).to.have.length(1);
     });

+    it('should find a React Component based on a component props via css selector', () => {
+      class Foo extends React.Component {
+        render() { return <div />; }
+      }
+
+        <div>
+          <Foo bar="baz" />
+        </div>
+      );
+
+      expect(wrapper.find('Foo[bar="baz"]')).to.have.length(1);
+      expect(wrapper.find('[bar]')).to.have.length(1);
+    });
+
     it('should find components with multiple matching react props', () => {
       function noop() {}
       const wrapper = shallow(
```

```
  495 passing (2s)
  10 pending
  1 failing

  1) (uses jsdom) mount .find(selector) should find a React Component based on a component props via css selector:
     AssertionError: expected { Object (component, root, ...) } to have a length of 1 but got 0
      at Context.<anonymous> (ReactWrapper-spec.jsx:397:54)
```

Is this intentional for shallow but not for mount?

travisperson:
Selector such as `Foo[bar="baz"]` previously would not return any components when rendering with `mount`.

Resolves #534

### DIFF
--- a/src/MountedTraversal.js
+++ b/src/MountedTraversal.js
@@ -84,7 +84,6 @@ export function instHasType(inst, type) {
 }
 
 export function instHasProperty(inst, propKey, stringifiedPropValue) {
-  if (!isDOMComponent(inst)) return false;
   const node = getNode(inst);
   const nodeProps = propsOfNode(node);
   const descriptor = Object.getOwnPropertyDescriptor(nodeProps, propKey);
--- a/test/ReactWrapper-spec.jsx
+++ b/test/ReactWrapper-spec.jsx
@@ -384,6 +384,20 @@ describeWithDOM('mount', () => {
 
     });
 
+    it('should find React Component based on a components prop via css selector', () => {
+      class Foo extends React.Component {
+        render() { return <div />; }
+      }
+      const wrapper = mount(
+        <div>
+          <Foo bar="baz" />
+        </div>
+      );
+
+      expect(wrapper.find('[bar]')).to.have.length(1);
+      expect(wrapper.find('Foo[bar="baz"]')).to.have.length(1);
+    });
+
     it('should not find components with invalid attributes', () => {
       // Invalid attributes aren't valid JSX, so manual instantiation is necessary
       const wrapper = mount(
--- a/test/ReactWrapper-spec.jsx
+++ b/test/ReactWrapper-spec.jsx
@@ -386,16 +386,16 @@ describeWithDOM('mount', () => {
 
     it('should find React Component based on a components prop via css selector', () => {
       class Foo extends React.Component {
-        render() { return <div />; }
+        render() { return <div {...this.props} />; }
       }
       const wrapper = mount(
         <div>
-          <Foo bar="baz" />
+          <Foo role="button" />
         </div>
       );
 
-      expect(wrapper.find('[bar]')).to.have.length(1);
-      expect(wrapper.find('Foo[bar="baz"]')).to.have.length(1);
+      expect(wrapper.find('[role]')).to.have.length(2);
+      expect(wrapper.find('Foo[role="button"]')).to.have.length(1);
     });
 
     it('should not find components with invalid attributes', () => {
--- a/test/ShallowWrapper-spec.jsx
+++ b/test/ShallowWrapper-spec.jsx
@@ -483,16 +483,16 @@ describe('shallow', () => {
 
     it('should find a React Component based on a components prop via css selector', () => {
       class Foo extends React.Component {
-        render() { return <div />; }
+        render() { return <div {...this.props} />; }
       }
       const wrapper = shallow(
         <div>
-          <Foo bar="baz" />
+          <Foo role="button" />
         </div>
       );
 
-      expect(wrapper.find('[bar]')).to.have.length(1);
-      expect(wrapper.find('Foo[bar="baz"]')).to.have.length(1);
+      expect(wrapper.find('[role]')).to.have.length(1);
+      expect(wrapper.find('Foo[role="button"]')).to.have.length(1);
     });
 
     it('should find components with multiple matching react props', () => {
--- a/test/ShallowWrapper-spec.jsx
+++ b/test/ShallowWrapper-spec.jsx
@@ -481,6 +481,20 @@ describe('shallow', () => {
       expect(wrapper.find('[data-foo]')).to.have.length(1);
     });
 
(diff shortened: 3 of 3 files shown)


blainekasten:
lgtm

travisperson:
I think this is going to introduce similar issues as #377. I went a head and passed the props all the way down to the child div and the `ReactWrapper` test now fails as it finds two items.

```diff
diff --git a/test/ReactWrapper-spec.jsx b/test/ReactWrapper-spec.jsx
index abb1aee..305f388 100644
--- a/test/ReactWrapper-spec.jsx
+++ b/test/ReactWrapper-spec.jsx
@@ -386,16 +386,16 @@ describeWithDOM('mount', () => {
 
     it('should find React Component based on a components prop via css selector', () => {
       class Foo extends React.Component {
-        render() { return <div />; }
+        render() { return <div {...this.props} />; }
       }
       const wrapper = mount(
         <div>
-          <Foo bar="baz" />
+          <Foo role="button" />
         </div>
       );
 
-      expect(wrapper.find('[bar]')).to.have.length(1);
-      expect(wrapper.find('Foo[bar="baz"]')).to.have.length(1);
+      expect(wrapper.find('[role]')).to.have.length(1);
+      expect(wrapper.find('Foo[role="button"]')).to.have.length(1);
     });
 
     it('should not find components with invalid attributes', () => {
diff --git a/test/ShallowWrapper-spec.jsx b/test/ShallowWrapper-spec.jsx
index 6ec1331..5f2f753 100644
--- a/test/ShallowWrapper-spec.jsx
+++ b/test/ShallowWrapper-spec.jsx
@@ -483,16 +483,16 @@ describe('shallow', () => {
 
     it('should find a React Component based on a components prop via css selector', () => {
       class Foo extends React.Component {
-        render() { return <div />; }
+        render() { return <div {...this.props} />; }
       }
       const wrapper = shallow(
         <div>
-          <Foo bar="baz" />
+          <Foo role="button" />
         </div>
       );
 
-      expect(wrapper.find('[bar]')).to.have.length(1);
-      expect(wrapper.find('Foo[bar="baz"]')).to.have.length(1);
+      expect(wrapper.find('[role]')).to.have.length(1);
+      expect(wrapper.find('Foo[role="button"]')).to.have.length(1);
     });
 
     it('should find components with multiple matching react props', () => {
```

```
  1) (uses jsdom) mount .find(selector) should find React Component based on a components prop via css selector:
     AssertionError: expected { Object (component, root, ...) } to have a length of 1 but got 2
      at Context.<anonymous> (ReactWrapper-spec.jsx:397:46)
```

Aweary:
Seems like we have to choose between these two behaviors. I would say that returning `2` is the correct behavior as you do have two instances where `role` is set to `"button"`, but I know that's probably not what people would expect.

blainekasten:
Yeah, i'm with @aweary. I feel like that makes the most sense. If `.debug` printed out multiple items with a prop, but a selector doesn't find each of them, I'd feel mislead. 

I feel like we should probably get some input from @ljharb / @lelandrichardson, but if we do take this, it seems like it should probably be a major change as it might break some peoples tests.

evisong:
Hi, team, 

Is there any updates on this?

Thanks.

ljharb:
If this would make existing tests that pass start to fail, then I think we need to have examples of those tests in master first - such that this PR has to change them to pass, making it semver-major.

kcarnold:
This bit me in writing one of my first Enzyme tests: I was expecting to find a child React node via a prop selector, and nothing matched :( This in spite of the [documentation](http://airbnb.io/enzyme/docs/api/selector.html#4-object-property-selector) saying "Enzyme allows you to find **components** and nodes based on a subset of their properties".

Newbie thought process: I expect `wrapper.find(query)` to find "things" that match `query`. Since React encourages me to think of components like just slightly special nodes, I expect "things" to include both nodes and components. If I happen to have both a component and a node that have some property, I might not realize that my selector is going to match both of them, and when my test fails because it accidentally matched two elements I'd puzzle over it for a while, but I'd eventually realize what's going on without digging through the Enzyme testsuite (which is, uh, what I did...). I'd then probably try `wrapper.findComponent` or `wrapper.findNode` if I needed to be more specific. I thus propose adding those.

(Regarding the number of items matched: since I'm used to `querySelector` vs `querySelectorAll`, I excepted `find` to return the _first_ thing and `findAll` to give all of them. But I guess the design follows [jquery](https://api.jquery.com/find/).)

ljharb:
I don't think this is relevant any more now that we're on v3. Can someone confirm?

aweary:
@ljharb that's correct, this isn't relevant. We don't use `isDOMComponent` anymore. Thanks for doing this @travisperson, sorry we didn't get it in!

PR #563, opened 2016-08-26
Fix valid + falsy propvalues

### LINKED ISSUE #562: Attribute selector doesn't work for falsy prop values
For example, 
`wrapper.find('[data-page=0]')` - this selector returns all nodes with _data-page_ attribute, 
but `wrapper.find('[data-page=1]') ` works fine.

Looks like a problem with conditional statement (_instHasProperty_ function)
```javascript
if (propValue) {
    return nodePropValue === propValue;
}
return nodeProps.hasOwnProperty(propKey);
```

ljharb:
Fixes #562.

I also consolidated some code.

### DIFF
--- a/src/MountedTraversal.js
+++ b/src/MountedTraversal.js
@@ -2,7 +2,6 @@ import isEmpty from 'lodash/isEmpty';
 import isSubset from 'is-subset';
 import {
   internalInstance,
-  coercePropValue,
   nodeEqual,
   propsOfNode,
   isFunctionalComponent,
@@ -13,6 +12,9 @@ import {
   SELECTOR,
   nodeHasType,
 } from './Utils';
+import {
+  nodeHasProperty,
+} from './ShallowTraversal';
 import {
   isDOMComponent,
   isCompositeComponent,
@@ -85,26 +87,10 @@ export function instHasType(inst, type) {
 
 export function instHasProperty(inst, propKey, stringifiedPropValue) {
   if (!isDOMComponent(inst)) return false;
+
   const node = getNode(inst);
-  const nodeProps = propsOfNode(node);
-  const descriptor = Object.getOwnPropertyDescriptor(nodeProps, propKey);
-  if (descriptor && descriptor.get) {
-    return false;
-  }
-  const nodePropValue = nodeProps[propKey];
 
-  const propValue = coercePropValue(propKey, stringifiedPropValue);
-
-  // intentionally not matching node props that are undefined
-  if (nodePropValue === undefined) {
-    return false;
-  }
-
-  if (propValue || propValue === 0) {
-    return nodePropValue === propValue;
-  }
-
-  return Object.prototype.hasOwnProperty.call(nodeProps, propKey);
+  return nodeHasProperty(node, propKey, stringifiedPropValue);
 }
 
 // called with private inst
--- a/src/MountedTraversal.js
+++ b/src/MountedTraversal.js
@@ -11,10 +11,8 @@ import {
   AND,
   SELECTOR,
   nodeHasType,
-} from './Utils';
-import {
   nodeHasProperty,
-} from './ShallowTraversal';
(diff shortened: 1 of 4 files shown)


nfcampos:
is `''` a valid prop value as well?

ljharb:
@nfcampos good call, i'll cover that also.

nfcampos:
and while you're at it `null` and `NaN` as well?

ljharb:
Are `null` and `NaN` actually selectable propValues with the string selector syntax? This PR is quickly growing large in scope…

ljharb:
Doesn't React strip out null properties?

nfcampos:
only `undefined` I believe
https://esnextb.in/?gist=899331d29dda61430d88278ddb44a5d8

ljharb:
Looks like it doesn't strip out `false` either.

ljharb:
Ready for a stamp!

src/Utils.js
```diff
@@ -213,6 +213,19 @@ export function coercePropValue(propName, propValue) {
     return propValue;
   }
 
+  // can be the empty string
+  if (propValue === '') {
+    return propValue;
+  }
+
+  if (is('NaN', propValue)) {
```
nfcampos: any reason this isn't `propValue !== 'NaN'`?

nfcampos: I mean ===

src/Utils.js
```diff
@@ -213,6 +213,19 @@ export function coercePropValue(propName, propValue) {
     return propValue;
   }
 
+  // can be the empty string
```
nfcampos: should these three checks here or after the trim like the ones for true and false?

src/MountedTraversal.js
```diff
@@ -14,6 +13,9 @@ import {
   nodeHasType,
 } from './Utils';
 import {
+  nodeHasProperty,
```
nfcampos: should this go into Utils?

src/Utils.js
ljharb: i'll change it; i'd been using a literal `NaN` and then updated it to take in a string

src/MountedTraversal.js
ljharb: Sure, why not

src/Utils.js
ljharb: It didn't feel right having `' '` be treated as the empty string, or `'   NaN'` be treated as a literal NaN - I think the current behavior of throwing for those cases is appropriate.

nfcampos: makes sense, but shouldn't we do the same for true and false then?

test/ShallowTraversal-spec.jsx
```diff
@@ -112,6 +112,35 @@ describe('ShallowTraversal', () => {
...
+    it('should work with null', () => {
+      expect(nodeHasProperty(<div foo={null} />, 'foo', 'null')).to.equal(true);
+      expect(nodeHasProperty(<div foo={0} />, 'foo', 'null')).to.equal(false);
+    });
+
+    it('should work with false', () => {
+      expect(nodeHasProperty(<div foo={false} />, 'foo', 'false')).to.equal(true);
+      expect(nodeHasProperty(<div foo={0} />, 'foo', 'false')).to.equal(false);
+    });
+
```
nfcampos: does this work for `Infinity`?

src/Utils.js
ljharb: Yes, but not in this PR, because that would make it a breaking change

nfcampos: yep I know, but is it actually worth it to change it seeing as it'd be a such a minor, yet breaking, change?

test/ShallowTraversal-spec.jsx
ljharb: Yes; added a test to confirm.

nfcampos:
LGTM

PR #571, opened 2016-08-31
Remove redundant `exposedProperties` array

jacobwalton:
I couldn't see any reason for the `exposedProperties` array in `withDom.js`. Seems to be redundant so this pull request removes it. 

/cc @lelandrichardson


Aweary:
Yeah it seems to be unneeded as far as I can tell, good catch! LGTM

PR #576, opened 2016-09-01
Fixed issue with undefined being wrapped as nodes while constructing …

### LINKED ISSUE #554: ref('...').isEmpty() is always false
Example of `ref()` usage with unexpected behavior:

```js
const wrapper = mount(<Component />);
const x = wrapper.ref('definitely-not-a-ref');
x.node // => undefined
x.nodes // => actual [undefined], expected []
x.length // => actual 1, expected 0
x.isEmpty() // => actual false, expected true
```

Contrast with behavior of `find`

```js
const wrapper = mount(<Component />);
const x = wrapper.find('.definitely-not-a-class');
x.node // => undefined
x.nodes // => []
x.length // => 0
x.isEmpty() // => true
```

Andarist:
…ReactWrapper (fixes #554)

### DIFF
--- a/test/ReactWrapper-spec.jsx
+++ b/test/ReactWrapper-spec.jsx
@@ -3005,5 +3005,18 @@ describeWithDOM('mount', () => {
         expect(wrapper.name()).to.equal('div');
       });
     });
+
+    describe('.ref()', () => {
+      it('unavailable ref should return empty nodes', () => {
+        class WithoutRef extends React.Component {
+          render() { return <div />; }
+        }
+        const wrapper = mount(<WithoutRef />);
+        const ref = wrapper.ref('not-a-ref');
+
+        expect(ref.length).to.equal(0);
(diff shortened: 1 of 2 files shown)


test/ReactWrapper-spec.jsx
```diff
@@ -3005,5 +3005,18 @@ describeWithDOM('mount', () => {
...
+        class WithoutRef extends React.Component {
+          render() { return <div />; }
+        }
+        const wrapper = mount(<WithoutRef />);
+        const ref = wrapper.ref('not-a-ref');
+
+        expect(ref.length).to.equal(0);
+        expect(ref.isEmpty()).to.equal(true);
+      });
+    });
```
Aweary: Can you name the top level `describe` block `.ref()` so we can put future tests for `ref` in there too? Thanks!

Andarist:
@Aweary corrected with your suggestion

Aweary:
LGTM, thanks

PR #580, opened 2016-09-05
Add documentation note clarifying shallow wrapper prop functions

### LINKED ISSUE #397: Documentation ambiguous about what .props are.
wrapper.state() returns the this.state from your component so logically one assumes that wrapper.props() returns the this.props from your component, but no it returns the HTML element properties from the root element.
You obtain the this.props from wrapper.instance().props

Confusing and should be pointed out in the documentation.

kevinzwh:
Fixes #397

As many others have stated, the documentation for .props & .prop(key) are fairly ambiguous on which props are being return for shallow wrappers. It's unclear that the props that are actually being returned are from the **parent element of the component**, since it's easy to mistakenly assume that `props()` would return the props you would recieve from a React component as with `this.props`

Changes proposed:

- Add notes to both `.prop(key)` and `.props()` documentation in the ShallowWrapper API warning newcomers about what **props are being returned & what props are not**

- Add references to `instance()` to show how to retrieve the instance props as with the usual React `this.props` pattern.

- Update examples for both pages to demonstrate the key difference between the props returned & the instance props.

### DIFF
--- a/docs/api/ShallowWrapper/prop.md
+++ b/docs/api/ShallowWrapper/prop.md
@@ -3,14 +3,15 @@
 Returns the prop value for the root node of the wrapper with the provided key.
 `.prop(key)` can only be called on a wrapper of a single node.
 
-NOTE: When called on a shallow wrapper, `.prop(key)` will only return values for props that exist for the parent element of the component, not the whole React component. To return the props for the
-entire React component, use `wrapper.instance().props`. See [`.instance() =>
-ReactComponent`](instance.md)
+NOTE: When called on a shallow wrapper, `.prop(key)` will return values for
+props on the root node that the component *renders*, not the component itself.
+To return the props for the entire React component, use `wrapper.instance().props`.
+See [`.instance() => ReactComponent`](instance.md)
 
 #### Arguments
 
 1. `key` (`String`): The prop name such that this will return value will be the `this.props[key]`
-of the parent element of the component.
+of the root node of the component.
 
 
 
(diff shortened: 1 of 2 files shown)


docs/api/ShallowWrapper/prop.md
```diff
@@ -1,22 +1,44 @@
 # `.prop(key) => Any`
 
-Returns the prop value for the node of the current wrapper with the provided key.
+Returns the prop value for the root node of the wrapper with the provided key.
+`.prop(key)` can only be called on a wrapper of a single node.
 
-NOTE: can only be called on a wrapper of a single node.
+NOTE: When called on a shallow wrapper, `.prop(key)` will only return values for props that exist for the parent element of the component, not the whole React component. To return the props for the
```
ljharb: this might be clearer stated as `… will return values for props on the root node that the component *renders*, not the component itself. To return the props…`

docs/api/ShallowWrapper/prop.md
```diff
@@ -1,22 +1,44 @@
...
+entire React component, use `wrapper.instance().props`. See [`.instance() =>
+ReactComponent`](instance.md)
 
 #### Arguments
 
 1. `key` (`String`): The prop name such that this will return value will be the `this.props[key]`
-of the component instance.
+of the parent element of the component.
```
ljharb: similarly, "parent element" isn't really the right verbage here - it's "of the root node that the component *renders*"

ljharb:
Added a couple comments - I'd like to remove the confusing term "parent element" since that describes something different.

kevinzwh:
Thanks @ljharb for the comments! Went ahead and added your suggestions in the latest commit ea7eb4e to correct the terminology used by referring to "root node" instead.

kevinzwh:
Revisiting this PR! Thoughts on this, @ljharb?

ljharb:
@kevinzwh if you'd mind rebasing onto latest master, so there's no merge commits, I think this should be good to go.

kevinzwh:
@ljharb sure - rebased onto latest master without merge commits ✅

ljharb:
LGTM

kamalchaturvedi15:
For a shallow wrapper, you can call a wrapper.instance().props.propName to get the value of respective prop.

ljharb:
@kamalchaturvedi15 or just `wrapper.prop('propName')` which is more reliable.

Issue #570, opened 2016-08-31
Cannot fetch prop value on shallow render


vrinek:
https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/prop.md contains the following example: 

```jsx
const wrapper = shallow(<MyComponent foo={10} />);
expect(wrapper.prop('foo')).to.equal(10);
```

When running it though, it does not work.

## Full example used

```jsx
import React from "react";

class MyComponent extends React.Component {
    render() {
        return (
            <div>
                {this.props.foo}
            </div>
        );
    }
}

MyComponent.propTypes = {
    foo: React.PropTypes.number
};

import {mount, shallow} from 'enzyme';
import {expect} from 'chai';

describe.only('enzyme test', () => {
    let wrapper;

    context('mount', () => {
        beforeEach(() => {
            wrapper = mount(<MyComponent foo={10} />);
        });

        it('debug works fine', () => { // this works
            expect(wrapper.debug()).to.contain('10');
        });

        it('prop works fine', () => { // this works
            expect(wrapper.prop('foo')).to.equal(10);
        });
    });

    context('shallow', () => {
        beforeEach(() => {
            wrapper = shallow(<MyComponent foo={10} />);
        });

        it('debug works fine', () => { // this works
            expect(wrapper.debug()).to.contain('10');
        });

        it('prop works fine', () => { // this does not work
            expect(wrapper.prop('foo')).to.equal(10);
        });
    });
});
```

(Potentially relevant discussion: https://github.com/airbnb/enzyme/issues/331)

ljharb:
The wrapper represents what MyComponent renders, not MyComponent itself. The div it renders doesn't have any props.

There's no value anyway in rendering something with a prop, and then immediately testing that it has that prop - you aren't validating anything except the immediately preceding line of code in your test.

Aweary:
Yeah this is expected behavior, see https://github.com/airbnb/enzyme/issues/331#issuecomment-230578503. We should update that example though since it's incorrect. @vrinek want to open a PR with a better example?

vrinek:
@ljharb the fact that `shallow` & `mount` behave differently is alarming to me. I would expect them to represent something on the same level. Instead it seems like one represents the actual MyComponent and the other what's inside MyComponent. For reference, my understanding on the distinction of shallow vs mount is that mount will render every child component while shallow will not.

[...]

PR #583, opened 2016-09-06
Replace shallow wrapper .prop() example

vrinek:
The old example was not working

Fixes https://github.com/airbnb/enzyme/issues/570

### DIFF
--- a/docs/api/ShallowWrapper/prop.md
+++ b/docs/api/ShallowWrapper/prop.md
@@ -2,7 +2,7 @@
 
 Returns the prop value for the node of the current wrapper with the provided key.
 
-NOTE: can only be called on a wrapper of a single node. Cannot be called on the root wrapper because that represents what the component renders, not the component itself.
+NOTE: can only be called on a wrapper of a single node. In contrast to a mounted component's `.prop()`, this one operates on the **root element** of the component, not the component itself.
 
 #### Arguments
 
@@ -11,10 +11,10 @@ of the component instance.
 
 #### Example
 
-
 ```jsx
-const wrapper = shallow(<div><MyComponent foo={10} /></div>);
-expect(wrapper.find('MyComponent').prop('foo')).to.equal(10);
+const MyComponent = () => <div bar={10}/>;
+const wrapper = shallow(<MyComponent/>);
+expect(wrapper.find('MyComponent').prop('bar')).to.equal(10);
 ```
 
 #### Related Methods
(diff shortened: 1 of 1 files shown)


docs/api/ShallowWrapper/prop.md
```diff
@@ -2,24 +2,21 @@
 
 Returns the prop value for the node of the current wrapper with the provided key.
 
-NOTE: can only be called on a wrapper of a single node.
+NOTE: can only be called on a wrapper of a single node. Cannot be called on the root wrapper because that represents what the component renders, not the component itself.
```
ljharb: It actually can be called on the root wrapper, but it will return the props on the rendered root element.

docs/api/ShallowWrapper/prop.md
```diff
@@ -2,24 +2,21 @@
...
-
-
 #### Example
 
 
 ```jsx
-const wrapper = shallow(<MyComponent foo={10} />);
-expect(wrapper.prop('foo')).to.equal(10);
+const wrapper = shallow(<div><MyComponent foo={10} /></div>);
+expect(wrapper.find('MyComponent').prop('foo')).to.equal(10);
```
ljharb: Instead, we should leave line 16 as it was, but include an example `MyComponent` implementation - and the root node it renders should have a different "foo" prop - and we should have an assertion that shows that `wrapper.prop('foo')` is what's in the `MyComponent` implementation, and another that shows that `wrapper.instance().prop('foo')` is `10`.

docs/api/ShallowWrapper/prop.md
vrinek: I'm confused. In the issue (linked above) I demonstrated that this is not the case with a test. What am I missing?

docs/api/ShallowWrapper/prop.md
vrinek: :+1: on demonstrating this with a prop on the root node of `MyComponent`.

About the second one, I get "TypeError: wrapper.instance(...).prop is not a function" which is to be expected as `.instance()` returns the ReactComponent, not the wrapper and `.prop()` is a function on the wrapper. Could you clarify what you meant by it?

docs/api/ShallowWrapper/prop.md
vrinek: Never mind, in combination with your next comment, this makes sense now.

docs/api/ShallowWrapper/prop.md
ljharb: An sorry. `wrapper.instance().props.foo` is what I meant.

vrinek:
It looks like this has been fixed by https://github.com/airbnb/enzyme/pull/580. I'm closing this one.

PR #595, opened 2016-09-12
fixes #590: change isCompoundSelector to not match prop selector

### LINKED ISSUE #590: find by attribute doesn't work if string has a dot
Given this component:

```js
import React from 'react';
export class Test extends React.Component {
  render() {
    return (
    <div>
      <span data-ex="works"/>
      <span data-ex="doesnt.work"/>
    </div>
    );
  }
}
```

The method `find` correctly identifies the first span with a data attribute (no dot in the string) but fails to identify the second one (data attribute has a dot)

```js
import React from 'react';
import { Test } from './Test';

describe("Test", ()=>{
  it("has first span", ()=>{
    const rendered = shallow(<Test />);
    const elements = rendered.find('[data-ex="works"]');
    expect(elements.length).toBe(1); // Success
  })

    it("has second span", ()=>{
    const rendered = shallow(<Test />);
    const elements = rendered.find('[data-ex="doesnt.work"]');
    expect(elements.length).toBe(1); // Fails
  })
})
```

joeduncan:


Aweary:
Thanks for this @joeduncan, all the extra tests are awesome. Since all of our existing tests look good as well, this LGTM.

I'd like to keep this open for review since I'm not 100% with regexes, so maybe someone like @ljharb can take a look as well.

src/Utils.js
```diff
@@ -189,7 +189,7 @@ export function selectorError(selector, type = '') {
   );
 }
 
-export const isCompoundSelector = /([a-z]\.[a-z]|[a-z]\[.*\]|[a-z]#[a-z])/i;
+export const isCompoundSelector = /^[\.#]?-?[_a-z]+[_a-z0-9-]*[\.\[#]/i;
```
ljharb: i wish we had commented regexps for situations like this. Let's see:
```js
/
  ^               # start
  [\.#]?          # anything but a hash, 0 or 1
  -?              # a hyphen, 0 or 1
  [_a-z]+         # any letter or underscore, 1 or more
  [_a-z0-9-]*     # any letter, underscore, number, or hyphen, 0 or more
  [\.\[#]         # any dot, left bracket, or hash
/i
```

Even after that I'm not really sure how to reason out what it's matching against. I wonder if we could write this as multiple different regexes, instead of trying to combine it into one case?

Aweary: > Even after that I'm not really sure how to reason out what it's matching against. I wonder if we could write this as multiple different regexes, instead of trying to combine it into one case?

Agreed, that's the approach I took with https://github.com/airbnb/enzyme/pull/591 because using a single regex was just too confusing. Breaking this into multiple regexes seems reasonable, but at a certain point it feels like we'd be implementing our own parser and I wonder if it would be better to just put the work into making https://github.com/airbnb/enzyme/issues/534.

ljharb: That also seems reasonable.

nfcampos: `[\.#]?          # anything but a hash, 0 or 1`
not sure it matters here but I think this one is actually '. or a #, 0 or 1'

ljharb: Whoops, you're right. Will update my comment

joeduncan:
Is there anything I can do here?

src/Utils.js
lencioni: We could bring in [xregexp](http://xregexp.com/) if we wanted to make this regexp more readable with comments. I generally think that regexps should have comments, but I honestly don't find this one to be too bad.

Also, inside of brackets, I don't think you need to escape things like `.` and `[`, so you could remove a couple of backslashes to clean it up a bit.

Overall, this seems to be an improvement so I think we might as well merge it in. In the long run, if we want to support the full range of CSS selectors, I agree with @jazen that we should bring in a parser.

lencioni:
@joeduncan Thanks for your contribution! When you have a moment, will you please rebase this onto the latest master branch?

src/Utils.js
Aweary: @lencioni we have an open PR for implementing a parser https://github.com/airbnb/enzyme/pull/458

joeduncan:
Rebased and ready for merge.

PR #608, opened 2016-09-26
Document CSS selectors when using shallow mounting

Wilfred:
I was a little confused by this today, and realised that selectors don't apply to rendering of child components. As far as I can see, this isn't currently documented in the selectors docs.

I propose adding a note to the docs about this. Alternatively, we could add the note to `.find` on https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/find.md instead. Let me know what you think.

### DIFF
--- a/docs/api/selector.md
+++ b/docs/api/selector.md
@@ -14,6 +14,34 @@ follows:
 - id syntax (`#foo`, `#foo-bar`, etc.)
 - prop syntax (`[htmlFor="foo"]`, `[bar]`, `[baz=1]`, etc.);
 
+**Note -- shallow rendering**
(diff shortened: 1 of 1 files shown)


ljharb:
We can add this if other contributors feel it necessary, but this is kind of the sole purpose of shallow rendering - to render shallowly, ie, not to render child components.

nfcampos:
hmm yes, I agree with @ljharb this is the whole point of the shallow renderer, if this is confusing we should instead better explain the difference between shallow and mount

Aweary:
I agree, if anything we should clarify shallow rendering elsewhere, or link to resources that explain it. I don't think the selectors API doc is the place to do that, but we appreciate the PR @Wilfred!

PR #617, opened 2016-09-30
Fixes #509: pass callback on setState and setProps

### LINKED ISSUE #509: Pass callback to wrapper.setState
React's `setState` method allows to pass a callback called after the state has been changed.

I think it would be useful to be able to pass a callback the same way for asynchronous testing purpose.

Example :

```javascript
it("should render childComponent when state says so", (done) => {
    const wrapper = shallow(<ParentComponent />);
    wrapper.setState({renderChild : true}, () => {
      expect(wrapper.find(childComponent)).to.have.length(1);
      done();
    });
  });
```

alecrobins:
Resolves https://github.com/airbnb/enzyme/issues/509

Due to the asynchronous nature of setState, we can't assume a `wrapper.setState` or `wrapper.setProps` (which uses `setState` in the `setChildProps` call) will act synchronously.

This PR adds callback functions to the `stateState` and `setProps` functions for full DOM rendering and `setState` for shallow rendering. 

Example: 

```
  constructor(props) {
    super(props);
    this.state = { id: 'foo' };
  }
  render() {
    return (
      <div className={this.state.id} />
    );
  }
}
expect(wrapper.state()).to.eql({ id: 'foo' });
wrapper.setState({ id: 'bar' }, () => {
  expect(wrapper.state()).to.eql({ id: 'bar' });
});
```


src/ReactWrapper.jsx
```diff
@@ -224,13 +225,14 @@ export default class ReactWrapper {
...
+   * @param {Function} cb - callback function
    * @returns {ReactWrapper}
    */
-  setState(state) {
+  setState(state, cb) {
     if (this.root !== this) {
       throw new Error('ReactWrapper::setState() can only be called on the root');
     }
-    this.instance().setState(state);
+    this.instance().setState(state, cb);
```
ljharb: does react handle throwing if `cb` isn't a function?

src/ReactWrapper.jsx
```diff
@@ -224,13 +225,14 @@ export default class ReactWrapper {
    * NOTE: can only be called on a wrapper instance that is also the root instance.
    *
    * @param {Object} state to merge
+   * @param {Function} cb - callback function
    * @returns {ReactWrapper}
    */
-  setState(state) {
+  setState(state, cb) {
```
ljharb: since this param is optional, it should default to `undefined` so it doesn't count towards the function's length (same throughout)

test/ReactWrapper-spec.jsx
```diff
@@ -968,6 +968,23 @@ describeWithDOM('mount', () => {
...
+              {this.props.id}
+            </div>
+          );
+        }
+      }
+      const wrapper = mount(<Foo id="foo" />);
+      expect(wrapper.find('.foo').length).to.equal(1);
+      wrapper.setProps({ id: 'bar', foo: 'bla' }, () => {
+        expect(wrapper.find('.bar').length).to.equal(1);
+      });
```
ljharb: is there a way we could add a sync assertion that the length is NOT equal to 1, to prove that the callback is doing its job?

src/ReactWrapper.jsx
alecrobins: Yea seems like it's taken care of here: 
https://github.com/facebook/react/blob/master/src/renderers/shared/stack/reconciler/ReactUpdateQueue.js#L257

ljharb: Awesome - it would still be good to add tests for that behavior, in case a future version of react changes it

alecrobins: Yea sounds good. I'll add a test.

test/ReactWrapper-spec.jsx
alecrobins: It is difficult to force `setState` to act asynchronously. In most cases it does act synchronously it's just not a guarantee.

ljharb: hm - the motivation for adding this callback is those cases where it does act async. there must be some likely usage pattern that forces it?

src/ReactWrapperComponent.jsx
```diff
@@ -32,9 +32,9 @@ export default function createWrapperComponent(node, options = {}) {
       };
     },
 
-    setChildProps(newProps) {
+    setChildProps(newProps, cb = undefined) {
```
Aweary: Wont `cb` be `undefined` by default if its not passed in?

Aweary: What are the advantages of doing that here? Specifically for `setState` which is meant to shadow the React API, which does not make this explicit.

ljharb: Function lengths are intended to only count required arguments - the language follows this for the most part. You're right that React doesn't make this explicit, but I'd prefer to correct their mistake for our API.

test/ReactWrapper-spec.jsx
koba04: I'm not sure usage patterns that setState does act async in testing. but I think you can use `batchedUpdates`  like this.

```js
import { batchedUpdates } from '../src/react-compat';

batchedUpdates(() => {
  wrapper.setProps({ id: 'bar', foo: 'bla' }, () => {
    expect(wrapper.find('.bar').length).to.equal(1);
  });
  expect(wrapper.find('.bar').length).to.equal(0);
});
```

alecrobins: thanks @koba04. Using `batchedUpdates` worked.

aronwoost:
Awesome work! Well, should it be shipped?

alecrobins:
@ljharb I added a test where the callback is called asynchronously ( https://github.com/airbnb/enzyme/pull/617/commits/7b94db81420d9db71703dddfbff0f5739ecb374e#diff-2f849fa739461853c70ddd193891074fR976 )

ljharb:
@alecrobins awesome, thanks! Would you mind rebasing so as to remove all merge commits, and then this is good to go.

lencioni:
@alecrobins can you rebase this again for us when you have a moment? Thanks!

docs/README.md
```diff
@@ -57,7 +57,7 @@
     * [render()](/docs/api/ShallowWrapper/render.md)
     * [setContext(context)](/docs/api/ShallowWrapper/setContext.md)
     * [setProps(nextProps)](/docs/api/ShallowWrapper/setProps.md)
-    * [setState(nextState)](/docs/api/ShallowWrapper/setState.md)
+    * [setState(nextState[, cb])](/docs/api/ShallowWrapper/setState.md)
```
lencioni: It would be nice to use `callback` for this variable (here and elsewhere), for maximum clarity.

alecrobins:
@lencioni Just rebased and changed the variable name to `callback`.

lencioni:
LGTM

kentor:
when will this get released?

Looking forward to do this in tests:
```js
flushSetState = (wrapper) => new Promise(resolve => wrapper.setState({}, resolve));
...
await flushSetState();
expect(wrapper.state(...))
```

koba04:
@kentor This has been released in `2.5.0`.

* https://github.com/airbnb/enzyme/blob/master/CHANGELOG.md#250-october-17-2016

PR #618, opened 2016-09-30
[New] `shallow`: add `.dive()`

ljharb:
The goal of this method is to make testing of HOCs (higher-order components) easier - [withStyles](https://www.npmjs.com/package/react-with-styles), flux-store-connected components, etc.

Typically, the workarounds are either:
 - `wrapper.find('StringNameOfComponent').shallow()`: brittle, because it's using strings (strings are bad); also annoying to type out.
 - `wrapper.childAt(0).shallow()`: doesn't check if there's more than one child, and is annoying to type out.

In addition, if suddenly there's a DOM element in there, you might suddenly start shallow-rendering it without knowing, and your tests might erroneously be passing.

This method allows `wrapper.dive()` - ie, it requires there be one non-DOM child, and shallow-renders it for you.

If you need multiples, you chain them - `wrapper.dive().dive()` - for now we have intentionally omitted the ability to dive multiple levels at once, so that tight coupling to HTML structure is discouraged.

Longer term, we have a much better solution in mind for HOCs, but this is still useful in the interim.

### DIFF
--- a/docs/api/ShallowWrapper/dive.md
+++ b/docs/api/ShallowWrapper/dive.md
@@ -1,0 +1,52 @@
+# `.dive([options]) => ShallowWrapper`
+
+Shallow render the one non-DOM child of the current wrapper, and return a wrapper around the result.
+
+NOTE: can only be called on wrapper of a single non-DOM component element node.
+
+
+#### Arguments
+
+1. `options` (`Object` [optional]):
+- `options.context`: (`Object` [optional]): Context to be passed into the component
+
+
+
+#### Returns
+
+`ShallowWrapper`: A new wrapper that wraps the current node after it's been shallow rendered.
+
+
+
+#### Examples
+
+```jsx
+class Bar extends React.Component {
+  render() {
+    return (
+      <div>
+        <div className="in-bar" />
+      </div>
+    );
+  }
+}
+```
+
+```jsx
+class Foo extends React.Component {
+  render() {
+    return (
+      <div>
+        <Bar />
+      </div>
+    );
+  }
+}
+```
+
+```jsx
+const wrapper = shallow(<Foo />);
+expect(wrapper.find('.in-bar')).to.have.length(0);
+expect(wrapper.find(Bar)).to.have.length(1);
+expect(wrapper.dive().find('.in-bar')).to.have.length(1);
+```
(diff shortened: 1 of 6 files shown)


lelandrichardson:
LGTM!

lelandrichardson:
(also, I like "dive"!)

docs/api/shallow.md
```diff
@@ -207,3 +207,6 @@ Returns whether or not all of the nodes in the wrapper match the provided select
 
 #### [`.everyWhere(predicate) => Boolean`](ShallowWrapper/everyWhere.md)
 Returns whether or not all of the nodes in the wrapper pass the provided predicate function.
+
+#### [`.dive([options]) => ShallowWrapper`](ShallowWrapper/dive.md)
+Throws unless the wrapper contains only one non-DOM Component child; otherwise, returns a shallow wrapper around it.
```
lelandrichardson: I feel like we are focusing too much on the condition in which it throws. The fact that it throws should just be a note.

The description should be something like "shallow render the one child of the current wrapper, and return a wrapper around the result"

docs/api/ShallowWrapper/dive.md
```diff
@@ -0,0 +1,52 @@
+# `.dive([options]) => ShallowWrapper`
+
+Throws unless the wrapper contains only one non-DOM Component child; otherwise, returns a shallow wrapper around it.
```
lelandrichardson: same here.

ljharb:
@Aweary doh! thanks, will add.

ljharb:
@Aweary added :-)

docs/api/ShallowWrapper/dive.md
```diff
@@ -0,0 +1,52 @@
...
+    );
+  }
+}
+```
+
+```jsx
+const wrapper = shallow(<Foo />);
+expect(wrapper.find('.in-bar')).to.have.length(0);
+expect(wrapper.find(Bar)).to.have.length(1);
+expect(wrapper.dive().find('.in-bar')).to.have.length(1);
```
matthieuprat: I suspect this would fail because `wrapper` wraps a DOM element and by design, `dive` (currently?) throws in that case.

Don't you mean `expect(wrapper.children().dive().find('.in-bar')).to.have.length(1);` instead?

ljharb: Oops, thanks, I did mean that :-) fixed in e7226255bded93e285e36b50e16b59c8dc2458f9

PR #627, opened 2016-10-12
Fixed bug with containsAllMatchingElements and nested elements

squidarth:
Hi there! I noticed an issue with `containsAllMatchingElements`, in which if the nodes being matched have different parent elements, `containsAllMatchingElements` returns `false`.

### Approach

I solved this by using a very similar approach to `containsAnyMatchingElements` (except using `every` instead of `some`). Let me know if I misunderstood the intended function of `containsAllMatchingElements`, or if there are performance issues with this.

Thank you for taking a look at this, and for making this awesome library!


nfcampos:
Actually I'm not sure about the intended behaviour here, the behaviour after this change sounds more intuitive to me but the current behaviour is not wrong per se. Thoughts @Aweary @blainekasten @ljharb @lelandrichardson ?

Aweary:
`containsAllMatchingElements` shouldn't require that the elements be siblings. I would say the current behavior is wrong (especially since its undocumented) and this new behavior is correct 👍

squidarth:
Cool! I just cleaned up the git history, let me know if there are any other changes you need me to make before this is mergeable.

Thanks for taking a look!

src/ReactWrapper.jsx
```diff
@@ -362,9 +362,7 @@ class ReactWrapper {
    * @returns {Boolean}
    */
   containsAllMatchingElements(nodes) {
-    const invertedEquals = (n1, n2) => instEqual(n2, n1, (a, b) => a <= b);
-    const predicate = other => containsChildrenSubArray(invertedEquals, other, nodes);
-    return findWhereUnwrapped(this, predicate).length > 0;
+    return Array.isArray(nodes) && nodes.every(node => this.containsMatchingElement(node));
```
ljharb: i think perhaps we should be throwing a `TypeError` here if `nodes` is not an array, rather than just silently returning false.

src/ShallowWrapper.js
```diff
@@ -378,9 +378,7 @@ class ShallowWrapper {
    * @returns {Boolean}
    */
   containsAllMatchingElements(nodes) {
-    const invertedEquals = (n1, n2) => nodeEqual(n2, n1, (a, b) => a <= b);
-    const predicate = other => containsChildrenSubArray(invertedEquals, other, nodes);
-    return findWhereUnwrapped(this, predicate).length > 0;
+    return Array.isArray(nodes) && nodes.every(node => this.containsMatchingElement(node));
```
ljharb: here too

squidarth:
@ljharb thanks for the feedback, just addressed those chnages

squidarth:
@ljharb let me know what you think of those last changes, I can go ahead and merge master back in

pmeinhardt:
Hi there, any updates on this? Who else needs to review this @ljharb?

ljharb:
@squidarth would you mind rebasing?

ljharb:
I've gone ahead and done the rebase.

squidarth:
Great, thanks for merging this!

PR #641, opened 2016-10-24
Add yarn

ggarnier:
Add `yarn.lock` file and update install instructions.


ljharb:
Thanks! `yarn` isn't quite ready for production use yet - we're sticking with `npm` for the time being.

Issue #658, opened 2016-11-05
Cannot find stateless components in IE11


dompuiu:
Hi,

It seems that stateless components are not found in IE11 (and probably the same is true for lower versions of IE). You can find an expample in the snippet below:

```
class MyClassComponent extends Component {
  render() {
    return (<div>MyClassComponent</div>);
  }
}

const MyStatelessComponent = () => (
  <div>MyStatelessComponent</div>
);

const MainComponent = () => (
  <div>
    <MyStatelessComponent />
    <MyClassComponent />
  </div>
);

const instance = mount(<MainComponent />);
console.log(instance.find(MyStatelessComponent).length); // This returns 0, which is incorrect.
console.log(instance.find(MyClassComponent).length); //This returns 1, which is correct.
```

It seems the problem lies in the [`isFunctionalComponent`](https://github.com/airbnb/enzyme/blob/47b44363d815b3668a956d4b9503ee4f37c2b219/src/Utils.js#L27) method from `Utils.js`. The constructor name does not exist in IE11. And because of that, the stateless components are not detected.

One possible solution would be to transform the function to string and use a regex to detect the if the function name is `StatelessComponent`. I am not sure if this is the best solution, or even if  you would consider fixing this issue. In case you think it's a good solution, let me know and I could try to submit a PR for this.

ljharb:
This is because IE 9 - 11 doesn't support the "name" property on functions.

Your workaround options are:
1) manually add `.displayName =`
2) [polyfill](https://www.npmjs.com/package/function.prototype.name) function names

All that said, it makes sense that enzyme should be able to use https://www.npmjs.com/package/function.prototype.name to work in IE. I'd be happy to accept a PR that adds that dependency and uses it, instead of `.name`, throughout the lib.

dompuiu:
This was fixed by #659.

PR #659, opened 2016-11-05
Use shim to detect constructor function name.

dompuiu:
This will make enzyme detect stateless components correctly also in
Internet Explorer.


dompuiu:
This is a tentative to fix #658.

dompuiu:
I have updated all the places where `.name` was used.

PR #662, opened 2016-11-06
Skip undefined props when comparing nodes

### LINKED ISSUE #656: `equals` should skip null and undefined props
This is a continuation of #151, which fixed the skipping of *nodes*, but the same issue exists for *props*.

Here's an example to demonstrate the problem:

```jsx
import React from 'react';
import {shallow} from 'enzyme';

const Box = ({className}) => (
  <div className={className}/>
);

shallow(<Box/>).equals(<div/>); //=> false
shallow(<Box/>).equals(<div className={undefined}/>); //=> true
shallow(<Box className="awesome"/>).equals(<div className="awesome"/>); //=> true
```

`nodeEqual()` fails, because it expects `className={undefined}` property in the other component (`<div/>`). This is happening here https://github.com/airbnb/enzyme/blob/master/src/Utils.js#L94. Tested using the latest enzyme available on npm (`2.5.1`).

I think this case should be covered in enzyme by ignoring `null` and `undefined` props, which is what React does too. If everyone agrees, I can submit a PR to fix it ;)

vadimdemedes:
Fixes #656.

This PR ensures `null` and `undefined` props are skipped when comparing nodes:

```js
expect(nodeEqual(
   <div className={null} id={undefined} />,
   <div />
)).to.equal(true);
```

A helper function `cleanProps` removes all `null` and `undefined` props. I wanted to do that in `propsOfNode()`, but there are multiple `return` statements, which would result in a lot of repetition.

Let me know if this PR needs modification or anything else ;)

### DIFF
--- a/docs/api/ShallowWrapper/equals.md
+++ b/docs/api/ShallowWrapper/equals.md
@@ -32,4 +32,4 @@ expect(wrapper.equals(<div className="foo bar" />)).to.equal(true);
 when you are calling it you are calling it with a ReactElement or a JSX expression.
 - Keep in mind that this method determines equality based on the equality of the node's children as
 well.
-
+- Following React's behavior, `.equals()` ignores properties whose values are `undefined`.
--- a/packages/enzyme/src/RSTTraversal.js
+++ b/packages/enzyme/src/RSTTraversal.js
@@ -1,4 +1,5 @@
 import flatten from 'lodash/flatten';
+import entries from 'object.entries';
 import isSubset from 'is-subset';
 import functionName from 'function.prototype.name';
 import { nodeHasProperty } from './Utils';
@@ -98,8 +99,17 @@ export function nodeHasId(node, id) {
 
 export { nodeHasProperty };
 
+const CAN_NEVER_MATCH = {};
+function replaceUndefined(v) {
+  return typeof v !== 'undefined' ? v : CAN_NEVER_MATCH;
+}
+function replaceUndefinedValues(obj) {
+  return entries(obj)
+    .reduce((acc, [k, v]) => ({ ...acc, [k]: replaceUndefined(v) }), {});
+}
+
 export function nodeMatchesObjectProps(node, props) {
-  return isSubset(propsOfNode(node), props);
+  return isSubset(propsOfNode(node), replaceUndefinedValues(props));
 }
 
 export function getTextFromNode(node) {
--- a/packages/enzyme/src/RSTTraversal.js
+++ b/packages/enzyme/src/RSTTraversal.js
@@ -40,11 +40,11 @@ export function treeFilter(tree, fn) {
  * To support sibling selectors we need to be able to find
  * the siblings of a node. The easiest way to do that is find
  * the parent of the node and access its children.
- * 
+ *
  * This would be unneeded if the RST spec included sibling pointers
  * such as node.nextSibling and node.prevSibling
- * @param {*} root 
- * @param {*} targetNode 
+ * @param {*} root
+ * @param {*} targetNode
  */
 export function findParentNode(root, targetNode) {
   const results = treeFilter(
@@ -113,7 +113,7 @@ export function nodeMatchesObjectProps(node, props) {
 }
 
 export function getTextFromNode(node) {
-  if (node === null || node === undefined) {
+  if (node == null) {
     return '';
   }
 
--- a/packages/enzyme/src/Utils.js
+++ b/packages/enzyme/src/Utils.js
@@ -23,7 +23,9 @@ export function isCustomComponentElement(inst, adapter) {
 }
 
 function propsOfNode(node) {
-  return (node && node.props) || {};
+  return entries((node && node.props) || {})
+    .filter(([, value]) => typeof value !== 'undefined')
+    .reduce((acc, [key, value]) => ({ ...acc, [key]: value }), {});
 }
 
 export function typeOfNode(node) {
--- a/packages/enzyme/src/Utils.js
+++ b/packages/enzyme/src/Utils.js
@@ -25,7 +25,7 @@ export function isCustomComponentElement(inst, adapter) {
 function propsOfNode(node) {
   return entries((node && node.props) || {})
     .filter(([, value]) => typeof value !== 'undefined')
-    .reduce((acc, [key, value]) => ({ ...acc, [key]: value }), {});
+    .reduce((acc, [key, value]) => Object.assign(acc, { [key]: value }), {});
 }
 
 export function typeOfNode(node) {
@@ -141,7 +141,7 @@ function childrenToArray(children) {
   const result = [];
 
   const push = (el) => {
-    if (el === null || el === false || el === undefined) return;
+    if (el === null || el === false || typeof el === 'undefined') return;
     result.push(el);
   };
 
@@ -161,7 +161,7 @@ export function childrenToSimplifiedArray(nodeChildren) {
     const child = childrenArray[i];
     const previousChild = simplifiedArray.pop();
 
-    if (previousChild === undefined) {
+    if (typeof previousChild === 'undefined') {
       simplifiedArray.push(child);
     } else if (isTextualNode(child) && isTextualNode(previousChild)) {
       simplifiedArray.push(previousChild + child);
@@ -220,11 +220,11 @@ export function nodeHasProperty(node, propKey, propValue) {
   }
   const nodePropValue = nodeProps[propKey];
 
-  if (nodePropValue === undefined) {
+  if (typeof nodePropValue === 'undefined') {
     return false;
   }
 
-  if (propValue !== undefined) {
+  if (typeof propValue !== 'undefined') {
     return is(nodePropValue, propValue);
   }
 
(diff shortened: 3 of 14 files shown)


src/Utils.js
```diff
@@ -83,13 +84,13 @@ export function nodeEqual(a, b, lenComp = is) {
...
-  const left = propsOfNode(a);
+  const left = cleanProps(propsOfNode(a));
   const leftKeys = Object.keys(left);
-  const right = propsOfNode(b);
+  const right = cleanProps(propsOfNode(b));
   for (let i = 0; i < leftKeys.length; i += 1) {
     const prop = leftKeys[i];
     // we will check children later
-    if (prop === 'children') {
+    if (prop === 'children' || !left[prop]) {
```
ljharb: shouldn't this be `left[prop] == null`, since `0` is falsy too?

src/Utils.js
```diff
@@ -1,5 +1,6 @@
 /* eslint no-use-before-define:0 */
 import isEqual from 'lodash/isEqual';
+import isNil from 'lodash/isNil';
```
ljharb: i'm not sure why we need a dependency for `== null`?

src/Utils.js
```diff
@@ -114,7 +115,7 @@ export function nodeEqual(a, b, lenComp = is) {
   }
 
   if (!isTextualNode(a)) {
-    const rightKeys = Object.keys(right);
+    const rightKeys = Object.keys(right).filter(key => !isNil(right[key]));
```
ljharb: let's just do `key => right[key] != null` - it's clearer and built-in.

vdemedes: Sure, agree.

src/Utils.js
ljharb: a non-strict check against `null` is the sole exception :-)

vdemedes:
Apparently, moving `cleanProps()` to `propsOfNode()` breaks this test: https://github.com/airbnb/enzyme/blob/master/test/ShallowTraversal-spec.jsx#L134. I guess `cleanProps()` should stay inside `nodeEqual()`. @ljharb What do you think?

ljharb:
If we're going to ignore `null`, we need to ignore it consistently. In other words, searching for a value of `null` should be the same as a value of `undefined`, or as "absence".

In other words, if we're asking if a node has property "foo" equal to `null`, and it doesn't have the property "foo", then I'd expect it to pass.

src/Utils.js
vdemedes: Ahh, leftover from debugging.

vdemedes:
Ok, I see your point!

Regarding this test: https://github.com/airbnb/enzyme/blob/master/test/ShallowWrapper-spec.jsx#L577, let me clarify one last thing. If there are nodes with props equal to `null` or `undefined` and we're looking for them (like in the test), these nodes shouldn't be returned, because we treat those props as missing. 

```js
expect(wrapper.find({ prop: undefined })).to.have.length(0);
expect(wrapper.find({ prop: null })).to.have.length(0);
```

ljharb:
yeah, that's a good point/wrinkle. It seems like the options are, when finding `{ foo: null, bar: 3 }`, to find anything that *omits* foo and *has* bar equal to 3 - or, to simply find anything that has bar equal to 3.

If we choose the former path, `.find({ prop: null })` will find everything. If we choose the latter, `.find({ prop: null })` will find nothing. I'm not sure if either makes sense.

vdemedes:
I think there's a 3rd option. Check props in `find()` and throw if there are `null` or `undefined` values.

```js
find(props) {
  const hasNilProps = Object.keys(props).some(key => props[key] == null);

  if (hasNilProps) {
    throw new TypeError('Props can\'t have `null` or `undefined` values.');
  }

  // ...
}
```

What do you think?

ljharb:
I think that's probably best, but then that means that it's a breaking change.

@lelandrichardson @Aweary @blainekasten @nfcampos what do you think?

nfcampos:
this 3rd option is definitely the best of the three. this change should be accompanied with a note on the docs for `find()` saying that if you want to find nodes whose props are null you can use `findWhere`. maybe even the error message should mention this. what do you think?

vdemedes:
I think @nfcampos's proposal is reasonable and makes this change clear to users.

Aweary:
> I think @nfcampos's proposal is reasonable and makes this change clear to users.

Agreed 👍

vdemedes:
Glad we agree, I'll update the PR!

ljharb:
ok, so it looks like React ignores `undefined` but not `null` - http://stackoverflow.com/questions/34809178/react-js-default-prop-is-not-used-with-null-is-passed - so we should change this PR to continue to treat `null` as a valid, existing value.

vdemedes:
@ljharb Good catch, will do!

vdemedes:
PR updated, addressing all of the feedback above. Thanks to everyone!

One question: I put the warning about an error when using `.find()` and having some `undefined` props in the `docs/selector.md`. Should I put it in API docs for `.find()` or leave it there?

docs/api/ShallowWrapper/equals.md
```diff
@@ -32,4 +32,4 @@ expect(wrapper.equals(<div className="foo bar" />)).to.equal(true);
 when you are calling it you are calling it with a ReactElement or a JSX expression.
 - Keep in mind that this method determines equality based on the equality of the node's children as
 well.
-
+- Following React's behavior, `.equals()` ignores properties, whose values are `undefined`.
```
nfcampos: please remove the comma after 'properties'

vdemedes:
All good now, thanks guys for the feedback ;)

src/ComplexSelector.js
```diff
@@ -78,6 +78,13 @@ export default class ComplexSelector {
       return this.handleSelectors(selectors, wrapper);
     }
 
+    if (typeof selector === 'object' && selector !== null) {
+      const hasUndefinedValues = Object.keys(selector).some(key => selector[key] === undefined);
```
ljharb: `import values from 'object.values'` and you can do `values(selector).some(value => value === undefined)`

vdemedes: Good suggestion, updated!

src/Utils.js
```diff
@@ -147,6 +147,18 @@ function isTextualNode(node) {
...
+function cleanProps(props) {
+  const newProps = {};
+
+  Object.keys(props)
+    .filter(key => props[key] !== undefined)
+    .forEach((key) => {
+      newProps[key] = props[key];
+    });
+
+  return newProps;
```
ljharb: alternatively:
```js
return entries(props)
  .filter(([, value]) => value !== undefined)
  .reduce((acc, [key, value]) => assign({}, acc, { [key]: value }), {});
```

nfcampos:
@ljharb should we get this in for v3?

ljharb:
I've rebased this on top of latest master; @aweary @nfcampos would you mind rereviewing?

packages/enzyme/src/selectors.js
```diff
@@ -141,6 +142,10 @@ export function buildPredicate(selector) {
   // If the selector is an non-empty object, treat the keys/values as props
   if (typeof selector === 'object') {
     if (!Array.isArray(selector) && selector !== null && !isEmpty(selector)) {
+      const hasUndefinedValues = values(selector).some(value => value === undefined);
```
nfcampos: Why is this one a test of the value itself but the others are testing typeof?

packages/enzyme/src/Utils.js
```diff
@@ -23,7 +23,9 @@ export function isCustomComponentElement(inst, adapter) {
 }
 
 function propsOfNode(node) {
-  return (node && node.props) || {};
+  return entries((node && node.props) || {})
+    .filter(([, value]) => typeof value !== 'undefined')
+    .reduce((acc, [key, value]) => ({ ...acc, [key]: value }), {});
```
nfcampos: This is rather ineffecient, creating as many objects as you have keys, I know this is a testing library, but if you have a lot of tests it starts to add up

packages/enzyme/src/RSTTraversal.js
```diff
@@ -98,8 +99,17 @@ export function nodeHasId(node, id) {
...
+  return typeof v !== 'undefined' ? v : CAN_NEVER_MATCH;
+}
+function replaceUndefinedValues(obj) {
+  return entries(obj)
+    .reduce((acc, [k, v]) => ({ ...acc, [k]: replaceUndefined(v) }), {});
+}
+
 export function nodeMatchesObjectProps(node, props) {
-  return isSubset(propsOfNode(node), props);
+  return isSubset(propsOfNode(node), replaceUndefinedValues(props));
```
nfcampos: I don’t understand why we need this change. Don’t we throw if the received props contain undefined values?

packages/enzyme/src/selectors.js
ljharb: My habit is to always use typeof for safety; babel makes this unnecessary. I'll make them all be the same.

packages/enzyme/src/Utils.js
ljharb: Sure, I could mutate the accumulator since the overall reduce is pure.

packages/enzyme/src/RSTTraversal.js
ljharb: In `find`, yes, but I don't think that's the only place this is called.

packages/enzyme/src/Utils.js
ljharb: Done.

packages/enzyme/src/selectors.js
ljharb: It's now consistent in the project.

PR #667, opened 2016-11-09
fix error on reduce and reduceRight doc

junpos:
1) on first iteration `amount` is `undefined` which results in `NaN` for total.

```js
it('reduce', function () {
  const total = wrapper.find(Bar).reduce((amount, n) => {
    console.log(amount, n.prop('amount'));
    return amount + n.prop('amount');
  });
  expect(total).to.equal(16);
});
```

```shell
AssertionError: expected NaN to equal 16
  + expected - actual

    -NaN
    +16

// log
undefined 2
NaN 4
NaN 8
```
It needs a sanity check whether `amount` is defined, or have an initial value.
```js
wrapper.find(Bar).reduce((acc, n) => {
  const amount = n.prop('amount');
  return acc 
    ? acc + amount
    : amount;
});
//or
wrapper.find(Bar).reduce((amount, n) => amount + n.prop('amount'), 0);
```

2. And the total should be `14` instead of `16`

### DIFF
--- a/docs/api/ReactWrapper/reduce.md
+++ b/docs/api/ReactWrapper/reduce.md
@@ -41,8 +41,8 @@ class Foo extends React.Component {
 
 ```jsx
 const wrapper = mount(<Foo />);
-const total = wrapper.find(Bar).reduce((amount, n) => amount + n.prop('amount'));
-expect(total).to.equal(16);
+const total = wrapper.find(Bar).reduce((amount, n) => amount ? amount + n.prop('amount') : n.prop('amount'));
+expect(total).to.equal(14); // 2 + 4 + 8
 ```
 
 
--- a/docs/api/ReactWrapper/reduceRight.md
+++ b/docs/api/ReactWrapper/reduceRight.md
@@ -41,8 +41,8 @@ class Foo extends React.Component {
 
 ```jsx
 const wrapper = mount(<Foo />);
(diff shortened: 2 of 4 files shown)


docs/api/ReactWrapper/reduce.md
```diff
@@ -41,8 +41,8 @@ class Foo extends React.Component {
 
 ```jsx
-const total = wrapper.find(Bar).reduce((amount, n) => amount + n.prop('amount'));
-expect(total).to.equal(16);
+const total = wrapper.find(Bar).reduce((amount, n) => amount ? amount + n.prop('amount') : n.prop('amount'));
```
ljharb: actually, on the first iteration `amount` is the first item - that's the behavior of `reduce` when omitting the initial value.

ljharb: in other words, `[1, 2, 3].reduce(function (prev, item) { console.log([prev, item]); });` logs `[1, 2]` and then `[undefined, 3]`

ljharb:
If enzyme's `reduce` fails to have the same behavior as `Array#reduce`, that's a bug we need to fix.

ljharb:
I've put up #673 to fix this bug, thanks for the PR!

nfcampos:
@ljharb since we have #673 shouldn't this one be closed?

ljharb:
@nfcampos good call

PR #670, opened 2016-11-09
Fix ID selectors

ljharb:
This fixes the unintentional breaking change in 2.4.x (and 2.5.x), by committing horrible sins with regular expressions. This also does not revert the new behavior added in 2.4.x (that caused the breakage).

We desperately need to replace this with a proper CSS parser, but this should suffice for now.

Once merged, I'll also backport this to the 2.4.x line (and 2.5.x if master already has some semver-minor stuff).


blainekasten:
LGTM

Issue #361, opened 2016-05-03
Easier way to access child refs


joshma:
Once in awhile we'll have a test that renders a component P with child C:

```
<P><C>
  <div></div>
  <div></div>
</C></P>
```

To get the second `<div>` in a test, we'll add a ref to it:

```
<P><C>
  <div></div>
  <div ref={'__testDiv'}></div>
</C></P>
//
TestUtils.findRenderedComponentWithType(p, C).refs.testDiv
```

With enzyme, though, it doesn't seem to be easy to get to that ref

```
pWrapper.find(C).ref('__testDiv') // Error: ReactWrapper::ref(refname) can only be called on the root
```

The only alternative has to use css classes, which isn't ideal since that shows up in the final markup. The div itself is pretty hard to select for - is there an easier way I'm missing?

ljharb:
If you're using `shallow`, does `pWrapper.find(C).shallow().ref('__testDiv')` work for you?

joshma:
@ljharb `shallow` doesn't actually render / have refs, does it? (`.ref` is undefined). (I tried using `mount` instead of `shallow` and calling `pWrapper.find(C).mount().ref('__testDiv')`, but you can only call `.mount()` on the root it says

Aweary:
@ljharb is there a way to make a child component returned via `find` into a new `root` component?

ljharb:
Ah, right, sorry - refs are only with `mount`.

I'd expect `.mount()` to re-establish what the root is while re-mounting the entire tree from that point.

Aweary:
@ljharb so conceptually, `ReactWrapper.mount` doesn't to correspond to the top-level `mount` utility. It just looks like it causes `ReactWrapperComponent` to render the component if it wasn't already rendered (by `unmount`).

Would it make sense to return a new instance of `ReactWrapper` from `ReactWrapper.mount()` then?

ljharb:
That would make sense to me, since I also assume `.shallow()` returns a new instance of `ShallowWrapper`. @lelandrichardson?

Aweary:
Would `unmount` return a new `ReactWrapper` too? It just seems potentially confusing since `ReactWrapper.mount` is paired with `ReactWrapper.unmount`.

ljharb:
That's a good point too.

joshma:
[...]

PR #672, opened 2016-11-10
Add string-refs support with shallowWrapper

yormi:
In response to https://github.com/airbnb/enzyme/issues/361 and my own need :)

I find it extremely useful to have a unique, local-scoped, way to identify my React Element.

Works as expected `shallowWrapper.ref('stringRef') => ShallowWrapper` or throws if more than one element has the wanted ref.

### DIFF
--- a/src/ShallowTraversal.js
+++ b/src/ShallowTraversal.js
@@ -151,3 +151,23 @@ export function getTextFromNode(node) {
     .join('')
     .replace(/\s+/, ' ');
 }
+
+export const isAReactComponentNode = node => typeof node.type === 'function';
+
+export const findRefInOwnedChildren = (node, ref, isRootNode) => {
+  const result = [];
+
+  if (node.ref === ref) {
+    result.push(node);
+  }
+
+  if (!isAReactComponentNode(node) || isRootNode) {
+    const children = childrenOfNode(node);
+
+    children.forEach((c) => {
+      result.push(...findRefInOwnedChildren(c, ref));
+    });
+  }
+
+  return result;
+};
--- a/src/ShallowTraversal.js
+++ b/src/ShallowTraversal.js
@@ -155,22 +155,50 @@ export function getTextFromNode(node) {
     .replace(/\s+/, ' ');
 }
 
-const findRef = (node, ref, isRootNode) => {
-  const result = [];
+const getOwnedReferenceNodes = (node, isRootNode) => {
+  const referencedNodes = [];
 
-  if (node.ref === ref) {
-    result.push(node);
+  if (!node) {
+    return referencedNodes;
+  }
+
+  if (node.ref) {
+    referencedNodes.push(node);
   }
 
   if (isDOMComponentElement(node) || isRootNode) {
     const children = childrenOfNode(node);
 
     children.forEach((c) => {
-      result.push(...findRef(c, ref));
+      const recursiveNodes = getOwnedReferenceNodes(c);
+      referencedNodes.push(...recursiveNodes);
     });
   }
 
-  return result;
+  return referencedNodes;
 };
 
-export const findRefInOwnedChildren = (node, ref) => findRef(node, ref, true);
+export const getNodeRefs = (node) => {
+  const callbackNodes = [];
+  const stringNodes = [];
+
+  const referencedNodes = getOwnedReferenceNodes(node, true);
+
+  referencedNodes.forEach((n) => {
+    switch (typeof n.ref) {
+      case 'function':
+        callbackNodes.push(n);
+        break;
+      case 'string':
+        stringNodes.push(n);
+        break;
+      default:
+        throw new Error(`${n.ref} is an invalid ref. Valid refs are string or function type`);
+    }
+  });
+
+  return {
+    callbackNodes,
+    stringNodes,
+  };
+};
--- a/src/ShallowTraversal.js
+++ b/src/ShallowTraversal.js
@@ -155,7 +155,7 @@ export function getTextFromNode(node) {
     .replace(/\s+/, ' ');
 }
 
-export const findRefInOwnedChildren = (node, ref, isRootNode) => {
+const findRef = (node, ref, isRootNode) => {
   const result = [];
 
   if (node.ref === ref) {
@@ -166,9 +166,11 @@ export const findRefInOwnedChildren = (node, ref, isRootNode) => {
     const children = childrenOfNode(node);
 
     children.forEach((c) => {
-      result.push(...findRefInOwnedChildren(c, ref));
+      result.push(...findRef(c, ref));
     });
   }
 
   return result;
 };
+
+export const findRefInOwnedChildren = (node, ref) => findRef(node, ref, true);
--- a/src/ShallowTraversal.js
+++ b/src/ShallowTraversal.js
@@ -12,6 +12,9 @@ import {
   nodeHasType,
   nodeHasProperty,
 } from './Utils';
+import {
+  isDOMComponentElement,
+} from './react-compat';
 
 
 export function childrenOfNode(node) {
@@ -152,8 +155,6 @@ export function getTextFromNode(node) {
     .replace(/\s+/, ' ');
 }
 
-export const isAReactComponentNode = node => typeof node.type === 'function';
-
 export const findRefInOwnedChildren = (node, ref, isRootNode) => {
   const result = [];
 
@@ -161,7 +162,7 @@ export const findRefInOwnedChildren = (node, ref, isRootNode) => {
     result.push(node);
   }
 
-  if (!isAReactComponentNode(node) || isRootNode) {
+  if (isDOMComponentElement(node) || isRootNode) {
     const children = childrenOfNode(node);
 
     children.forEach((c) => {
--- a/src/ShallowWrapper.js
+++ b/src/ShallowWrapper.js
@@ -153,8 +153,7 @@ class ShallowWrapper {
           throw new Error('There is more than one component own by the tree root with the same ref string');
         }
 
-        refObject[n.ref] = n
-        return refObject
+        return Object.assign({}, refObject, { [n.ref]: n });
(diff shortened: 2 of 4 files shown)


src/ShallowTraversal.js
```diff
@@ -125,7 +124,7 @@ export function buildPredicate(selector) {
         return node => nodeMatchesObjectProps(node, selector);
       }
       throw new TypeError(
-        'Enzyme::Selector does not support an array, null, or empty object as a selector',
+        'Enzyme::Selector does not support an array, null, or empty object as a selector'
```
ljharb: this trailing function comma is required, why did you remove it?

src/ShallowTraversal.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
 import isEmpty from 'lodash/isEmpty';
 import isSubset from 'is-subset';
-import functionName from 'function.prototype.name';
```
ljharb: i think perhaps you included some unintentional changes - have you rebased on the latest master?

src/ShallowTraversal.js
yormi: I'm not sure what happened but it seems that I used some kind of auto-format that I'm not aware of... sorry for that.

Correcting that !

src/ShallowTraversal.js
```diff
@@ -151,3 +151,23 @@ export function getTextFromNode(node) {
     .join('')
     .replace(/\s+/, ' ');
 }
+
+export const isAReactComponentNode = node => typeof node.type === 'function';
```
ljharb: Instead of defining your own function, please use `isFunctionalComponent`, `isCustomComponentElement`, and `isDOMComponentElement` which are already present in the file.

src/ShallowWrapper.js
```diff
@@ -454,6 +456,32 @@ class ShallowWrapper {
...
+   */
+  ref(ref) {
+    return this.single('ref', (node) => {
+      if (!isAReactComponentNode(this.unrendered)) {
+        throw new Error('The current node must be a React component.');
+      }
+
+      const result = findRefInOwnedChildren(node, ref, true);
+
+      if (result.length === 0) return false;
```
ljharb: shouldn't this return `null` when there's no result?

src/ShallowWrapper.js
```diff
@@ -454,6 +456,32 @@ class ShallowWrapper {
...
+      if (result.length === 1) return new ShallowWrapper(result[0], this.root);
+      if (result.length > 1) {
+        throw new Error('There is more than one component own by the tree root with the same ref string');
+      }
+
+      return false;
```
ljharb: when is this line hit?

yormi:
I still wonder why they've been deprecated for callbacks. I don't remember reading about the reason behind it.

I personally use the string-refs to access my react-element in integration-test. I use them to pilot my tests rather than testing the ref.

I haven't found another way that is as elegant(local-scoped, unique) yet... I'm open to other solutions though.

src/ShallowWrapper.js
yormi: In case of many "owned" element with the same ref. Shouldn't happen but if an error occurs, I think it'll be better to know it instead of returning the first occurrence.

ljharb: The previous lines though cover all possibilities for `result.length

src/ShallowWrapper.js
yormi: You are right !

src/ShallowWrapper.js
yormi: Oupsi... you meant the `return false`. I thought the thrown error. You are right again !

src/ShallowTraversal.js
```diff
@@ -151,3 +154,21 @@ export function getTextFromNode(node) {
     .join('')
     .replace(/\s+/, ' ');
 }
+
+export const findRefInOwnedChildren = (node, ref, isRootNode) => {
```
ljharb: this being exported, having `isRootNode` be provideable by the consumer seems brittle. Could we split this up into separate functions, so that the recursion implementation details are hidden?

Aweary:
> It's worth discussing that string refs are deprecated in React (in favor of callback refs)

They're not technically deprecated just yet, just considered legacy. The point still stands though, it's not a pattern that's recommended anymore.

Good work on the PR @yormi, but I would have to say that I'm not in favor of this. Adding`.ref()` to the API solely to support a legacy implementation seems confusing to me.

yormi:
First of all, thank you :)

Secondly, I understand your concerns @Aweary. I would prefer avoiding legacy/deprecated features but I haven't found other ways as clear and simple to do so.

I would be very interested in your suggestions on a (hopefully) local-scoped and unique way of accessing an element to pilot tests ?

Aweary:
> I would be very interested in your suggestions on a (hopefully) local-scoped and unique way of accessing an element to pilot tests ?

If you use ref callbacks you should be able to access them on the instance

```js
  render() {
    return <div ref={n => this.node = n}/>
  }
}

const div = wrapper.instance().node; // should be the HTMLDivElement instance
```

I don't think shallow should support refs, let alone string refs.

blainekasten:
Just to ignite a flame, I would actually advocate for support for strong refs right now. Though they are legacy, they are still a supported feature. String refs are used wildly in the wild, so without the feature we're making every consumer write brittle and confusing tests to get access to their refs. Or they just don't write tests because it's too difficult. Both are problems this library should be trying to help with.

We can remove the feature as react removes it. Which would make more sense to do then as consumers updating to newer versions of react would HAVE to change their refs.

ljharb:
@blainekasten I don't mind supporting string refs - but i *do* mind having a nicer API for string refs than for callback refs. We should be providing API that encourages people to do the right thing, even if we also provide API that enables people to test a legacy pattern.

blainekasten:
@ljharb that totally makes sense. Do we even have a designated API to work with callback refs?

src/ShallowTraversal.js
yormi: Yeah, good idea. I'm on it. Thanks for the feedback btw. I haven't had the chance to have my code reviewed much.

yormi:
@Aweary I agree that works (or should work... I think I saw an issue about it !?) with `mount`. I think it should be possible to use refs with `shallow` as well. If we find a way to use callback refs with `shallow`, I'd be happy to change my string refs.

I think it'll be possible:
By getting all the refs of the root (pretty much as I did) and executing them in the shallowWrapper constructor we could access the "refs" on the instance as with `mount`.

The problems I foresee are:
- if a callback is more complex than `c => { this.foo = c }` it might have code that needs to have the whole thing rendered... I guess it shouldn't happen though
- we need to find a way not to shallow render all the elements with a ref

What do you guys think ?

Aweary:
>if a callback is more complex than c => { this.foo = c } it might have code that needs to have the whole thing rendered... I guess it shouldn't happen though`

Correct, ref callbacks shouldn't do anything to affect the render output, and it should almost always just be attaching the node to the instance.

I'm OK with adding support for refs with `shallow`, but we need to support string refs and callback refs if we do. Supporting only a legacy implementation isn't good.

> Couldn't we just let the same API resolve for both string and callbacks? I'm not sure if it's safe to assume that refs always get stored on the component instance, but we could do something like this:

The shallow renderer doesn't actually process refs, so unfortunately we can't.
>Shallow testing currently has some limitations, namely not supporting refs.


[ - source](https://facebook.github.io/react/docs/test-utils.html#shallow-rendering)

yormi:
Ok so I went on with the idea but since I didn't find a way to make the callback refs work within `enzyme`, I dive into `react` code: https://github.com/facebook/react/pull/8311

yormi:
Ok since the react team seems to have a lot to do with Fiber and everything, maybe it'll be simpler to have the feature implemented here. I might have found a way to make it work with callback refs as well.

The problem I'm facing is that callback refs are bound to the instance and it seems that the instance is undefined when calling it from enzyme(i.e. not directly during rendering in react ShallowRenderer).

We can still copy the render method so when the callback ref are rendered, they have a different `this`. This is possible only if the render method is not fat arrow function. I don't quite understand why but it works when it's bound using `.bind` but not when using "fat-arrow-binding".

@Aweary do you think it's fair to assume that the render method is not or shouldn't be bound using a fat-arrow function ? Too big assumption ? I hate to have limitation like that but on the other hand, I don't see why we would need an arrow function there.

Aweary:
@yormi I went ahead and reviewed your PR on the React repo. If this is implemented in React it shouldn't require any changes to the core reconciler modules. It might be better to implement this in enzyme, if you've found a valid approach that supports callback refs.

yormi:
Ok my laptop HDD just drown in the snowstorm here... I just replaced it. I'll try to implement it in enzyme in the next 2 days. Thanks for the feedback :)

yormi:
Ok sorry for the delay ! Semester's over now so here's my new proposal.

Refs are accessible directly on the instance so it's easily accessible to just check the props. It's also easy to build upon if you want to add to to the API to dive in the tree... even though that's not the point of shallow rendering, I think.

yormi:
Actually, I found a "bug"... The refs are actualized at every render, thanks to `componentDidUpdate`.

This hook is called when we update a component state through a button handler for instance but not when using `wrapper.setProps` or anything using the `ShallowWrapper.prototype.rerender`. This cause the props of the refs to not be in sync. If `lifecycleExperimental` is turned on, the `componentDidUpdate` fires and everything is alright.

So what is the purpose of having `lifecycleExperimental` flag turned off ?

Do you see a better way to listen on the component render ?

ljharb:
Now that createRef/forwardRef is the preferred way forward for refs - even though string and callback refs are unlikely to go away - I don't think this is a feature worth making first-class in shallow rendering.

What would the ref even return for the cases where in `mount` it returns a DOM element?

I'm going to close this; but we can reopen if needed.

PR #673, opened 2016-11-11
[Fix] `reduce`/`reduceRight`: follow `Array#reduce` when omitting initialValue

ljharb:
Closes #667.

I've marked this as a patch - but it might potentially be breaking if someone was relying on omitting the required (incorrectly documented as optional) initialValue being `undefined`.

Thoughts?

### DIFF
--- a/docs/api/ReactWrapper/reduce.md
+++ b/docs/api/ReactWrapper/reduce.md
@@ -12,8 +12,7 @@ following arguments:
   - `node` (`ReactWrapper`): A wrapper around the current node being processed
   - `index` (`Number`): The index of the current node being processed
 
-2. `initialValue` (`T` [optional]): If provided, this will be passed in as the first argument to the
-first invocation of the reducing function.
+2. `initialValue` (`T` [optional]): If provided, this will be passed in as the first argument to the first invocation of the reducing function. If omitted, the first `node` will be provided and the iteration will begin on the second node in the collection.
 
 
 
--- a/docs/api/ReactWrapper/reduceRight.md
+++ b/docs/api/ReactWrapper/reduceRight.md
@@ -12,8 +12,7 @@ following arguments:
   - `node` (`ReactWrapper`): A wrapper around the current node being processed
   - `index` (`Number`): The index of the current node being processed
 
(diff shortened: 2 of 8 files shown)


src/ReactWrapper.jsx
```diff
@@ -722,11 +722,19 @@ class ReactWrapper {
    * @param {*} initialValue - the initial value
    * @returns {*}
    */
-  reduce(fn, initialValue) {
-    return this.getNodes().reduce(
-      (accum, n, i) => fn.call(this, accum, this.wrap(n), i),
-      initialValue,
-    );
+  reduce(fn) {
```
blainekasten: could we maybe add the old argument back within comment blocks so we see that it is part of the API. Yes jsdoc helps above, but it gets a bit confusing when see this and then L729

blainekasten: same for the reduceRight

Aweary: I believe @ljharb is trying to preserve `function.length`, which should only count required arguments. I think elsewhere we just use a default value of `undefined`, which might be good here.

```js
reduce(fn, initialValue = undefined) {
```

ljharb: I can't do that because the native reduce/reduceRight distinguish between `undefined` and "absent" - or rather, id still have to check `arguments.length` and ignore `initialValue` when `arguments.length` is < 2

ljharb:
@blainekasten bumping versions is indeed cheap, but it's nice to avoid bumping majors because of semver ranges.

src/ReactWrapper.jsx
blainekasten: yes yes, i was merely suggesting this: `reduce(fn /*, initialValue */) { ... }`

ljharb: I'll do one of the two - as long as the behavior and function length are correct, I'm happy.

ljharb: Updated.

blainekasten:
I think it's smarter to prevent code from breaking without users expecting it than to worry users will be on old versions of enzyme due to ranges..

nfcampos:
@ljharb should we get this in for v3?

ljharb:
@nfcampos good call; I'll rebase.

PR #677, opened 2016-11-12
Make hasClass work for `mount`ed composite components

christian-schulze:
This (hopefully?) address the long running issue #134, where `wrapper.hasClass` always returns false on `mount`ed composite components. The reason for the failure is the following block in `MountedTraversal.js#instHasClassName`:
```js
  if (!isDOMComponent(inst)) {
    return false;
  }
```

Simply removing this block breaks many things, as the method is used internally in `MountedTraversal.js`. My naive solution is to split the method in two, so we retain the exported `instHasClassName` function for use by `ReactWrapper.hasClass`, and extract a private `hasClassName` function for use internally in `MountedTraversal.js`.

From here we can remove the `!isDOMComponent` early exit code from `instHasClassName`, and instead add it to `hasClassName` to maintain existing functionality.

### DIFF
--- a/test/ReactWrapper-spec.jsx
+++ b/test/ReactWrapper-spec.jsx
@@ -2073,17 +2073,37 @@ describeWithDOM('mount', () => {
   });
 
   describe('.hasClass(className)', () => {
-    it('should return whether or not node has a certain class', () => {
-      const wrapper = mount(
-        <div className="foo bar baz some-long-string FoOo" />,
-      );
+    context('When using a DOM component', () => {
+      it('should return whether or not node has a certain class', () => {
+        const wrapper = mount(
+          <div className="foo bar baz some-long-string FoOo" />,
+        );
 
-      expect(wrapper.hasClass('foo')).to.equal(true);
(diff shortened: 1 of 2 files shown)


test/ReactWrapper-spec.jsx
```diff
@@ -2073,17 +2073,37 @@ describeWithDOM('mount', () => {
   });
 
   describe('.hasClass(className)', () => {
-    it('should return whether or not node has a certain class', () => {
-      const wrapper = mount(
-        <div className="foo bar baz some-long-string FoOo" />,
-      );
+    context('When using a DOM component', () => {
```
Aweary: @ljharb are you OK with using `context` blocks here? I don't think we use them anywhere else, but it makes enough sense.

ljharb: `context` and `describe` are identical - either is fine, but i have a light preference for `describe`.

Aweary: Right, I know they're identical, it was just a question of preference.

PR #679, opened 2016-11-13
Get outer DOM node from composite component wrapper

christian-schulze:
As outlined in #623 and #446, implement `ReactWrapper#getDOMNode` which provides access to the outer most DOMComponent of a mounted composite component.

If you are happy for this to exist, I will add api documentation.

PR #680, opened 2016-11-13
Fix bug introduced in #677

nfcampos:
PR #677 introduced a bug where calling `hasClass` on a component that renders `null` would throw an exception. This fixes that.

### DIFF
--- a/src/MountedTraversal.js
+++ b/src/MountedTraversal.js
@@ -48,7 +48,7 @@ export function instEqual(a, b, lenComp) {
 
 export function instHasClassName(inst, className) {
   const node = findDOMNode(inst);
-  if (!node) { // inst renders null
+  if (node === null) { // inst renders null
     return false;
   }
   if (node.classList) {
--- a/src/MountedTraversal.js
+++ b/src/MountedTraversal.js
@@ -48,6 +48,9 @@ export function instEqual(a, b, lenComp) {
 
 export function instHasClassName(inst, className) {
   const node = findDOMNode(inst);
+  if (!node) { // inst renders null
+    return false;
+  }
   if (node.classList) {
     return node.classList.contains(className);
   }
--- a/test/ReactWrapper-spec.jsx
+++ b/test/ReactWrapper-spec.jsx
@@ -3387,12 +3387,12 @@ describeWithDOM('mount', () => {
       }
     }
 
-    it('should return the outer most DOMComponent of the root wrapper', () => {
+    it('should return the outermost DOMComponent of the root wrapper', () => {
       const wrapper = mount(<Test />);
       expect(wrapper.getDOMNode()).to.have.property('className', 'outer');
     });
 
-    it('should return the outer most DOMComponent of the inner div wrapper', () => {
+    it('should return the outermost DOMComponent of the inner div wrapper', () => {
       const wrapper = mount(<Test />);
       expect(wrapper.find('.inner').getDOMNode()).to.have.property('className', 'inner');
     });
@@ -3412,12 +3412,12 @@ describeWithDOM('mount', () => {
         </div>
       );
 
-      it('should return the outer most DOMComponent of the root wrapper', () => {
+      it('should return the outermost DOMComponent of the root wrapper', () => {
         const wrapper = mount(<SFC />);
         expect(wrapper.getDOMNode()).to.have.property('className', 'outer');
       });
 
-      it('should return the outer most DOMComponent of the inner div wrapper', () => {
+      it('should return the outermost DOMComponent of the inner div wrapper', () => {
         const wrapper = mount(<SFC />);
         expect(wrapper.find('.inner').getDOMNode()).to.have.property('className', 'inner');
       });
--- a/test/ReactWrapper-spec.jsx
+++ b/test/ReactWrapper-spec.jsx
@@ -2126,6 +2126,42 @@ describeWithDOM('mount', () => {
         expect(wrapper.hasClass('doesnt-exist')).to.equal(false);
       });
     });
+
+    context('When using nested composite components', () => {
+      it('should return whether or not node has a certain class', () => {
+        class Foo extends React.Component {
+          render() {
+            return (<div className="foo bar baz some-long-string FoOo" />);
+          }
+        }
+        class Bar extends React.Component {
+          render() {
+            return <Foo />;
+          }
+        }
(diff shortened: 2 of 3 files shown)


src/MountedTraversal.js
```diff
@@ -47,6 +47,9 @@ export function instEqual(a, b, lenComp) {
 
 export function instHasClassName(inst, className) {
   const node = findDOMNode(inst);
+  if (!node) { // inst renders null
```
nfcampos: Is it better to check `!node` or `node === null` here?

ljharb: If the instance renders "nothing", it could return an empty array, or `null`, or `undefined`… is there perhaps a react method we could use, like `React.Children.count` or something, that would cover us here?

nfcampos: But can a react component return anything other than `null` or a react element? Besides, what we're talking about here is the output of `findDOMNode`, can it be anything other than a dom element or `null`, I dont know...

ljharb: That's a good point - https://facebook.github.io/react/docs/react-dom.html#finddomnode makes it seem like it's only `null` or a node. I wonder if then it would be better to use `=== null` specifically so that it blows up if React changes the behavior?

nfcampos: I agree, I've changed it to `=== null`

nfcampos:
same thing as in https://github.com/airbnb/enzyme/pull/679#issuecomment-260498491 applies here I believe, after #677 `hasClass` can be used with composites as long as they're class based (because its implementation relies on findDOMNode).
Should we introduce the logic similar to #679 or should we try to re-use it (eg. by calling `getDOMNode` in `hasClass`)?

ljharb:
I think the two PRs can coexist - `hasClass` is a boolean method and it seems reasonable for it to return `false` here, and for `getDOMNode` to throw.

nfcampos:
@ljharb You wouldn't expect this test to pass?
```js
const Foo = () => <div className="bar" />
const wrapper = mount(<Foo />)
expect(wrapper.hasClass('bar')).to.equal(true)
```

ljharb:
hmm, that's a good point. no, i wouldn't; that should work.

nfcampos:
So what should we do here, throw if used on a SFC?

ljharb:
Sorry, my comment was confusing. I would indeed expect that test to pass - so SFCs should work, and not throw.

If they won't work properly, then they should throw, of course.

nfcampos:
@ljharb sorry for the late reply, I don't think there's any way to get a ref to the dom node rendered for a stateless component so I'll update the PR to make the method throw when used on a stateless component.

nfcampos:
So, apparently the react documentation is not up-to-date or I misread it because findDOMNode does work on functional components, as such, I've updated the PR with tests for hasClass on SFCs and removed the previously added check on getDOMNode for SFCs. Thoughts?

test/ReactWrapper-spec.jsx
```diff
@@ -3318,11 +3368,28 @@ describeWithDOM('mount', () => {
...
+        </div>
+      );
+
+      it('should return the outer most DOMComponent of the root wrapper', () => {
+        const wrapper = mount(<SFC />);
+        expect(wrapper.getDOMNode()).to.have.property('className', 'outer');
+      });
 
-      it('should throw when wrapping an SFC', () => {
+      it('should return the outer most DOMComponent of the inner div wrapper', () => {
```
ljharb: "outermost" is one word :-p

nfcampos: indeed it is ;) updated

test/ReactWrapper-spec.jsx
```diff
@@ -3318,11 +3368,28 @@ describeWithDOM('mount', () => {
...
-      it('should throw when wrapping an SFC', () => {
+      it('should return the outermost DOMComponent of the inner div wrapper', () => {
         const wrapper = mount(<SFC />);
-        expect(() => wrapper.getDOMNode()).to.throw(TypeError, 'Method “getDOMNode” cannot be used on functional components.');
+        expect(wrapper.find('.inner').getDOMNode()).to.have.property('className', 'inner');
+      });
+
+      it('should throw when wrapping multiple elements', () => {
+        const wrapper = mount(<SFC />).find('span');
+        expect(() => wrapper.getDOMNode()).to.throw(Error);
```
Aweary: Can we test the error message like we had before? Testing that it throws any error seems brittle.

ljharb: Probably both would be good; but if we want our error tests to be more robust we should be making error subclasses, not doing string comparisons.

Aweary: Error subclasses are a better solution, I'm on board with that approach. But a string comparison would be better than nothing, so I'm fine with doing that for now too.

nfcampos:
FYI @ljharb @Aweary https://twitter.com/dan_abramov/status/808781309745201152
given this, should we support these methods for SFCs while they work, potentially seeing unexpectedly them break in a way we can't fix with a future React release?

ljharb:
Yes, if it works, we should support it for SFCs.

If it breaks in a major version (since it'd be a breaking change), we'll have to conditionally support it based on the react version.

nfcampos:
@ljharb since we're relying on "accidental" non-public behaviour of the ReactDOM API it could theoretically break in a patch version

ljharb:
Sadly "if it's reachable, it's public" isn't a philosophy React tends to follow, but in this case I think we can, since it would probably cause *lots* of breakage across the React ecosystem. I think we should be optimistic here that React will follow semver.

nfcampos:
@ljharb I think we might be pretty alone in relying on this particular behaviour of findDOMNode, (especially because the usefulness of using this very convoluted way (it requires rendering the SFC as a child of a class component, ReactWrapperComponent for us, and accessing it from there) of accessing dom nodes for SFCs is next to null outside of testing (but very useful in testing obviously))

nfcampos:
@ljharb @aweary I've added the error message testing, this should now be ready to merge

devrelm:
@ljharb Can we get a patch release to npm with these changes? We're going to have to pin to 2.6.0 as a few of our tests are currently broken due `hasClass` not working on empty nodes.

ljharb:
Sure, I should be able to get one out today or tomorrow.

PR #686, opened 2016-11-17
Display a useful error when `withDom` fails to find "jsdom".

indexzero:
Lost a bunch of time on this today because I thought it was a compatibility issue around the new `react@15.4.0` release ...

... instead of just my own error of forgetting to install `jsdom` in a new module. 🙀 🙀 🙀 

Hopefully this helps others avoid my mistake. Tried to make the message displayed to users as clear as possible. 💯 ✨ 👍 💪 😼 

screen shot 2016-11-16 at 11 32 42 pm

_
> _Testability is tricky on this since the `enzyme` module itself depends on `jsdom` so the error will never be produced. I could introduce `proxyquire` and force `require('jsdom')` to throw, but that seemed like overkill. Hopefully the change is obvious and small enough._

### DIFF
--- a/withDom.js
+++ b/withDom.js
@@ -15,5 +15,12 @@ if (!global.document) {
     };
   } catch (e) {
     // jsdom is not supported...
+    if (e.message === "Cannot find module 'jsdom'") {
+      console.error('[enzyme/withDom] Error: missing required module "jsdom"');
+      console.error('[enzyme/withDom] To fix this you must run:');
+      console.error('[enzyme/withDom]   npm install jsdom --save-dev');
+    } else {
+      console.error('[enzyme withDom] ' + (e.stack || e.message));
+    }
   }
 }


withDom.js
```diff
@@ -15,5 +15,12 @@ if (!global.document) {
     };
   } catch (e) {
     // jsdom is not supported...
+    if (e.message === "Cannot find module 'jsdom'") {
+      console.error('[enzyme/withDom] Error: missing required module "jsdom"');
+      console.error('[enzyme/withDom] To fix this you must run:');
+      console.error('[enzyme/withDom]   npm install jsdom -D');
```
ljharb: let's stick to `--save-dev` rather than using the shortcuts.

indexzero: Sure. I didn't feel strongly either way.

indexzero:
@ljharb updated to `--save-dev` following your feedback.

indexzero:
Sure I can rebase and squash it.

indexzero:
All set.

ljharb:
Thanks!

PR #690, opened 2016-11-17
Update FindWhere.md

JonasMS:
Changed documentation to fix a broken example:
typeof n.type() !== 'string' fails while n.type() !== 'string' works


ljharb:
Thanks!

Issue #693, opened 2016-11-18
Attribute selector doesn't work on attributes that have numbers in them


raduanastases:
In [ShallowTraversal.js](https://github.com/airbnb/enzyme/blob/master/src/ShallowTraversal.js#L113) the selector for the attribute doesn't work on attributes that have numbers in them.
`const propKey = selector.split(/\[([a-zA-Z-]*?)(=|])/)[1];`

Instead, it should be:
`const propKey = selector.split(/\[([a-zA-Z-0-9-]*?)(=|])/)[1];`

ljharb:
@raduanastases it'd be great if you could file a PR with a test case!

ljharb:
(also, please note that attributes that start with numbers are invalid)

raduanastases:
sure, I'll get to it as soon as possible.

jsmey:
I made this change and added a test https://github.com/airbnb/enzyme/pull/700

ljharb:
Closed in #1086

PR #700, opened 2016-11-23
[fix] bug #693 allow numbers in attributes

jsmey:
Updated the regex in  ShallowTraversal.js to allow numbers in attribute name.


src/ShallowTraversal.js
```diff
@@ -110,7 +110,7 @@ export function buildPredicate(selector) {
           return node => nodeHasId(node, selector.slice(1));
 
         case SELECTOR.PROP_TYPE: {
-          const propKey = selector.split(/\[([a-zA-Z-]*?)(=|])/)[1];
+          const propKey = selector.split(/\[([a-zA-Z-0-9-]*?)(=|])/)[1];
```
ljharb: please also include tests that show that an attribute named `1-foo` would be invalid, per https://github.com/airbnb/enzyme/issues/693#issuecomment-261472688

jsmey: Something like this?
expect(wrapper.find('span[1-foo=1]')).to.have.length(0);

We can't render tags with attributes that start with numbers. <span 1-foo={1} /> will throw a parse error. Is a simple test like that sufficient or are you expecting ShallowTraversal to throw an error?

ljharb: I definitely expect ShallowTraversal to throw an error - if you try to use an invalid selector, we should blow up instead of silently giving you a result.

src/ShallowTraversal.js
```diff
@@ -111,6 +111,10 @@ export function buildPredicate(selector) {
 
         case SELECTOR.PROP_TYPE: {
           const propKey = selector.split(/\[([a-zA-Z-0-9-]*?)(=|])/)[1];
+          const pattern = /^[a-zA-Z][a-zA-Z-0-9-]*?/;
```
ljharb: it might be simpler to change the `propKey` splitting regex to be `/\[([a-z]+[a-z-0-9-]*?)(=|])/i`, and do a bit more checking on the split result?

ljharb:
(before pinging that this is ready for a rereview, please rebase so as to remove all merge commits)

src/ShallowTraversal.js
```diff
@@ -110,10 +110,13 @@ export function buildPredicate(selector) {
           return node => nodeHasId(node, selector.slice(1));
 
         case SELECTOR.PROP_TYPE: {
-          const propKey = selector.split(/\[([a-zA-Z-]*?)(=|])/)[1];
+          const propKey = selector.split(/\[([a-z]+[a-z-0-9-]*?)(=|])/i);
+          if (propKey.length === 1) {
+            throw new TypeError('Enzyme::Invalid attribute selector');
```
blainekasten: It'd be helpful if you printed some details in this to help users track down which test broke. It'd be helpful if you printed the selector and the root's component name

jsmey: ok, will do

jsmey:
I added logging of the selector in the error message. I didn't see any easy way to log the root node without passing it to function.

src/ShallowTraversal.js
```diff
@@ -110,10 +110,13 @@ export function buildPredicate(selector) {
           return node => nodeHasId(node, selector.slice(1));
 
         case SELECTOR.PROP_TYPE: {
-          const propKey = selector.split(/\[([a-zA-Z-]*?)(=|])/)[1];
+          const propKey = selector.split(/\[([a-z]+[a-z-0-9-]*?)(=|])/i);
+          if (propKey.length === 1) {
+            throw new TypeError(`Enzyme::Invalid attribute selector ('${selector}')`);
+          }
           const propValue = selector.split(/=(.*?)]/)[1];
```
ljharb: `const [, propValue] = selector.split(/=(.*?)]/);` (i know this is existing code)

src/ShallowTraversal.js
```diff
@@ -110,10 +110,13 @@ export function buildPredicate(selector) {
...
 
-          return node => nodeHasProperty(node, propKey, propValue);
+          return node => nodeHasProperty(node, propKey[1], propValue);
```
ljharb: similarly, it'd be nice to destructure instead of index into propKey - ie, `const [, secondPropKey] = propKey;` or something

jsmey:
I updated the propKey and propValue:
const [, propKey] = selectorSplit;
const [, propValue] = selector.split(/=(.*?)]/);

test/ShallowWrapper-spec.jsx
```diff
@@ -565,6 +565,36 @@ describe('shallow', () => {
...
 
+    it('should support attributes with numbers', () => {
+      const wrapper = shallow(
+        <div>
+          <span value-1={1} />
+          <a value-1={false} />
+        </div>,
+      );
+
+      expect(wrapper.find('span[value-1=1]')).to.have.length(1);
```
aweary: `span[value-1=1]` is not a valid CSS selector. Attribute values must be either a valid selector (which cannot start with a number) or a string ([reference](https://www.w3.org/TR/css3-selectors/#attribute-substrings)).

`span[value-1=1]` should fail and `span[value-1="1"]` should pass.

jsmey: OK, I'll add that

jsmey: Would that be a breaking change? This existing test would fail.
` it('should support boolean and numeric values for matching props', () => {
      const wrapper = shallow(
        <div>
          <span value={1} />
          <a value={false} />
        </div>,
      );

      expect(wrapper.find('span[value=1]')).to.have.length(1);
`

currently expect(wrapper.find('span[value="1"]')).to.have.length(1); would fail

ljharb: I'd call it a bugfix.

ljharb:
I don't think this is relevant any more now that we're on v3. @aweary, can you confirm?

aweary:
@ljharb correct, I verified it handles attribute selectors with numbers in the attribute name. Thanks for doing this @jsmey, sorry we didn't get it in!

PR #708, opened 2016-11-28
Update jsdom.md to work with react-native

connected-mgosbee:
The original example does not work with react-native because of (for example) TimerMixin sets a variable `GLOBAL` based on whether or not `window` or `global` exists. If `window` exists, it chooses that to look for `requestAnimationFrame`, `setTimeout` and `setInterval`. I then tries to use `GLOBAL.setTimeout` which doesn't exist.

### DIFF
--- a/docs/guides/jsdom.md
+++ b/docs/guides/jsdom.md
@@ -17,7 +17,7 @@ As a result, a standalone script like the one below is generally a good approach
 var jsdom = require('jsdom').jsdom;
 
 global.document = jsdom('');
-global.window = document.defaultView;
+global.window = global;
 Object.keys(document.defaultView).forEach((property) => {
   if (typeof global[property] === 'undefined') {
     global[property] = document.defaultView[property];


docs/guides/jsdom.md
```diff
@@ -17,7 +17,7 @@ As a result, a standalone script like the one below is generally a good approach
 var jsdom = require('jsdom').jsdom;
 
 global.document = jsdom('');
-global.window = document.defaultView;
+global.window = global;
 Object.keys(document.defaultView).forEach((property) => {
```
ljharb: this is an important part of jsdom (using `document.defaultView`).

This loop should ensure that `setTimeout` etc are copied from `document.defaultView` to `global`.

connected-mgosbee: Then should a separate section be dedicated for react-native support?

ljharb: Yes, node indeed has `setTimeout`.

Also, the point of `jsdom` is to provide all the browser globals to `node`, so as to fake that it's a browser.

connected-mgosbee: Yes, node has setTimeout, but `document.defaultView` does not (on Node, atleast). So TimerMixin looks for setTimeout on `window.setTimeout` which does not exist.

ljharb: ah - i think perhaps we should keep `window` as `document.defaultView`, but add an additional loop that copies things on `global` to the new `window`.

acidhax: Okay, I can work on that.  Should it be a PR on the readme or something else?

acidhax: Sorry, commenting from my personal account now.

ljharb: I think a PR on the readme is fine for now - I don't think we have the code living anywhere else.

nfcampos: we do here I think https://github.com/airbnb/enzyme/blob/master/withDom.js

acidhax: ^ that still doesn't solve the conflict between React TimerMixin though.

PR #713, opened 2016-12-01
mount: fix broken `.parents()`

just-boris:
Actually, this is a bug report. Test should pass, but it doesn't.

When elements tree has more than one branch, `.parents()` method returns some nodes, which are not  parent of the element.

### DIFF
--- a/src/MountedTraversal.js
+++ b/src/MountedTraversal.js
@@ -161,15 +161,77 @@ export function childrenOfInst(node) {
   return childrenOfInstInternal(internalInstanceOrComponent(node));
 }
 
+// This function should be called with an "internal instance". Nevertheless, if it is
+// called with a "public instance" instead, the function will call itself with the
+// internal instance and return the proper result.
+function findAllInRenderedTreeInternal(inst, test) {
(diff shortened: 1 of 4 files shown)


just-boris:
Currently, I don't have time to look at this deeply.
If somebody will provide an idea how to fix this, I will update the PR.

ljharb:
Thanks! I added the same test case to `shallow` (which wasn't broken), and then adapted the ShallowTraversal solution to work for MountedTraversal.

src/MountedTraversal.js
```diff
@@ -161,15 +161,77 @@ export function childrenOfInst(node) {
...
+        inst._renderedComponent,
+        test,
+      ),
+    );
+  }
+  return ret;
+}
+
+// This function could be called with a number of different things technically, so we need to
+// pass the *right* thing to our internal helper.
```
Aweary: nit: this comment feels too vague, what are things and what is the right thing?

ljharb: This comment is already in master; I just moved the function higher up due to no-use-before-define

PR #714, opened 2016-12-01
Add `getClasses()` to mount and shallow API's

happycollision:
Returns an array containing all the classes for the current node. This can be used in a similar
fashion as `.hasClass`, but can return more interesting information in certain
test runners on failure, since it will include all class names, and not just a boolean.

I admittedly do not know all the ins and outs of the traversal vs wrapper files, but I think I've got it.

In addition to adding the functions, I have also used the added function inside the `hasClass()` function. That way all the tests for `hasClass()` are also, by proxy, testing `getClasses()`.

Of course, `getClasses()` has its own tests as well.

### DIFF
--- a/docs/api/ReactWrapper/getClasses.md
+++ b/docs/api/ReactWrapper/getClasses.md
@@ -1,0 +1,18 @@
+# `.getClasses() => Array<String>`
+
+Returns an array containing all the classes for the current node. This can be used in a similar
+fashion as [`.hasClass`](hasClass.md), but can return more interesting information in certain
+test runners on failure, since it will include all class names, and not just a boolean.
+
+
+#### Returns
+
+`Array<String>`: an array containing all the classes for the current node.
+
+
+#### Example
+
+```jsx
+const wrapper = mount(<MyComponent />);
+expect(wrapper.find('.main-navigation-link').getClasses()).to.include('active');
+```
--- a/docs/api/ShallowWrapper/getClasses.md
+++ b/docs/api/ShallowWrapper/getClasses.md
@@ -1,0 +1,18 @@
+# `.getClasses() => Array<String>`
+
+Returns an array containing all the classes for the current node. This can be used in a similar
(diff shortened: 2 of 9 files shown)


ljharb:
Does `.render().attr('class').split(' ')` not suffice?

happycollision:
The reason I did this is because I couldn't find any way of getting the all the classes of a node in the documentation. This would make getting all the class names of a node a "first class citizen" method, instead of using the API on a Cheerio Wrapper.

Personally, I've never used `render()` in any of my tests because I saw that it returned a Cheerio Wrapper, which I've avoided because I know nothing about it. Can you interact with a CheerioWrapper? Simulate events/pass in new props? I have always avoided it because the Enzyme documentation for `render()` says "Static Rendering" and I just wasn't sure and didn't want to look over yet another library's documentation. Maybe others would feel the same?

Anyway, it seems to me that it is missing from the current API, but that was just a sense I got.

ljharb:
Sure - a CheerioWrapper is just what cheerio returns, which is similar to what jQuery returns. It is static rendering, but that's what you're doing if you're getting all the class names.

happycollision:
I guess I thought there was something inherently different about it. Like it was not as dynamic as Enzyme's wrappers. I guess then, I am unclear as to when I should use `render()` vs. `shallow()` vs. `mount()`. Perhaps I need to look into that.

Thanks!

ljharb:
It's definitely not as dynamic as those wrappers - but getting a list of class names is not a dynamic operation.

happycollision:
This also might be out of the scope of what you are doing with Enzyme, but it might be nice to have a page in the documentation describing "common operations" or something like that. I'm not sure how often you see things like "how do I test this?" or "how do I get that?", but I know I for one—as someone who is only a few months into TDD/BDD and is barely familiar with the tools—would have been happy to see a section called "getting a list of classes for a node".

:wink:

Anyway, I love what you have all done with Enzyme. I am retro-fitting it and Jest into all the React projects I have.

PR #717, opened 2016-12-05
Update webpack usage with Webpack 2

jackfranklin:
Webpack 2 changes what's allowed in the externals part of the config, so I updated the docs to have Webpack 1 and Webpack 2 examples.

The example WP2 change is what I use on a project and it all works just great.


blainekasten:
Thanks @jackfranklin !

Issue #718, opened 2016-12-08
Document .isEmptyRender


prayogoa:
It's not listed in the docs

jordann-nulogy:
on it

benlakey:
I think this can be closed because it was added in https://github.com/airbnb/enzyme/commit/80a74b45a5b5aeafd1cab7658e080c0a21a535c1

PR #719, opened 2016-12-09
Add docs for isEmptyRender()

blackpost38:
I added docs `isEmptyRender()` api.

Could you check this out?

> ref #718


ljharb:
Seems like you both put up a PR at the same time :-) I'm going to close this in favor of #720.

Issue #705, opened 2016-11-26
New API: Alias `exists` to `isEmpty`


christemple:
**Question: Should isEmpty() be replaced by an exist() function?**

**My personal experience**
So I have been trying to test a React component that can return `false` in the render function [1].

After reading the docs I seen `isEmpty`and thought this would be a perfect fit, until I read the example of how to use `isEmpty` [2]

It almost seems like `isEmpty()` in this scenario is closer to `exists()` for example, the docs demonstrate searching for a child node that does not exist, and that is the only time `isEmpty()` will return true.

The scenario I was hoping for is given the following component:
```jsx
class Example extends React.Component {
  render() {
    return false;
  }
}
```
and the following test:
```js
it('is empty', () => {
  const wrapper = mount(<Example />);
  expect(wrapper.isEmpty()).to.equal(true);
});
```

This results in a failure expecting false to equal true, because the `isEmpty()` call doesn't check what I assumed it to, it checks the length of the node, of which every wrapper never has 0, unless you do a `.find` for a node within it that does not exist.

Even doing the following still does not result in behaviour that I'd expect, it still returns `false` for `isEmpty()`:
```js
it('is empty', () => {
  const wrapper = mount(<Example />);
  const example = wrapper.find(Example);
  expect(example.isEmpty()).to.equal(true);
});
```

After a while of reading through the docs I found a way of accomplishing what I was looking to do:
```js
it('is empty', () => {
  const wrapper = mount(<Example />);
  expect(wrapper.html()).to.equal(null);
});
```
So, to wrap up...
* Does anyone else find `isEmpty()` misleading?
* Would `exist()` be more intuitive for ehat `isEmpty()` is currently doing?
* Would the `isEmpty()` implementation suit better to checking `.html() === null`?

Thanks

_Disclaimer: I have only started using enzyme and think it's a fantastic library_

[1] https://facebook.github.io/react/docs/react-component.html#render
[2] http://airbnb.io/enzyme/docs/api/ReactWrapper/isEmpty.html

blainekasten:
@chriscartlidge Thanks for a fantastic write up! You raised good points. I'll try and consolidate my thoughts.

1. `exists` is a much better name than `isEmpty` for checking if a query found nodes.
2. `isEmpty` still doesn't feel right for checking `.html() === null`, and I'm not sure that use-case warrants an explicit API.

[...]

PR #722, opened 2016-12-11
New API change: Delegate isEmpty() to exists() and begin deprecation of using isEmpty()

christemple:
To address the issue whereby `isEmpty()` being confusing from an API perspective when checking if a node `exists()`.
https://github.com/airbnb/enzyme/issues/705

Includes:
 - addition of `exists()` function and delegating `isEmpty()` to `exists()`
 - deprecation warnings for using `isEmpty()`
 - updated docs for using `exists()` and deprecation docs for `isEmpty()`


christemple:
I've also fixed the gitbook prism plugins version number to the latest that works.  

When I tried running the docs it kept failing, which I narrowed down to the latest release of the prism plugin.

ljharb:
(would be nice to rebase down to a single commit)

christemple:
@ljharb done 👍

src/ReactWrapper.jsx
```diff
@@ -885,12 +885,23 @@ class ReactWrapper {
...
   /**
-   * Returns true if the current wrapper has no nodes. False otherwise.
+   * Delegates to exists()
    *
    * @returns {boolean}
    */
   isEmpty() {
-    return this.length === 0;
+    // eslint-disable-next-line no-console
+    console.warn('Deprecated method isEmpty() called, use exists() instead.');
```
blainekasten: Is it worth ensuring we only log this once? I imagine anyone using `isEmpty` getting 100s of these warnings..

blainekasten: I think it would be nice to prefix this with `Enzyme::` just so people know where the warning is coming from.

christemple: Yeah sounds like a good idea, I'll update this tomorrow, thanks

ljharb: Actually I'm wondering if we should hold off on the deprecation notice in-code for now, until we've been able to make PRs into things like `chai-enzyme` and others - ie, give a whole release before adding the warning message in?

blainekasten: @ljharb up to you. But I always just put deprecation warnings in on minor versions and pull them out in next major versions. I think react adds deprecations in majors only?

blainekasten: With a library like this, I don't see it making a difference. Aliasing an API without a log is not going to encourage anyone to remove it from existing tests and probably won't even affect them writing new tests. The odds of developers stumbling upon the documentation to see it's deprecated and actually act on that is unlikely.

If we add the log, it lets us remove the tech debt sooner. But I do think we should change this to only log once.

christemple: @blainekasten I definitely think prefixing with `Enzyme::` makes sense, however after thinking more about it, I'm not so sure about logging it only once.

If someone uses `isEmpty()` a lot isn't it better to make noise to encourage people to make the switch sooner rather than later so the tech debts lifespan is short?

The noise was very helpful for me, after switching off mocha `--reporter dot`, to identify and locate specs using `isEmpty` so that I could update them, as `isEmpty()` is the equivalent  of `!exists()` it wasn't a matter of doing a global search and replace on the specs.

ljharb: I'm not sure how we'd "only log once" - store a global? what's "once" - if i'm using `--watch` with my tests, should only the first run log it?

I think if it logs, it has to log every time.

Aweary: >I'm not sure how we'd "only log once" - store a global?

React uses a scoped flag to only log warnings once, but I think in our case it makes the most sense to just log every time. Test output can be rather verbose, making a single warning potentially hard to distinguish. `isEmpty` isn't likely used with extreme frequency in the average test suite either, so in most cases we're probably not going to be outputting too much noise.

Also, in regards to versioning: semver specifies that deprecation warnings [should go in minor releases](https://github.com/mojombo/semver/blob/master/semver.md#how-should-i-handle-deprecating-functionality) so I'm 👍 with that.

christemple:
I've just done the update to prefix `Enzyme::` in the warning

src/ReactWrapper.jsx
christemple: I've just done the update to prefix Enzyme:: in the warning

blainekasten:
lgtm! Thanks

PR #723, opened 2016-12-12
Add some small clarification around `.simulate()`

lukerollans:
I felt like it may be useful to expand on the second "gotcha" of `.simulate()`. One may initially think it only supports native events such as `onClick`, `onChange` etc, but is in fact generic enough to work with any prop.

### DIFF
--- a/docs/api/ShallowWrapper/simulate.md
+++ b/docs/api/ShallowWrapper/simulate.md
@@ -55,4 +55,4 @@ expect in a real environment. As a result, one must call `.simulate()` on the ac
 the event handler set.
 - Even though the name would imply this simulates an actual event, `.simulate()` will in fact 
 target the component's prop based on the event you give it. For example, `.simulate('click')` will
-actually get the `onClick` prop and call it.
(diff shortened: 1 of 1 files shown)


blainekasten:
This is not entirely accurate. It is currently only accurate for `shallow`. If you try and simulate an arbitrary prop in mount it will throw. In `shallow` it uses props. Though it should probably actually throw just so we have unity between the renderers.

I'm assuming we don't want to document this as it's not the current intentions of the feature. I'll defer to @ljharb.

If anything we need to clarify this is for `shallow` only. But based on discussions we've had in other issues, I think that we are not wanting people to use the  `simulate` API for arbitrary props.

ljharb:
Yes, the goal is to lock it down so that .simulate *cant* be used for arbitrary props.

blainekasten:
Alright, Sorry @lukerollans. I'm going to have to close this one. I created #725 to track keeping shallow wrapper from simulating non-standard events.

lukerollans:
Right, that makes sense. No worries at all. Failing this, I'm wondering what the recommended way of testing arbitrary callbacks/actions on props would be? I suppose simply grabbing the prop myself and executing will have the exact same effect, but is this part of the library anywhere that I'm not aware of?

ljharb:
Yes, you'd manually call `wrapper.props().onWhatever()`, or `wrapper.instance().onWhatever()`, etc.

Issue #715, opened 2016-12-02
Support compatible component libraries with defined spec


trueadm:
I was thinking, given that Inferno and Preact both aim to support the React API – would it be in the interest of this project to add a layer to enable support of alternative rendering engines other than React? Both libraries offer compatibility packages, i.e. `inferno-compat` and `preact-compat` that aim to provide backwards support for React via alias changes.

blainekasten:
I don't know much about Inferno, but Preact is gaining a lot of popularity. I'm not sure how well it would work as enzyme dives deep into unsupported React internals. So if it's possible with something like `preact-compat` to do I think it would be worth considering. Though if it adds too much weight to maintain both interops it's likely not worth it..

I'd be curious on hearing from @ljharb

trueadm:
@blainekasten Inferno and Preact have similar goals – if you have time, please also look into [Inferno](https://github.com/trueadm/inferno). I know the author of Preact well and I'm certain @developit would love this to be landed if possible. 

In terms of using React internals, that will soon have to change as React pulls away internal access using internal APIs (even more so with the upcoming React Fibre) so maybe now would be a good time to look how these things can overlap and help one another.

Aweary:
I've talked with @developit briefly about supporting Preact. I think initially any interop would be supported via those `*-compat` packages, which would need to support `ReactTestUtils` and the various internal APIs we use.

Long term it would make more sense for Enzyme to expose a way for consumers to configure the rendering engine, instead of supporting Inferno and Preact directly. This might mean you'd need to provide modules like `inferno-shallow-renderer`, for example.

developit:
I would love to have support for this and I'm willing to help same as @trueadm. 

@Aweary - how plausible do you think it would be to make the shallow/full/static renderers pluggable/injectable?

Aweary:
I think it's pretty plausible, a lot of what we pull from React is isolated in [src/react-compat.js](https://github.com/airbnb/enzyme/blob/master/src/react-compat.js), which is where we require in most (all?) the React packages we consume. If we can decide on a config API we could possibly read it from there and require user defined packages instead that export the same APIs.

That would take care of a sizeable portion of the work needed to support this level of interop, though its likely not as easy as it sounds.

Ultimately it depends on whether this is even something we can agree is a goal for Enzyme. @ljharb would be the authority on that.

ljharb:
[...]

PR #742, opened 2016-12-20
[RFC] Enzyme Adapter + React Standard Tree Proposal

lelandrichardson:
to: @ljharb @aweary @nfcampos @blainekasten 
cc: @cpojer @sebmarkbage @spicyj @developit @trueadm @bvaughn

This is a proposal and I was hoping to get some feedback from people in this PR. The enzyme maintainers met to discuss #715 and what we can do to support this and also improve on the maintainability and usefulness of enzyme.

I'm looking for any and all feedback. Thanks!

### DIFF
--- a/docs/future/compatibility.md
+++ b/docs/future/compatibility.md
@@ -61,6 +61,8 @@ type SourceLocation = {|
   lineNumber: number 
 |}
 
+type NodeType = 'class' | 'function' | 'host';
+
 // An RSTNode has this specific shape
 type RSTNode = {|
   // Either a string or a function. A string is considered a "host" node, and
@@ -68,8 +70,8 @@ type RSTNode = {|
   // an SFC in the case of a function.
   type: string | function;
 
-  // Whether or not this node is a "Host" node.
-  host: boolean;
+  // This node's type
+  nodeType: NodeType;
 
   // The props object passed to the node, which will include `children` in its raw form,
   // exactly as it was passed to the component.
--- a/docs/future/compatibility.md
+++ b/docs/future/compatibility.md
@@ -53,49 +53,55 @@ type LiteralValue = string | number
 // A "node" in an RST is either a LiteralValue, or an RSTNode
 type Node = LiteralValue | RSTNode
 
+// if node.type
+type RenderedNode = RSTNode | [Node]
+
+type SourceLocation = {| 
+  fileName: string
+  lineNumber: number 
+|}
+
 // An RSTNode has this specific shape
 type RSTNode = {|
   // Either a string or a function. A string is considered a "host" node, and
   // a function would be a composite component. It would be the component constructor or
   // an SFC in the case of a function.
-  type: string | function
-  
+  type: string | function;
+
+  // Whether or not this node is a "Host" node.
+  host: boolean;
+
   // The props object passed to the node, which will include `children` in its raw form,
   // exactly as it was passed to the component.
-  props: object
-  
+  props: object;
+
   // The backing instance to the node. Can be null in the case of "host" nodes and SFCs.
-  // React alternative libs can choose to provide whatever concept here applies.  If it is
-  // provided, Enzyme will expect it to follow the ComponentInterface type specified below
-  instance: ComponentInstance?
-  
-  // This is a transformed version of `props.children`. It is an array or null. If it's an array,
-  // it's an array of Nodes, meaning that they are LiteralNodes or RSTNodes. This means that 
-  // non-rendering values such as `false`, `null`, and `[]` wouldn't make it in here. This is also
-  // a "flat" array.
-  children: [Node]?
-  
-  // This can feel similar to children, but it's not the same thing at all. For a given node,
-  // this corresponds to the result of the `render` function with the provided props, but transformed
-  // into an RST.  For "host" nodes, this will always be `null`.
-  rendered: Node?
+  // Enzyme will expect instances to have the _public interface_ of a React Component, as would
+  // be expected in the corresponding React release returned by `getTargetVersion` of the
+  // renderer. Alternative React libraries can choose to provide an object here that implements 
+  // the same interface, and Enzyme functionality that uses this will continue to work (An example 
+  // of this would be the `setState()` prototype method).
+  instance: ComponentInstance?;
+
+  // For a given node, this corresponds roughly to the result of the `render` function with the 
+  // provided props, but transformed into an RST. For "host" nodes, this will always be `null` or
+  // an Array. For "composite" nodes, this will always be `null` or an `RSTNode`.
+  rendered: RenderedNode?;
+
+  // an optional property with source information (useful in debug messages) that would be provided
+  // by this babel transform: https://babeljs.io/docs/plugins/transform-react-jsx-source/
+  __source?: SourceLocation;
 |}
 ```
 
+Thoughts:
 
-A `ComponentInstance` is going to be expected by enzyme to follow a certain interface in order
-for certain API methods to actually work, but they don't *need* to. Those methods just won't be
-guaranteed to work in those cases.
-
-
-So far, that interface is:
-```js
-type ComponentInstance = {
-  state: object?
-  context: object?
-  setState: function
-}
-```
+- If an `RSTNode` has `host: true`, `rendered` will always be an `Array` or `null`. (though this 
+might change with Fiber I guess).
+- `rendered` is a better property name than `children`, as the meaning of `children` is ambiguous
+here since it is defined on props/elements etc.
+- If a node has `host: true`, loosely speaking that means that an instance of that component
+- For "host" nodes, can `instance` ever be non-null? Could it be the DOM element? Yes/No?
 
 
 ### Enzyme Adapter Protocol
@@ -108,28 +114,40 @@ react case, this would be the data structure returned from `React.createElement`
 
 ```js
 type RendererOptions = {
-  // I'm not 100% sure about this, but I think with an API like this, we could implement
-  // `mount` and `shallow` with the same renderer.
-  ?isLeaf(Element): bool
+  // An optional predicate function that takes in an `Element` and returns
+  // whether or not the underlying Renderer should treat it as a "Host" node
+  // or not. This function should only be called with elements that are 
+  // not required to be considered "host" nodes (ie, with a string `type`),
+  // so the default implementation of `isHost` is just a function that returns
+  // false.
+  ?isHost(Element): boolean;
 }
 
 type EnzymeAdapter = {
+  // This is a method that will return a semver version string for the _react_ version that
+  // it expects enzyme to target. This will allow enzyme to know what to expect in the `instance`
+  // that it finds on an RSTNode, as well as intelligently toggle behavior across react versions
+  // etc. For react adapters, this will likely just be `() => React.Version`, but for other
+  // adapters for libraries like inferno or preact, it will allow those libraries to specify
+  // a version of the API that they are committing to.
+  getTargetApiVersion(): string;
+
   // Provided a bag of options, return an `EnzymeRenderer`. Some options can be implementation
   // specific, like `attach` etc. for React, but not part of this interface explicitly.
-  createRenderer(RendererOptions?): EnzymeRenderer
+  createRenderer(?options: RendererOptions): EnzymeRenderer;
 
   // converts an RSTNode to the corresponding JSX Pragma Element. This will be needed
   // in order to implement the `Wrapper.mount()` and `Wrapper.shallow()` methods, but should
   // be pretty straightforward for people to implement.
-  nodeToElement(RSTNode): Element
+  nodeToElement(RSTNode): Element;
 }
 
 type EnzymeRenderer = {
   // both initial render and updates for the renderer.
-  render(Element): void
+  render(Element): void;
   
   // retrieve a frozen-in-time copy of the RST.
-  getNode(): RSTNode?
+  getNode(): RSTNode?;
 }
 ```
 
@@ -150,8 +168,7 @@ Enzyme.configure({ adapter: ThirdPartyEnzymeAdapter });
 Additionally, each wrapper Enzyme exposes will allow for an overriding `adapter` option that will use a
 given adapter for just that wrapper:
 
-```js
-
+```jsx
 import { shallow } from 'enzyme';
 import ThirdPartyEnzymeAdapter from 'third-party-enzyme-adapter';
 
--- a/docs/future/compatibility.md
+++ b/docs/future/compatibility.md
@@ -3,8 +3,8 @@
 
 ## Motivation
 
-This proposal is attempting to address a hand full of pain points that Enzyme has been
-suspect to for quite a while.  This proposal has resulted mostly [#715](https://github.com/airbnb/enzyme/issues/715),
+This proposal is attempting to address a handful of pain points that Enzyme has been
+subject to for quite a while.  This proposal has resulted mostly [#715](https://github.com/airbnb/enzyme/issues/715),
 and a resulting discussion among core maintainers of this project.
 
 The desired results of this proposal are the following:
@@ -94,16 +94,6 @@ type RSTNode = {|
 |}
 ```
 
-Thoughts:
-
-- If an `RSTNode` has `host: true`, `rendered` will always be an `Array` or `null`. (though this 
-might change with Fiber I guess).
-- `rendered` is a better property name than `children`, as the meaning of `children` is ambiguous
-here since it is defined on props/elements etc.
-- If a node has `host: true`, loosely speaking that means that an instance of that component
-- For "host" nodes, can `instance` ever be non-null? Could it be the DOM element? Yes/No?
-
-
 ### Enzyme Adapter Protocol
 
 **Definitions:**
@@ -175,7 +165,8 @@ import ThirdPartyEnzymeAdapter from 'third-party-enzyme-adapter';
 shallow(<Foo />, { adapter: ThirdPartyEnzymeAdapter });
 ```
 
-By default, Enzyme will ship with adapters for all major versions of React since React 0.13:
+Enzyme will build adapters for all major versions of React since React 0.13, though will deprecate
+adapters as usage of a particular major version fades.
 
 ```js
 import React13Adapter from 'enzyme-adapter-react-13';
--- a/docs/future/compatibility.md
+++ b/docs/future/compatibility.md
@@ -1,0 +1,174 @@
+# Enzyme Adapter & Compatibility Proposal
+
+
+## Motivation
+
+This proposal is attempting to address a hand full of pain points that Enzyme has been
+suspect to for quite a while.  This proposal has resulted mostly [#715](https://github.com/airbnb/enzyme/issues/715),
+and a resulting discussion among core maintainers of this project.
+
+The desired results of this proposal are the following:
+
+1. Cleaner code, easier maintenance, less bug prone.
+
+By standardizing on a single tree specification, the implementation of Enzyme would no longer have
+to take into account the matrix of supported structures and nuanced differences between different
+versions of React, as well as to some extent the differences between `mount` and `shallow`.
+
+2. Additional libraries can provide compatible adapters
+
+React API-compatible libraries such as `preact` and `inferno` would be able to provide adapters to Enzyme
+for their corresponding libraries, and be able to take full advantage of Enzyme's APIs.
+
+3. Better user experience (ie, bundlers won't complain about missing deps)
+
+Enzyme has had a long-standing issue with static-analysis bundlers such as Webpack and Browserify because
+of our usage of internal React APIs. With this change, this would be minimized if not removed entirely,
+since these things can be localized into the adapter modules, and users will only install the ones they need.
+
+Additionally, we can even attempt to remove the use of internal react APIs by lobbying for react-maintained packages
+such as `react-test-renderer` to utilize the React Standard Tree (RST) format (details below).
+
+4. Standardization and interopability with other tools
+
+If we can agree on the tree format (specified below as "React Standard Tree"), other tools can start to use and
+understand this format as well. Standardization is a good thing, and could allow tools to be built that maybe
+don't even exist yet.
+
+
+## Proposal
+
+
+### React Standard Tree (RST)
+
+This proposal hinges on a standard tree specification. Keep in mind that this tree needs to account for more
+than what is currently satisfied by the output of something like `react-test-renderer`, which is currently
+only outputting the "host" nodes (ie, HTML elements). We need a tree format that allows for expressing a full
+react component tree, including composite components.
+
+```js
+// Strings and Numbers are rendered as literals.
+type LiteralValue = string | number
+
+// A "node" in an RST is either a LiteralValue, or an RSTNode
+type Node = LiteralValue | RSTNode
+
+// An RSTNode has this specific shape
+type RSTNode = {|
+  // Either a string or a function. A string is considered a "host" node, and
+  // a function would be a composite component. It would be the component constructor or
+  // an SFC in the case of a function.
+  type: string | function
+  
+  // The props object passed to the node, which will include `children` in its raw form,
+  // exactly as it was passed to the component.
+  props: object
+  
+  // The backing instance to the node. Can be null in the case of "host" nodes and SFCs.
+  // React alternative libs can choose to provide whatever concept here applies.  If it is
+  // provided, Enzyme will expect it to follow the ComponentInterface type specified below
+  instance: ComponentInstance?
+  
+  // This is a transformed version of `props.children`. It is an array or null. If it's an array,
+  // it's an array of Nodes, meaning that they are LiteralNodes or RSTNodes. This means that 
+  // non-rendering values such as `false`, `null`, and `[]` wouldn't make it in here. This is also
+  // a "flat" array.
+  children: [Node]?
+  
+  // This can feel similar to children, but it's not the same thing at all. For a given node,
+  // this corresponds to the result of the `render` function with the provided props, but transformed
+  // into an RST.  For "host" nodes, this will always be `null`.
+  rendered: Node?
+|}
+```
+
+
+A `ComponentInstance` is going to be expected by enzyme to follow a certain interface in order
+for certain API methods to actually work, but they don't *need* to. Those methods just won't be
+guaranteed to work in those cases.
+
+
+So far, that interface is:
+```js
+type ComponentInstance = {
+  state: object?
+  context: object?
+  setState: function
+}
+```
+
+
+### Enzyme Adapter Protocol
+
+**Definitions:**
+
+An `Element` is considered to be whatever data structure is returned by the JSX pragma being used. In the 
+react case, this would be the data structure returned from `React.createElement`
+
+
+```js
+type RendererOptions = {
+  // I'm not 100% sure about this, but I think with an API like this, we could implement
+  // `mount` and `shallow` with the same renderer.
+  ?isLeaf(Element): bool
+}
+
+type EnzymeAdapter = {
+  // Provided a bag of options, return an `EnzymeRenderer`. Some options can be implementation
+  // specific, like `attach` etc. for React, but not part of this interface explicitly.
+  createRenderer(RendererOptions?): EnzymeRenderer
+
+  // converts an RSTNode to the corresponding JSX Pragma Element. This will be needed
+  // in order to implement the `Wrapper.mount()` and `Wrapper.shallow()` methods, but should
+  // be pretty straightforward for people to implement.
+  nodeToElement(RSTNode): Element
+}
+
+type EnzymeRenderer = {
+  // both initial render and updates for the renderer.
+  render(Element): void
+  
+  // retrieve a frozen-in-time copy of the RST.
+  getNode(): RSTNode?
+}
+```
+
+
+### Using different adapters with Enzyme
+
+At the top level, Enzyme would expose a `configure` method, which would allow for an `adapter`
+option to be specified and globally configure Enzyme's adapter preference:
+
+```js
+import Enzyme from 'enzyme';
+import ThirdPartyEnzymeAdapter from 'third-party-enzyme-adapter';
+
+Enzyme.configure({ adapter: ThirdPartyEnzymeAdapter });
+
+```
+
+Additionally, each wrapper Enzyme exposes will allow for an overriding `adapter` option that will use a
+given adapter for just that wrapper:
+
+```js
+
+import { shallow } from 'enzyme';
+import ThirdPartyEnzymeAdapter from 'third-party-enzyme-adapter';
+
+shallow(<Foo />, { adapter: ThirdPartyEnzymeAdapter });
+```
+
+By default, Enzyme will ship with adapters for all major versions of React since React 0.13:
+
+```js
+import React13Adapter from 'enzyme-adapter-react-13';
+import React14Adapter from 'enzyme-adapter-react-14';
+import React15Adapter from 'enzyme-adapter-react-15';
+// ...
+```
+
+### Validation
+
+Enzyme will provide an `validate(node): Error?` method that will traverse down a provided `RSTNode` and 
+return an `Error` if any deviations from the spec are encountered, and `null` otherwise. This will 
+provide a way for implementors of the adapters to determine whether or not they are in compliance or not.
--- a/docs/README.md
+++ b/docs/README.md
@@ -129,4 +129,5 @@
   * [Selectors](/docs/api/selector.md)
 * [Change Log](/CHANGELOG.md)
 * [Future](/docs/future.md)
+  * [Adapter & Compatibility Proposal](/docs/future/compatibility.md)
 * [Contributing Guide](/CONTRIBUTING.md)
--- a/docs/future.md
+++ b/docs/future.md
@@ -7,6 +7,11 @@ There are several things we'd like to address with Enzyme that often get asked. 
 of projects that we plan on addressing in the near future:
 
 
+### Enzyme Adapter & Compatibility
+
+[See the full proposal](future/compatibility.md)
+
+
 #### Improved CSS Selector Support
 
 Currently, "hierarchical" CSS selectors are not supported in Enzyme. That means that the only CSS
--- a/docs/future/rst_examples.js
+++ b/docs/future/rst_examples.js
@@ -65,11 +65,11 @@ renderer.render(<Qoo />);
 // tree = renderer.getNode();
 tree = {
   type: Qoo,
-  host: false,
+  nodeType: 'function',
   props: {},
   rendered: {
     type: 'span',
-    host: true,
+    nodeType: 'host',
     props: { className: 'Qoo' },
     rendered: ['Hello World!'],
   },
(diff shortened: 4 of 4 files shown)


docs/future/compatibility.md
```diff
@@ -0,0 +1,174 @@
...
+
+If we can agree on the tree format (specified below as "React Standard Tree"), other tools can start to use and
+understand this format as well. Standardization is a good thing, and could allow tools to be built that maybe
+don't even exist yet.
+
+
+## Proposal
+
+
+### React Standard Tree (RST)
```
ljharb: i'd thought of COM - Component Object Model - but this is probably better :-p

lelandrichardson:
@cpojer FYI, I have a branch of react that makes `react-test-renderer` return something kind of similar to what i'm proposing here. I m happy refactor it to actually implement this spec if we agree on it.

docs/future/compatibility.md
```diff
@@ -0,0 +1,174 @@
...
+type Node = LiteralValue | RSTNode
+
+// An RSTNode has this specific shape
+type RSTNode = {|
+  // Either a string or a function. A string is considered a "host" node, and
+  // a function would be a composite component. It would be the component constructor or
+  // an SFC in the case of a function.
+  type: string | function
+  
+  // The props object passed to the node, which will include `children` in its raw form,
```
ljharb: what about the context being passed to the node?

docs/future/compatibility.md
```diff
@@ -0,0 +1,174 @@
...
+A `ComponentInstance` is going to be expected by enzyme to follow a certain interface in order
+for certain API methods to actually work, but they don't *need* to. Those methods just won't be
+guaranteed to work in those cases.
+
+
+So far, that interface is:
+```js
+type ComponentInstance = {
+  state: object?
+  context: object?
```
ljharb: `context` seems like it should be alongside props, not alongside state

cpojer:
@lelandrichardson please consider the downstream effects of a change like this; if you break pretty-format, you will break every single snapshot test that's out there, so we need to make sure we have a proper upgrade path rather than making all tests fail on a react update with no new version of Jest or something. pretty-format is here: https://github.com/facebook/jest/tree/master/packages/pretty-format

lelandrichardson:
@cpojer definitely. I was actually envisioning it as a separate method (ie, `toRSTNode()` or something in addition to `toJSON`.

We aren't losing any information here, only gaining, so it wouldn't be hard to create a function that transformed an `RSTNode` to the current "JSON" value that react-test-renderer returns. That way, we could just implement the renderer once, but refactor `toJSON()` to return what it's currently returning.

I'm certain i'm not considering everything, but that's the kind of feedback I was hoping to get in this thread as well.

docs/future/compatibility.md
blainekasten: I do like subbing React for Component as it isn't tied to React but just the idea of a Component.

docs/future/compatibility.md
```diff
@@ -0,0 +1,174 @@
...
+
+```js
+// Strings and Numbers are rendered as literals.
+type LiteralValue = string | number
+
+// A "node" in an RST is either a LiteralValue, or an RSTNode
+type Node = LiteralValue | RSTNode
+
+// An RSTNode has this specific shape
+type RSTNode = {|
```
thysultan: suggest adding [nodeType<int>](https://developer.mozilla.org/en/docs/Web/API/Node/nodeType) to the shape that can be used to differentiate between non-element nodes like text nodes and fragment nodes.

cpojer:
yeah that's great, as long as enzyme objects at the end also provide a `toJSON` feature the get back that same React tree so it can be passed on to the pretty-printer, I think this is all sound from my point of view.

docs/future/compatibility.md
ljharb: CST already means "Concrete Syntax Tree" though (a superset of AST)

docs/future/compatibility.md
lelandrichardson: technically, `props` is also here, i'm just not listing it as we won't be using it directly. I'm actually tempted to just completely remove `context` here in this proposal

docs/future/compatibility.md
lelandrichardson: `nodeType` is specific to DOM, and something i'd prefer not to have in this proposal. I'm not exactly sure it would gain us anything either, as strings are represented in this tree as string literals, not nodes.

docs/future/compatibility.md
developit: - Virtual Syntax Tree (VST, already taken)
- VTree

docs/future/compatibility.md
ljharb: Why is it likely to change? Do Fiber, Inferno, or Preact not implement it?

docs/future/compatibility.md
ljharb: Technically it's not components, it's elements - Element Something Tree, or Something Element Tree?

docs/future/compatibility.md
thysultan: @lelandrichardson something like nodeType is not necessarily specific to the DOM. Inferno uses flags<int> to represent something similar for better performance inline with monomorphism.

docs/future/compatibility.md
lelandrichardson: I was kind of copying flux's "FSA" ("Flux Standard Action") with "React Standard Tree".

What's wrong with "RST"?  Redux !== Flux but people have no problem using FSA with redux.  We could use SET (Standard Element Tree), but I don't really like the aesthetics of it as much as RST because SET looks like a SQL statement more than an acronym, lol.

docs/future/compatibility.md
ljharb: Also, since v13, 14, and 15 all use context, I don't think we can just pretend it doesn't exist. Even if v16 removed it (which I doubt) we'd still need this spec to support it.

docs/future/compatibility.md
lelandrichardson: Also, thanks for weighing in! I hope i'm not dismissing a good idea too quickly!

docs/future/compatibility.md
Andarist: its often mentioned that the `context` was a feature implemented kinda in a spur of the moment and is not ideal, however it will stay as is until a better alternative is invented

docs/future/compatibility.md
ljharb: Both "react" and "component" aren't quite accurate - I think using "element" as well as having a good initialism are important. Agree that "set" could be confusing.

docs/future/compatibility.md
thysultan: @lelandrichardson So the idea was instead of using type for identifying a host/composite node via function vs string there could be a dedicated flag for this, in this case i suggested the name `nodeType` but it could be anything that is dedicated to identify the type of node, for example...

```
{
nodeType: 1   // host node - elements
nodeType: 2   // composite node - component constructors
nodeType: 3   // text node 
nodeType: 11 // fragment node - similar to a host node but different from element nodes
}
```
of course this will probably remove the need for `LiteralValue` since text nodes will be represented a `RSTNode` with a specific flag.

gaearon: Note that we plan to support more node types (e.g. Portals and Coroutines) in the future.
https://github.com/facebook/react/blob/b49210f4986a9ad9db0f7dadee568d4c74aa5231/src/renderers/shared/fiber/isomorphic/ReactTypes.js#L18

docs/future/compatibility.md
gaearon: Abstract Component Tree?

Not a fan of “virtual” suggestions since IMO there’s nothing “virtual” about component trees. “Virtual DOM” was just a gimmicky way to describe how React works but I don’t think this description helps at this point.

docs/future/compatibility.md
```diff
@@ -0,0 +1,174 @@
...
+  // The backing instance to the node. Can be null in the case of "host" nodes and SFCs.
+  // React alternative libs can choose to provide whatever concept here applies.  If it is
+  // provided, Enzyme will expect it to follow the ComponentInterface type specified below
+  instance: ComponentInstance?
+  
+  // This is a transformed version of `props.children`. It is an array or null. If it's an array,
+  // it's an array of Nodes, meaning that they are LiteralNodes or RSTNodes. This means that 
+  // non-rendering values such as `false`, `null`, and `[]` wouldn't make it in here. This is also
+  // a "flat" array.
+  children: [Node]?
```
gaearon: Could you explain why we need to duplicate information that’s always derivable from the `props`? Is this a convenience?

docs/future/compatibility.md
```diff
@@ -0,0 +1,174 @@
...
+  
+  // This can feel similar to children, but it's not the same thing at all. For a given node,
+  // this corresponds to the result of the `render` function with the provided props, but transformed
+  // into an RST.  For "host" nodes, this will always be `null`.
+  rendered: Node?
```
gaearon: In Fiber (and Preact/Inferno?) a composite may return a fragment (multiple elements).

docs/future/compatibility.md
lelandrichardson: @gaearon because we don't have the knowledge of how to transform an `Element` to a `Node`. That's the responsibility of the adapter.

`RSTNode.children !== RSTNode.props.children`. The `props.children` will be completely untouched (it could be anything, really). This also abstracts away the flattening etc. that react would do.

docs/future/compatibility.md
lelandrichardson: good point. I think in this case we would define `Node` as a recursive type:

```js
type Node = Literal | RSTNode | [Node]
```

Would that work?

thysultan: the nodeType flag could handle such cases since fragments could just be another type of Node i.e

```
{
nodeType: 11,
type: 'fragment',

...

children: [Node]
}
```

docs/future/compatibility.md
gaearon: If I have `<div><Foo /></div>`, and `<Foo />` renders `<Bar />`, what's going to be in `children` of `div`?

If it's truly a "transformed version of `props.children`", then it will be shallow `Foo` and which itself won't link to `Bar` RST node. Because `props.children` is shallow and does not contain rendering result.

However this doesn't make sense to me because per next comment, `rendered` is `null` for hosts. So we end up not having any way to traverse from `div` to `Bar`.

Coming back to this, it seems like `children` is not merely a "transformed version of `props.children`", but actual RST nodes corresponding to them. At which point I don't see why separate them from `rendered`. It's the same concept: an array of children nodes.

Imagine React DevTools tree. It has a universal concept of "children" (that's why it's a tree). Composites "contain" what they render, and hosts contain what's inside of them. `<div><Foo /></div>` is displayed in its fully expanded form: `<div><Foo><Bar><h1>I am bar content</h1></Bar></Foo></div>`.

I think for the same reason distinction between `rendered` and `children` is unnecessary. It is always up to traversing code "where to stop" unwrapping the tree. Composite component's children are what they render, and host component's children are based directly on what was passed to them. And you can always check if something is a composite or not by its type.

docs/future/compatibility.md
lelandrichardson: true. we could get rid of the idea of `Node` vs `RSTNode` and instead just represent an `RSTNode` as a disjoint union of types, hinging on `nodeType`

At the moment I don't have any context into what coroutines and portals would mean for this proposal... might have to do some reading into that...

docs/future/compatibility.md
lelandrichardson: OK, I think i see what you are saying now @thysultan , your link to the MDN docs was throwing me off at first.  I think this could be a reasonable way to define `RSTNode` as a disjoint union type instead of multiple types. 👍

docs/future/compatibility.md
```diff
@@ -0,0 +1,174 @@
...
+// A "node" in an RST is either a LiteralValue, or an RSTNode
+type Node = LiteralValue | RSTNode
+
+// An RSTNode has this specific shape
+type RSTNode = {|
+  // Either a string or a function. A string is considered a "host" node, and
+  // a function would be a composite component. It would be the component constructor or
+  // an SFC in the case of a function.
+  type: string | function
+
```
gaearon: Not sure if it's useful for your use case but we also keep track of "source location" for components in development. We read it from the `element.__source` property when [this transform](https://babeljs.io/docs/plugins/transform-react-jsx-source/) is enabled. We use it for element stacks in warnings but I imagine you could use this for friendlier errors in assertions maybe?

docs/future/compatibility.md
```diff
@@ -0,0 +1,174 @@
...
+An `Element` is considered to be whatever data structure is returned by the JSX pragma being used. In the 
+react case, this would be the data structure returned from `React.createElement`
+
+
+```js
+type RendererOptions = {
+  // I'm not 100% sure about this, but I think with an API like this, we could implement
+  // `mount` and `shallow` with the same renderer.
+  ?isLeaf(Element): bool
```
nfcampos: hmm, not sure this works, since you have no access to depth, so how would you implement the current shallow renderer, `{ isLeaf: () => true }`? does that mean this function would be called for all elements other than the one you pass to `EnzymeRenderer.render`?

docs/future/compatibility.md
gaearon: Is it useful to represent fragments as nodes? Or should their output be "inlined" into the thing that renders a fragment? I'm not sure what's more useful in your use case.

docs/future/compatibility.md
lelandrichardson: Hmm.  You're right that this proposal has a "bug" in it when I commented that `rendered` is `null` for host nodes. I think that's incorrect.

I see what you are saying w.r.t. a discrepency here with `children` and `rendered`.

I think it would be good for me to write out some example components and what the corresponding RST would be so that we can discuss in more concrete terms.

In your example:

```jsx
<div>
  <Foo>
    <Bar>
      <h1>I am bar content</h1>
    </Bar>
  </Foo>
</div>
```

Since this is a true (ie, one concept of children), we've lost the concept of ownership here, so we aren't able to identify that `<Foo />` was rendered with no children, unless we keep track of the "prop" `children` separate from the concept of true children:

What I'm trying to accomplish is a data structure that allows us to implement shallow rendering + full rendering and anywhere in-between. Let me try to add some examples to this document, as I'm reaching the limit of what I can reason about in my head :)

gaearon: I think there was some markup that disappeared in your comment.
What I meant to say is you can always grab original `children` from the props. So that should satisfy:

>we aren't able to identify that <Foo /> was rendered with no children

docs/future/compatibility.md
lelandrichardson: I think because they would have independent backing instances, we would want them to be separate nodes

docs/future/compatibility.md
thysultan: The way i see rendered being different is that given when type is a function(Component constructor) rendered will represent the node the component returns.

docs/future/compatibility.md
lelandrichardson: Yeah, I was considering adding `key` here, but i think maybe it isn't actually needed for enzyme's purposes. Perhaps it is for others?

lelandrichardson: 👍  Will this just show up as a `prop` though? In that case we wouldn't need to make it part of this proposal, but enzyme could indeed look for this prop to give better errors if it exists

docs/future/compatibility.md
gaearon: I understand but the distinction is already derivable from the type in cases where it's important (and I'm not sure how important it is in practice when writing APIs like `find()`).

I'd aim for avoiding duplicate/derivable information in the tree representation.

docs/future/compatibility.md
lelandrichardson: I should cook up an example for this as well. As i'm thinking about this more, I don't think this really enables us to emulate `shallow` exactly, but is definitely some functionality i've been wanting for configurable depth rendering. I'll have to think about this a little bit more. If you have any ideas def. let me know.

docs/future/compatibility.md
gaearon: Fragments wouldn't have backing instances. They are "invisible" to the application developer. Just like `<div>{[[[<div />], <span />], <p />]}</div>` would be "the same" as `<div><div /><span /><p /></div>` to anything outside the component itself.

docs/future/compatibility.md
lelandrichardson: > I'd aim for avoiding duplicate/derivable information in the tree representation.

Agreed!

The reason I still feel like we need a standalone `children` property is because I want to be able to use this representation to recreate the current behavior that enzyme's `shallow` API has. This uses the shallow renderer, which enzyme basically uses to traverse the result of `render` of a single component.

The result of `render` is a tree of `ReactElement`s. With this proposal in place, enzyme wouldn't know how to traverse a tree of `ReactElement`s, but it *would* know how to traverse a tree of `Node`s.

I'm not sure if my comments were accurate or not, and I'm going to give this more thought, but I'm pretty sure `children` is necessary if we aim to reimplement `shallow` on top of this spec.

docs/future/compatibility.md
gaearon: It seems like you want `resolveType(type) => type` that is identity by default. Then shallow rendering of `Foo` is `resolveType = type => type === Foo ? ShallowWrapper : type`. And `ShallowWrapper` would render `null` by default but its node would still have the original `type` so it would "appear" to be `Foo` in traversal.

lelandrichardson: @gaearon Yeah, I agree. Shallow rendering is effectively the same as mocking.

Would it render `null` or `this.props.children` though?

docs/future/compatibility.md
lelandrichardson: so if I have a component `Foo` that renders `[<div />, <span />]`, is the right mental model to change that in my head to something like `<react:fragment>{[<div />, <span />]}</react:fragment>`?

If so, I wasn't understanding earlier, as I was thinking it would have been the node with `type: Foo` that would have `nodeType: ReactFragment`.

docs/future/compatibility.md
gaearon: I think you might be on point with `props.children`.

docs/future/compatibility.md
gaearon: > is the right mental model to change that in my head to something like `<react:fragment>{[<div />, <span />]}</react:fragment>`?

This is the right model.

docs/future/compatibility.md
milesj: For what it's worth, context support was added to React fiber. I don't think they're going away anytime soon.

ljharb: If I write a component that relies on `context`, and the tests pass for React 15, I should be able to switch the AST generator to, say, inferno, and have it work identically (assuming that inferno has the concept of "context", in this hypothetical). If `context` isn't in this spec, how can that be achieved?

docs/future/compatibility.md
nfcampos: I always imagined shallow rendering as effectively treating composite components as host components, so it would make sense for it to "render `props.children`" I think

docs/future/compatibility.md
blainekasten: So, stepping back to the definition. We are asking for a tree, describing the components that would be rendered. While Element is sort of correct, element is loaded with terms in HTML itself. So In my head, that represents a few different key words:

- Abstract
- Component
- Tree
- Description

I think we use those words in some order, all or part of them.

ACT (Abstract Component Tree)
CDT (Component Description Tree)
ACDT (Abstract Component Description Tree)

ljharb: Yes but in react, "component" is the element constructor, and a div isn't a component.

aweary: And an element isn't technically a "component", so using either element or component excludes the other by that definition. If you wanted to be as descriptive as possible you could do Abstract Component and Element Tree (ACET).

Is there a term that describes both elements and components? If not we may just have to adjust our definition of "component" or "element" within the context of the spec, and just be explicit about it.

docs/future/compatibility.md
aweary: Would all possible `NodeType`s be defined by the spec? Things like Coroutines are very specific to React right now and defining that in the spec would feel like we're coupling it too heavily to one implementation.

docs/future/compatibility.md
```diff
@@ -0,0 +1,174 @@
...
+### Using different adapters with Enzyme
+
+At the top level, Enzyme would expose a `configure` method, which would allow for an `adapter`
+option to be specified and globally configure Enzyme's adapter preference:
+
+```js
+import Enzyme from 'enzyme';
+import ThirdPartyEnzymeAdapter from 'third-party-enzyme-adapter';
+
+Enzyme.configure({ adapter: ThirdPartyEnzymeAdapter });
```
aweary: I think the explicit API is best, I'm not a huge fan of "magical" auto-configuration approaches like that. What happens if they update from React 14 to 15 and forget to remove `enzyme-adapter-react-14` from their dependencies?

Maybe that could be a long-term goal once we're more familiar with actual use patterns.

docs/future/compatibility.md
ljharb: Nothing in this tree is a component tho - it's an instance of a component, which is an "element" - the similarity to HTML is no doubt intentional, but also conceptually accurate, and I think that's something we should follow.

docs/future/compatibility.md
lelandrichardson: I'm more partial to the explicit API as well. It also makes it really clear how to use different adapters in the same test suite.  Though I am kind of bummed we will be joining the ranks of tools that don't "just work" with one npm install :/

docs/future/compatibility.md
lelandrichardson: @aweary I agree. I'm not sure how I feel about the node types that don't seem to apply to enzyme... but I need to understand them more. If they become a part of react and enzyme needs to know about them in order to provide useful testing APIs, we might still need them.

docs/future/compatibility.md
lelandrichardson: As far as this spec is currently written, both `context` and `state` are things that will "just work" so long as the libraries implement their public facing API the same.  I'm not really seeing these two concepts as belonging in this spec, as they are not relevant to traversal. Even most props don't really matter to this proposal... the only reason `props` are here is because we use them for traversal. We don't use `context` or `state` for any such things.

I think we shouldn't add things to this spec that aren't needed?

docs/future/compatibility.md
lelandrichardson: `node.instance` is an instance of a component, but nothing else in this proposal is.  As far as react-isms go, these nodes are closest semantically to what React calls "Elements".

How about removing both terms though?

"SRT": "Standard Render Tree"
"RSRT": "React Standard Render Tree"
"RST": "React Standard Tree"
"RRT": "React Render Tree"
"ART": "Abstract Render Tree" (ART is ambiguous with React ART)
"ARRT": "Abstract React Render Tree"

Honestly,  I don't see a problem with using "React" in the name. It doesn't mean that non-react libraries can't use it, it just means that they are following that spec.

sigh. this is such a 🚲 🏠  lol

docs/future/compatibility.md
ljharb: I guess I wasn't under the impression that this spec was just for traversal - I was thinking it would be a way to render the entire tree with 100% fidelity.

docs/future/compatibility.md
aweary: >Nothing in this tree is a component tho - it's an instance of a component, which is an "element" 

`RSTNode.type` can be a function which represents a component, so there are nodes in the tree that correspond to components that we treat as components with APIs like `find(FooComponent)`.

docs/future/compatibility.md
lelandrichardson: I think if we want to iron down things about context, that we can do so by specifying it in the `ComponentInstance` type and the `EnzymeRenderer` API.... but I don't think it needs to be part of the tree structure outside of that.

I could see more APIs being put on `EnzymeRenderer` like `provideContextForNode` and `setStateForNode` or something like that, but I was trying to make the API as minimal as possible, and saw that as potentially unneeded.

I think the tree structure (ie, `RSTNode`) is really just about traversal though.

docs/future/compatibility.md
nfcampos: how about Renderer Standard Tree for the name?

gaearon: >Nothing in this tree is a component tho - it's an instance of a component, which is an "element" 

This sounds wrong to me.

An element in React is just a plain object representing what you want to see.
Like `<div />` or `<Foo />`.

An element is not "expanded". It is always shallow. 

The tree we are discussing is not a tree of elements. We are discussing a fully "expanded" tree where renderer has *read* the elements and produced a tree based on those by recursively "expading" them.

gaearon: RCT: Reactish Component Tree. 😄

aweary: "Expanded" Component Tree? 😉

ljharb: @gaearon yes but those aren't components anymore - a component is what you use to *create* that plain object (the react element).

gaearon: Right! I mean that those are neither components nor elements really. Colloquially we often say "component tree" though. "Element tree" sounds more technical to me, and misleading for this reason: element tree is exactly what is being returned from render. Anyway, I don't really care.

ljharb: That's a good point too - I'd rather invent a new name than pick a name that isn't accurate, or could be misleading. lots of 🚲 🏠 on this one, for sure.

"Concrete Render Tree"? i dunno

docs/future/compatibility.md
```diff
@@ -0,0 +1,174 @@
...
+  // it's an array of Nodes, meaning that they are LiteralNodes or RSTNodes. This means that 
+  // non-rendering values such as `false`, `null`, and `[]` wouldn't make it in here. This is also
+  // a "flat" array.
+  children: [Node]?
+  
+  // This can feel similar to children, but it's not the same thing at all. For a given node,
+  // this corresponds to the result of the `render` function with the provided props, but transformed
+  // into an RST.  For "host" nodes, this will always be `null`.
+  rendered: Node?
+|}
```
blainekasten: Are we missing `refs` on purpose?

docs/future/compatibility.md
```diff
@@ -0,0 +1,174 @@
...
+for certain API methods to actually work, but they don't *need* to. Those methods just won't be
+guaranteed to work in those cases.
+
+
+So far, that interface is:
+```js
+type ComponentInstance = {
+  state: object?
+  context: object?
+  setState: function
```
blainekasten: Are we missing `refs` on purpose?

lelandrichardson: i don't think we should standardize on `refs`. As it stands right now, it's a behavior that's being "deprecated" in favor of people handling refs with functions themselves

blainekasten: I think that's fair. But what impact will that have on the Enzyme API. Do we have to deprecate the `.ref` api if we do not have a _safe_ way to look it up?

gaearon: Can you provide an API to read arbitrary fields from the instance? That's what callback refs set.

ljharb: How would we be able to know, from a callback ref, what changes were made? Sometimes people store the refs in an object, and sometimes they're not stored at all :-/

lelandrichardson: This leaves the world in no different of a place than where we already are. As it stands, the `ref` API on enzyme only works with people using the string ref API, and using `instance().nameOfProperty` works and always has. This proposal needs only be about traversing the tree. Everything else can and will work identically to how it has.

docs/future/compatibility.md
```diff
@@ -0,0 +1,191 @@
+# Enzyme Adapter & Compatibility Proposal
+
+
+## Motivation
+
+This proposal is attempting to address a hand full of pain points that Enzyme has been
```
ljharb: `s/hand full/handful`

docs/future/compatibility.md
```diff
@@ -0,0 +1,191 @@
...
+suspect to for quite a while.  This proposal has resulted mostly [#715](https://github.com/airbnb/enzyme/issues/715),
```
ljharb: `s/suspect/subject` (also, no need to put two spaces after a period unless on a typewriter ;-) )

docs/future/compatibility.md
```diff
@@ -0,0 +1,191 @@
...
+given adapter for just that wrapper:
+
+```jsx
+import { shallow } from 'enzyme';
+import ThirdPartyEnzymeAdapter from 'third-party-enzyme-adapter';
+
+shallow(<Foo />, { adapter: ThirdPartyEnzymeAdapter });
+```
+
+By default, Enzyme will ship with adapters for all major versions of React since React 0.13:
```
ljharb: if React 16 comes out before this is done, then we'd probably support 0.14 - 16 instead? 

Perhaps more accurate to state "the latest, and previous two, major versions of React"?

lelandrichardson:
I've revised this proposal, considering some of the feedback given here.  I've chatted offline with @ljharb about this at length, and think that this is generally moving in the right direction.

Some thoughts around the changes:

 **1. Do we need a `nodeType` property?**  

This is still debatable. I don't think we need to care about "fragment" types in the way the spec is now, but we could alter the proposal to wrap literal values and other things into nodes of the graph with a specified `nodeType`, and represent an `RSTNode` as a disjoint union type. I feel like the current approach is pretty reasonable though. 

  **2. What about `state`, `context`, `refs`, etc.?**

  These APIs are not needed for traversal and are not going to be included in this spec. This spec is explicitly stating that the `instance` property on an `RSTNode` implements the public API of a  React Component in the corresponding release signified by `getTargetVersion`. As a result, we are effectively saying that other than the specified structure of the `RSTNode`, Enzyme will be following whatever other API React chooses to go with.   

**3. Do we need both `rendered` and `children`?**  

Ultimately I decided that Dan was right about this. I do think there is more information provided by `rendered` and `children` combined, but I don't actually think that it's useful to us, and ends up creating a LOT of duplicate data in the tree. I've decided to use `rendered` as the namespace because this is semantically more accurate, and not as ambiguous. With the setup I've laid out here, "Host" nodes will end up having `rendered` be an Array, whereas "Composites" will end up having `rendered` be another `RSTNode`. When you think about it, this makes a bit of sense: "Composites" are in effect "wrapping" Host nodes.   

**4. What about Portals and Coroutines?**

  I still don't know what these mean for this proposal. If anyone has thoughts on this, please let me  know.   

**5. Can we achieve `shallow` with just `isHost(Element): boolean`?**

  At the moment, I think the answer to this is "sort of, but with a caveat". I think we certainly can if we say that the `isHost` function is *not* called on the root element that gets passed into the renderer. In this case, `isHost: () => true` would be all we would need to implement the shallow renderer. This actually seems pretty reasonable to me as a solution.

  **6. For "host" nodes, can `instance` ever be non-null?** 

 I haven't specified this yet. It doesn't have a `ComponentInstance`, but we could have it return  the actual DOM node if we are working with something like `mount`. Could it be the DOM element? Yes/No?? 

----

Let me know what ya'll think. If we can get alignment on this, I think we can start working on implementation in this branch.  My hope is that we can get something like this done before React 16 so we can be sure there are no big barriers for people migrating.

docs/future/compatibility.md
lelandrichardson: I'm actually maybe rethinking this. Perhaps enzyme shouldn't ship with any adapter? Those are all added via a second dependency and import?

As for supporting react 0.13. Structuring the project this way should make supporting react 0.13 much much easier.

docs/future/compatibility.md
```diff
@@ -0,0 +1,191 @@
...
+|}
+```
+
+Thoughts:
+
+- If an `RSTNode` has `host: true`, `rendered` will always be an `Array` or `null`. (though this 
+might change with Fiber I guess).
+- `rendered` is a better property name than `children`, as the meaning of `children` is ambiguous
+here since it is defined on props/elements etc.
+- If a node has `host: true`, loosely speaking that means that an instance of that component
```
aweary: I think this point got cut off?

docs/future/compatibility.md
blainekasten: I also think adapters should not be shipped automatically. For users of inferno/preact it becomes bloat. Even for users of React on a specific version, it becomes bloat getting adapters for other versions.

I'm thinking we have a specific `react-enzyme-adapter` that matches react versions. I don't think we should have different repos for each version of react.

blainekasten:
> 2. What about state, context, refs, etc.?
> ...Enzyme will be following whatever other API React chooses to go with.   

I think it's worth documenting this API regardless so adapter implementors have to only look at the spec to build an adapter.

lelandrichardson:
@blainekasten I hesitate to document it because I don't want to make this spec more constraining for the react core team than it has to be. It is their responsibility to decide what the public API of react is, and I don't want enzyme to be sending different messages of them. I see this as our way of saying enzyme is following the changes of react, not the other way around (and thus adapter maintainers would be expected to do the same).

docs/future/compatibility.md
```diff
@@ -0,0 +1,182 @@
...
+
+// An RSTNode has this specific shape
+type RSTNode = {|
+  // Either a string or a function. A string is considered a "host" node, and
+  // a function would be a composite component. It would be the component constructor or
+  // an SFC in the case of a function.
+  type: string | function;
+
+  // Whether or not this node is a "Host" node.
+  host: boolean;
```
gaearon: For supporting coroutines and portals, you might want to add a `nodeType` after all (we call it `tag` internally in Fiber). It lets you keep track of future React changes. It's fine if it's just `'host' | 'class' | 'function'` for now, and adds `'portal' | 'coroutine'` at some point later.

lelandrichardson: this is good feedback, thanks. What do you think about also having a `literal` `nodeType` that wraps things like numbers/strings/etc.?

gaearon: In Fiber, we only have string literals at the renderer level, numbers get converted to strings before that.

I don't see a benefit in wrapping all literals into nodes. Let's just always use strings. Technically we *do* create fibers for literals when there are several of them but it's an implementation detail and depends on the renderer. For example, DOM renderer inlines a single text child, but test renderer may choose not to.

gaearon: Also, regarding the above comment: I meant you probably want to replace `host: boolean` with `nodeType/tag: union`. Since that covers `host: boolean` already.

docs/future/compatibility.md
ljharb: Yeah I'm fine not shipping any adapter directly, but we'd have to be very very clear with instructions and error messages so people know how to add one.

docs/future/compatibility.md
lelandrichardson: @gaearon understood. thanks. I'll make some updates to use `nodeType`.

danielbayerlein:
Awesome! I'm very excited to use it! 💃

docs/future/compatibility.md
blainekasten: RAT: Reactish Abstract Tree.

It's a tree of abstraction over the elements produced by reactish environments

IgnusG:
Is this proposal still valid or has it been postponed/scraped?

ljharb:
@IgnusG still valid.

ljharb:
Since #1007 is merged, I think this can be closed. Feel free to reopen if not.

PR #755, opened 2017-01-02
Fix entry point for TypeScript transpile

kenju:
Though this is not this library's problem, here is a tweak fix necessary for TypeScript project.

## Documents

"Untyped imports" Section:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html

```ts
// Succeeds if `node_modules/asdf/index.js` exists
import { x } from "asdf";
```

## FYI

- Common PR ( resolved ) in other library: https://github.com/reactjs/react-modal/pull/294

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "enzyme",
   "version": "2.7.0",
   "description": "JavaScript Testing utilities for React",
-  "main": "build",
+  "main": "build/index.js",
   "scripts": {
     "prepublish": "not-in-publish || (npm run clean && npm run build && safe-publish-latest)",
     "preversion": "npm run clean && npm run check",


ljharb:
This change should absolutely not be necessary, since it's how node's module resolution algorithm has always worked and will always work.

The best practice in JS - which TypeScript deviates from at its peril - is to always omit `index` filenames, as well as extensions.

kenju:
Okay, thanks for yoru quick feedback.

For those who may have the same problem, here is a temporar workaround:

```
import { shallow } from 'enzyme/build/index.js';
```

kenju:
@ljharb 

here is a related issue, and its milestone says this bug will be fixed in TypeScript 2.2
https://github.com/Microsoft/TypeScript/issues/12839

PR #756, opened 2017-01-03
[docs] add link to Cheerio API and another example

ChrisWren:
I wasn't clear on what the API was for a wrapper returned from using `render` so I thought this might be helpful.

### DIFF
--- a/docs/api/render.md
+++ b/docs/api/render.md
@@ -10,7 +10,7 @@ disservice.
 
 For the purposes of this documentation, we will refer to Cheerio's constructor as
 `CheerioWrapper`, which is to say that it is analogous to our `ReactWrapper` and `ShallowWrapper`
-constructors.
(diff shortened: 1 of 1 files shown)


docs/api/render.md
```diff
@@ -11,7 +11,7 @@ disservice.
 
 For the purposes of this documentation, we will refer to Cheerio's constructor as
 `CheerioWrapper`, which is to say that it is analogous to our `ReactWrapper` and `ShallowWrapper`
-constructors.
+constructors. [Here](https://github.com/cheeriojs/cheerio#api) is a link to the API available from a `CheerioWrapper` instance.
```
aweary: Can we rephrase this to:

> You can reference the [Cheerio API docs](https://github.com/cheeriojs/cheerio#api) for methods available on a `CheerioWrapper` instance.

I'm not a big fan of providing links with "here" if we can integrate it into a natural sentence otherwise

ljharb:
This may need to be updated, since v3 uses cheerio v1.

PR #761, opened 2017-01-05
[Docs] Fix alphabetical order of API lists

koba04:
I suppose these API lists are sorting as alphabetical order, aren't they?
This PR is to fix the order.

BTW, Is the following list sorting as some intentional order?

http://airbnb.io/enzyme/docs/api/shallow.html#shallowwrapper-api

Thanks.


ljharb:
I don't think either list is intended to be sorted in any particular order. I don't mind alphabetizing it, but I also don't think it matters - people will use "find in page" to find something specific, or will browse through them all.

koba04:
I think [ShallowWrapper API](http://airbnb.io/enzyme/docs/api/shallow.html#shallowwrapper-api) looks good to me.

But [Table of Contents](https://github.com/airbnb/enzyme/blob/master/docs/README.md) is cofusing because it's almost alphabetical order, and it's not same as [ShallowWrapper API](http://airbnb.io/enzyme/docs/api/shallow.html#shallowwrapper-api).
(ex. `is(selector)` and `isEmpty()` are not adjacent)

When I write tests with enzyme, I look for a suitable API in the API list of Table of Contents.
I think it would be nice if the list is sorted as some rule.

blainekasten:
I am 👍 for alphabetical lists.

blainekasten:
can you rebase this over master so we can merge it?

koba04:
@blainekasten Thank you! Rebased!

blainekasten:
Thank you @koba04 !

koba04:
Thanks!

Issue #410, opened 2016-05-24
TypeError: Cannot read property 'reverse' of null


vlkosinov:
Thank you very much for this library! ❤️ 

I'm trying to test a library which provides some Higher-order Component, but I get a weird bug.
#### Example:

**Enhancer:**

``` js
function enhance(ComposedComponent) {
    class Enhancer extends Component {
        render() {
            const newProps = Object.assign({}, this.props, {
                someAdditionalProp: true    
            });
            return createElement(WrappedComponent, newProps);
        }
    }

    return Enhancer;
}
```

**Component to enchace:**

``` js
class Foo extends Component {
    render() {
        return (<div>Bla bla</div>);
    }
}
const FooEnhanced = enhance(Foo);
```

**Main component:**

``` js
class App extends React.Component {
    render() {
        return (
            <div>
                <FooEnhanced />
            </div>
        );
    }
}
```

I want to ensure that the Enhancer adds additional prop 'someAdditionalProp' to composed components, so I'm doing the following:

``` js
const wrapper = mount(<App />);
const enhancedComponents = wrapper
    .find(App)
    .children()
    .findWhere(n => typeof n.type() !== 'string')
    .findWhere(n => n.parent().is(FooEnhanced));

enhancedComponents.forEach(node => {
    expect(node.prop('someAdditionalProp')).to.be.a('boolean');
});
```

And get error message:

``` bash
/Users/vladkosinov/projects/test/node_modules/enzyme/build/MountedTraversal.js:199
  return pathToNode(inst, root).reverse();
                               ^

TypeError: Cannot read property 'reverse' of null
    at parentsOfInst (/Users/vladkosinov/projects/test/node_modules/enzyme/build/MountedTraversal.js:199:32)
    at ReactWrapper.<anonymous> (/Users/vladkosinov/projects/test/node_modules/enzyme/build/ReactWrapper.js:841:54)
    at ReactWrapper.single (/Users/vladkosinov/projects/test/node_modules/enzyme/build/ReactWrapper.js:1297:19)
    at ReactWrapper.parents (/Users/vladkosinov/projects/test/node_modules/enzyme/build/ReactWrapper.js:840:41)
    at ReactWrapper.<anonymous> (/Users/vladkosinov/projects/test/node_modules/enzyme/build/ReactWrapper.js:860:21)
    at /Users/vladkosinov/projects/test/node_modules/enzyme/build/ReactWrapper.js:1166:21
    at Array.map (native)
[...]

PR #769, opened 2017-01-12
Add tests on broken .parent() use-case

just-boris:
When I traverse through React components in a mounted tree, it can't find parent and fails with the error

```
TypeError: Cannot read property 'reverse' of null
   at parentsOfInst (src/MountedTraversal.js:255:10)
   at ReactWrapper.<anonymous> (src/ReactWrapper.jsx:623:35)
   at ReactWrapper.single (src/ReactWrapper.jsx:916:21)
```

See added test as a reproduction example.

As far as I could track, it happens, because the node can't find itself because component wrapper the wrong children `wrapper.children()` in this test returns an empty array, however we have a node inside.

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -2422,6 +2422,20 @@ describeWithDOM('mount', () => {
       expect(parents.at(0).hasClass('foo')).to.equal(true);
       expect(parents.at(1).hasClass('bax')).to.equal(true);
     });
+
+    it('should work with components in the tree', () => {
+      const Foo = createClass({
+        render() {
+          return <div className="bar" />;
+        },
+      });
+      const wrapper = mount((
+        <div className="root">
+          <Foo />
+        </div>
+      ));
+      expect(wrapper.find('.bar').parents('.root')).to.have.length(1);
+    });
   });
 
   describe('.parent()', () => {
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -2434,7 +2434,14 @@ describeWithDOM('mount', () => {
           <Foo />
         </div>
       ));
-      expect(wrapper.find('.bar').parents('.root')).to.have.length(1);
+      const root = wrapper.find('.root');
+      expect(root).to.have.lengthOf(1);
+      expect(root.hasClass('root')).to.equal(true);
+      expect(root.hasClass('bar')).to.equal(false);
+
+      const bar = root.find('.bar');
+      expect(bar).to.have.lengthOf(1);
+      expect(bar.parents('.root')).to.have.lengthOf(1);
     });
   });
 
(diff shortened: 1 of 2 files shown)


ljharb:
(linking to #410)

test/ReactWrapper-spec.jsx
```diff
@@ -2051,6 +2051,12 @@ describeWithDOM('mount', () => {
       expect(parents.at(1).hasClass('bar')).to.equal(true);
       expect(parents.at(2).hasClass('bax')).to.equal(true);
     });
+
+    it('should work with components in the tree', () => {
+      const Foo = () => (<div className="bar" />);
+      const wrapper = mount(<Foo />);
+      expect(wrapper.find('.bar').parent()).to.equal(wrapper.find('Foo'));
```
ljharb: I'm not convinced this should actually work - `Foo` isn't supposed to be in the tree at all; only what Foo *renders* is.

just-boris: Maybe, not sure what should be at the right side. I haven't had a chance to look what it returns because it fails now.

ljharb: Could you add a similar test to `ShallowWrapper`, so we can compare the behaviors?

just-boris: Done. Shallow wrapper returns empty wrapper in this case.

Also, I have updated test for mounted wrapper. Now it expects that it has`.root` as a parent. It is not possible to do the same for shallow because it doesn't render components deeply.

just-boris:
Hello, @ljharb 
I have updated the test and added a new one for shallow render. 
Any more feedback?

test/ShallowWrapper-spec.jsx
```diff
@@ -1897,6 +1897,16 @@ describe('shallow', () => {
...
     });
+
+    it('should work with component', () => {
+      const Foo = React.createClass({
+        render() {
+          return <div className="bar" />;
+        },
+      });
+      const wrapper = shallow(<Foo />);
+      expect(wrapper.find('.bar').parent()).to.have.length(0);
```
ljharb: In this case, `.find('.bar')` *is* the root - I'm not even sure `.find('.bar')` will find anything. Could you add an assertion that `wrapper.find('.bar')` has length 1?

just-boris: Done.
Shallow test is passing, mounted is not.

src/MountedTraversal.js
```diff
@@ -257,7 +257,7 @@ export function pathToNode(node, root) {
 }
 
 export function parentsOfInst(inst, root) {
-  return pathToNode(inst, root).reverse();
+  return (pathToNode(inst, root) || []).reverse();
```
blainekasten: it seems to me that pathToNode should just always return an array instead of having this `|| []` logic.

ljharb: When I change that, another test breaks that I can't figure out. I'll push that commit up so you can play with it.

alexreardon:
any reason this was abandoned?

ljharb:
@alexreardon it's not abandoned, it's just that nobody's provided a fix yet. I updated the PR just a few days ago

alexreardon:
I missed that. Cheers

benlakey:
@alexreardon if you're looking for a quick (bad, slow) workaround:

```javascript
const parent = ofNode => ofNode.root.findWhere(n => n.node == ofNode.node.parentNode);
```

**warning: you probably shouldn't use this because its a pretty gross hack and quite slow**

izverev86:
any updates?

ljharb:
No updates; but I think this is the same as #1235

ljharb:
This appears to have been addressed in v3, as the tests are now passing. I'm going to merge the tests and close the linked issue.

PR #771, opened 2017-01-13
[Fix] `dive` should pass along options from `shallow`

### LINKED ISSUE #770: `shallow(<Component />, options1).dive(options2)` should merge and pass options through
https://github.com/airbnb/enzyme/blob/master/src/ShallowWrapper.js#L1033 - currently `.dive()` takes options and passes those along, but it doesn't inherit the original options passed to `shallow`.

It should.

blackpost38:
I made it, could you check this source?

I only added a logic that merges options to `dive()`

refs: #770

### DIFF
--- a/src/ShallowWrapper.js
+++ b/src/ShallowWrapper.js
@@ -3,6 +3,7 @@ import flatten from 'lodash/flatten';
 import unique from 'lodash/uniq';
 import compact from 'lodash/compact';
 import cheerio from 'cheerio';
+import assign from 'object.assign';
 
 import ComplexSelector from './ComplexSelector';
 import {
@@ -1022,7 +1023,7 @@ class ShallowWrapper {
    * @param options object
    * @returns {ShallowWrapper}
    */
-  dive(options) {
+  dive(options = {}) {
     const name = 'dive';
     return this.single(name, (n) => {
       if (isDOMComponentElement(n)) {
@@ -1031,7 +1032,7 @@ class ShallowWrapper {
       if (!isCustomComponentElement(n)) {
         throw new TypeError(`ShallowWrapper::${name}() can only be called on components`);
       }
-      return new ShallowWrapper(n, null, options);
+      return new ShallowWrapper(n, null, assign({}, this.options, options));
     });
   }
 }
--- a/test/ShallowWrapper-spec.jsx
+++ b/test/ShallowWrapper-spec.jsx
@@ -3842,11 +3842,18 @@ describe('shallow', () => {
         return <RendersDOM />;
       }
     }
+    WrapsRendersDOM.contextTypes = { foo: React.PropTypes.string };
     class DoubleWrapsRendersDOM extends React.Component {
       render() {
         return <WrapsRendersDOM />;
       }
     }
+    class ContextWrapsRendersDOM extends React.Component {
+      render() {
+        return <WrapsRendersDOM />;
+      }
+    }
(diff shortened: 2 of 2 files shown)


src/ShallowWrapper.js
```diff
@@ -3,6 +3,7 @@ import flatten from 'lodash/flatten';
 import unique from 'lodash/uniq';
 import compact from 'lodash/compact';
 import cheerio from 'cheerio';
+import merge from 'lodash/merge';
```
ljharb: we have `object.assign` available, please use that instead of adding a new dependency.

src/ShallowWrapper.js
```diff
@@ -1023,14 +1024,15 @@ class ShallowWrapper {
...
+    const originalOptions = this.options;
     return this.single(name, (n) => {
       if (isDOMComponentElement(n)) {
         throw new TypeError(`ShallowWrapper::${name}() can not be called on DOM components`);
       }
       if (!isCustomComponentElement(n)) {
         throw new TypeError(`ShallowWrapper::${name}() can only be called on components`);
       }
-      return new ShallowWrapper(n, null, options);
+      return new ShallowWrapper(n, null, merge({}, originalOptions, options));
```
ljharb: I think this can use `this.options` inline - the arrow function will ensure `this` is correct.

blackpost38: you're right. okay, I'll fix it.

src/ShallowWrapper.js
blackpost38: I think we needs `merge` because `options` might be nested object.

```js
this.options = { context: {foo: 'hello'}};
options = {context: {bar: 'enzyme!'}};
assign({}, this.options, options)  // It will be {context: {bar: 'enzyme!'}}
```

nfcampos: deep merging the options sounds like too much, i think eg merging the context as in your example should be done manually by the user. thoughts?

nfcampos:
@ljharb this might be a major, no? it can break existing tests for components that depend on context being available/not available.

src/ShallowWrapper.js
ljharb: Yes, I think we want to avoid deep merging.

ljharb:
@nfcampos my thinking is that a component that expects context would never be used or tested without context - thus, what would be the use case for having a test that this change breaks?

This also will pass through the lifecycle flag, but we're planning to enable that by default in the next major anyways.

src/ShallowWrapper.js
blackpost38: oh, okay. i got it. I'll fix this and then send PR 😄 

thanks for your feedbacks 👍

blackpost38:
I fixed it. I tried to apply your feedbacks.

Could you check it out? thanks!

nfcampos:
@ljharb only got to this now — ive been without internet at home for the better part of two weeks :( — thinking about this some more I could agree with this being a patch, arguably the previous behaviour was wrong, the expectation in react for a child to receive the same thing in context as its parent, what do you think?

ljharb:
I think I agree that it should be a patch.

nfcampos:
@blackpost38 do you mind reabsing this on latest master?

blackpost38:
@nfcampos It's done.

nfcampos:
Thanks @blackpost38 !

PR #776, opened 2017-01-17
[Fix] mount/shallow: "match" should not differentiate nullary props

ljharb:
This PR fixes the matching APIs.

`<div>` and `<div id={undefined} />` absolutely should match, because React, most of JS, and human intuition does not distinguish between "absent" and "undefined". This is inarguably a bug fix.

This PR also asserts that `<div>` and `<div id={null} />` should match. JS doesn't treat them this way - and React treats them differently only in the case of defaultProps. However, by the time `enzyme` is attempting to match elements, any defaultProps have been applied - and my contention is that absent a defaultProp, `null`, `undefined`, and "absent" should all be treated the same.

This came up specifically because I attempted to add missing defaultProps, set to `null`, on a component, and tests using enzyme's matching APIs began to fail. I don't believe they should have failed.

I consider null-matching expected behavior from these APIs, and I think that this is unlikely to cause any tests to fail, or falsely pass - thus I'm convinced it's a semver-patch.

### DIFF
--- a/test/Utils-spec.jsx
+++ b/test/Utils-spec.jsx
@@ -9,6 +9,7 @@ import {
   coercePropValue,
   getNode,
   nodeEqual,
+  nodeMatches,
   isPseudoClassSelector,
   propFromEvent,
   SELECTOR,
@@ -21,7 +22,6 @@ import { REACT013 } from '../src/version';
 describe('Utils', () => {
 
   describeWithDOM('getNode', () => {
-
     it('should return a DOMNode when a DOMComponent is given', () => {
       const div = mount(<div />).getNode();
       expect(getNode(div)).to.be.instanceOf(window.HTMLElement);
@@ -45,27 +45,21 @@ describe('Utils', () => {
   });
 
   describe('nodeEqual', () => {
-
     it('should match empty elements of same tag', () => {
-
       expect(nodeEqual(
         <div />,
         <div />,
       )).to.equal(true);
-
     });
 
     it('should not match empty elements of different type', () => {
-
       expect(nodeEqual(
         <div />,
         <nav />,
       )).to.equal(false);
-
     });
 
     it('should match basic prop types', () => {
-
       expect(nodeEqual(
         <div className="foo" />,
         <div className="foo" />,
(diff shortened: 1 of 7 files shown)


ljharb:
Anyone care to review this? :-)

test/Utils-spec.jsx
```diff
@@ -229,26 +217,237 @@ describe('Utils', () => {
...
+          <div />
+        </div>,
+      )).to.equal(false);
+    });
+
+    it('should test deepEquality with object props', () => {
+      expect(nodeMatches(
+        <div foo={{ a: 1, b: 2 }} />,
+        <div foo={{ a: 1, b: 2 }} />,
+      )).to.equal(true);
```
aweary: Does it test equality for arbitrarily nested properties? 

```jsx
 <div foo={{ a: 1, b: 2, c: { d: 3 } }} />
```

test/Utils-spec.jsx
```diff
@@ -229,26 +217,237 @@ describe('Utils', () => {
...
+          <div className="foo">other child</div>,
+        )).to.equal(false);
+
+        expect(nodeMatches(
+          <div className="foo">child</div>,
+          <div className="foo">other child</div>,
+        )).to.equal(false);
+      });
+
+      it('should match nodes when children are different but falsy', () => {
```
aweary: Can we add tests to verify both falsy children and props are correctly matched?

```jsx
// should not match
<div foo="" />
<div foo={0} />

// should not match
<div>""</div>
<div>{0}</div>

test/Utils-spec.jsx
ljharb: I copied the cases directly from the nodeEqual ones - I'm happy to add nested cases to both but I don't think that needs to go in this PR necessarily.

test/Utils-spec.jsx
ljharb: Sure, will do

test/Utils-spec.jsx
aweary: Cool, doing it in a follow-up PR sounds good to me.

test/Utils-spec.jsx
ljharb: updated!

Issue #678, opened 2016-11-12
enable `lifecycleExperimental` by default?


ljharb:
https://github.com/airbnb/enzyme/issues/34#issuecomment-260141576 reminded me that perhaps for v3, we should enable this by default.

Any objections? @Aweary @blainekasten @nfcampos @lelandrichardson

Aweary:
Haha, I was just about to open this same issue. We haven't heard any reports of issues with it, though it's hard to say if that's due to the implementation's correctness or lack of awareness.

Either way, I think it's a good idea to enable it by default with the next major release.

nfcampos:
I agree we should enable it by default in the next major. (do we keep the flag?)
(I've added a couple more test cases for this that I felt were missing in https://github.com/airbnb/enzyme/pull/681)

laumair:
:+1: I've been using it extensively for over a month now I presume and it works great. I can maybe provide some examples as well as I've used it in many places. 
However, even if used with `mount` it just work. I think it should only work with `shallow` and should atleast display some sort of a warning when used with `mount` @ljharb @Aweary

nfcampos:
@laumair thanks for telling us your experience with it. glad to see it's been useful! 
I agree that we should validate the options object and throw or warn when unknown keys are passed in.

laumair:
@nfcampos Thank YOU for this awesome library. Yeah validation would be much helpful if added.

ljharb:
@nfcampos i'd say we maybe rename the flag to `deprecatedDisableLifecycleMethods` or something, and keep it, but remove that in the next major.

jwbay:
I've also used this for months and seen no problems. I can work on a PR to make this the default.

@ljharb From your last comment, it sounds like you want to add a flag to *suppress* the lifecycle methods? A rename with the same functionality seems like it would be `deprecatedEnableLifecycleMethods`?

ljharb:
Yes, I think that would be the safe approach.

In fact, I'd prefer one semver-minor PR that adds "deprecatedDisableLifecycleMethods" (that right now is a no-op) and to publish that, and then later, a semver-major PR that changes the default behavior to enable the lifecycle methods and removes `lifecycleExperimental`, but respects the explicit disable flag as well.

jwbay:
Opened #7912 with some questions.

ljharb:
Done in #1140.

PR #789, opened 2017-01-28
add disableLifecycleMethods for shallow

jwbay:
As part of the plan to switch on all lifecycle methods for `shallow` in the next major, introduce a no-op flag ahead of time that turns that off.

Ref: https://github.com/airbnb/enzyme/issues/678#issuecomment-274949209

~~Since there's no options validation, I'm not sure any code changes are necessary.~~ (added) It may be confusing if someone sets this flag to explicitly false and expects their lifecycle methods to be called. Should we set lifecycleExperimental to true in that case?

I didn't prefix the option with `deprecated` because it seemed this flag is going to be sticking around as the official, inverted version of lifecycleExperimental. I may have misunderstood something though.

The name feels a bit off since it's not disabling *all* the lifecycle methods, just two of them. Should it be something like `disableFullLifecycle`?

### DIFF
--- a/src/ShallowWrapper.js
+++ b/src/ShallowWrapper.js
@@ -63,12 +63,24 @@ function filterWhereUnwrapped(wrapper, predicate) {
   return wrapper.wrap(compact(wrapper.getNodes().filter(predicate)));
 }
 
+function validateOptions(options) {
+  if (
+    options.lifecycleExperimental != null &&
+    options.lifecycleExperimental === options.disableLifecycleMethods
+  ) {
+    throw new Error(
+      'lifecycleExperimental and disableLifecycleMethods cannot be set to the same value',
+    );
+  }
+}
+
 /**
  * @class ShallowWrapper
  */
 class ShallowWrapper {
 
   constructor(nodes, root, options = {}) {
+    validateOptions(options);
     if (!root) {
       this.root = this;
       this.unrendered = nodes;
--- a/src/ShallowWrapper.js
+++ b/src/ShallowWrapper.js
@@ -63,10 +63,34 @@ function filterWhereUnwrapped(wrapper, predicate) {
   return wrapper.wrap(compact(wrapper.getNodes().filter(predicate)));
 }
 
+/**
+ * Ensure options passed to ShallowWrapper are valid. Throws otherwise.
+ * @param {Object} options
+ */
 function validateOptions(options) {
+  const { lifecycleExperimental, disableLifecycleMethods } = options;
   if (
-    options.lifecycleExperimental != null &&
-    options.lifecycleExperimental === options.disableLifecycleMethods
+    typeof lifecycleExperimental !== 'undefined' &&
+    lifecycleExperimental !== !!lifecycleExperimental
+  ) {
+    throw new Error(
+      'lifecycleExperimental must be either true or false if provided',
+    );
+  }
+
+  if (
+    typeof disableLifecycleMethods !== 'undefined' &&
+    disableLifecycleMethods !== !!disableLifecycleMethods
+  ) {
+    throw new Error(
+      'disableLifecycleMethods must be either true or false if provided',
+    );
+  }
+
+  if (
+    lifecycleExperimental != null &&
+    disableLifecycleMethods != null &&
+    lifecycleExperimental === disableLifecycleMethods
   ) {
     throw new Error(
       'lifecycleExperimental and disableLifecycleMethods cannot be set to the same value',
--- a/src/ShallowWrapper.js
+++ b/src/ShallowWrapper.js
@@ -71,7 +71,7 @@ function validateOptions(options) {
   const { lifecycleExperimental, disableLifecycleMethods } = options;
   if (
     typeof lifecycleExperimental !== 'undefined' &&
-    lifecycleExperimental !== !!lifecycleExperimental
+    typeof lifecycleExperimental !== 'boolean'
   ) {
     throw new Error(
       'lifecycleExperimental must be either true or false if provided',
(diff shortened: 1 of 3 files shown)


jwbay:
Added some validation.

src/ShallowWrapper.js
```diff
@@ -63,12 +63,24 @@ function filterWhereUnwrapped(wrapper, predicate) {
   return wrapper.wrap(compact(wrapper.getNodes().filter(predicate)));
 }
 
+function validateOptions(options) {
+  if (
+    options.lifecycleExperimental != null &&
+    options.lifecycleExperimental === options.disableLifecycleMethods
```
ljharb: i think this comparison should probably just `‼` both options, and throw if they're the same, as long as *both* of them aren't `== null`?

test/ShallowWrapper-spec.jsx
```diff
@@ -2684,6 +2684,18 @@ describe('shallow', () => {
...
+    it('throws when used with disableLifecycleMethods in invalid combinations', () => {
+      expect(() => shallow(<div />, {
+        lifecycleExperimental: true,
+        disableLifecycleMethods: true,
+      })).to.throw(/same value/);
+
+      expect(() => shallow(<div />, {
+        lifecycleExperimental: false,
+        disableLifecycleMethods: false,
+      })).to.throw(/same value/);
```
ljharb: let's also add tests where both are omitted, and where either aren't a boolean (ie, it should throw if a non-boolean non-undefined value is provided)

test/ShallowWrapper-spec.jsx
```diff
@@ -2683,7 +2683,131 @@ describe('shallow', () => {
...
+        ]);
+      });
+
+      it('calls expected methods when receiving new props', () => {
+        wrapper.setProps({ foo: 'foo' });
+        expect(spy.args).to.deep.equal([
+          ['componentWillReceiveProps'],
+          ['shouldComponentUpdate'],
+          ['componentWillUpdate'],
+          ['render'],
```
jwbay: I'm not sure whether `componentDidUpdate` is supposed to be called here, but it's not currently. If this behavior is correct it's going to make this flag + documentation pretty confusing since, per next test down, setState *does* trigger `componentDidUpdate` for shallow rendered components (even without enzyme in the picture).

The flag is looking more like it should be disableDidMountAllTheTimeAndDidUpdateSomeOfTheTime. 😅 

cc @koba04 any thoughts on correctness here? Note this is all without lifecycleExperimental set to true.

src/ShallowWrapper.js
jwbay: Makes sense, done

test/ShallowWrapper-spec.jsx
jwbay: Done

test/ShallowWrapper-spec.jsx
koba04: @jwbay 
Currently, `lifecycleExperimental` does complement lifecycle methods that ShallowRenderer is missing, which are `componentDidMount` and `componentDidUpdate` in `setProps` and `setContext` https://github.com/airbnb/enzyme/pull/318#issuecomment-225148673.

In `setState`, `componentDidUpdate` is always called because ShallowWrapper calls `instance.setState()` directly. 😅 

To be clear, if `disableLifecycleMethods` set to true, `componentDidMount` and `componentDidUpdate`(in `setProps` and `setContext`) are not called on the component.

src/ShallowWrapper.js
```diff
@@ -64,11 +64,47 @@ function filterWhereUnwrapped(wrapper, predicate) {
...
 
 /**
+ * Ensure options passed to ShallowWrapper are valid. Throws otherwise.
+ * @param {Object} options
+ */
+function validateOptions(options) {
+  const { lifecycleExperimental, disableLifecycleMethods } = options;
+  if (
+    typeof lifecycleExperimental !== 'undefined' &&
+    lifecycleExperimental !== !!lifecycleExperimental
```
koba04: `typeof lifecycleExperimental !== 'boolean'` is more explicitly?

test/ShallowWrapper-spec.jsx
```diff
@@ -2777,13 +2777,27 @@ describe('shallow', () => {
...
+          expect(spy.args).to.deep.equal([
+            ['componentWillReceiveProps'],
+            ['shouldComponentUpdate'],
+            ['componentWillUpdate'],
+            ['render'],
+          ]);
+        });
+      });
+
+      describeIf(REACT014, 'setContext', () => {
```
ljharb: I'm glad we're testing this on all React versions - could you elaborate on why `componentWillReceiveProps` fires in 13 and 15, but not 14?

jwbay: In React 13 and 14, the check for whether willReceiveProps is called is just a strict equality check in ReactCompositeComponent between the previous and next elements. In React 15, they added an additional equality check between previous and next context.

willReceiveProps is called in 13 because Enzyme clones the element on each render in createRendererCompatible, failing the equality check.

It's not called in 14 because the nodes are equal.

It's called in 15 because of the new check for context.

ljharb:
@aweary because then people who depend on the `false` behavior have a way to explicitly preserve it when they upgrade (which will flip the default, making `true` the noop instead of `false`), without having to change their code.

aweary:
It just sounds unlikely that anyone is going enable a no-op flag from a minor release when their tests are all working as they would expect. What's more likely is that they upgrade to the next major release, have breakage, and then enable the flag. 

But either way, it's not like this change will hurt anyone so 👍

ljharb:
I envision "upgrade, breakage, downgrade, gradually migrate their tests to use an explicit flag, upgrade smoothly later" :-)

blainekasten:
Sooo. Sounds like this is good to merge then? @jwbay mind rebasing?

jwbay:
@blainekasten rebased

blainekasten:
cc @ljharb do you want these all rebased down to 1 commit or is this fine?

ljharb:
This is fine, as long as there's no merge commits :-)

blainekasten:
Thanks for the work on this @jwbay ! This is big!

PR #791, opened 2017-02-01
Use webpack IgnorePlugin instead of externals

remcohaszing:
The documentation describes that optional dependencies should be marked as externals in webpack. This means that they will still show up in the source map.

The new setup simply ignores them, which is more correct. This was already reflected by Enzyme's own karma webpack configuration.

The React 0.14 ignore list was also written as a webpack config for consistency.

### DIFF
--- a/docs/guides/webpack.md
+++ b/docs/guides/webpack.md
@@ -10,8 +10,7 @@ compatible with React 0.13 and React 0.14.
 Unfortunately, these conditional requires mean there is a bit of extra setup with bundlers like
 webpack.
 
-In your webpack configuration, you simply need to make sure that the following files are
-labeled as "external", which means they will be ignored:
+In your webpack configuration, you simply need to make sure that the following files will be ignored:
 
 ```
 cheerio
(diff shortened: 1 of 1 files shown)


aweary:
Thanks for the PR! @remcohaszing have you verified that this setup works with Webpack 1/2 in all our supported versions of React?

remcohaszing:
I didn't test it with older versions of React, but I don't see why it wouldn't work.

The old docs stated: 

> you simply need to make sure that the following files are labeled as "external", which means they will be ignored:

By labeling them as "external" they're not really ignored, just undefined. Now they truly are ignored.

docs/guides/webpack.md
```diff
@@ -20,44 +19,34 @@ react/lib/ReactContext
...
-```
-cheerio
-react-dom
-react-dom/server
-react-addons-test-utils
+```js
+/* webpack.config.js */
+// ...
+plugins: [
+  new webpack.IgnorePlugin(/cheerio/),
```
nfcampos: why do we need to ignore `cheerio` and `react-dom/server`? our own `karma.conf.js` seems to work ok without ignoring these

docs/guides/webpack.md
```diff
@@ -20,44 +19,34 @@ react/lib/ReactContext
...
+  new webpack.IgnorePlugin(/react-dom/),
```
nfcampos: I believe this list of ignore plugins is actually for react 13 not 14, they're swapped for some reason?

remcohaszing: I just converted the list to a list of `webpack.IgnorePlugin` instances. If the list is wrong, I'd say that's unrelated to this patch. I've only used react 0.15, so I don't know what should be ignored exactly.

ljharb:

PR #807, opened 2017-02-13
Fix a bug that doesn't update props with lifecycleExperimental: true and sCU: false

### LINKED ISSUE #805: Props not updated when shouldComponentUpdate returns false (lifecycleExperimental: true)
When shallow rendering a component using lifecycleExperimental: true, it seems that it's props are not updated if it's shouldComponentUpdate()-method returns false. Since the props are updated in production and when using standard shallow rendering, I'm guessing that this would be the desired behavior.

```
import React from 'react';
import chai from 'chai';

class MyDummyComponent extends React.Component {

    shouldComponentUpdate() {
        return false;
    }

    render() {
        return (<div>hello world.</div>)
    }
}

describe('<MyDummyComponent/>', function () {

    it('fails to update props with lifecycleExperimental: true', function () {
        let props = {
            foo: 'bar'
        }

        const wrapper = shallow(<MyDummyComponent {...props} />, { lifecycleExperimental: true });

        chai.expect(wrapper.instance().props.foo).to.equal('bar');

        props.foo = 'baz';

        wrapper.setProps(props);

        // fails when shouldComponentUpdate returns false, succeeds otherwise
        chai.expect(wrapper.instance().props.foo).to.equal('baz');
    });

    it('updates props with "normal" shallow rendering', function () {
        let props = {
            foo: 'bar'
        }

        const wrapper = shallow(<MyDummyComponent {...props} />);

        chai.expect(wrapper.instance().props.foo).to.equal('bar');

        props.foo = 'baz';

        wrapper.setProps(props);

        chai.expect(wrapper.instance().props.foo).to.equal('baz');
    });
});
```

I am using React 15.4.2 & Enzyme 2.7.1.

koba04:
This PR is to fix #805.
When `lifecycleExperimental` is `true` and `shouldComponentUpdate` returns `false`, ShallowWrapper should update instances's props with passed props.

I've also added some unit tests for to assert updating `props` and `state` and `context` whether `shouldComponentUpdate` returns `true` or not.

### DIFF
--- a/src/ShallowWrapper.js
+++ b/src/ShallowWrapper.js
@@ -230,6 +230,9 @@ class ShallowWrapper {
               instance.componentDidUpdate(prevProps, state, prevContext);
             }
             this.update();
+          // If it doesn't need to rerender, update only its props.
+          } else if (props) {
+            instance.props = props;
           }
           if (originalComponentWillReceiveProps) {
             instance.componentWillReceiveProps = originalComponentWillReceiveProps;
--- a/test/ShallowWrapper-spec.jsx
+++ b/test/ShallowWrapper-spec.jsx
@@ -2862,8 +2862,11 @@ describe('shallow', () => {
         }
 
         const wrapper = shallow(<Foo foo="bar" />, { lifecycleExperimental: true });
+        expect(wrapper.instance().props.foo).to.equal('bar');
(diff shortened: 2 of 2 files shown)


koba04:
Thanks!

PR #810, opened 2017-02-15
Reorganizes project with Lerna

jondubin:
### What?

First stab at reorganizing the project to use Lerna to manage multiple packages in the same repo. Most of the files were moved into `/packages/enzyme`. The ones that weren't are generally applicable to the project. Happy to reorganize if there are misplaced files.

Now the root directory looks like this:
- CONTRIBUTING.md
- INTHEWILD.md
- lerna.json
- LICENSE.md
- MAINTAINERS
- package.json
- packages/
- README.md
- .eslintrc
- .git
- .gitignore
- .lgtm
- .npmignore

to: @lelandrichardson


jondubin:
@ljharb My understanding is that setting lerna up requires this sort of restructuring. Is it possible to set a lerna project up and still keep enzyme at the root?

ljharb:
I don't have any interest in using lerna on enzyme, so if that's a requirement to use lerna, then let's not.

jondubin:
@ljharb Ahh, ok I understand the concern. I think the plan for the rewrite was to use Lerna to support the repo housing "enzyme-core" and the react 13/14/15 adapters. 

I did just discover that there is a `lerna import` command that will preserve the commit history. Closing this PR. Thanks for flagging this.

ljharb:
Can you point me to that plan? I'm not aware that we'd agreed to use a monorepo for any part of enzyme - I'd want to see the react adapters in separate repos.

jondubin:
@lelandrichardson Can probably provide better context

lelandrichardson:
@ljharb I think using lerna to handle enzyme + adapters is a good and sensible use case for lerna. What's the problem with doing this?

lelandrichardson:
For additional context, jonathan is going to help out refactoring enzyme following the new adapter plan. This PR (and all following PRs until all tests pass) would happen in the enzyme-harmony branch

lelandrichardson:
Also all of the files are being followed by git so the git blame should be followed in this case

ljharb:
@lelandrichardson it's the same problem we've talked about before; having the adapters in a primary repo "blesses" them, and creates a situation where people who make other adapters will want us to host them.

Let's chat about it in person a bit.

PR #841, opened 2017-03-07
[Docs] Update AVA example for better use of API

alexrussell:
AVA has `t.true` and `t.false` for `true`/`false` checking, so you don't need `t.is(..., true)`.

I couldn't see the same in tape's docs (other than `t.ok` and `t.notOk`, but they test for truthy/falsy, not `true` and `false` specifically), so I haven't updated the tape example as I wasn't sure that'd be precise enough for you.

### DIFF
--- a/docs/guides/tape-ava.md
+++ b/docs/guides/tape-ava.md
@@ -40,13 +40,13 @@ import Foo from '../path/to/foo'
 
 test('shallow', t => {
   const wrapper = shallow(<Foo />)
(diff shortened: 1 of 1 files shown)


ljharb:
`.true` and `.false` are syntax errors in ES3, which is still an environment people may want to run tests in.

alexrussell:
Ahh didn't think of that (though I'd be interested to know who's running AVA tests using enzyme in ES3).

PR #855, opened 2017-03-23
Fix an issue w/ cleaning up global.document.

gdborton:
Except in the most recent versions of node, developers are unable to delete
things from global when running code in a node vm.  This means that enzyme is
failing to clean up after itself when temporarily defining global.document for
setState calls.

This papers over the issue by explicitly setting global.document to undefined
if it failed to be deleted.

Relevant jest issue - https://github.com/facebook/jest/issues/3152

Relevant node pr - https://github.com/nodejs/node/pull/11266

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -221,6 +221,9 @@ export function withSetStateAllowed(fn) {
   }
   fn();
   if (cleanup) {
+    // This works around a bug in node/jest in that developers aren't able to
+    // delete things from global when running in a node vm.
+    global.document = undefined;
     delete global.document;
   }
 }


src/Utils.js
```diff
@@ -222,6 +222,11 @@ export function withSetStateAllowed(fn) {
   fn();
   if (cleanup) {
     delete global.document;
+    // This works around a bug in node/jest in that developers aren't able to
```
ljharb: can this be done *before* deleting it, so that it's always set to undefined, and then deleted when possible?

Issue #858, opened 2017-03-27
CDN version?


trusktr:
It would be super nice to have an AMD or UMD release of this. I, for example, have a "legacy" RequireJS environment that can't load CommonJS-style node_modules.

ljharb:
This is basically a duplicate of #806 and #264 - the gist from my comments both is that you should `npm install` `enzyme`, and then use your own browserify transform or webpack plugin to convert commonJS to requireJS - and if you don't have that as part of your build process, do that locally and commit the result.

The vast majority of JS libs are (or have become) commonJS-only, and this trend will only continue at a faster pace - we wouldn't be doing your legacy environment any favors by enabling any further delaying of the ability to consume CJS deps.

trusktr:
> then use your own browserify transform or webpack plugin to convert commonJS to requireJS

That's [much more difficult](https://github.com/airbnb/enzyme/issues/47) that it should be. Can you please fix that?

ljharb:
Yes, that should be fixed - at the least, by documentation. However, a UMD build isn't the proper fix.

I'm going to close this issue, and we can continue discussion on #47.

Issue #806, opened 2017-02-10
browser build of enzyme


sylvain-hamel:
Hi, I'm looking for a browser ready build of enzyme. I found [this article](https://gist.github.com/jonmbake/6362a9f54f57a4506057166aa5ec7027) to make one myself but I'd rather use a build that is updated regularly. Is there such a thing?

ljharb:
There isn't any official browser build, no.

sylvain-hamel:
Would you consider tagging this issue as a feature request? This would make fiddling with the react/enzyme in a plain project (no build process) much easier and faster with a simple link to unpkg.com.

ljharb:
I don't mind including a browser build, provided that it can only be consumed by `browserify`. However, I wouldn't want one that served AMD, UMD, or created browser globals.

finnigantime:
We had to browserify enzyme so we could use it with our other AMD modules to run our tests in the browser. We had to add a script to do some post-processing on the browserified file, to avoid collisions with the global require, to remove some problematic polyfills that came from pulling in sub-libraries, and to inject the license into the src file.

@ljharb why do you not want one that serves AMD or UMD?

ljharb:
Those who need AMD can use a browserify transform to produce it, but I don't want to encourage an obsolete module format with the default build.

sylvain-hamel:
I too need an AMD module. The extra steps to make it work (as @finnigantime described) are not trivial. Adding the AMD build to the package would make enzime accessible to a broader audience. 

@finnigantime, until then, would you mind sharing your build scripts here? They could then be used to document a "How to build an AMD version of enzime?" help page.

ljharb:
@sylvain-hamel making things accessible to people *still* using AMD in 2017 isn't doing them any favors; if you insist on remaining an AMD user, you *must* learn how to adapt all the CJS modules out there to your build system; if you don't know how to do that, you should hit a wall sooner rather than later.

sylvain-hamel:
Hi I ended up building it myself with browserify. 

from my project folder:

- `cd node_modules\enzyme`
- `npm install`
- `npm run build`
- `cd build`
- `browserify index.js -o enzyme.2.7.1.js -x react/addons -x react/lib/ReactContext -x react/lib/ExecutionEnvironment -s enzyme -d`
- move `enzyme.2.7.1.js` where I need it and use it

You can close this issue.

Thanks

sylvain-hamel:
[...]

Issue #264, opened 2016-03-16
Using enzyme with react-rails (asset pipeline)


ryanwood:
This is more of an inquiry. We have a fairly large Rails application and are transitioning much of the front end work to React. This means that we are using the Rails asset pipeline rather than a module system. :( 

I really like the API of enzyme and would love to find a way to use it within my front end testing suite. I'm using [Teaspoon](https://github.com/modeset/teaspoon) with Jasmine underneath. I'm wondering if there is any way to utilize enzyme in a non-module (globals) environment without access to `require`?

Any help is appreciated. Thanks.

lelandrichardson:
@ryanwood we had discussed having a separate UMD build that would end up in something like `/dist/enzyme.js` and `/dist/enzyme.min.js`, but never got around to it.  If you would like to PR something that does this, we will likely accept it.

ljharb:
I would prefer to not include a UMD at all - it's legacy baggage that's not needed.

@ryanwood - can you not run enzyme through `browserify --standalone` and stick that in a vendor directory?

Most npm modules don't have a UMD build, so you'll need to integrate some kind of module system into your Rails app.

lelandrichardson:
@ljharb didn't we have a whole conversation about this last time? I think if we provide a UMD build (one that's not normally required), that's the best of both worlds.

lelandrichardson:
for reference: https://github.com/airbnb/enzyme/pull/90

ljharb:
You're right, we did, I'd forgotten.

As we discussed there, it certainly can be included, but I still don't think it adds any value. Enabling people to continue using legacy JS module systems (or none at all) isn't doing anyone any favors.

BernardDoherty:
Hi all. I am having the same issue as Ryan. We have to continue to use the Rails asset pipeline as we transition parts of our app to a React. Any chance of adding the /dist file to the project or provide some info on how to use browserify --standalone to create stand alone file? Thanks!

jeffmicklos:
Same boat here, React + Rails. Would love a UMD package with a global exposed!

What @ljharb said kind of bums me out:

> Enabling people to continue using legacy JS module systems (or none at all) isn't doing anyone any favors.

By definition, it is doing a favor to those people using `legacy JS module systems`; this thread proves that is absolutely a segment of the world. ¯\_(ツ)_/¯

jellymann:
[...]

PR #861, opened 2017-03-27
Add browserify and build-global script to make UMD version

trusktr:


ljharb:
See #858, #806, and #264 for why this is not desired.

trusktr:
> Everything should only be available as node-style CJS

I strongly disagree. There's plenty of people out there that use plain JavaScript (script tags) with react and no JSX. It's unfair to treat everyone as skilled in module bundler tools.

I strongly believe library authors can take extra care to make sure their library is as consumable as possible so other people can simply plop the library into their project and have it simply work.

As you see in #47 (which also applies to browserify), it's not as simple as just "go build this with webpack or browserify" because... it fails in ways that most people probably don't even know how to deal with when using webpack and browserify in standard ways.

ljharb:
> I strongly believe library authors can take extra care to make sure their library is as consumable as possible so other people can simply plop the library into their project and have it simply work.
I totally agree with this sentiment.

However, enzyme is not a production library - it's a testing library. People still using script tags are unlikely to even be writing tests at all, let alone trying to run them *in the browser*.

It's 2017. Everyone needs to *become* skilled in module bundler tools - it's how modern web dev is done, full stop. It's not doing anyone any favors to make it easier for them to remain in the stone ages.

trusktr:
Imagine a brand new programmer learning React, using globals, okay not using JSX for now, then wants to use this awesome "enzyme" test tool to test out a component. That user will be stuck without being able to experience the awesome, unless they now stop what they're doing and go learn build tools (which isn't a bad thing, but maybe not what they wanted to do at that moment, and maybe not what they are capable of yet, you'd be surprised how many people start programming and making awesome things without even knowing how the filesystem or command line works).

trusktr:
>  it's how modern web dev is done, full stop

In companies, yes. Someone learning JavaScript with the plethora of online web-based in-browser editors? Not necessarily.

That's besides the point. I am familiar with build tool. However, I really do enjoy being able to open a website like jsfiddle, plopping a `<script>` up in that thing, and getting to coding small examples without worrying about a build tool.

Suppose I'd like to test out a test library in jsfiddle, for example. It's relatively easy for a library author to ship a global. The time that it takes to do that gets distributed to one person instead of many many users.

trusktr:
Suppose I'd like to show my friend, "hey, look how awesome this test tool enzyme is, here open this fiddle", and I was able to do it because of a simple global script tag.

ljharb:
Also a fair point, and having a nice enzyme template jsfiddle would be super helpful.

Issue #823, opened 2017-02-23
Simulate events returning promises?


jdgreenberger:
Hi there - really enjoying the library! I have a question about testing `simulate` events. Many of the events in my code trigger functions that wait for promises.
```javascript
onClick = async (e) => {
   data = await fetchData(e.target.value)
   this.setState({data})
}
```
In the live react, events clearly don't return promises, but I'm not sure if enzyme's `simulate` is also event-driven. Is there a way that I can wait for a simulate event to return a promise? I know there is an alternative, but it's a little bit more overhead.
```javascript
someFunc = async () => {
  data = await fetchData(e.target.value)
   this.setState({data})
}
onClick = (e) => {
   someFunc()
}
// test onClick calls someFunc()
// test someFunc()
```

ljharb:
No, there's no way to await the outcome of an event listener. Separately, `this.setState` is async and takes a callback, so for your promise to work properly it'd need to pass a callback and resolve the wrapping `new Promise` when setState is done.

ljharb:
What I'd recommend is having your event listener call into an instance method; spy on the instance method to test the event listener, and then unit-test the instance method.

yutin1987:
https://github.com/airbnb/enzyme/blob/master/src/ShallowWrapper.js#L620

this code should return a Promise, and then testing
`await component.find('TouchableOpacity').simulate('press');`

just a idea 😄

ljharb:
Events are not appropriate to model as Promises. Events happen multiple times; a promise resolves once.

Also, events are synchronous in the browser.

jwbay:
@jdgreenberger If you've got your fetchData mocked such that it returns an immediately resolved promise, this test approach might work for you: https://github.com/airbnb/enzyme/blob/7cd2e9a8d03d32d3518918c446906c82c68be052/test/ShallowWrapper-spec.jsx#L4181-L4188

It would rely on the fact that resolved Promise callbacks fire before the setImmediate callback. This is currently a safe assumption in v8, but less so in other JS implementations.

yutin1987:
I have a case.

```

**Comment by yutin1987** [comment]
I have a case
```
  onPress = async () => {
    const { client, onUpload } = this.props;

    const body = await Cropper.getPhotoFromAlbum();
    const reply = await client.fetch('/upload', {
      method: 'POST',
      body: JSON.stringify({ body }),
    });

    onUpload(reply.key, `data:image/jpeg;base64,${body}`);
  }
```

testing
[...]

PR #867, opened 2017-04-01
feat(simulate): support promise event on simulate

yutin1987:
#823 simulate events returning promises

### DIFF
--- a/src/ShallowWrapper.js
+++ b/src/ShallowWrapper.js
@@ -606,18 +606,20 @@ class ShallowWrapper {
    * @returns {ShallowWrapper}
    */
   simulate(event, ...args) {
+    const promise = Promise.resolve();
     const handler = this.prop(propFromEvent(event));
     if (handler) {
       withSetStateAllowed(() => {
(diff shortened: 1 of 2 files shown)


nfcampos:
I don't think I agree with this. Always returning a promise from `simulate` seems to suggest the event system is somehow asynchronous (which it itsn't) and users of the library should always wait for the promise to resolve before proceeding with their tests.

For the use case you exemplify in the test you added, why not do this?
```js
    it('should simulate promise events', (done) => {
        ...
      }

      expect(wrapper.find('.clicks-0').length).to.equal(1);
      wrapper.prop('onClick')().then(() => {
        expect(wrapper.find('.clicks-1').length).to.equal(1);
        done();
      });
    });
```

ie. select the value of the `onClick` prop and call it directly

ljharb:
Events are not appropriate to model as Promises. Events happen multiple times; a promise resolves once.

Also, yes, events are synchronous in the browser.

You can wrap yourself with `new Promise` to avoid the `done` pattern.

PR #869, opened 2017-04-03
Update everyWhere.md

halfninja:
_All_ of the nodes need to match in `everyWhere`. May have been copied from `someWhere` and not fully updated.


halfninja:
That should be done now

PR #871, opened 2017-04-04
Update shallow docs with lifecycleExperimental

serut:
As far as I know, the `options.lifecycleExperimental` is not documented yet.  
I just found this pull request https://github.com/airbnb/enzyme/pull/318 and that's the only place where we see that option

### DIFF
--- a/docs/api/shallow.md
+++ b/docs/api/shallow.md
@@ -46,6 +46,7 @@ describe('<MyComponent />', () => {
   - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, `componentDidMount`
 is not called on the component, and `componentDidUpdate` is not called after
 [`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md).
+  - `options.lifecycleExperimental`: (`Boolean` [optional]): If set to true, the entire lifecycle (`componentDidMount` and `componentDidUpdate`) of the React component is called
 
 #### Returns
 
--- a/docs/api/shallow.md
+++ b/docs/api/shallow.md
@@ -43,10 +43,12 @@ describe('<MyComponent />', () => {
 1. `node` (`ReactElement`): The node to render
 2. `options` (`Object` [optional]):
   - `options.context`: (`Object` [optional]): Context to be passed into the component
-  - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, `componentDidMount`
-is not called on the component, and `componentDidUpdate` is not called after
-[`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md).
-  - `options.lifecycleExperimental`: (`Boolean` [optional]): If set to true, the entire lifecycle (`componentDidMount` and `componentDidUpdate`) of the React component is called. The current default value is `false` with enzyme v2, but the next major version will flip the default value to `true`
+  - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, most of the React
+lifecycle methods will never be executed: `componentDidMount`, `componentWillReceiveProps`,
+`shouldComponentUpdate`, `componentWillUpdate` and `componentDidUpdate`. Default to `false`.
+  - `options.lifecycleExperimental`: (`Boolean` [optional]): If set to true, the entire lifecycle
+(`componentDidMount` and `componentDidUpdate`) of the React component is called. The current default value
+is `false` with enzyme v2, but the next major version will flip the default value to `true`.
 
 #### Returns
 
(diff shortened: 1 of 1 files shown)


docs/api/shallow.md
```diff
@@ -50,6 +50,7 @@ describe('<MyComponent />', () => {
   - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, `componentDidMount`
 is not called on the component, and `componentDidUpdate` is not called after
 [`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md).
+  - `options.lifecycleExperimental`: (`Boolean` [optional]): If set to true, the entire lifecycle (`componentDidMount` and `componentDidUpdate`) of the React component is called
```
ljharb: Let's also note that the current default is `false`; the next major version will flip the default to `true`; and setting an explicit `true` or `false` here will mean the default change will not effect you.

serut: Yes you're right !

serut:
Do you know why the generated documentation http://airbnb.io/enzyme/docs/api/shallow.html#arguments does not contain the `options.disableLifecycleMethods` property ?

audiolion:
👍 for this, I had to find this PR to figure out how to do this

ljharb:
@serut the official docs should be updated with every npm release.

serut:
@ljharb The official doc is ok right now, but last week it was not updated ..!

ljharb:
@serut because last week, there hadn't been a release in awhile.

nfcampos:
I think this is confusing, it's not that clear that using both `lifecycleExperimental` and `disableLifecycleMethods` doesn't make sense, ie. it's not clear these two options are the opposite of each other. Could we try to change the wording to make it clearer that
1. you should use `{disableLifecycleMethods: true}` to ensure your test will not use lifecycle methods once you update to the next major of `enzyme`
2. you should use `{lifecycleExperimental: true}` to have your test call lifecycle methods in the current release of of `enzyme` (which will be the default in the next major)

serut:
Can I get some feedback about this ?

docs/api/shallow.md
```diff
@@ -47,9 +47,12 @@ describe('<MyComponent />', () => {
...
 2. `options` (`Object` [optional]):
   - `options.context`: (`Object` [optional]): Context to be passed into the component
-  - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, `componentDidMount`
-is not called on the component, and `componentDidUpdate` is not called after
-[`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md).
+  - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, most of the React
+lifecycle methods will never be executed: `componentDidMount`, `componentWillReceiveProps`,
+`shouldComponentUpdate`, `componentWillUpdate` and `componentDidUpdate`. Default to `false`.
+  - `options.lifecycleExperimental`: (`Boolean` [optional]): If set to true, the entire lifecycle
+(`componentDidMount` and `componentDidUpdate`) of the React component is called. The current default value
```
nfcampos: shouldn't both options call out a similar list of lifecycle events?

nfcampos:
@serut thanks for the reminder, sorry about the wait. left you a comment, other than LGTM

nfcampos:
(and we'll need a rebase before merging as well)

docs/api/shallow.md
serut: Not really. If you do not specify any boolean setContext, setState and setProps will trigger  part of React lifecycle methods. If you disableLifecycleMethods nothing is executed whereas if you use lifecycleExperimental all the react lifecycle can be called.

serut:
I still think this pr can be merged. 

Should I provide the code that can prove what is called on the React component lifecycle ?

ljharb:
@serut can you rebase this? if so, i can merge ASAP

docs/api/shallow.md
nfcampos: @serut That's incorrect. Please refer to the test cases for `disableLifecycleMethods` https://github.com/airbnb/enzyme/blob/master/test/ShallowWrapper-spec.jsx#L2722
As tested there, if `disableLifecycleMethods` is `true` most lifecycle methods are called, `componentDidUpdate` and `componentDidMount` are the only two that are not called.

@ljharb this PR cannot be merged until the docs specify the correct list of lifecycle methods affected by either option.

ljharb:
This rebase seems to have gone awry; do you need help fixing it?

ljharb:
(I've fixed the rebase)

serut:
Yes ! The rebase ended up with a merge with master, so I force pushed without that merge and it looks better. I think I made a commit directly on Github.com and I forgot to pull it before starting to rebase... sadness !

I read tests @nfcampos and you're totally right ! I though `disableLifecycleMethods` was more impactfull.

Anyway, when we read the `disableLifecycleMethods` and `lifecycleExperimental` documentation do you understand these are two opposite options ? I looks good to me even if it's implicit.

ljharb:
They are opposite for the time being; in the next breaking version we're planning on flipping the default so that lifecycle methods are enabled, which is why an explicit `disableLifecycleMethods` exists now.

nfcampos:
Thanks @serut !

PR #874, opened 2017-04-06
Add beforeInput to simulated event map.

seejee:
For the purposes of testing an editor based on https://github.com/facebook/draft-js, I'd like to be able to simulate the `beforeinput` event.

I didn't see tests to make sure that each event is mapped properly, so I didn't add one. If I missed it, please point me the proper place, and I'm happy to add one.

Thanks for enzyme! It's fantastic!

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -412,6 +412,7 @@ export function mapNativeEventNames(event) {
     ratechange: 'rateChange',
     timeupdate: 'timeUpdate',
     volumechange: 'volumeChange',
+    beforeinput: 'beforeInput',
   };
 
   if (!REACT013) {


ljharb:
Thanks! Can you point to where this is defined in React itself? (in other words, does React support this event natively)

aweary:
@ljharb there is [an event plugin](https://github.com/facebook/react/blob/b1b4a2fb252f26fe10d29ba60d85ff89a85ff3ec/src/renderers/dom/shared/eventPlugins/BeforeInputEventPlugin.js) that handles `beforeinput`

aweary:
All event plugins are always present, so there would be no way to do that kind of detection.

PR #876, opened 2017-04-07
[Compat] support React@15.5

### LINKED ISSUE #875: Support react-dom/test-utils
[Today's announcement](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html) moves `react-addons-test-utils` to `react-dom/test-utils`. I'm not sure I understand all the intricacies of the test utils requirements from enzyme, but I expect that something may be able to be done about the requirement for `react-addons-test-utils`. Any chance this could be revisited?

kentcdodds:
A few deprecation warnings were added in React 15.5. This supports the new APIs that React recommends.

Closes #875


README.md
```diff
@@ -60,14 +60,22 @@ npm i --save-dev enzyme
...
 npm i --save-dev react-addons-test-utils
 npm i --save-dev react-dom
 ```
 
+If you are using `React >=15.5`, in addition to `enzyme`, you will have to ensure that you also have
+the following npm modules installed if they were not already:
+
+```bash
+npm i --save-dev react-dom
+npm i --save-dev prop-types
```
ljharb: Does `prop-types` work with every version of React?

If it does, then let's just depend on it explicitly instead of making it funky. `airbnb-prop-types` will likely depend on it very soon anyways, and enzyme will be depending on that soon if it doesn't already :-)

src/react-compat.js
```diff
@@ -150,6 +157,16 @@ if (REACT013) {
   };
 }
 
+if (REACT155) {
+  // eslint-disable-next-line import/no-extraneous-dependencies
+  PropTypes = require('prop-types');
+  // eslint-disable-next-line import/no-extraneous-dependencies
+  createClass = require('create-react-class');
```
ljharb: does this package work with React versions prior to 15.5? 15.0-15.4, or 14, or 13?

README.md
kentcdodds: Sounds great to me :+1:

src/react-compat.js
kentcdodds: I think so. Maybe we should get @acdlite involved here.

ljharb: @acdlite fwiw, if both of these packages happened to work down to react 0.13, we could drop a lot of our crappy conditional requires here :-)

src/react-compat.js
```diff
@@ -95,8 +97,13 @@ if (REACT013) {
...
+    if (REACT155) {
+      // eslint-disable-next-line import/no-extraneous-dependencies
+      TestUtils = require('react-dom/test-utils');
+    } else {
+      // eslint-disable-next-line import/no-extraneous-dependencies
+      TestUtils = require('react-addons-test-utils');
+    }
   } catch (e) {
     // eslint-disable-next-line no-console
     console.error(
```
aweary: Does this need to be in the `try` block? The error logged in `catch` only mentions `react-addons-test-utils` so it might be confusing if requiring `react-dom/test-utils` throws for whatever reason.

src/react-compat.js
acdlite: `create-class` should work, but I think the signature for `prop-types` may have changed recently. I'll check.

src/react-compat.js
ljharb: The error message should be conditional just like the require is, good catch.

src/react-compat.js
ljharb: Thanks!

kentcdodds:
I realize this is kinda high priority, but I can't work on it any more. If we need this fixed asap, someone else will have to take it from here. Good luck friends!

ljharb:
I filed https://github.com/facebook/react/issues/9371 to fix the fact that `prop-types` does not properly specify its dependencies.

bvaughn:
Nice work @kentcdodds !

There's also a small issue with `ShallowWrapper` which uses `react-compat` which calls `TestUtils.createRenderer`. The shallow renderer has also moved to `react-test-renderer/shallow` in 15.5

ljharb:
**note**: If a non-collaborator wants to add to this PR, please push some commits up to your fork, and simply paste a link to them here - I'll be happy to cherry-pick them into the PR so we can all get this done together.

bvaughn:
Created a patch locally with the changes I mentioned above but notice that we seem to still have problems with the new shallow renderer import. 😦

Using `TestUtils.createRenderer` logs a deprecation warning but otherwise works. Using `react-test-renderer/shallow` causes 5 tests to fails. It's getting late here now so I'm going to call it a night. Will look more tomorrow.

package.json
```diff
@@ -23,7 +23,7 @@
     "react:clean": "rimraf node_modules/react node_modules/react-dom node_modules/react-addons-test-utils",
     "react:13": "rimraf node_modules/.bin/npm && npm run react:clean && npm i react@0.13 && npm install",
     "react:14": "rimraf node_modules/.bin/npm && npm run react:clean && npm i react@0.14 react-dom@0.14 react-addons-test-utils@0.14 && npm install",
-    "react:15": "rimraf node_modules/.bin/npm && npm run react:clean && npm i react@15 react-dom@15 react-addons-test-utils@15 && npm install",
+    "react:15": "rimraf node_modules/.bin/npm && npm run react:clean && npm i react@15 react-dom@15 react-addons-test-utils@15 prop-types@15 create-react-class@15 && npm install",
```
nfcampos: shouldn't we be removing react-addons-test-utils from this line?

src/react-compat.js
```diff
@@ -18,6 +18,8 @@ let childrenToArray;
 let renderWithOptions;
 let unmountComponentAtNode;
 let batchedUpdates;
+let PropTypes;
+let createClass;
```
nfcampos: should we start importing createClass from here in the tests that use it? otherwise we'll have plenty of warnings

ljharb: yes, all the tests should pull it from here too.

gaearon:
Shallow rendering turns out to be trickier to decouple from ReactDOM than we expected.

We didn’t know that people use `shallow()` together with `Simulate`. In React documentation they are not described together, but it seems like people already rely on this coupling and so any attempts to separate shallow rendering from ReactDOM will break that.

Perhaps we should reverse the course on that one and revisit it in 16.

gaearon:
Never mind my previous comment, I misunderstood how Enzyme works. I was reading `ReactWrapper` but should have looked at `ShallowWrapper`. Looks like the issue is fixable.

bvaughn:
A problem was discovered while testing the new shallow renderer but has since been fixed by @gaearon (facebook/react/pull/9382) and will be re-released with a point update shortly.

Consider adding [this commit](https://github.com/bvaughn/enzyme/commit/dbb029f427713dab4ba661638bad60674b880136) to pull the shallow renderer from the correct place and [this commit](https://github.com/bvaughn/enzyme/commit/366bf476fdccc5bd1ec8c9f7c149d42ffa0f2bb3) to accompany Dan's above PR change.

ljharb:
@bvaughn thanks, I've cherry-picked in those two commits!

gaearon:
Please bump `react-test-renderer` to 15.5.4 and Brian's changes should work. I'll deprecate versions 15.5.x until 15.5.4.

ljharb:
@gaearon the install command installs `15` so it should automatically grab that. I'm rerunning the build now; but it appears there's still some failures.

nfcampos:
shouldn't we start testing both 15.4 and 15.5, if not on CI at least have the command to do it locally?

README.md
```diff
@@ -60,12 +60,18 @@ npm i --save-dev enzyme
...
-npm i --save-dev react-addons-test-utils
-npm i --save-dev react-dom
+npm i --save-dev react-addons-test-utils react-dom
+```
+
+If you are using `React >=15.5`, in addition to `enzyme`, you will have to ensure that you also have
+the following npm modules installed if they were not already:
+
+```bash
+npm i --save-dev react-dom create-react-class
```
nfcampos: can't we use create-react-class for all versions, like prop-types?

bvaughn:
Script should probably be updated to test both 15.4.2 and 15.5+ yeah.

README.md
ljharb: Most likely yes https://github.com/airbnb/enzyme/pull/876#discussion_r110502646

ljharb:
Let's definitely both 15.4 and 15.5 on CI; local testing that's not enforced by CI is useless :-)

README.md
bvaughn: There's one subtle difference. If you pair create-react-class with react-dom < 15.5.0 then the callback param of `replaceState` won't work.

Otherwise I've tested back to 0.14.0 and I think it's okay.

ljharb: @bvaughn is that because that's a new feature in v15.5, or would that work with `React.createClass` prior to 15.5?

bvaughn: | react-dom | create-react-class | works? |
|---|---|---|
| 15.4.2 | N/A | ✓ |
| 15.4.2 | 15.5+ |   |
| 15.5+ | 15.5+ | ✓ |

bvaughn: PS I'm falling asleep. Will check this PR again in the morning. 👋

nfcampos:
@ljharb ive pushed a commit testing both react 15 versions

README.md
ljharb: @bvaughn thanks; if that's the case, then we shouldn't use `create-react-class` at all prior to react 15.5 :-/

Is this a bug we could file on the `create-react-class` repo and get fixed?

src/ShallowWrapper.js
```diff
@@ -142,6 +142,16 @@ class ShallowWrapper {
     this.complexSelector = new ComplexSelector(buildPredicate, findWhereUnwrapped, childrenOfNode);
   }
 
+  batchedUpdates(fn) {
+    const renderer = this.root.renderer;
+    if (renderer.unstable_batchedUpdates) {
+      // React 15.5+ exposes batching on shallow renderer itself
```
ljharb: should this use `REACT155` (or other version constants) so that it's only checked there?

src/ShallowWrapper.js
```diff
@@ -142,6 +142,16 @@ class ShallowWrapper {
     this.complexSelector = new ComplexSelector(buildPredicate, findWhereUnwrapped, childrenOfNode);
   }
 
+  batchedUpdates(fn) {
```
ljharb: instead of exposing another prototype method, what do you think about using a standalone function?

src/version.js
```diff
@@ -4,3 +4,4 @@ export const VERSION = React.version;
 export const REACT013 = VERSION.slice(0, 4) === '0.13';
 export const REACT014 = VERSION.slice(0, 4) === '0.14';
 export const REACT15 = VERSION.slice(0, 3) === '15.';
+export const REACT155 = VERSION.slice(0, 4) === '15.5';
```
ljharb: maybe `= REACT15 && VERSION.slice(3, 5) === '5.';` would be more accurate? I don't want this matching `15.50` one day ;-)

README.md
bvaughn: No, unfortunately. 15.5 required a change to continue supporting that param. It's not backwards compatible.

src/version.js
nfcampos: well wouldn't your suggestion match `16.5.0` which is more likely than `15.50`? ;)

ljharb: no, `REACT15 &&` ensures it starts with `15.`.

nfcampos: ah sorry didn't notice that. then i agree

src/ShallowWrapper.js
gaearon: And pass `root` there?

gaearon:
Moved to a function. https://github.com/gaearon/enzyme/commit/9042abd5fef083ccc500943dbee50c1754d6bae5

src/ShallowWrapper.js
gaearon: Addressed below

README.md
aweary: Could we just stop using `React.createClass` internally so that only users actually using `React.createClass` need to add it to their dependencies?

nfcampos: i'd agree, but most of our usage is in tests of createClass components, so the only option there would be to do away with those tests altogether 

(its use in  ReactWrapperComponent can be replaced with an ES6 class easily)

bvaughn: That sounds even better!

aweary: @nfcampos for the tests can't we just import conditionally `create-react-class` ourselves? If the main reason for keeping it is for our tests we should restructure it so our testing requirements aren't being leaked to consumers

koba04: > (its use in ReactWrapperComponent can be replaced with an ES6 class easily)
 
I've submitted a PR for that. #877
It doesn't depend on this PR, but can I change the target branch to here?

src/ShallowWrapper.js
ljharb: Yes, exactly - pass `root`, or `this`

README.md
nfcampos: @aweary if we want to keep running our tests from 0.13 to 15.5 with createClass components we could create something like the react-compat file just for the tests and import createClass from there. makes sense?
cc @ljharb

ljharb: That seems better - making a `test/react-compat` file that handles `createClass`, and we can import that for tests, and add `create-react-class` as a dev dependency.

nfcampos:
@ljharb the tests that are failing in this PR shouldn't exist IMO, see explanation at #878 (TLDR: they're testing `babel-plugin-transform-react-display-name`, not `enzyme`, should we remove them?

README.md
nfcampos: @ljharb i've done that in my latest commit (also removed changes to src/ReactWrapperComponent.jsx now that that's being done in #877)

ljharb:
@nfcampos i'm not sure why they're testing that? `displayName` is a property users can (and do) manually add to any of the three kinds of components.

nfcampos:
because we have two test cases that test createClass components that *do not* have `displayName` set, eg. https://github.com/kentcdodds/enzyme/blob/pr/support-15.5/test/ReactWrapper-spec.jsx#L3309, these test cases pass because that babel plugin adds displayName equal to the name of the variable (or it did when we're using React.createClass)

ljharb:
aha, then yes, those tests should have explicit `displayName`s added instead, and we should disable that particular babel transform entirely.

nfcampos:
i'll remove the tests because we already identical test cases with displayName
as for disabling the transform, it is part of babel-preset-react, so should we really be disabling it? most users will have it enabled

ljharb:
I'm not convinced most users use that preset, and yes, I think we should be using the most basic jsx transforms.

ljharb:
Although, for now, let's not change the transforms; just removing the tests is fine.

nfcampos:
ok, done
the build for this PR should now be green, I think ;)

nfcampos:
if someone does the two changes you requests around version detection this should now be ready to go in I think.
one last question is have is whether it'd be better to make our version check look for `>15.5`?

src/ShallowWrapper.js
```diff
@@ -98,6 +98,17 @@ function validateOptions(options) {
   }
 }
 
+
+function performBatchedUpdates(wrapper, fn) {
+  const renderer = wrapper.root.renderer;
+  if (renderer.unstable_batchedUpdates) {
```
ljharb: I still think we should check `REACT155` here instead of just ducktyping.

src/react-compat.js
ljharb: This error message still needs updating.

src/react-compat.js
```diff
@@ -107,6 +113,26 @@ if (REACT013) {
...
+      // eslint-disable-next-line import/no-extraneous-dependencies
+      shallowRendererFactory = require('react-test-renderer/shallow').createRenderer;
+    } else {
+      // eslint-disable-next-line import/no-extraneous-dependencies
+      shallowRendererFactory = TestUtils.createRenderer;
+    }
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.error(
+      'react-test-renderer is an implicit dependency in order to support react@15.5+. ' +
```
ljharb: can this use a multiline template string (or a single-line string) instead of a multiline string?

src/ShallowWrapper.js
ljharb: (per https://github.com/airbnb/enzyme/pull/876#issuecomment-292875234) yes, I think `REACT155` should probably be `>= 15.5` instead.

src/react-compat.js
kentcdodds: If you do that, I recommend `oneLine` from [`common-tags`](https://www.npmjs.com/package/common-tags).

th3fallen:
Any updates here, not to rush anyone but this is blocking the upgrade for my teams :( or anything i can do to assist?

src/react-compat.js
ljharb: I'd rather the line just be long. Line length limits are just a poor proxy for limiting code complexity.

src/ShallowWrapper.js
nfcampos: @ljharb i've addressed both

src/react-compat.js
nfcampos: @ljharb I'd rather not change this in this PR seeing as the existing error messages also use multiline strings

nfcampos:
@ljharb I believe all your comments have now been addressed, this should be ready to 🚢

karma.conf.js
```diff
@@ -4,6 +4,36 @@ require('babel-register');
...
+  if (REACT013 || REACT155) {
+    plugins.push(new IgnorePlugin(/react-addons-test-utils/));
+  }
+  if (!REACT155) {
+    plugins.push(new IgnorePlugin(/react-test-renderer/));
+    plugins.push(new IgnorePlugin(/react-dom\/test-utils/));
+    plugins.push(new IgnorePlugin(/create-react-class/));
+  }
+
+  return plugins
```
ljharb: this line is missing a semicolon

karma.conf.js
```diff
@@ -29,11 +59,7 @@ module.exports = function karma(config) {
     reporters: ['dots'],
 
     files: [
-      'test/*.{jsx,js}',
-    ],
-
-    exclude: [
-      'test/_*.{jsx,js}',
+      'test/*-spec.{jsx,js}',
```
ljharb: everything in the test folder, whether it has a "-spec" suffix or not, should be included, except the helpers that start with _ - let's revert this part for now.

src/version.js
```diff
@@ -1,6 +1,10 @@
...
 
 export const VERSION = React.version;
+
+const [major, minor] = VERSION.split('.');
+
 export const REACT013 = VERSION.slice(0, 4) === '0.13';
 export const REACT014 = VERSION.slice(0, 4) === '0.14';
-export const REACT15 = VERSION.slice(0, 3) === '15.';
+export const REACT15 = major === '15';
+export const REACT155 = REACT15 && +minor >= 5;
```
ljharb: the `+` isn't necessary; `>=` casts to a number.

kentcdodds:
To bring it back full circle, I _think_ I've addressed the final comments on the PR. Let me know if I messed anything up 😅

karma.conf.js
kentcdodds: Just to be clear, you wanted to revert changes to the `files` and `exclude` properties right? I've changed it back to what it was originally.

ljharb: yes, exactly :-) thanks!

nfcampos: there was a reason for that change: the new file `test/react-compat.js` was, under the old code, treated as a test file, which is incorrect and fails the karma tests (one hopes)

so, either the new file is renamed to `_react-compat` (which is strange but fine with me) or we accept that all our test files indeed have `-spec` in the filename (which is the case so far) and make that the rule. 

what do you guys think?

ljharb: Hmm.

tbh I'd rather make a `_helpers` dir, put all the helpers and react-compat under that, and then exclude `_helpers` entirely. I'm not a fan of the "-spec" convention, and if we want it enshrined in config, we'd need to have a way to have it break if someone forgets to add the suffix.

nfcampos: good point. then i'd rather just rename the file to `test/_react-compact.js`, no?

ljharb: For this PR, sure, let's just do that. Thanks for following up!

nfcampos: no problem. I don't have acces to a laptop atm so one of you'll have to do the renaming

ljharb:
I'll try to cut a release in the next day or two; hopefully @lelandrichardson or @gdborton can do so sooner.

bvaughn:
Great work @ljharb, @nfcampos, @kentcdodds and @gaearon!

vikkio88:
Hello everyone, great work, thanks. there will be a release which includes this fix any time soon?

kentcdodds:
As @ljharb noted:

> I'll try to cut a release in the next day or two; hopefully @lelandrichardson or @gdborton can do so sooner.

lelandrichardson:
I will cut a release asap, thanks

sprjr:
Thanks so much, Chimi <3

lelandrichardson:
2.8.1 is now published!

FezVrasta:
```
Cannot find module 'react-test-renderer/shallow' from 'react-compat.js'
```

?

ljharb:
Per the readme, you need to manually install that.

FezVrasta:
I already did, I simply upgraded Enzyme on an existing project

kentcdodds:
Actually, this should probably have been a breaking change release 🤔

ljharb:
@FezVrasta you can't simply upgrade enzyme; check the readme and be sure you have the right implicit deps installed as top level deps.

kentcdodds:
Consider this: A project with `react@^15.0.0`, `react-addons-test-utils@^15.0.0` `react-dom@^15.0.0`, `enzyme@^2.0.0` a week ago would work, but now would not work (because you'd need to also install `react-test-renderer`). So technically this should probably be a new major version...

ljharb:
It wouldn't have worked after 15.5 but before the enzyme release; so react should maybe have been a major; this release fixes that problem but still requires an extra step.

I see the argument, but strict semver is tough to apply when react considers warnings non-major, and when npm lacks the facility for us to express our implicit dependencies statically :-/

FezVrasta:
Take this for what is it, but this library is becoming a mess with all those "untold" dependencies...

ljharb:
@FezVrasta that is the fate of any library supporting a peer dep like react across major versions. It's not unique to enzyme or even react.

th3fallen:
i hope i'm missing something but after updating `react` `react-dom` `enzyme` `react-test-utils` `react-test-renderer` im still getting the proptypes warnings dispite updaitng my components?

bvaughn:
@th3fallen It's possible that other 3rd party libs you use still have references to `React.PropTypes`

th3fallen:
@bvaughn hmm that's unfortunate but it appears you're right, thanks

lelandrichardson:
> but this library is becoming a mess with all those "untold" dependencies...

For what it's worth, we are working right now on rewriting enzyme to use an "adapter" pattern which will make it so that enzyme no longer has any "implicit" dependencies like this, but instead "adapters" will rely on specific versions of react:

```js
const Enzyme = require('enzyme');
Enzyme.use({ adapter: require('enzyme-adapter-react-16') });
```
This pattern will allow adapters to pin to evolve with react (along with version numbers), and have the enzyme core library evolve independently, and require no nasty branching between dependencies for different versions of react.

plemarquand:
@ljharb @kentcdodds This should have been a major release. I upgraded to React 15.5 after it was released while watching this thread for the corresponding fix to enzyme. Because I was using Enzyme `^2.8.0`, when you released `2.8.1` it broke my builds because `react-test-renderer` became an implicit dependency that I did not have.

ljharb:
@plemarquand see https://github.com/airbnb/enzyme/issues/892#issuecomment-293745578

lelandrichardson:
I'm not opposed to publishing a `2.8.2` with a revert and a `3.0.0` with this change in it, FWIW.

Semver decisions are always a little bit subjective. In this case I think we should do whatever presents the best UX for consumers of enzyme.

ljharb:
@lelandrichardson that would be fine too! My comments are meant to indicate that I think using a patch is defensible; not that I think reverting and shipping it as a major is untenable.

plemarquand:
@ljharb This is probably just my ignorance speaking but how was React 15.5 a breaking release beyond emitting warnings? I know some very strict build systems are configured to fail on warnings but I'd wager those are probably not the norm.

gaearon:
FWIW we're going to use `console.warn()` rather than `console.error()` in 15.6 specifically for deprecation warnings so that you can you can distinguish error-ish warnings from warn-ish warnings on CIs.

wilbert-abreu:
@gaearon Thank you! Also, it would be a big help if there be an indication of where the warnings are coming exactly, meaning my local repo vs one of my packages. Right now, I'm having to investigate which packages are causing the warnings.

ljharb:
@plemarquand Warnings come from propTypes; tests should fail from propType warnings. Ideally propTypes should throw in dev, not just warn.

@gaearon thanks, that's useful.

plemarquand:
@ljharb Thanks, that makes sense. I agree I'd rather see PropType validation errors throw in dev instead of them using console.error/warn. I wouldn't enable build failures for warnings but do still want failures for PropType validation failures.

jwalton:
I'll propose a more complete version of the fix I proposed in #892:

```js
 // We require the testutils, but they don't come with 0.14 out of the box, so we
 // require them here through this node module. The bummer is that we are not able
 // to list this as a dependency in package.json and have 0.13 work properly.
 // As a result, right now this is basically an implicit dependency.
 try {
   if (REACT155) {
     // eslint-disable-next-line import/no-extraneous-dependencies
     try {
       TestUtils = require('react-dom/test-utils');
     } catch(err) {
       // If the user hasn't done all the upgrades to react@15.5 yet, this will print a nice deprecation warning.
       TestUtils = require('react-addons-test-utils');
     }
   } else {
     // eslint-disable-next-line import/no-extraneous-dependencies
     TestUtils = require('react-addons-test-utils');
   }
 } catch (e) {
   if (REACT155) {
     console.error( // eslint-disable-line no-console
       'react-dom@15.5+ is an implicit dependency when using react@15.5+ with enzyme. ' +
       'Please add the appropriate version to your devDependencies. ' +
       'See https://github.com/airbnb/enzyme#installation',
     );
   } else {
     console.error( // eslint-disable-line no-console
       'react-addons-test-utils is an implicit dependency in order to support react@0.13-14. ' +
       'Please add the appropriate version to your devDependencies. ' +
       'See https://github.com/airbnb/enzyme#installation',
     );
   }
   throw e;
 }

 // Shallow renderer is accessible via the react-test-renderer package for React 15.5+.
 // This is a separate package though and may not be installed.
 try {
   if (REACT155) {
     // eslint-disable-next-line import/no-extraneous-dependencies
     try {
         shallowRendererFactory = require('react-test-renderer/shallow').createRenderer;
     } catch(err) {
         // Backwards compatible for users who have upgraded to 15.5 but haven't done all the fidly bits yet.
         shallowRendererFactory = TestUtils.createRenderer;
     }
   } else {
     // eslint-disable-next-line import/no-extraneous-dependencies
     shallowRendererFactory = TestUtils.createRenderer;
   }
 } catch (e) {
   // eslint-disable-next-line no-console
   console.error(
     'react-test-renderer is an implicit dependency in order to support react@15.5+. ' +
     'Please add the appropriate version to your devDependencies. ' +
     'See https://github.com/airbnb/enzyme#installation',
   );
   throw e;
}
```

This makes enzyme do "the right thing" in all cases.  It uses the 15.5 version of the libraries if they are there, it falls back to the old-style versions with deprecation warnings from react if they are not.

ljharb:
@jwalton would you mind making a PR, so we can discuss that there?

Regardless of varying opinions on semver and "what enzyme should have done", applying your proposal in another patch isn't out of the question.

jwalton:
I can do that, but it'll have to wait for morning.  :)

gaearon:
>Also, it would be a big help if there be an indication of where the warnings are coming exactly, meaning my local repo vs one of my packages

This is not really possible for us to determine, is it? If you have the stack you should be able to check this yourself though.

tleunen:
We don't always have the stack.

Here's what I see from a storyshots test for example:
```
    console.error node_modules/fbjs/lib/warning.js:36
      Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
    console.info node_modules/@kadira/storybook/dist/server/babel_config.js:69
      => Loading custom .babelrc
    console.error node_modules/fbjs/lib/warning.js:36
      Warning: A Component: React.createClass is deprecated and will be removed in version 16. Use plain JavaScript classes instead. If you're not yet ready to migrate, create-react-class is available on npm as a drop-in replacement.
```

I don't really know in which file/project this is from (might be from storyshots or any of its dependencies). So unless I do a huge search under node_modules, it's impossible to quickly see that and help the authors of these libs to upgrade.

@gaearon Does an error always contain a filename? If that's the case, a "hack" would be to throw an error, and then send the warning once you catch your fake error.

jwalton:
@tleunen The ugly hack I use is:

// Make React warnings into errors
const originalError = console.error;
console.error = message => {
    if(message.startsWith("Warning: ")) {
        throw new Error(message);
    } else {
        originalError.call(console, message);
    }
};

This will make it an error, and get you a stack trace.  It would be really slick if react had a hook for this.  Something like `react.onWarning(warn => ...)`.

gaearon:
I don’t know about tests, but can’t you run the code in the browser? It should display the stack for warnings in Chrome.

plemarquand:
@gaearon Enzyme tests are frequently run in Node, thats where I have trouble getting the stack and have to fallback to hacks like @jwalton suggested.

gaearon:
@plemarquand I understand—just saying that it might be easier to find deprecation sites while working on the app than from running tests.

plemarquand:
@gaearon In the case of this patch/issue a lot of people like myself fixed all the deprecations inside their running app, but were still getting them when running their tests because enzyme (and/or other testing libraries) were/are emitting them.

gaearon:
Oh I see. To be honest I’m a little surprised third party components have already updated. Are you using very little of them?

travi:
as mentioned above, getting these to throw and show the full stack trace when running tests is to cause `console.error` to throw:

```js
console.error = warning => { throw new Error(warning); };
```

i set this up in any test framework that i use and the framework will typically catch the thrown error the same way that it would an assertion error, causing a failure and showing the full stack trace

plemarquand:
@gaearon Yeah I only ran in to it with enzyme, and only for a day before this patch was merged. Most if not all of the major libraries have updated by now. Serves me right for being on the bleeding edge, typically I like to wait a week or two before upgrading but it was a slow day at work...

kopax:
Is this merged in enzyme v2.8.2 ? I still get some warnings : 

```

console.error node_modules/fbjs/lib/warning.js:36
  Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
console.error node_modules/fbjs/lib/warning.js:36
  Warning: ReactTestUtils has been moved to react-dom/test-utils. Update references to remove this warning.
```

kentcdodds:
Hi @kopax,
Like other comments above, this is likely coming from some other package you're using.

enzyme folks, it may not be a bad idea to add a final comment explaining this and locking this issue. In any case, I'll just unsubscribe from updates.

kopax:
@kentcdodds do you have an idea on how I could trace this error ? I am trying to find which library is creating this warning. I am not using a lot but all the maintainers says they have corrected it. Would be just nice to know how to produce a more verbose message. I have tried to set `"jest":{ "verbose": true }` in my `package.json` but it did not change.

kentcdodds:
I would try going to where that warning log is (`node_modules/fbjs/lib/warning.js:36`) and add a `console.log(new Error('here I am'))` which will give you a stacktrace (which you'll normally have in the browser). Good luck.

bvaughn:
@kopax You could add something like this to the header of your file:
```js
const decorated = console.error;
console.error = message => decorated(new Error(message).stack);
```

kopax:
@kentcdodds I haven't try with it but it didn't point a lot of my files except the line of the file from the test : 

 PASS  src/components/Modal/tests/Fade.test.js
  ● Console

    console.log node_modules/fbjs/lib/warning.js:36
      Error: here I am
          at printWarning (/workspace/module.kopaxgroup.com/styled-components/bootstrap-styled/node_modules/fbjs/lib/warning.js:36:21)
           ...
          at Object.<anonymous> (/workspace/module.kopaxgroup.com/styled-components/bootstrap-styled/src/components/Modal/tests/Fade.test.js:6:15)
          ...

Where the line 6:15 of that file is `import { mount } from 'enzyme';`

peterheard01:
I'm getting this error too. React 15.5.

Ballpin:
@peterheard01 For me it was resolved by adding react-test-renderer as a dependency.

Source https://github.com/airbnb/enzyme/issues/875
Check what **OscarBarrett** comment.

donaldpipowitch:
@kopax @kentcdodds @bvaughn 

Could it be that the `Shallow renderer has been moved to react-test-renderer/shallow.` warning is still shown in certain webpack based projects?

@kopax says he gets an error like `Cannot find module 'react-addons-test-utils' from 'react-compat.js'`. I get something similar `Can't resolve 'react-addons-test-utils' in 'build/react-compat.js'.`

I think it comes from [this `try/catch`-block](https://github.com/airbnb/enzyme/blob/master/src/react-compat.js#L111). It looks like `require`ing `react-dom/test-utils` and/or `react-test-renderer/shallow` fail while bundling (or: they are bundled at build time and something fails inside this `try/catch` at runtime), so it tries to get `react-addons-test-utils`. Or does webpack try to bundled things inside a `catch` by default?

ljharb:
@donaldpipowitch please file a new issue; this one's pretty long already and posting here pings a lot of people. It's certainly possible tho that webpack detects the `require`s and bundles them up anyways; you need to configure webpack to ignore all the dynamically required things in enzyme. It should all be in the docs.

donaldpipowitch:
Okay, done: https://github.com/airbnb/enzyme/issues/968 👍

PR #877, opened 2017-04-10
Convert React.createClass to React.Component

koba04:
In react@15.5, `React.createClass` is deprecated, So we have to use `create-react-class` instead. #876

In this PR, I've converted `React.createClass` to `React.Component` except some tests for `React.createClass`. 
Should Enzyme support `create-react-class`? If it isn't necessary, I can remove all `React.createClass` from the tests.

* https://github.com/airbnb/enzyme/blob/master/test/ReactWrapper-spec.jsx#L3270
* https://github.com/airbnb/enzyme/blob/master/test/ReactWrapper-spec.jsx#L3306
* https://github.com/airbnb/enzyme/blob/master/test/ShallowWrapper-spec.jsx#L1273
* https://github.com/airbnb/enzyme/blob/master/test/ShallowWrapper-spec.jsx#L3909
* https://github.com/airbnb/enzyme/blob/master/test/ShallowWrapper-spec.jsx#L3955

After this PR is merged, Enzyme can put `create-react-class` in `devDependencies` instead of `dependencies`.

### DIFF
--- a/docs/api/ShallowWrapper/matchesElement.md
+++ b/docs/api/ShallowWrapper/matchesElement.md
@@ -21,16 +21,20 @@ render tree.
 
 
 ```jsx
-const MyComponent = React.createClass({
+class MyComponent extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleClick = this.handleClick.bind(this);
+  }
   handleClick() {
     ...
-  },
+  }
   render() {
     return (
       <div onClick={this.handleClick} className="foo bar">Hello</div>
     );
   }
-});
+}
 
 const wrapper = shallow(<MyComponent />);
 expect(wrapper.matchesElement(
--- a/src/ReactWrapperComponent.jsx
+++ b/src/ReactWrapperComponent.jsx
@@ -12,37 +12,19 @@ import objectAssign from 'object.assign';
  * pass new props in.
  */
 export default function createWrapperComponent(node, options = {}) {
-  const spec = {
-
-    propTypes: {
-      Component: PropTypes.oneOfType([PropTypes.func, PropTypes.string]).isRequired,
-      props: PropTypes.object.isRequired,
-      context: PropTypes.object,
-    },
-
-    getDefaultProps() {
-      return {
-        context: null,
-      };
-    },
-
-    getInitialState() {
-      return {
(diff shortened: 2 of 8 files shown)


src/ReactWrapperComponent.jsx
```diff
@@ -12,63 +12,54 @@ import objectAssign from 'object.assign';
...
-
-    getDefaultProps() {
-      return {
-        context: null,
-      };
-    },
-
-    getInitialState() {
-      return {
+  class WrapperComponent extends React.Component {
```
ljharb: we definitely shouldn't change the specs here; we should continue to use `createClass`.

koba04: @ljharb What's the reason to continue to use `createClass`? It seems to be able to replace with ES2015 classes. Is there any problems?

koba04: I think it would be nice if Enzyme users don't have to install `create-react-class` to use it.

ljharb: oops, you're right, this one isn't a spec file - this can stay as-is :-)

docs/api/ShallowWrapper/matchesElement.md
```diff
@@ -21,16 +21,16 @@ render tree.
 
 
 ```jsx
-const MyComponent = React.createClass({
+class MyComponent extends React.Component {
   handleClick() {
     ...
```
nfcampos: this component doesn't really work now that it's a class because methods aren't auto bound, so if `this` is used in handleClick it won't work. 

do we want to have an "invalid" component as an example in the docs?

ljharb: good catch - let's add the best practice of a manual constructor binding to it.

koba04: That makes sense. I've fixed it.

koba04:
Ok, I've reverted changes for tests.

But I think it's good to use `React.createClass` only in tests that `React.createClass` is required because React is going to separate `React.createClass` from React core.

nfcampos:
LGTM

koba04:
Thanks!

PR #878, opened 2017-04-10
Make displayNameOfNode return function.name even if it is falsey

nfcampos:
previously it would return the function itself, which is strange, `displayNameOfNode` should return either a string or null

@ljharb this can (should?) be merged before #876 

found this because the changes in #876 madeus start using `createClass` instead of `React.createClass` in the tests to create createClass components, which made this babel plugin https://babeljs.io/docs/plugins/transform-react-display-name/ stop adding the variable name as displayName to these components and thus those tests started failing with `AssertionError: expected [Function] to equal 'Foo'` (example failure at https://travis-ci.org/airbnb/enzyme/jobs/220197356#L4836)

(the tests should probably be removed, because they seem to be testing a babel plugin instead of enzyme, but anyway `displayNameOfNode` shouldn't return the function when the function doesn't have a name)


test/Utils-spec.jsx
```diff
@@ -598,6 +598,14 @@ describe('Utils', () => {
         expect(displayNameOfNode(<Foo />)).to.equal('Foo');
       });
 
+      it('should return the name even if it is falsey', () => {
```
ljharb: `falsey → falsy`

Issue #847, opened 2017-03-13
Mis-leading documentation


joetidee:
I am using Webpack 2 and React 15.x in my application. Your documentation gives two different configurations for `externals`:

https://github.com/airbnb/enzyme/blob/master/docs/guides/webpack.md

So which one do i use?

ljharb:
It looks like you want to use the "webpack 2" one, but omit the "cheerio" line.

If that works, then a PR to update the readme would be very appreciated :-)

joetidee:
Having problems with this setting. Getting the following errors on this repo:

https://github.com/joetidee/erl-example

Errors:

```
ERROR in ./~/enzyme/build/react-compat.js
Module not found: Error: Can't resolve 'react/addons' in '/mnt/c/ubuntu/erl-example/node_modules/enzyme/build'
resolve 'react/addons' in '/mnt/c/ubuntu/erl-example/node_modules/enzyme/build'
  Parsed request is a module
  using description file: /mnt/c/ubuntu/erl-example/node_modules/enzyme/package.json (relative path: ./build)
    Field 'browser' doesn't contain a valid alias configuration
  after using description file: /mnt/c/ubuntu/erl-example/node_modules/enzyme/package.json (relative path: ./build)
    resolve as module
      /mnt/c/ubuntu/erl-example/node_modules/enzyme/build/node_modules doesn't exist or is not a directory
      /mnt/c/ubuntu/erl-example/node_modules/enzyme/node_modules doesn't exist or is not a directory
      /mnt/c/ubuntu/erl-example/node_modules/node_modules doesn't exist or is not a directory
      /mnt/c/ubuntu/node_modules doesn't exist or is not a directory
      /mnt/c/node_modules doesn't exist or is not a directory
      /mnt/node_modules doesn't exist or is not a directory
      /node_modules doesn't exist or is not a directory
      looking for modules in /mnt/c/ubuntu/erl-example/node_modules
        using description file: /mnt/c/ubuntu/erl-example/package.json (relative path: ./node_modules)
          Field 'browser' doesn't contain a valid alias configuration
        after using description file: /mnt/c/ubuntu/erl-example/package.json (relative path: ./node_modules)
          using description file: /mnt/c/ubuntu/erl-example/node_modules/react/package.json (relative path: ./addons)
            as directory
              /mnt/c/ubuntu/erl-example/node_modules/react/addons doesn't exist
            no extension
              Field 'browser' doesn't contain a valid alias configuration
[...]

Issue #740, opened 2016-12-19
webpack externals are not complete


gilamran:
Hi,
In our situation we use webpack to bundle the application code into a bundle, and the external packages (Like `react`)  we load them using a `script` tag (CDN). We would like to mimic that also in our tests.
But when importing `enzyme` in our tests, it will bundle **react** and **react-dom** into the application bundle, even after we specifically mentioned them in our `externals` section.

To solve this, you have to add this to the externals:

```
    "react": 'React',
    "react-dom": 'ReactDOM',
    "react-dom/server": true,
    "react-dom/lib/ReactTestUtils": true
```

blainekasten:
Could you throw up a PR for updating the docs?

gilamran:
Done, #744

gilamran:
I've updated the issue description to explain the situation. (PR was not updated)
[Here](https://github.com/gilamran/enzyme_webpack_bug_example)'s a repo. to test the issue.
Tests are on the latest react `15.4.1`

ljharb:
I don't think this is relevant any more now that we're on v3.

Issue #309, opened 2016-04-11
_cheerio2.default.load is not a function


erwingo:
I'd appreciate any help with this: I'm using webpack + karma for my tests and did the setup as indicated in the installation guides. But when I try to use `enzyme.render(reactElement)` I got that error. This is my `karma.config.js`:

``` javascript
module.exports = function (config) {
  config.set({
    basePath: '',
    frameworks: ['mocha', 'chai'],
    port: 9876,
    browsers: ['Chrome'],

    files: [
      'node_modules/babel-polyfill/dist/polyfill.min.js',
      './src/**/*.test.js'
    ],

    preprocessors: {
      './src/**/*.test.js': ['webpack']
    },

    plugins: [
      // Browsers
      'karma-phantomjs-launcher',
      'karma-chrome-launcher',
      'karma-mocha',
      'karma-chai',
      'karma-webpack'
    ],

    webpack: {
      resolve: {
        modules: [require('path').resolve(__dirname, './src'), 'node_modules']
      },
      module: {
        loaders: [
          {
            test: /\.js$/,
            loader: 'babel-loader',
            exclude: /node_modules/
          }
        ]
      },
      externals: {
        cheerio: 'window',
        'react/addons': true,
        'react/lib/ReactContext': true,
        'react/lib/ExecutionEnvironment': true
      }
    },
    webpackMiddleware: { quiet: false }
  });
};
```

and `component.test.js`:

``` javascript
import React from 'react';
import * as enzyme from 'enzyme';
import SidebarMenu from './sidebar-menu';

describe.only('<SidebarMenu />', function () {
  it('should render correctly', () => {
    const reactElement = <SidebarMenu title="Menu" />;
    const element = enzyme.render(reactElement);
    expect(element.find('.sidebar-menu')).to.have.length(1);
  });
});
```

ttrmw:
Seeing the same thing with Jest, see #286

rnsloan:
I remember running into this error with webpack (not using karma). I cannot remember what I did specifically to fix it but here are the parts of my webpack config for testing to getting cheerio working:

```
module.exports = {
  resolve: { 
    extensions: ['.json']
  },
  externals: {
    'react/lib/ExecutionEnvironment': true,
    'react/lib/ReactContext': 'window',
  },
  module: {
    loaders: [
      {test: /\.json$/, loaders: ['json']},
    ]
  }
};
```

Maybe one of these addresses the issue.

erwingaitano:
[...]

PR #895, opened 2017-04-13
Add enzyme/webpack entrypoint with getPluginsForInstalledReact function

### LINKED ISSUE #881: Expose function that returns list of IgnorePlugins needed for using enzyme with webpack
@ljharb @lelandrichardson @aweary @blainekasten 

What do you guys think about exposing a function that returns an array of the IgnorePlugins needed to use enzyme with webpack and the currently installed react version, something like https://github.com/kentcdodds/enzyme/blob/296dc3090dd05c20a799af7d1478752a100e1a04/karma.conf.js#L9

It would make our docs for using enzyme with webpack / karma easier to maintain and always up-to-date (which it seems they rarely are). This function would be necessarily up-to-date because it'd be used in our own karma tests.

The documentation for eg. use with webpack would become something like:

---
# Using Enzyme with Webpack

If you are using a test runner that runs code in a browser-based environment, you may be using
[webpack]() in order to bundle your React code.

Webpack uses static analysis to create a dependency graph at build-time of your source code to
build a bundle. Enzyme has a handful of conditional `require()` calls in it in order to remain
compatible with React 0.13 and React 0.14.

Unfortunately, these conditional requires mean there is a bit of extra setup with bundlers like
webpack.

In your webpack configuration, you simply need to make sure that you include `IgnorePlugin`s for the conditional dependencies not needed for your version of `react`. You can do this easily with the following additions to your `webpack` config:
```
{
  plugins: require('enzyme/webpack').getPluginsForInstalledReact(),
  loaders: {
    {
      test: /\.json$/,
      loader: 'json-loader',
    },
  }
}
```

---

nfcampos:
Motivation
- make using enzyme with webpack easier
- make sure the documentation in the webpack guide doesn't go out of date

Changes
- export getPluginsForInstalledReact function
- update our karma config to use it
- update documentation guides to use it

`getPluginsForInstalledReact` is currently tested with webpack 1 and all supported versions of react because it is used in our karma tests

closes #881 #740 #744 #847 #309 #791

### DIFF
--- a/src/webpack.js
+++ b/src/webpack.js
@@ -1,0 +1,32 @@
+import { IgnorePlugin } from 'webpack';
+
+import { REACT013, REACT155 } from './version';
+
+export function getPluginsForInstalledReact() {
+  var plugins = [];
+
+  /*
+  this list of conditional IgnorePlugins mirrors the conditional
+  requires in src/react-compat.js and exists to avoid error
+  output from the webpack compilation
+  */
+
+  if (!REACT013) {
+    plugins.push(new IgnorePlugin(/react\/lib\/ExecutionEnvironment/));
+    plugins.push(new IgnorePlugin(/react\/lib\/ReactContext/));
+    plugins.push(new IgnorePlugin(/react\/addons/));
+  }
+  if (REACT013) {
+    plugins.push(new IgnorePlugin(/react-dom/));
+  }
+  if (REACT013 || REACT155) {
+    plugins.push(new IgnorePlugin(/react-addons-test-utils/));
+  }
+  if (!REACT155) {
+    plugins.push(new IgnorePlugin(/react-test-renderer/));
+    plugins.push(new IgnorePlugin(/react-dom\/test-utils/));
+    plugins.push(new IgnorePlugin(/create-react-class/));
+  }
+
+  return plugins;
+}
--- a/src/webpack.js
+++ b/src/webpack.js
@@ -1,9 +1,11 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
 import { IgnorePlugin } from 'webpack';
 
 import { REACT013, REACT155 } from './version';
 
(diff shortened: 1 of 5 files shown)


src/webpack.js
```diff
@@ -0,0 +1,33 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { IgnorePlugin } from 'webpack';
```
ljharb: Unfortunately this means we have yet another implicit dependency :-/

Theoretically it's a safe one, if someone's importing the "webpack" entry point they're likely to have webpack available, but there's no way to declare version compatibility.

src/webpack.js
```diff
@@ -0,0 +1,33 @@
...
+
+import { REACT013, REACT155 } from './version';
+
+// eslint-disable-next-line import/prefer-default-export
+export function getPluginsForInstalledReact() {
```
ljharb: Do we have any future plans for any other exports from this entry point, such that this shouldn't be the default export?

nfcampos: not really, i made it a named export to make the name of the function clarify what it does

src/webpack.js
nfcampos: I know :(
Another concern is that we'd be introducing a new API that we must remove once we release the rewrite with adapters...

nfcampos:
@aweary @blainekasten @lelandrichardson id like to have one more opinion before merging this, if any of you guys have time to take a look it'd be great :)

nfcampos:
should I close this now @ljharb ?

ljharb:
We'll probably still want to incorporate it once #1007 lands.

nfcampos:
but after #1007 we will have no need for custom webpack configs i hope, no?

ljharb:
ah that's true :-)

Yeah, let's close this, but make sure to revisit just in case after the next major.

PR #896, opened 2017-04-13
Make enzyme work for react ^15.4.x.

### LINKED ISSUE #892: 2.8.1 fails with "Cannot find module 'react-test-renderer/shallow'"
Downgrading to 2.8.0 fixes it.

jwalton:
Users with package.json files that include:

```
    "react": "^15.4.0",
    "react-dom": "^15.4.0",
    "enzyme": "^2.8.0",
```

will have their tests break after https://github.com/airbnb/enzyme/pull/876.  This PR fixes it so enzyme "does the right thing" in the case where the user is running react 15.4, running react 15.5 but hasn't done any of the "upgrade work" yet, and in the case where they are actually running 15.5.

Verified by running [this test script](https://gist.github.com/jwalton/efda0a6d2b966fb8e6431f3fb864295b#file-test-sh).  If you want to run this, you need a folder called ~/Development/enzyme which has a built copy of this branch, and you need to run it in some other directory.  (Need the built copy instead of just `npm link`ing because we need to pick up the right versions from the test project.)  The output of this script on my machine is [here](https://gist.github.com/jwalton/efda0a6d2b966fb8e6431f3fb864295b#file-output).

Closes #892.

### DIFF
--- a/src/ShallowWrapper.js
+++ b/src/ShallowWrapper.js
@@ -102,7 +102,7 @@ function validateOptions(options) {
 
 function performBatchedUpdates(wrapper, fn) {
   const renderer = wrapper.root.renderer;
-  if (REACT155) {
+  if (REACT155 && renderer.unstable_batchedUpdates) {
     // React 15.5+ exposes batching on shallow renderer itself
     return renderer.unstable_batchedUpdates(fn);
   }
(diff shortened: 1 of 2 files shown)


jwalton:
I'm going to update my test script to try 0.14.8 as well...

jwalton:
Thank-you, sir.  :)

ljharb:
Thank you!

fillippeyton:
Awesome!

ljharb:
v2.8.2 published.

PR #904, opened 2017-04-16
Added missing space after a "your" in react-compat

JoshuaKGoldberg:


src/react-compat.js
```diff
@@ -118,7 +118,7 @@ if (REACT013) {
     if (REACT155) {
       console.error( // eslint-disable-line no-console
         'react-dom@15.5+ and react-test-renderer are implicit dependencies when using' +
```
tleunen: The space is also missing after `using` ;)

ljharb:
@JoshuaKGoldberg Let's update that one as well ^ :-)

nfcampos:
Thanks @JoshuaKGoldberg and @tleunen!

Issue #912, opened 2017-04-21
Docs link in PR placeholder doesn't go anywhere


hakunin:
https://github.com/airbnb/enzyme/commit/c2545d9d8637d2b6a4d1f2d117b45a70f578baf5#commitcomment-21854609

blainekasten:
Mind putting together a quick PR to fix it?

hakunin:
I can try :) where did you intend the docs link to go?
I don't see a common issues page here http://airbnb.io/enzyme/docs/guides.html

blainekasten:
https://github.com/airbnb/enzyme/blob/master/docs/common-issues.md

blainekasten:
Has there not been a deploy to the docs since that was added? @ljharb

ljharb:
@blainekasten there's a deploy every release; that file was last updated prior to v2.8.0, so it should be deployed. We may not add every file into gitbook tho

hanumanthan:
Fix in PR [912](https://github.com/airbnb/enzyme/pull/913).

Sorry @hakunin If i jumped in and opened PR. If have already worked on it, happy to close my PR.

hakunin:
No worries :) thanks for jumping in @hanumanthan

ljharb:
Merged in #913.

PR #913, opened 2017-04-23
Common issues link in issue template

hanumanthan:
Linking common issues in issue template.

Fix for [912](https://github.com/airbnb/enzyme/issues/912)

@blainekasten please review

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -3,11 +3,11 @@ Before submitting, please read over our commonly reported issues to prevent dupl
 
 ### All common issues
 
-* [common issues](/docs/common-issues)
+* [common issues](../blob/master/docs/common-issues.md)
 
 ### Notoriously common issues
 
-* [Webpack build issues](/docs/common-issues#webpackbuildissues)
-* [Cannot find module 'react-dom/lib/ReactTestUtils'](/docs/common-issues)
-* [Query Selector fails](/docs/common-issues#queryselectorfails)
-* [Testing third party libraries](/docs/common-issues#testingthirdpartylibraries)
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -3,11 +3,11 @@ Before submitting, please read over our commonly reported issues to prevent dupl
 
 ### All common issues
 
-* [common issues](/docs/common-issues)
+* [common issues](/docs/common-issues.md)
```
ljharb: My concern is that in the current [issue template](https://github.com/airbnb/enzyme/issues/new), this link goes to https://github.com/docs/common-issues - this change only makes it go to https://github.com/docs/common-issues.md, which is also incorrect.

I think that none of the links can have leading slashes, because that makes them absolute links on github.com.

hanumanthan: @ljharb Isnt it ok to be absolute links on github as this is an issue template which shows up in github when you raise an issue.

Or do we want to show the links instead of markdown style?

ljharb: @hanumanthan I think you misunderstood my comment - if you try to open a new issue, and click preview, and try to click one of the links, you will see that it goes to the wrong place.

In other words, I think this needs to be `../blob/master/docs/common-issues.md` in order for it to work properly.

hanumanthan: gotcha @ljharb. Will fix it

hanumanthan: @ljharb Updated.

hanumanthan:
@ljharb Squashed to single commit

Issue #919, opened 2017-04-27
Converting instance of React component to shallow wrapper


dentuzhik:
I assume I missed some bit from documentation, but so far my searches didn't achieve anything.
Let's say I have a very simple setup of component, which looks like this, and I want to test it:
```javascript
import React, { Component } from 'react';
import Button from 'components/Button';

class MyComponent extends Component {
    render() {
    	const { shouldRenderButton } = this.props;

    	return (<div>
    	    <h1>My Amazing Title<h1>
    	    {shouldRenderButton && this.renderButton()}
    	</div>);
    }

    renderButton() {
    	return (<Button 
    	    outlined={true} 
    	    amazing={'dinosaur'}
    	/>);
    }
}
```

The common approach to test in react is to invoke the shallow render on the root of the component, and then start testing internals. In this particular case, this will work nice, but if your application grows, writing tests like this will become cumbersome and you will start splitting your components into additional subrenders.

The way how I want to test my component in this case, is I want to test both `render` and `renderButton` methods separately (stubbing where necessary), while still using enzyme. The reasons for doing this is a separate topic, but basically to achieve better tests, which are less fragile to other changes.

And here I'm stuck, because when invoking this code, I have no idea how to use enzyme anymore:
```javascript
const wrapper = shallow(<MyComponent shouldRenderButton />;
const resultOfSubrender = wrapper.instance().renderButton();
// How to use enzyme API on the resultOfSubrender
```

Here's how assertion, which I want to write looks like:
```javascript
describe('renderButton', () => {
    it('should have an amazing prop of dinosaur', () => {
        const wrapper = shallow(<MyComponent shouldRenderButton />);
        const renderedButton = wrapper.instance.renderButton();
        let buttonShallowWrapper;
        // Here I do some magick with renderedButton to convert it into shallow wrapper

        // Assuming I use chai-enzyme
        // https://github.com/producthunt/chai-enzyme#propkey-val
        expect(buttonShallowWrapper).to.have.prop('amazing', 'dinosaur');
    })
})
```

My question is how can I convert the instance of rendered react component into a shallow wrapper, after it happened, and whether it's even possible to do this. So far I've been relying on `react-dom/test-utils` and just direct assertions agains the properties of the instance, but it would be nice to be able to use enzyme.

[...]

PR #920, opened 2017-04-27
Add wrap utility helper documentation

dentuzhik:
See https://github.com/airbnb/enzyme/issues/919 for my monolog of finding this in source code :)
I know explanation in documentation may not be the best, but it's better then nothing.

### DIFF
--- a/docs/api/ShallowWrapper/wrap.md
+++ b/docs/api/ShallowWrapper/wrap.md
@@ -1,0 +1,25 @@
+# `.wrap(node) => ShallowWrapper`
+Helpful utility method to create a new wrapper with the same root as the current wrapper, with any nodes passed in as the first parameter automatically wrapped.
+
+#### Arguments
+
+
+1. `node` (`ReactElement`): The node, which will be the root of your shallow wrapper
+
+
+#### Returns
+
+`ShallowWrapper`: A new wrapper that wraps the provided node.
+
+
+
+#### Examples
+
+```jsx
+const wrapper = shallow(<MyComponent />);
+const myOtherWrapper = wrapper.wrap(<MyOtherComponent />);
+
+expect(myOtherWrapper.instance()).to.be.instanceOf(MyOtherComponent);
+```
+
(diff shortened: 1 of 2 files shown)


docs/api/ShallowWrapper/wrap.md
```diff
@@ -0,0 +1,25 @@
...
+
+
+
+#### Examples
+
+```jsx
+const wrapper = shallow(<MyComponent />);
+const myOtherWrapper = wrapper.wrap(<MyOtherComponent />);
+
+expect(myOtherWrapper.instance()).to.be.instanceOf(MyOtherComponent);
```
ljharb: this could probably be explained a bit better by providing `.debug()` output for both wrappers?

dentuzhik: Fair enough, I'll update it tomorrow.

xaka:
@dentuzhik `wrap` is we've been looking for! thanks for discovering it :) 

it's been a while, but do you think you'll have time to address that minor PR comment, rebase and get it merged? thanks again

lelandrichardson:
@xaka @dentuzhik i'm curious... what is it you're using `wrap` for?  It's traditionally been thought of as an internal method and was thus not documented. I'm a bit reluctant to make it part of our public API, but am curious to know why you found it useful?

dentuzhik:
I have described my usage pattern in the related issue, basically it makes it possible to test chunks of html extracted from a main render method to other component methods (like "renderButton").

I mostly use this scenario when I am not entirely sure that this chunk of markup is reusable, but don't want to leave it uncovered. This way I have a method, which is almost prepared to extraction to the component with an isolated test, which will not fail if the tests for the main render will.

I am willing to rebase PR and provide addition description as requested.

dentuzhik:
After a few years of keeping this PR opened, for those who will come by I will say that I agree with original reasoning of @ljharb regarding SFCs.

Especially after having changed context to TS, where type safety of SFC is way more straightforward than for component methods which return React elements.

So @lelandrichardson I think it shouldn't be in public API :)

PR #921, opened 2017-04-28
[docs] Update "jsdom" guides for v10 and later

### LINKED ISSUE #942: Enzyme + jsdom -> TypeError: jsdom is not a function
I am trying to get Enzyme up and running with jsdom as describe here:

http://airbnb.io/enzyme/docs/guides/jsdom.html

```
/* setup.js */

var jsdom = require('jsdom').jsdom;

global.document = jsdom('');
global.window = document.defaultView;
Object.keys(document.defaultView).forEach((property) => {
  if (typeof global[property] === 'undefined') {
    global[property] = document.defaultView[property];
  }
});

global.navigator = {
  userAgent: 'node.js'
};
```

However, this gives the following:

`TypeError: jsdom is not a function` from line 3 `global.document = jsdom('');`, why? And how to resolve this?

kenju:
# What

Updated "jsdom" guides which is based on "jsdom" v10 and later API

# Why

`jsdom` has recently updated to v10, but the enzyme's guides is based on ["old-api"](https://github.com/tmpvar/jsdom/blob/master/lib/old-api.md). `jsdom` says it supports the old api yet, so I did not delete the old code block but instead add a new code block.

> The old API is still supported for now

Fixes #942.

### DIFF
--- a/docs/guides/jsdom.md
+++ b/docs/guides/jsdom.md
@@ -11,22 +11,52 @@ gets run *before* React's code is run.
 
 As a result, a standalone script like the one below is generally a good approach:
 
+`jsdom v10~`:
+
 ```js
 /* setup.js */
 
-var jsdom = require('jsdom').jsdom;
+const { JSDOM } = require('jsdom');
+const jsdom = new JSDOM('<!doctype html><html><body></body></html>');
+const { window } = jsdom;
 
-global.document = jsdom('');
-global.window = document.defaultView;
-Object.keys(document.defaultView).forEach((property) => {
-  if (typeof global[property] === 'undefined') {
-    global[property] = document.defaultView[property];
-  }
-});
+function copyProps(src, target) {
+  const props = Object.getOwnPropertyNames(src)
+    .filter(prop => typeof target[prop] === 'undefined')
+    .map(prop => Object.getOwnPropertyDescriptor(src, prop));
+  Object.defineProperties(target, props);
+}
 
+global.window = window;
+global.document = window.document;
 global.navigator = {
   userAgent: 'node.js'
 };
+copyProps(window, global);
+```
+
+Here is the sample of [jsdom old API](https://github.com/tmpvar/jsdom/blob/master/lib/old-api.md) as well.
+
+`jsdom ~<v10`:
+
+```js
+/* setup.js */
+
+const jsdom = require('jsdom').jsdom;
+
+global.document = jsdom('');
+global.window = document.defaultView;
+global.navigator = {
+  userAgent: 'node.js'
+};
+
+function copyProps(src, target) {
+  const props = Object.getOwnPropertyNames(src)
+    .filter(prop => typeof target[prop] === 'undefined')
+    .map(prop => Object.getOwnPropertyDescriptor(src, prop));
+  Object.defineProperties(target, props);
+}
+copyProps(document.defaultView, global);
 ```
 
 


docs/guides/jsdom.md
```diff
@@ -11,6 +11,31 @@ gets run *before* React's code is run.
 
 As a result, a standalone script like the one below is generally a good approach:
 
+`jsdom v10~`:
+
+```js
+/* setup.js */
+
+var { JSDOM } = require('jsdom');
+var jsdom = new JSDOM('');
```
kenju: nits chage though. the latter parts is the same, so maybe better way to slim up?

eddies: I don't think this is enough, either. With the proposed doc changes, I see the same errors as reported here: https://github.com/tmpvar/jsdom/issues/1827.

docs/guides/jsdom.md
```diff
@@ -11,6 +11,31 @@ gets run *before* React's code is run.
...
+
+global.document = jsdom;
```
eddies: Assuming the above (`var jsdom = new JSDOM('');`), this should be:
```javascript
global.document = jsdom.window.document;
```

kenju: Thank you for reviewing. 

I fixed the doc changes with https://github.com/airbnb/enzyme/pull/921/commits/35c2f5e29e9f942975b5fc0838b2838bbd0c90de

I created a working sample repository with `jsdom ^10.1.0`.
https://github.com/kenju/enzyme-draftjs-sample/blob/master/test/setup.js

docs/guides/jsdom.md
```diff
@@ -11,6 +11,33 @@ gets run *before* React's code is run.
...
+/* setup.js */
+
+var { JSDOM } = require('jsdom');
+var jsdom = new JSDOM('<!doctype html><html><body></body></html>');
+var { window } = jsdom;
+
+global.window = window;
+global.document = window.document;
+
+Object.keys(window).forEach((property) => {
```
ljharb: this should use `Object.getOwnPropertyDescriptors` (or `Object.getOwnPropertyNames`), not `Object.keys`, to ensure non-enumerables are grabbed as well.

docs/guides/jsdom.md
```diff
@@ -11,6 +11,33 @@ gets run *before* React's code is run.
...
+  if (typeof global[property] === 'undefined') {
+    global[property] = window[property];
```
ljharb: this should use `defineProperty` with a property descriptor to maintain nonenumerability and nonwritability when applicable.

docs/guides/jsdom.md
kenju: Thanks for the advise, that way would be better!

Fixed with: https://github.com/airbnb/enzyme/pull/921/commits/57dc37a58ea17a690895b82702a0a983a74b7ebd

docs/guides/jsdom.md
```diff
@@ -11,22 +11,52 @@ gets run *before* React's code is run.
...
 /* setup.js */
 
-var jsdom = require('jsdom').jsdom;
+const { JSDOM } = require('jsdom');
+const jsdom = new JSDOM('<!doctype html><html><body></body></html>');
+const { window } = jsdom;
+
+function copyProps(src, target) {
+  const props = Object.getOwnPropertyNames(src)
+    .filter(prop => typeof target[prop] === 'undefined')
```
ljharb: i think you want to keep the ones where it's *not* undefined, no?

docs/guides/jsdom.md
```diff
@@ -11,22 +11,52 @@ gets run *before* React's code is run.
...
-  }
-});
-
 global.navigator = {
   userAgent: 'node.js'
 };
+
+function copyProps(src, target) {
+  const props = Object.getOwnPropertyNames(src)
+    .filter(prop => typeof target[prop] === 'undefined')
```
ljharb: also here

docs/guides/jsdom.md
kenju: no, actually I meant to filter out ones *which are undefined at the `target` object*.

this is equivalent with below code (would you prefer this style?):

```js
  Object.getOwnPropertyNames(src).forEach((prop) => {
    if (typeof target[prop] === 'undefined') { // if not defined yet as the 'target's props
      Object.getOwnPropertyDescriptor(src, prop);
    }
  });
```

ljharb: Ah, you're right, thanks for clarifying :-)

eddies:
I meant to create a minimal test case to replicate, but haven't had a chance yet. In the meantime, however, I wanted to report that with this latest version of the documentation, I get `XMLHttpRequest is not defined` in my project's tests.

kenju:
@eddies 

Does your code premise that `XMLHttpRequest` exists in `global` object?

`XMLHttpRequest` seems to be attached to `jsdom.window`.
https://github.com/tmpvar/jsdom/search?utf8=%E2%9C%93&q=XMLHttpRequest&type=

Does attaching `jsdom.window.XMLHttpRequest` to `global` help for your case like below code?

```js
global.XMLHttpRequest = jsdom.window.XMLHttpRequest;
```

or,

```js
global.XMLHttpRequest = require('xmlhttprequest').XMLHttpRequest;
```

ljharb:
It's a reasonable assumption for browser code to treat all `window` properties as globals; and for jsdom to be properly simulating a browser, all `window` properties should be installed on the global object.

ljharb:
@kenju would you mind rebasing this?

kenju:
@ljharb Sure I did ->
https://github.com/airbnb/enzyme/pull/921/commits/a6933271ae690b118911566059f1c5af799fbef1

kenju:
@eddies 

what is the version of your `jsdom`? would you try below script out if `XMLHttpRequest` property can be found. `XMLHttpRequest` can be found in [this sample repository](https://github.com/kenju/enzyme-draftjs-sample/blob/master/package.json#L13), with `jsdom ^v10.1.0`:

```js
  const props = Object.getOwnPropertyNames(src)
    .filter(prop => typeof target[prop] === 'undefined')
    .map(prop => {
      if (/XMLHttpRequest/.test(prop)) { // checks if 'XMLHttpRequest'-like props found or not
        console.log(`${prop} found`);
      }
      return Object.getOwnPropertyDescriptor(src, prop);
    });
```

ljharb:
@kenju it'll need another rebase; see "This branch is out-of-date with the base branch" below

kenju:
@ljharb Oops I have missed that, now updated with the latest master branch

mattnedrich:
It looks like while this is merged, it has yet to be released - is that correct? The docs still look old - http://airbnb.io/enzyme/docs/guides/jsdom.html.

I ran into an issue getting Enzyme + ReactResponsive to work in a test. Glad I came across this PR -nice work, it fixed the issue I was having.

ljharb:
Yes, that's correct.

PR #926, opened 2017-05-03
Fix #925 - indentation is broken for debug() on shallow wrapper

### LINKED ISSUE #925: indentation is broken for debug() on shallow wrapper
### Repro
1. render a shallow wrapper of an element that has children
2. call .debug() on the wrapper

expected: each level of children is on a new line and intended 2 spaces
actual: no levels of children are indented

finnigantime:
### SUMMARY
- Fixes #925, the issue where .debug() on a shallow wrapper does not indent child nodes

### DETAILS
The problem is with the use of `.map()`:
```
export function debugNodes(nodes) {
  return nodes.map(debugNode).join('\n\n\n');
 }
```

[Array.map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map?v=example) passes 3 args to the callback, the second of which is the current index (0 in this case since there's only one top-level node), which overrides the default arg value `indentLength = 2` in debugNode.

The fix here adds a callback that explicitly only takes one parameter and then calls debugNode explicitly with that single parameter: `nodes.map(node => debugNode(node))`

### DIFF
--- a/test/Debug-spec.jsx
+++ b/test/Debug-spec.jsx
@@ -4,8 +4,9 @@ import {
   spaces,
   indent,
   debugNode,
+  debugNodes,
 } from '../src/Debug';
-import { mount } from '../src/';
+import { mount, shallow } from '../src/';
 import {
   describeWithDOM,
   describeIf,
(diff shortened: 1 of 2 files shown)


finnigantime:
@ljharb Thanks for looking. I pushed an additional commit to skip these new tests on v0.13 where they were failing.

test/Debug-spec.jsx
```diff
@@ -447,9 +448,104 @@ describe('debug', () => {
...
+  <Foo baz="bax">
+    <span>
+      From Bar
+    </span>
+  </Foo>
+</div>`);
+    });
+  });
+
+  describeIf(!REACT013, 'debugNodes', () => {
```
ljharb: instead of skipping these tests on react 13, can we just not use an SFC?

finnigantime: Ah, got it. Good idea, I can try that.

finnigantime:
@ljharb updated

ljharb:
@finnigantime please don't use the update branch button in Github's UI - this needs a rebase on the command line.

finnigantime:
Should be squashed and rebased against the latest master commit now.

ljharb:
The tests are failing due to npm v4.6 breaking on node < 1.0. I'll fix in master, after which you can rebase.

ljharb:
@finnigantime k, if you rebase once more it should pass

finnigantime:
@ljharb Thanks! Green build. G2G?

PR #945, opened 2017-05-16
Add ShallowWrapper#invoke() to invoke event handlers and return the handlers value

milesj:
cc: @kesne @alecklandgraf

This updates `ShallowWrapper#simulate()` to return the value of the executed handler if available. This allows us to easily test changes that occur in the next tick, like promise chains.

```js
// handler
onClick() {
  return Promise.resolve()
    .then(() => {
      this.setState({ foo: true });
    });
}

// test
wrapper.simulate('click').then(() => {
  expect(wrapper.state('foo')).toBe(true);
});
```

### DIFF
--- a/docs/api/ShallowWrapper/invoke.md
+++ b/docs/api/ShallowWrapper/invoke.md
@@ -1,0 +1,39 @@
+# `.invoke(event[, ...args]) => Any`
+
+Invokes an event handler (a prop that matches the event name).
+
+#### Arguments
+
+1. `event` (`String`): The event name to be invoked
+2. `...args` (`Any` [optional]): Arguments that will be passed to the event handler
+
+#### Returns
+
+`Any`: Returns the value from the event handler..
+
+#### Example
+
+```jsx
+class Foo extends React.Component {
+  loadData() {
+    return fetch();
+  }
+  render() {
+    return (
+      <a onClick={() => this.loadData()}>
+        Load more
+      </a>
+    );
+  }
+}
+
+const wrapper = shallow(<Foo />);
+
+wrapper.invoke('click').then(() => {
+  // expect()
+});
+```
+
+#### Related Methods
+
+- [`.simulate(event[, data]) => Self`](simulate.md)
--- a/docs/api/shallow.md
+++ b/docs/api/shallow.md
@@ -180,6 +180,9 @@ Returns the key of the current node.
 #### [`.simulate(event[, data]) => ShallowWrapper`](ShallowWrapper/simulate.md)
 Simulates an event on the current node.
 
+#### [`.invoke(event[, ...args]) => Any`](ShallowWrapper/invoke.md)
+Invokes an event handler on the current node and returns the handlers value.
+
 #### [`.setState(nextState) => ShallowWrapper`](ShallowWrapper/setState.md)
 Manually sets state of the root component.
 
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -48,6 +48,7 @@
     * [hasClass(className)](/docs/api/ShallowWrapper/hasClass.md)
     * [hostNodes()](/docs/api/ShallowWrapper/hostNodes.md)
     * [html()](/docs/api/ShallowWrapper/html.md)
+    * [invoke(event[, ...args])](/docs/api/ShallowWrapper/invoke.md)
     * [instance()](/docs/api/ShallowWrapper/instance.md)
     * [is(selector)](/docs/api/ShallowWrapper/is.md)
     * [isEmpty()](/docs/api/ShallowWrapper/isEmpty.md)
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -1203,6 +1203,7 @@ describe('shallow', () => {
     'hostNodes',
     'html',
     'instance',
+    'invoke',
     'is',
     'isEmpty',
     'isEmptyRender',
--- a/packages/enzyme-test-suite/test/shared/methods/invoke.jsx
+++ b/packages/enzyme-test-suite/test/shared/methods/invoke.jsx
@@ -1,0 +1,224 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { expect } from 'chai';
+import wrap from 'mocha-wrap';
+import sinon from 'sinon-sandbox';
+import { Portal } from 'react-is';
+
+import { render } from 'enzyme';
+import getAdapter from 'enzyme/build/getAdapter';
+import {
+  ITERATOR_SYMBOL,
+  sym,
+} from 'enzyme/build/Utils';
+
+import {
+  describeIf,
+  itIf,
+} from '../../_helpers';
+import realArrowFunction from '../../_helpers/realArrowFunction';
+import { getElementPropSelector, getWrapperPropSelector } from '../../_helpers/selectors';
+import {
+  is,
+  REACT16,
+} from '../../_helpers/version';
+
+import {
+  createClass,
+  createPortal,
+  createRef,
+  Fragment,
+} from '../../_helpers/react-compat';
+
+export default function describeInvoke({
+  Wrap,
+  WrapRendered,
+  Wrapper,
+  WrapperName,
+  isShallow,
+  isMount,
+  makeDOMElement,
+}) {
+  describe('.invoke(eventName, ..args)', () => {
+    it('should return the handlers return value', () => {
+      const spy = sinon.stub().returns(123);
+      class Foo extends React.Component {
+        render() {
+          return (<a onClick={spy}>foo</a>);
+        }
+      }
+
+      const wrapper = shallow(<Foo />);
+      const value = wrapper.invoke('click');
+
+      expect(value).to.equal(123);
+      expect(spy).to.have.property('callCount', 1);
+    });
+
+    it('should invoke event handlers without propagation', () => {
+      class Foo extends React.Component {
+        constructor(props) {
+          super(props);
+          this.state = { count: 0 };
+          this.incrementCount = this.incrementCount.bind(this);
+        }
+
+        incrementCount() {
+          this.setState({ count: this.state.count + 1 });
+        }
+
+        render() {
+          const { count } = this.state;
+          return (
+            <div onClick={this.incrementCount}>
+              <a
+                className={`clicks-${count}`}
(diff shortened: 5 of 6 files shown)


src/ShallowWrapper.js
```diff
@@ -614,21 +614,23 @@ class ShallowWrapper {
...
         performBatchedUpdates(this, () => {
-          handler(...args);
+          response = handler(...args);
         });
         this.root.update();
+        return response;
       });
     }
-    return this;
+    return null;
```
ljharb: this should return `undefined` rather than `null`.

src/ShallowWrapper.js
```diff
@@ -614,21 +614,23 @@ class ShallowWrapper {
...
     const handler = this.prop(propFromEvent(event));
     if (handler) {
-      withSetStateAllowed(() => {
+      return withSetStateAllowed(() => {
         // TODO(lmr): create/use synthetic events
         // TODO(lmr): emulate React's event propagation
+        let response;
         performBatchedUpdates(this, () => {
-          handler(...args);
+          response = handler(...args);
```
ljharb: I wonder if it would be better to return the equivalent of `Promise.try(() => handler(...args))`, to cleanly encapsulate any exceptions - the only downside being that you couldn't get at the result synchronously. Are there use cases for a synchronous non-Promise return value?

src/Utils.js
```diff
@@ -218,13 +218,14 @@ export function withSetStateAllowed(fn) {
     cleanup = true;
     global.document = {};
   }
-  fn();
+  const response = fn();
```
ljharb:
```js
return try { fn(); } finally {
  if (cleanup) {
      // This works around a bug in node/jest in that developers aren't able to
      // delete things from global when running in a node vm.
      global.document = undefined;
      delete global.document;
    }
};
```

(this is arguably what it should have been doing all along, in case `fn` throws)

nfcampos:
Linking to two previous PRs doing similar things https://github.com/airbnb/enzyme/pull/867 https://github.com/airbnb/enzyme/pull/898

As per the discussion on the other two PRs, events don't have a single result, because they fire zero, one or more event listeners. (so if you start exposing the return value of a handler, which of the several handlers do you choose?) The fact that our current implementation of `ShallowWrapper#simulate` does not properly propagate events, thus firing at most a single event handler, is a shortcoming that we probably want to address at some point (see https://github.com/airbnb/enzyme/pull/368 for an implementation of just that).

As such, this gets a 👎 from me, what do the other maintainers think?

cc @ljharb

nfcampos:
In other words, if we get this in we're accepting that the `simulate` method on `shallow` will be semantically different from the `simulate` method on `mount` going forward, and we should close #368

milesj:
@nfcampos In the context of a React component, there is only a single event with a single return value, so that shouldn't be an issue.

But yes, this would deviate mount and shallow. Perhaps we can add a new method and keep existing functionality, like `simulateAndReturn`?

nfcampos:
@milesj events in browsers bubble through the tree of rendered elements, ie. in
```
<div onClick={doSomethingElse}>
  <div onClick={doSomething} />
</div>
```
if you click on the inner div both onClick handlers fire.
Our current implementation of `shallow` happens to not respect that (`mount` does) so currently if you do `.simulate('click')` on that inner div using the shallow renderer only `doSomething` will be invoked.

milesj:
Correct, and that's a single event. That's what I'm getting at with shallow.

nfcampos:
yes, it's a single event, invoking two event handlers, so how do you choose which one to return the return value for?

milesj:
Shallow doesn't invoke 2 event handlers -- you literally just said this above. It executes a prop of the same name, for the element that is currently found. That is 1 event with 1 return value.

We're talking in the context of shallow, not mount.

nfcampos:
yes, and as I said above, that's something we've been thinking about changing for quite some time (see open PR at https://github.com/airbnb/enzyme/pull/368/files#diff-cc80d1de64f4765743cbe5179e51ff36, or in fact just look at @lelandrichardson's comment right above one of the lines you changed https://github.com/airbnb/enzyme/pull/945/files#diff-00a0c52d2a0eac5ec0025c1d32767a95R624).

So, re-iterating, if we do bring your change in, we'll have to abandon our previous effort at introducing event propagation for `shallow.simulate` because the two concepts are not compatible.

I'm not strictly against either option, but I do believe we should have that trade-off in mind while considering either option.

milesj:
Which is why I suggested an additional method that doesn't use propagation.

I'd wager that triggering a prop function on a React component has much higher usage than triggering an actual event.

nfcampos:
like https://github.com/airbnb/enzyme/pull/368/files#diff-13d31897d7e0bc58111b00a5b11bee40 ?

milesj:
Yup. Would just need to return the value, or be wrapped in a promise, or else we'd still have the same next tick problems.

nfcampos:
do you want to cherry-pick the relevant commits out of that PR and change the return value?

milesj:
I can look into it since that PR is currently blocked.

milesj:
@nfcampos @ljharb @kesne

Updated the PR to pull `invoke` from https://github.com/airbnb/enzyme/pull/368

docs/api/ReactWrapper/isEmptyRender.md
```diff
@@ -0,0 +1,18 @@
+# `.isEmptyRender() => Boolean`
+
+Returns whether or not the current component returns a falsey value: `false`, `null`, `undefined`.
```
ljharb: `s/falsey/falsy/g`

ljharb: (altho actually i'll just cherry-pick this into master now and fix it inline; you can rebase after that and omit this commit)

milesj: 👍

milesj:
@ljharb It looks like simulate functionality from React test utils does not return the handlers value, so we're unable to do that at this time.

ljharb:
(you may want to retry the rebase, and this time excise the isEmptyRender commit entirely)

milesj:
@ljharb Rebased the last few commits and squashed them. Looks good now.

test/ShallowWrapper-spec.jsx
```diff
@@ -1006,6 +1006,189 @@ describe('shallow', () => {
...
+          this.setState({ count: this.state.count + 1 });
+        }
+
+        render() {
+          return (
+            <div onClick={this.incrementCount}>
+              <a
+                className={`clicks-${this.state.count}`}
+                onClick={this.incrementCount}
+              >foo</a>
```
ljharb: nbd, especially in a test, but i'd expect the linter to error out here and require:
```jsx
<a
  className={`clicks-${this.state.count}`}
  onClick={this.incrementCount}
>
  foo
</a>
```

milesj: Perhaps the eslint config is out of date?

ljharb: maybe, i also might be wrong and the linter doesn't catch this yet :-) no big deal tho, it'll get fixed eventually

milesj:
@ljharb This is where the simulate happens from test-utils: https://github.com/facebook/react/blob/master/src/renderers/dom/test/ReactTestUtils.js#L442

Events are queued and executed as a batch. Doesn't look easy to pull the handler value out of it. https://github.com/facebook/react/blob/master/src/renderers/shared/shared/event/EventPluginHub.js#L224

ljharb:
I wonder if we did something like, invoke react's simulate method, but not on the prop value - instead, on something like `(...args) => { const value = originalPropValue(...args); return value; }`?

For example, this could be done by temporarily mutating the prop value on the component, and then after the simulation was done, restoring it? The batched updates may make it harder tho.

milesj:
@ljharb When using React's simulate, we pass a DOM node, not the props. I'm not sure that will work.

docs/api/ShallowWrapper/invoke.md
```diff
@@ -0,0 +1,39 @@
...
+Invokes an event handler (a prop that matches the event name).
+
+#### Arguments
+
+1. `event` (`String`): The event name to be invoked
+2. `...args` (`Any` [optional]): Arguments that will be passed to the event handler
+
+#### Returns
+
+`Any`: Returns the value from the event handler..
```
nfcampos: two dots at the end

docs/api/shallow.md
```diff
@@ -172,6 +172,9 @@ Returns the key of the current node.
 #### [`.simulate(event[, data]) => ShallowWrapper`](ShallowWrapper/simulate.md)
 Simulates an event on the current node.
 
+#### [`.invoke(event[, ...args]) => Any`](ShallowWrapper/invoke.md)
+Invokes an event handler on the current node and returns the handlers value.
```
nfcampos: “and returns the handler’s return value.”

nfcampos:
is anything stopping us from getting implementing this in `mount` in the exact same way, ie. select the prop out of the component and invoke it?

koba04:
Currently, `.simulate()` causes many problems so I think enzyme should have an alternative API to replace `simulate()`.
I think this `invoke()` feels to be close it. 
What status is this PR?

docs/api/ShallowWrapper/invoke.md
```diff
@@ -0,0 +1,39 @@
+# `.invoke(event[, ...args]) => Any`
+
+Invokes an event handler (a prop that matches the event name).
```
ljharb: i kind of wonder if maybe this should just be a prop name rather than an event name, ie, `onClick`, not `click`? Then it can be used to invoke any function-valued prop.

nfcampos: Yes I think I agree

milesj:
This kind of got lost in the version 3 release. I'll dig into it again at some point.

docs/api/ShallowWrapper/invoke.md
koba04: +1 for `onClick` rather than `click`.

ljharb:
@milesj would you be open to rebasing this with `invoke` on both shallow and mount wrappers, with this signature/behavior?
```js
invoke(propName, ...args) {
  // throw if `this.prop(propName)` is not a function
  // call it with `args` list
  // returns the result
}
```

That way, we can conveniently get access to the return value, with a nicer syntax than `.prop('onClick')()`.

Thoughts?

milesj:
Yeah most definitely. Can jump on this again.

Issue #960, opened 2017-05-31
Request: add option to debug() to ignore props


finnigantime:
It would be nice to have an option in debug() that skips printing props in the debug string.

For large components with many props spread across the different elements in the render tree, if you want to focus on the structure of the tree when debugging, this option can hide all props so you can focus just on the hierarchy of elements.

This is particularly useful if matchesElement() is used in a test that fails. matchesElement() just returns a boolean that doesn't show the diff of what didn't match when it fails. Using debug() is the best way to debug this. However, matchesElement() ignores props in its comparison, so it would be nice to have a variant of debug() that also ignores props so that it can be used while debugging to figure out exactly what is the diff between actual and expected values that caused matchesElement()  to fail.

ljharb:
Seems reasonable; but it might be tricky in that I think in `debug` we just pass through what React gives us.

finnigantime:
Took a stab at it (wrote the code before I filed this issue), let me know what you think.

Also, this might be useful inhttps://github.com/blainekasten/enzyme-matchers/pull/95

PR #961, opened 2017-05-31
[enhancement] #960 allow debug() to exclude props

finnigantime:
Implements #960. Adds options object to debug(). Uses options.ignoreProps to opt-in to ignoring props from the debug() string output. This change is backwards-compatible so existing usages of debug() are unaffected.

### DIFF
--- a/src/Debug.js
+++ b/src/Debug.js
@@ -63,27 +63,28 @@ function indentChildren(childrenStrs, indentLength) {
     : '';
 }
 
-export function debugNode(node, indentLength = 2) {
+export function debugNode(node, indentLength = 2, options = {}) {
   if (typeof node === 'string' || typeof node === 'number') return escape(node);
   if (!node) return '';
 
-  const childrenStrs = compact(childrenOfNode(node).map(n => debugNode(n, indentLength)));
+  const childrenStrs = compact(childrenOfNode(node).map(n => debugNode(n, indentLength, options)));
   const type = typeName(node);
-  const props = propsString(node);
+
+  const props = options.ignoreProps ? '' : propsString(node);
   const beforeProps = props ? ' ' : '';
-  const nodeClose = childrenStrs.length ? `</${type}>` : '/>';
   const afterProps = childrenStrs.length
     ? '>'
     : ' ';
   const childrenIndented = indentChildren(childrenStrs, indentLength);
+  const nodeClose = childrenStrs.length ? `</${type}>` : '/>';
   return `<${type}${beforeProps}${props}${afterProps}${childrenIndented}${nodeClose}`;
 }
 
-export function debugNodes(nodes) {
-  return nodes.map(node => debugNode(node)).join('\n\n\n');
+export function debugNodes(nodes, options = {}) {
+  return nodes.map(node => debugNode(node, undefined, options)).join('\n\n\n');
 }
 
-export function debugInst(inst, indentLength = 2) {
+export function debugInst(inst, indentLength = 2, options = {}) {
   if (typeof inst === 'string' || typeof inst === 'number') return escape(inst);
   if (!inst) return '';
 
@@ -93,7 +94,7 @@ export function debugInst(inst, indentLength = 2) {
 
   if (!inst.getPublicInstance) {
     const internal = internalInstance(inst);
-    return debugInst(internal, indentLength);
+    return debugInst(internal, indentLength, options);
   }
   const publicInst = inst.getPublicInstance();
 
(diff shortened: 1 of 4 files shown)


src/Debug.js
```diff
@@ -63,24 +63,29 @@ function indentChildren(childrenStrs, indentLength) {
...
   const type = typeName(node);
-  const props = propsString(node);
-  const beforeProps = props ? ' ' : '';
-  const nodeClose = childrenStrs.length ? `</${type}>` : '/>';
+
+  let props = '';
+  let beforeProps = '';
+  if (!options.ignoreProps) {
+    props = propsString(node);
+    beforeProps = props ? ' ' : '';
```
ljharb: I think i'd prefer to change line 72 to `const props = options.ignoreProps ? '' : propsString(node);` and then none of the consts need to change to lets.

src/Debug.js
```diff
@@ -63,24 +63,29 @@ function indentChildren(childrenStrs, indentLength) {
...
     : ' ';
   const childrenIndented = indentChildren(childrenStrs, indentLength);
+  const nodeClose = childrenStrs.length ? `</${type}>` : '/>';
   return `<${type}${beforeProps}${props}${afterProps}${childrenIndented}${nodeClose}`;
 }
 
-export function debugNodes(nodes) {
-  return nodes.map(node => debugNode(node)).join('\n\n\n');
+export function debugNodes(nodes, options) {
+  return nodes.map(node => debugNode(node, undefined, options)).join('\n\n\n');
```
ljharb: This should probably be `options = {}`, since it's not a required argument

src/ShallowWrapper.js
```diff
@@ -1049,10 +1049,12 @@ class ShallowWrapper {
   /**
    * Returns an HTML-like string of the shallow render for debugging purposes.
    *
+   * @param {Object} options - (Optional) Property bag of additional options.
+   * options.ignoreProps - if true, props are omitted from the string.
    * @returns {String}
    */
-  debug() {
-    return debugNodes(this.getNodes());
+  debug(options) {
```
ljharb: Since this is optional, it should be defaulted (`= {}`), so it doesn't contribute to the function's length.

src/Debug.js
finnigantime: Ok sure, I will change this to keep these using const.

src/Debug.js
finnigantime: Sounds good. I defaulted it to empty object in debugNode(), but it makes the code more consistent to default it to empty object here too. Will do.

src/ShallowWrapper.js
finnigantime: Will do. What do you mean by contributing to the function's length?

ljharb: `function (options) {}.length` is `1`, `function (options = {}) {}.length` is `0` - a function's "length" property tells you how many required arguments it has.

finnigantime:
Feedback addressed.

src/ShallowWrapper.js
finnigantime: Makes total sense! Thanks for explaining.

src/Debug.js
```diff
@@ -63,24 +63,25 @@ function indentChildren(childrenStrs, indentLength) {
...
 
-  const childrenStrs = compact(childrenOfNode(node).map(n => debugNode(n, indentLength)));
+  const childrenStrs = compact(childrenOfNode(node).map(n => debugNode(n, indentLength, options)));
   const type = typeName(node);
-  const props = propsString(node);
-  const beforeProps = props ? ' ' : '';
-  const nodeClose = childrenStrs.length ? `</${type}>` : '/>';
+
+  const props = options.ignoreProps ? '' : propsString(node);
+  const beforeProps = (options.ignoreProps || !props) ? '' : ' ';
```
ljharb: if the `ignoreProps` option is set, then `props` will be the empty string, which is falsy - i think you can just revert this line.

finnigantime: Good point. Cleaned this up. Let me know if I should squash or of it's fine to have this as an additional commit.

ljharb: Either's OK, but it's always better to rebase down to atomic commits :-)

finnigantime:
Sure, that seems reasonable. I'll be out a week, but when I return I could do that either as part of this PR or a separate one. For the CI failure, I see:
```
01 06 2017 01:48:41.764:ERROR [framework.browserify]: bundle error
01 06 2017 01:48:41.766:ERROR [framework.browserify]: Error: Cannot find module 'react/addons' from 
```

Doesn't seem related to this change. Is there a way I can try kicking off another build?

ljharb:
I think it's related to npm v5 issues. I'll rebase your branch on top of my fixes and that should do it.

finnigantime:
@ljharb Added support for ReactWrapper.

ljharb:
@finnigantime once tests are passing, i'll add other collabs as reviewers so we can get this in. It looks like they're failing because jsdom fails pre-iojs, and `mount` only works with jsdom. Perhaps the shallow and mount assertions should be split into separate `it` tests, and you can use `describeWithDOM` to wrap the `mount` ones?

finnigantime:
@nfcampos Thanks for pointing that out. Done!

Issue #967, opened 2017-06-02
Accessing element properties after unmounting and remounting throws Invariant Violation


AkshatM:
Enzyme Version 2.7.1. Windows 7 with Node 6.10.3.

Code that demonstrates the issue:

```jsx
class WrappedComponent extends Component {
  constructor(props) {
    super(props);
    this.state = {
      currentText: DEFAULT_TEXT,
    };
  }

  render() {
    return (<div className="test">{this.state.currentText}</div>);
  }
}

/* In my actual tests folder using Jest */

const renderedComponent = mount(<WrappedComponent />);
expect(renderedComponent.text()).toBe(DEFAULT_TEXT);  // passes

renderedComponent.unmount();
renderedComponent.mount();
expect(renderedComponent.text()).toBe(DEFAULT_TEXT);  // throws exception
```

This raises the following exception: 
```
Invariant Violation: findDOMNode was called on an unmounted component.
  at invariant (node_modules\fbjs\lib\invariant.js:44:15)
  at findDOMNode (node_modules\react-dom\lib\findDOMNode.js:53:54)
  at ReactWrapper.<anonymous> (node_modules\enzyme\build\ReactWrapper.js:707:47)
  at ReactWrapper.single (node_modules\enzyme\build\ReactWrapper.js:1421:25)
  at ReactWrapper.text (node_modules\enzyme\build\ReactWrapper.js:706:21)
```

# Some Investigation

It looks like (surprise, surprise) `mount()` after `unmount()` does not actually yield anything:

```jsx
console.log(renderedComponent.debug());
// correctly renders expected structure, is consistent with .html()
// <WrappedComponent currentText="">
//    <div className="test">
//        Default text to display when no data is available.
//    </div>
// </WrappedComponent>

renderedComponent.unmount();
console.log(renderedComponent.debug());  // blank, accessing .html() throws error above

renderedComponent.mount();
console.log(renderedComponent.debug());  // blank, accessing .html() throws error above
```

I would surmise from [the documentation](http://airbnb.io/enzyme/docs/api/ReactWrapper/unmount.html) that this is not the expected behaviour.

ljharb:
This seems heavily related to #956.

However, I wouldn't expect a test to ever need to re-mount an unmounted component wrapper - although I'd expect it to work (i'd expect "testing unmounting" and "testing mounting" to always be done in separate "it" blocks).

AkshatM:
[...]

PR #969, opened 2017-06-02
fixes #956 wrapper.instance() throws error after wrapper.unmount()

### LINKED ISSUE #956: wrapper.instance() throws error after wrapper.unmount()
When calling `wrapper.instance()` after `wrapper.unmount()` it throws `TypeError: component.getPublicInstance is not a function`

The use-case for accessing the React component instance after the unmount is to check that any possible teardown in `componentWillUnmount()` has been successful. In my particular use-case I want to check that an RxJS Subscription has been unsubscribed.

Code to reproduce:
```
import React from 'react';
import Rx from 'rxjs';
import { mount } from 'enzyme';

class MyComponent extends React.Component {

  constructor() {
    super();
    this.state = {
      foo: 'initial'
    };
  }

  componentDidMount() {
    this.subscription = this.props.subject.subscribe(
      value => this.setState({ foo: value })
    );
  }

  componentWillUnmount() {
    this.subscription.unsubscribe();
  }

  render() {
    return <div>{ this.state.foo }</div>;
  }
}

it('unsubscribes on unmount', () => {
  const subject = new Rx.BehaviorSubject('new value');

  const wrapper = mount(
    <MyComponent subject={subject} />,
  );
  expect(wrapper.instance()).toHaveProperty('subscription.closed', false);
  wrapper.unmount();
  expect(wrapper.instance()).toHaveProperty('subscription.closed', true);
});
```

Test output:
```
FAIL  src/ui/memo/__tests__/AfterUnmount.test.js
  ● unsubscribes on unmount

    TypeError: component.getPublicInstance is not a function
      
      at WrapperComponent.getInstance (node_modules/enzyme/build/ReactWrapperComponent.js:81:32)
      at ReactWrapper.instance (node_modules/enzyme/build/ReactWrapper.js:219:31)
      at Object.<anonymous>.it (src/ui/memo/__tests__/AfterUnmount.test.js:37:18)
      at Promise.resolve.then.el (node_modules/p-map/index.js:42:16)
      at process._tickCallback (internal/process/next_tick.js:103:7)
```

There is a workaround by using `wrapper.node` instead of `wrapper.instance()` in the last expect: `expect(wrapper.node).toHaveProperty('subscription.closed', true);`. But this feels dirty because `.node` is an undocumented feature.

Used versions:
enzyme: 2.8.2
jest: 20.0.3
react: 15.5.4
react-scripts: 1.0.5
rxjs: 5.4.0
enzyme-to-json: 1.5.1 (I don't think this affects this particular test, but I had it installed anyway)

aaroncrows:
Fixes #956

### DIFF
--- a/src/ReactWrapper.jsx
+++ b/src/ReactWrapper.jsx
@@ -208,7 +208,10 @@ class ReactWrapper {
       throw new Error('ReactWrapper::unmount() can only be called on the root');
     }
     this.single('unmount', () => {
-      this.component.setState({ mount: false });
+      this.component.setState({
+        mount: false,
+        lastRender: this.component._reactInternalInstance._renderedComponent.getPublicInstance(),
+      });
     });
     return this;
   }
--- a/src/ReactWrapperComponent.jsx
+++ b/src/ReactWrapperComponent.jsx
@@ -17,6 +17,7 @@ export default function createWrapperComponent(node, options = {}) {
       super(...args);
       this.state = {
         mount: true,
+        lastRender: null,
         props: this.props.props,
         context: this.props.context,
       };
@@ -27,7 +28,9 @@ export default function createWrapperComponent(node, options = {}) {
     }
     getInstance() {
       const component = this._reactInternalInstance._renderedComponent;
-      const inst = component.getPublicInstance();
+      const { mount, lastRender } = this.state;
+
+      const inst = mount ? component.getPublicInstance() : lastRender;
       if (inst === null) {
         return component._instance;
       }
--- a/test/ReactWrapper-spec.jsx
+++ b/test/ReactWrapper-spec.jsx
@@ -208,6 +208,17 @@ describeWithDOM('mount', () => {
     });
   });
 
+  describe('.instance()', () => {
+    it('should return instance after unmount', () => {
+      const wrapper = mount(<div />);
+      const instance = wrapper.instance();
+
+      wrapper.unmount();
+
+      expect(wrapper.instance()).to.equal(instance);
+    });
+  });
+
   describe('.contains(node)', () => {
     it('should allow matches on the root node', () => {
       const a = <div className="foo" />;


src/ReactWrapper.jsx
```diff
@@ -208,7 +208,10 @@ class ReactWrapper {
       throw new Error('ReactWrapper::unmount() can only be called on the root');
     }
     this.single('unmount', () => {
-      this.component.setState({ mount: false });
+      this.component.setState({
+        mount: false,
+        lastRender: this.component._reactInternalInstance._renderedComponent.getPublicInstance(),
```
ljharb: i'm concerned that this could overwrite `this.state.lastRender` inside the component. Shouldn't we find a place to store this info that the component code can't interact with?

aaroncrows: That makes sense. It was my first thought to put it there because the state was being modified for `mount` so it seemed like the simplest way to handle it. It could also potentially be a property of the component itself something like:
`this.component._lastRender = ...`

ljharb: What about if we store it on `this.component._reactInternalInstance`, since that's not something component code can generally access?

aaroncrows: I gave that a shot and it's not working. What I think is happening is that `setState()` is triggering a rerender which is replacing `_reactInternalInstance`. I can't set the property immediately after  `setState()` because it's asynchronous, and apparently if a component unmounts callbacks passed to `setState()` don't fire.

ljharb: Another alternative is using a `WeakMap` with `this.component` as the key and the instance as the value.

aaroncrows: I like that idea but it leaves the question as to where to put it. Because of what the lifecycle does to `_reactInternalInstance` and the fact that `unmount()` is on `ReactWrapper` but `getInstance()` is on `ReactWrapperComponent` it seems to me like the only easy options are to put a property on `this.component`, which could certainly be a `WeakMap`, or to put it in the components state.

Without entirely knowing how they're being used elsewhere in the library there doesn't seem to be a particularly good reason that `getInstance()` and `getWrapperComponent()` are on the `ReactWrapperComponent` at all. They aren't a part of the public api and only seem to be getting used by `ReactWrapper`. With that I suppose another option would be to factor those methods out of `ReactWrapperComponent` which would allow for storing a reference to the last render on `ReactWrapper` instead. It would also make that same reference available to `mount()` which might help address #967.

I'd be happy to give that a shot but it seems a lot more involved and might reach a bit beyond a quick bug fix.

ljharb: Sorry I wasn't being clear. The `WeakMap` would be a module-level variable accessible to all instances via closure - ie, `const instanceMap = new WeakMap(); class ShallowWrapper { … }`

src/ReactWrapperComponent.jsx
```diff
@@ -17,6 +17,7 @@ export default function createWrapperComponent(node, options = {}) {
       super(...args);
       this.state = {
         mount: true,
+        lastRender: null,
```
ljharb: i think this can be removed?

src/ReactWrapperComponent.jsx
```diff
@@ -27,6 +28,7 @@ export default function createWrapperComponent(node, options = {}) {
     }
     getInstance() {
       const component = this._reactInternalInstance._renderedComponent;
+
```
ljharb: probably this too

src/ReactWrapper.jsx
```diff
@@ -204,10 +208,13 @@ class ReactWrapper {
    * @returns {ReactWrapper}
    */
   unmount() {
+    const { component } = this;
```
ljharb: might as well destructure `root` out also

src/ReactWrapper.jsx
```diff
@@ -204,10 +208,13 @@ class ReactWrapper {
...
+    const inst = component._reactInternalInstance._renderedComponent.getPublicInstance();
     if (this.root !== this) {
       throw new Error('ReactWrapper::unmount() can only be called on the root');
     }
     this.single('unmount', () => {
+      instanceMap.set(component, inst);
       this.component.setState({ mount: false });
```
ljharb: just `component`?

src/ReactWrapper.jsx
```diff
@@ -204,10 +208,13 @@ class ReactWrapper {
    * @returns {ReactWrapper}
    */
   unmount() {
+    const { component } = this;
+    const inst = component._reactInternalInstance._renderedComponent.getPublicInstance();
```
ljharb: Instead of repeating this logic, could we use `this.instance()` here?

src/ReactWrapper.jsx
aaroncrows: Yeah it's destructured out at the top of the function. The next line should actually be `component` too.

src/ReactWrapper.jsx
```diff
@@ -175,6 +177,8 @@ class ReactWrapper {
     if (this.root !== this) {
       throw new Error('ReactWrapper::instance() can only be called on the root');
     }
+    if (!this.component.state.mount) return instanceMap.get(this.component);
```
ljharb: I wonder if this should be `return this.component.getInstance() || instanceMap.get(this.component);`, to cover "not yet mounted", and to always prioritize the component's getInstance method?

aaroncrows: As it stands that would end up hitting the line that threw the error before in the components `getInstance()`. It could be added pretty easily by making `getInstance()` return `null` or something if the rendered component didn't have `getPublicInstance()`.

ljharb: That seems like a reasonable companion change anyways; `getInstance` shouldn't throw if it doesn't have an instance, it should just not return one.

aaroncrows: Doing so seems to cause it to fail in React 13. I'm not sure exactly why, though. Might have something to do with a change in the way empty components were handled?

ljharb: It's worth investigating :-)

aaroncrows:
This moves the check for the component's state into the components `getInstance()` method. It prioritizes the method, I'm not sure if it covers for "not yet mounted." What I was running into was that React changed the way it handled empty components after 13. I couldn't find a reliable way to check whether or not the current instance was React 13s version of an empty component without breaking a whole bunch of other tests.

ljharb:
@aaroncrows looks like this might need a fresh rebase, not sure if the test failure is a fluke or not

aaroncrows:
I believe the test failure is because one of the tests travis is running uses PhantomJS and the current version (2.1) doesn't fully support ES2015 so it doesn't know what a `WeakMap` is. All the tests I have use jsdom so I'm not sure where the inconsistency is. Apparently there's a beta out for Phantom 2.5 which supports it and there's also a polyfill option via babel.

ljharb:
ah, WeakMap can't actually be polyfilled; we should probably detect if WeakMap is available, and fall back to `Map` if not?

aaroncrows:
It appears Phantom 2.1 doesn't support `Map` either. The failing test is in the karma-webpack example. For what it's worth introducing the babel-polyfill to the karma configuration does get it to recognize `WeakMap`. I suppose a plain object could be used, but that has the drawbacks of needing a unique identifier for the key and holding references to objects that don't exist anywhere else. Or there's still the option of finding somewhere on the component to put the last render.

At some point tried to close over a reference to the last render in the `createWrapperComponent` function but I couldn't really get to it without introducing lifecycle methods to the component. I could look into that further, though.

ljharb:
I'm really loath to rely on babel-polyfill - we could add es6-shim to the karma build to get Map to work, though.

ljharb:
@aaron-j-krause hi! i know it's been a long while. If you're willing to update this PR and try again - https://npmjs.com/side-channel will probably be helpful.

JesusHV (approved):

PR #983, opened 2017-06-13
Change componentDidMount to componentWillMount in shallow options api doc

calinoracation:
In shallow rendering we do not call `componentDidMount`, this fixes the docs to reflect that.


lorenzos:
I'm confused. With `disableLifecycleMethods` do `componentDidMount` get called or not? In the comment here it's said it's not, but the commit itself clearly change the docs by making them say that is `componentWillMount` to not be called, not mentioning `componentDidMount` anymore...

ljharb:
DidMount should never be called when using shallow.

palaniichukdmytro:
But my case the componentDidMount is calledOnce with shallow, I use react16  and enzyme 3.x,

ljharb:
@palaniichukdmytro please file a new issue

PR #984, opened 2017-06-14
Make it consistent

joaojeronimo:


nfcampos:
@ljharb you might be interested to know that this is the result of running all code inside our markdown through eslint (*after* disabling some rules that arguably don't really apply to code snippets)
screen shot 2017-06-15 at 22 33 56

nfcampos:
so do we really want to fix these 6 semicolons but ignore the other +900 deviations from the style guide?

ljharb:
Let's fix one rule at a time, but yes, they should all be fixed.

It'd be great to add this check to the tests, so that incorrect code can be kept out of examples in the first place.

nfcampos:
@ljharb its trivial to add this check, the problem is we can't add this check without first fixing all existing errors

ljharb:
If you can point me to a branch I can take a look at that.

nfcampos:
here it is https://github.com/airbnb/enzyme/compare/eslint-markdown?expand=1

I've disabled some rules that arguably don't make too much sense for snippets (bringing the errors down to ~500)

nfcampos:
one thing to bear in mind is that eslint-plugin-markdown does not support `eslint --fix` so any fixes will either be manual or require some custom code

ljharb:
Thanks, I'll do it manually this weekend; I can fix 500 eslint errors manually with my eyes closed :-p

nfcampos:
eheh good luck :)

PR #988, opened 2017-06-17
lint our markdown

ljharb:
From https://github.com/airbnb/enzyme/pull/984#issuecomment-309115597

Now our docs will be linted.


.eslintrc-markdown
```diff
@@ -0,0 +1,21 @@
...
+  "root": true,
+  "plugins": [
+    "markdown"
+  ],
+  "rules": {
+    "class-methods-use-this": 0,
+    "import/no-unresolved": 0,
+    "import/no-extraneous-dependencies": 0,
+    "import/extensions": 0,
+    "max-len": 0,
```
nfcampos: why disable this one?

nfcampos:
only question is for me is what if someone adds a markdown with an extension other .md? should we do anything for that?

.eslintrc-markdown
ljharb: I don't really like this rule in general; it's basically a poor proxy for limiting complexity. All the lines it warned on weren't worth arbitrarily splitting up.

ljharb:
I'm not sure if there's anyway we can prevent that short of manual review.

Issue #364, opened 2016-05-05
Unable to change the value of form inputs


MoonTahoe:
First... let me mention that I love using enzyme.  Great work.  I am currently putting togther training materials for React that will use enzyme for testing.

#### Problem
I am having a problem testing a form.  Specifically changing input values on the form.  

`  wrapper.ref('first').simulate('change', {target: {value: 'David'}});  ` is not working
`  wrapper.find("input#first").simulate('change', {target: {value: 'David'}});  ` is not working

I have created a reduced test case that demonstrates the problem.  The output for this example looks like:

#### Output
```
  Testing a form
first name: Mike
last name: Tyson
    ✓ can fill out the form
```  

No matter what the default __Mike Tyson__ is always the name, despite the fact that I am chanign it.  Why does this happen?  Is there another way to change the values for the form?

#### Test
```javascript
import { Component } from 'react'
import { expect } from 'chai'
import { mount } from 'enzyme'

class TheForm extends Component {

    submit() {
        console.log('first name:', this.refs.first.value);
        console.log('last name:', this.refs.last.value);
    }

    render() {
        return (
            <form action="javascript:void(0)"
                  onSubmit={this.submit.bind(this)}>

                <div>
                    <input ref="first"
                           type="text"
                           placeholder="your first name..."
                           required="required"
                           defaultValue="Mike"/>
                </div>

                <div>
                    <input ref="last"
                           type="text"
                           placeholder="your last name..."
                           required="required"
                           defaultValue="Tyson" />
                </div>

                <button>ADD</button>

            </form>
        );
    }

}

describe("Testing a form", () => {

    it("can fill out the form", () => {
        const wrapper = mount(<TheForm />);
        wrapper.ref('first').simulate('change', {target: {value: 'David'}});
        wrapper.ref('last').simulate('change', {target: {value: 'Blane'}});
        wrapper.find('form').simulate('submit');
    });

});

```

also... wrapper.find('form').simulate('submit') works, but clicking the button does not.  Even though clicking the button submits the form in the browser.

Aweary:
>also... wrapper.find('form').simulate('submit') works, but clicking the button does not. Even though clicking the button submits the form in the browser.

FYI this has been reported previously here: https://github.com/airbnb/enzyme/issues/308

Aweary:
I'm able to reproduce your issue with the current release of enzyme, I'll look into how this is expected to work. For what it's worth, you can use controlled components to handle the input state, which works as you expected:

```js
[...]

PR #995, opened 2017-06-21
Problem with simulating a "change" with a mock event object

nerfologist:
Based on enzyme's doc we should be able to use simulate()'s second
argument this way:

simulate(event [, mock])
- mock (Object [optional]): A mock event object that will be merged with
the event object passed to the handlers.

This test demonstrates a problem trying to simulate a change event on
an input field: we are unable to override the `target` property.

The same goes for the `currentTarget` property (not demonstrated in this test case).

This PR adds a failing tests case which demonstrates the issue.

Thanks in advance for your great work! 👍

### DIFF
--- a/test/ReactWrapper-spec.jsx
+++ b/test/ReactWrapper-spec.jsx
@@ -1273,6 +1273,23 @@ describeWithDOM('mount', () => {
       expect(spy.args[0][0].someSpecialData).to.equal('foo');
     });
 
+    it('should let us override the event object', () => {
+      const spy = sinon.spy();
+      class Foo extends React.Component {
+        render() {
+          return (
+            <input type="text" onChange={spy} value="foo" />
+          );
+        }
+      }
+
+      const wrapper = mount(<Foo />);
+
+      wrapper.simulate('change', { target: { value: 'bar' } });
+      expect(spy.calledOnce).to.equal(true);
+      expect(spy.args[0][0].target.value).to.equal('bar');
+    });
+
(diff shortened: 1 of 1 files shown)


aweary:
This test appears to be failing in versions >=0.14. It's not actually demonstrating the issue that you believe it is either: even if you remove the event data from the `simulate` call, it fails with the same error.

```
TypeError: Cannot read property 'value' of null
      at Context.<anonymous> (test.js:37:14)
```

The reason for this is that events were not being persisted in `ReactTestUtils` before React 15. That means that when the event finished dispatching it will be released back to the pool and call a destructor which [nullifies every property from the event interface](https://github.com/facebook/react/blob/b1b4a2fb252f26fe10d29ba60d85ff89a85ff3ec/src/renderers/shared/shared/event/SyntheticEvent.js#L175-L187).

This includes the `target` property. That's why this test is failing, because the event isn't persisted and it's try to access `value` on `target`, which has been set to `null`. Note that this is specifically because you're attempting to assert on the value in the event itself, not of the input. I would recommend you instead query for the `input` and check its value, which should sidestep this problem (and also be a better test)

As of https://github.com/facebook/react/pull/6380 these events are persisted. That means this should work fine in React 15+

Since this isn't something Enzyme can control, I'm going to close this out. I recommend following the solutions outlined in https://github.com/airbnb/enzyme/issues/364 to handle this.

Hope that clarifies things!

nerfologist:
Hi @aweary, thanks for taking the time to clarify this. it all makes sense now 👍

PR #997, opened 2017-06-21
Lexicographically sort methods in Shallow.md

mcnichol:
The methods are sorted for ease of use to be used as coherent reference

Sorting Shallow.md lexicographically to be consistent with ordered API method listing in `/docs/README.md`

### DIFF
--- a/docs/api/shallow.md
+++ b/docs/api/shallow.md
@@ -54,23 +54,20 @@ is not called on the component, and `componentDidUpdate` is not called after
 
 ## ShallowWrapper API
 
-#### [`.find(selector) => ShallowWrapper`](ShallowWrapper/find.md)
-Find every node in the render tree that matches the provided selector.
+#### [`.at(index) => ShallowWrapper`](ShallowWrapper/at.md)
+Returns a wrapper of the node at the provided index of the current wrapper.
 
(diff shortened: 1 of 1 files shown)


mcnichol:
I feel there has to be something worth a laugh regarding the 99 line commit there...

*Committed 99 lines and a trailing blank line isn't one.*

mcnichol:
I've worked with engineers who preferred ordering their Java annotations based on the level of abstraction to the class they annotate, so I can respect the view of *importance*.

I just felt as someone learning the methods, it looked like a mistake that was made due to being sorted one way in `/docs`README.md` and another in the `docs/api/shallow.md`

ljharb:
Let's not close so hastily; I think it's worth discussing :-)

mcnichol:
Okay, I'll leave it open.  Happy to follow your lead with it :)

For what it's worth I did see myself jumping around a little bit, it wasn't to the point where I found myself extremely frustrated, only enough for me to throw it into my terminal and `parse | sort` it and then read the docs from my terminal.  

I had figured I was already that far so might as well toss it over if anyone else is interested.  I do understand the rationale between sorting by importance and have seen it drive some really good conversation.

PR #998, opened 2017-06-21
Update uuid 2.x -> 3.x

lencioni:
@airbnb/webinfra

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "object.entries": "^1.0.4",
     "object.values": "^1.0.4",
     "prop-types": "^15.5.10",
-    "uuid": "^2.0.3"
+    "uuid": "^3.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


tstirrat15:
@ljharb this has the potential to result in a broken build of enzyme.

The `3.0.0` version of `uuid` doesn't support the `uuid/v4` import style, and will throw an error on import. I ran into this problem in my own project. `3.1.0` doesn't experience this problem.

I'm not sure how `npm` determines which version of a package to pull when resolving dependencies beyond the first layer - this may be to blame?

lencioni:
Ah my bad. It looks like that was [added in 3.0.1](https://unpkg.com/uuid@3.0.1/). I'll open a PR to fix.

ljharb:
Thanks - it would only result in 3.0.0 if something else in the dep graph pinned to that version, but it's a good idea to fix it asap :-)

PR #1007, opened 2017-06-26
Refactor enzyme to use Adapters, initial React 16 support

lelandrichardson:
to: @ljharb @aweary @nfcampos @blainekasten 
cc: @sebmarkbage @developit @bvaughn

This PR is effectively a rewrite of Enzyme's internal machinery, per the discussion outlined in https://github.com/airbnb/enzyme/pull/742

I attempted to make this as small of a breaking change as pragmatically possible, and minimize the amount our current unit tests needed to be changed across the refactor.

## Motivation

This PR attempts to address a handful of pain points that Enzyme has been
subject to for quite a while.

The desired results are the following:

### Cleaner code, easier maintenance, less bug prone.

By standardizing on a single tree specification, the implementation of Enzyme would no longer have
to take into account the matrix of supported structures and nuanced differences between different
versions of React, as well as to some extent the differences between `mount` and `shallow`.

### Additional libraries can provide compatible adapters

React API-compatible libraries such as `preact` and `inferno` will be able to provide adapters to Enzyme
for their corresponding libraries, and be able to take full advantage of Enzyme's APIs.

### Better user experience (ie, bundlers won't complain about missing deps)

Enzyme has had a long-standing issue with static-analysis bundlers such as Webpack and Browserify because
of our usage of internal React APIs. With this change, this is minimized to the adapter implementations only, which will be imported separately, so bundlers will not have any of these problems.

### Standardization and interopability with other tools

With an agreed on tree format, other tools can start to use and
understand this format as well. Standardization is a good thing, and could allow tools to be built that maybe
don't even exist yet.

## Things left to do before merging this PR:

- [ ] clarify changes in documentation
- [ ] complete migration guide
- [x] add in support for keys

## Things left to do before publishing Enzyme 3.0:

1. Followup PR moving the adapters into their own packages
2. Publish alpha(s), get enzyme users to try it out and report back any issues (Airbnb at least will have to do this)
3. Implement new configurable top level API (debatable whether or not this is needed to publish 3.0)


docs/future/migration.md
```diff
@@ -0,0 +1,94 @@
...
+dont have "instances" associated with them, which means there are a couple of things
+that we used to be able to do with enzyme + SFCs that will just no longer work.
+
+We could fix a lot of this if there was a reliable way to get from an SFC "fiber" to
+the corresponding DOM element that it renders.
+
+## Strings vs. Numbers
+
+React 16 converts numbers to strings very early on. we can't change this. this will change
+some behavior in enzyme but we are considering this the "correct" behavior.
```
ljharb: I'm pretty concerned about this behavior; this does not seem correct to me, and being unable to make assertions on numeric props seems like a showstopper.

withDom.js
```diff
@@ -1,3 +1,5 @@
+require('raf/polyfill');
```
ljharb: should we just use airbnb-browser-shims here? (at the end, after jsdom has ran)

test/_helpers/react-compat.js
```diff
@@ -4,11 +4,11 @@
   import/prefer-default-export: 0,
 */
 
-import { REACT155 } from '../../src/version';
+import { REACT155, REACT16 } from '../../src/version';
 
 let createClass;
 
-if (REACT155) {
+if (REACT155 || REACT16) {
```
ljharb: we may just want a constant for "15.5 or higher"

test/ShallowWrapper-spec.jsx
```diff
@@ -4174,6 +4231,7 @@ describe('shallow', () => {
     it('should have updated output after child prop callback invokes setState', () => {
       const wrapper = shallow(<Test />);
       wrapper.find(Child).props().callback();
+      wrapper.update(); // TODO(lmr): this is a breaking change...
```
ljharb: is the breaking change here that `setState` doesn't automatically rerender?

test/ShallowWrapper-spec.jsx
```diff
@@ -3559,7 +3613,8 @@ describe('shallow', () => {
     });
   });
 
-  describe('.key()', () => {
+  // TODO(lmr): keys aren't included in RST Nodes. We should think about this.
+  describe.skip('.key()', () => {
```
ljharb: we should unskip this so tests do fail, in the meantime.

test/ShallowWrapper-spec.jsx
```diff
@@ -86,6 +90,22 @@ describe('shallow', () => {
...
+
+      itIf(REACT16, 'is not introspectable through context API', () => {
+        const SimpleComponent = (props, context) => (
+          <div>{context.name}</div>
+        );
+        SimpleComponent.contextTypes = { name: PropTypes.string };
+
+        const wrapper = shallow(<SimpleComponent />, { context });
+
+        expect(() => wrapper.context()).to.throw(Error,
```
ljharb: if the entire invocation can't fit on one line, all args need to be on a line by themselves - ie, `Error` needs to be on a separate line. same on line 105.

test/ShallowWrapper-spec.jsx
```diff
@@ -86,6 +90,22 @@ describe('shallow', () => {
         expect(wrapper.context().name).to.equal(context.name);
         expect(wrapper.context('name')).to.equal(context.name);
       });
+
+      itIf(REACT16, 'is not introspectable through context API', () => {
```
ljharb: why is this the case? if you can inspect the props passed to an SFC, why not the context?

test/ReactWrapper-spec.jsx
```diff
@@ -2163,12 +2201,21 @@ describeWithDOM('mount', () => {
...
-        expect(wrapper.hasClass('some-long-string')).to.equal(true);
-        expect(wrapper.hasClass('FoOo')).to.equal(true);
+        expect(wrapper.hasClass('foo')).to.equal(false);
+        expect(wrapper.hasClass('bar')).to.equal(false);
+        expect(wrapper.hasClass('baz')).to.equal(false);
+        expect(wrapper.hasClass('some-long-string')).to.equal(false);
+        expect(wrapper.hasClass('FoOo')).to.equal(false);
         expect(wrapper.hasClass('doesnt-exist')).to.equal(false);
+
+        // NOTE(lmr): the fact that this no longer works is a semantically
```
ljharb: can you elaborate on why these assertions can't be true?

src/render.jsx
```diff
@@ -1,6 +1,16 @@
...
-import { renderToStaticMarkup } from './react-compat';
+
+import configuration from './configuration';
+
+function getAdapter(options) {
+  if (options.adapter) {
+    return options.adapter;
+  }
+  const adapter = configuration.get().adapter;
+  // TODO(lmr): warn about no adapter being configured
```
ljharb: should this warn, or hard-throw?

src/adapters/elementToTree.js
```diff
@@ -0,0 +1,22 @@
...
+  const { type, props } = el;
+  const { children } = props;
+  let rendered = null;
+  if (Array.isArray(children)) {
+    rendered = flatten(children, true).map(elementToTree);
+  } else if (children !== undefined) {
+    rendered = elementToTree(children);
+  }
+  return {
+    nodeType: typeof type === 'string' ? 'host' : 'class',
```
ljharb: i assume "host" means "DOM element" for web; i'm confused tho, why would a non-host be a "class"? SFCs aren't classes, and really `React.createClass` components aren't either.

ReactFourteenAdapter defines "function" also; should that be available here?

src/adapters/ReactThirteenAdapter.js
```diff
@@ -0,0 +1,215 @@
...
+      },
+    };
+  }
+
+  // Provided a bag of options, return an `EnzymeRenderer`. Some options can be implementation
+  // specific, like `attach` etc. for React, but not part of this interface explicitly.
+  // eslint-disable-next-line class-methods-use-this, no-unused-vars
+  createRenderer(options) {
+    switch (options.mode) {
+      case 'mount': return this.createMountRenderer(options);
```
ljharb: should there be an exposed list of supported renderer modes on each adapter, so it can be reflected on?

src/adapters/ReactThirteenAdapter.js
```diff
@@ -0,0 +1,215 @@
...
+      },
+      batchedUpdates(fn) {
+        return withSetStateAllowed(() => batchedUpdates(fn));
+      },
+    };
+  }
+
+  createStringRenderer(/* options */) {
+    return {
+      render(el /* , context */) {
```
ljharb: no context for server rendering?

src/adapters/ReactThirteenAdapter.js
```diff
@@ -0,0 +1,215 @@
...
+}
+
+SimpleWrapper.propTypes = { node: PropTypes.node.isRequired };
+
+class ReactThirteenAdapter extends EnzymeAdapter {
+  createMountRenderer(options) {
+    const domNode = options.attachTo || global.document.createElement('div');
+    let instance = null;
+    return {
+      render(el/* , context */) {
```
ljharb: also no context?

src/adapters/ReactThirteenAdapter.js
```diff
@@ -0,0 +1,215 @@
...
+  render() {
+    return this.props.node || null;
+  }
+}
+
+SimpleWrapper.propTypes = { node: PropTypes.node.isRequired };
+
+class ReactThirteenAdapter extends EnzymeAdapter {
+  createMountRenderer(options) {
+    const domNode = options.attachTo || global.document.createElement('div');
```
ljharb: if `createMountRenderer` in a non-dom env is going to throw, can we have an explicit throw instead of accessing a property off of undefined?

src/adapters/ReactSixteenAdapter.js
```diff
@@ -0,0 +1,265 @@
...
+        throw new Error('Unrecognized mode');
+    }
+  }
+
+  // converts an RSTNode to the corresponding JSX Pragma Element. This will be needed
+  // in order to implement the `Wrapper.mount()` and `Wrapper.shallow()` methods, but should
+  // be pretty straightforward for people to implement.
+  // eslint-disable-next-line class-methods-use-this, no-unused-vars
+  nodeToElement(node) {
+    if (!node || typeof node !== 'object') return null;
```
ljharb: can a node not be a function?

src/adapters/ReactSixteenAdapter.js
```diff
@@ -0,0 +1,265 @@
+/* eslint no-use-before-define: 0 */
+import React from 'react';
+import ReactDOM from 'react-dom';
+import ReactDOMServer from 'react-dom/server';
+// import TestRenderer from 'react-test-renderer';
```
ljharb: let's make sure to do a sweep and remove all unneeded commented-out code before merging :-)

src/adapters/EnzymeAdapter.js
```diff
@@ -0,0 +1,24 @@
+function unimplementedError(methodName, classname) {
+  return new Error(
+    `${methodName} is a required method of ${classname}, but was not implemented.`,
+  );
+}
+
+class EnzymeAdapter {
```
ljharb: i don't see anywhere we're checking `instanceof EnzymeAdapter` - it seems like we'd want to enforce people use this interface.

src/ShallowWrapper.js
```diff
@@ -1078,13 +1100,14 @@ class ShallowWrapper {
   dive(options = {}) {
     const name = 'dive';
     return this.single(name, (n) => {
-      if (isDOMComponentElement(n)) {
+      if (n && n.nodeType === 'host') {
         throw new TypeError(`ShallowWrapper::${name}() can not be called on DOM components`);
```
ljharb: should this phrase - "DOM components" - come from the adapter, and not from enzyme itself?

src/ShallowWrapper.js
```diff
@@ -684,6 +699,11 @@ class ShallowWrapper {
         'a context option',
       );
     }
+    if (this.instance() === null) {
+      throw new Error(
+        'ShallowWrapper::context() can only be called on class components as of React 16',
```
ljharb: this seems like an error that should come from the react 16 adapter

src/ReactWrapper.jsx
```diff
@@ -243,11 +277,14 @@ class ReactWrapper {
...
    */
-  setProps(props, callback = undefined) {
+  setProps(props, callback = noop) {
     if (this.root !== this) {
       throw new Error('ReactWrapper::setProps() can only be called on the root');
     }
-    this.component.setChildProps(props, callback);
+    this.component.setChildProps(props, () => {
+      this.update();
+      callback();
```
ljharb: we should probably check that it's a function here, just like on line 309

docs/future/migration.md
gaearon: It's not about numeric props per se. It's about numeric children. `<div>{42}</div>` does not make a lot of sense on its own because it would get coerced to string anyway. Now we do this early.

docs/future/compatibility.md
```diff
@@ -0,0 +1,184 @@
+# Enzyme Adapter & Compatibility Proposal
+
+
+## Motivation
+
+This proposal is attempting to address a handful of pain points that Enzyme has been
+subject to for quite a while.  This proposal has resulted mostly [#715](https://github.com/airbnb/enzyme/issues/715),
```
nfcampos: the second sentence seems to be missing some word somewhere, maybe "from" after "mostly"

docs/future/compatibility.md
```diff
@@ -0,0 +1,184 @@
...
+
+```js
+import Enzyme from 'enzyme';
+import ThirdPartyEnzymeAdapter from 'third-party-enzyme-adapter';
+
+Enzyme.configure({ adapter: ThirdPartyEnzymeAdapter });
+
+```
+
+Additionally, each wrapper Enzyme exposes will allow for an overriding `adapter` option that will use a
```
nfcampos: in this sentence we're using `wrapper` to mean 2 different things, first it means the factory for wrappers, 2nd it means a thing returned by one the factories, can we rename one of them, or both?

docs/future/compatibility.md
```diff
@@ -0,0 +1,184 @@
...
+given adapter for just that wrapper:
+
+```jsx
+import { shallow } from 'enzyme';
+import ThirdPartyEnzymeAdapter from 'third-party-enzyme-adapter';
+
+shallow(<Foo />, { adapter: ThirdPartyEnzymeAdapter });
+```
+
+Enzyme will build adapters for all major versions of React since React 0.13, though will deprecate
```
nfcampos: why are we maintaining compatibility with 4 major react versions at the same time?

nfcampos: or maybe this will be released before we release support for react 16?

docs/future/compatibility.md
```diff
@@ -0,0 +1,184 @@
...
+```js
+import React13Adapter from 'enzyme-adapter-react-13';
+import React14Adapter from 'enzyme-adapter-react-14';
+import React15Adapter from 'enzyme-adapter-react-15';
+// ...
+```
+
+### Validation
+
+Enzyme will provide an `validate(node): Error?` method that will traverse down a provided `RSTNode` and
```
nfcampos: Enzyme will provide a~n~

nfcampos: should this be run when `Enzyme.configure` is called?
To avoid users ever seeing failing tests because of a broken adapter

docs/future/migration.md
```diff
@@ -0,0 +1,94 @@
+# Migration Guide (for React 0.13 - React 15.x)
+
+
+## Root Wrapper
+
+The initially returned wrapper used to be around the element passed
+into the `mount` API, and for `shallow` it was around the root node of the rendered output of the element passed in. After the upgrade, the
+two APIs are now symmetrical, starting off
```
nfcampos: this sentence is not complete

docs/future/migration.md
```diff
@@ -0,0 +1,94 @@
...
+const wrapper = mount(<Foo outer={x} />);
+```
+
+```js
+expect(wrapper.props()).to.deep.equal({ outer: x });
+```
+
+## Refs
+
+Refs no longer return a "wrapper". They return what the ref would actually be.
```
nfcampos: or a composite component instance?

docs/future/migration.md
```diff
@@ -0,0 +1,94 @@
...
+
+we need to keep in mind that `getElement()` will no longer be referentially equal to what it was before.
+
+## Updates are required
+
+```
+wrapper.find('.async-btn').simulate('click');
+setImmediate(() => {
+  // this didn't used to be needed
+  wrapper.update(); // TODO(lmr): this is a breaking change...
```
nfcampos: this is a pretty big breaking change that we should try very hard to avoid

docs/future/migration.md
```diff
@@ -0,0 +1,94 @@
...
+  expect(wrapper.find('.show-me').length).to.equal(1);
+  done();
+});
+```
+
+
+
+
+## Enzyme.use
```
nfcampos: what's this? Enzyme.configure?

docs/future/migration.md
```diff
@@ -0,0 +1,94 @@
...
+
+# Left to do:
+
+- move adapters into standalone packages
+- x create Enzyme.use API
+- x create api to inject adapter per use
+- x make sure all react dependence is moved into adapters
+- x make tests run for all adapters
+- export tests for 3rd party adapters to run
+- check the targetApiVersion returned by the adapter and use the semver library
```
nfcampos: use it for what? maintain a map of react api versions to available api methods? and warn/throw accordingly if an unavailable method is used?

package.json
```diff
@@ -21,13 +21,14 @@
     "test:watch": "mocha --recursive --watch test",
     "test:karma": "karma start",
     "test:env": "sh ./example-test.sh",
-    "test:all": "npm run react:13 && npm run test:only && npm run react:14 && npm run test:only && npm run react:15.4 && npm run test:only && npm run react:15 && npm run test:only",
+    "test:all": "npm run react:13 && npm run test:only && npm run react:14 && npm run test:only && npm run react:15.4 && npm run test:only && npm run react:15 && npm run test:only && npm run react:16 && npm run test:only",
     "clean-local-npm": "rimraf node_modules/.bin/npm node_modules/.bin/npm.cmd",
     "react:clean": "npm run clean-local-npm && rimraf node_modules/react node_modules/react-dom node_modules/react-addons-test-utils node_modules/react-test-renderer && npm prune",
```
nfcampos: add create-react-class

docs/future/compatibility.md
ljharb: One of the main benefits of the adapter pattern is that we don't have to *ever* drop support for a React major version, since the cost of maintaining the adapter is so minimal.

docs/future/migration.md
ljharb: @gaearon Since children propType validators can check the type - including ensuring that they're numbers - it does in fact make sense imo. In other words, the coercion needs to happen after propTypes run - and being able to make assertions on the *actual* value, not the final string value, is what's important in enzyme.

package.json
ljharb: doesn't create-react-class not need to be removed because it'd jut be ignored in older versions?

nfcampos: I hope that’s true yes, but I’d rather test on environments closer to what an actual person eg. using react 13 would have (ie. no create react class installed)

docs/future/compatibility.md
nfcampos: Good point, I agree

docs/future/migration.md
gaearon: This is related to host components only. Could you please clarify how propTypes are related to Enzyme (or to host components)? A code example where this matters would help.

setupAdapters.js
```diff
@@ -0,0 +1,25 @@
...
+ * version of React is loaded, and configures enzyme to use the right
+ * corresponding adapter.
+ */
+const Version = require('./src/version');
+const Enzyme = require('./src');
+
+let Adapter = null;
+
+if (Version.REACT013) {
+  Adapter = require('./src/adapters/ReactThirteenAdapter');
```
milesj: What if enzyme went the lerna approach and published individual packages for each version? So if I only wanted react15, I would install the `enzyme` and `enzyme-adapter-react15` packages?

nfcampos: isn't that the plan?

milesj: Perhaps? Doesn't look to be from this PR.

ljharb: It's not part of this PR, but it's absolutely the plan. See "Followup PR moving the adapters into their own packages" in the OP.

docs/future/migration.md
ljharb: @gaearon ah, thanks! That wasn't clear to me from the description. If it's only on host/DOM components, then it really doesn't matter much, and the ability to assert on non-string props isn't super important.

That said, though, I'd expect the following to pass:
```jsx
function Foo({ num }) {
  return <div><span children={num} /></div>;
}
const wrapper = shallow(<Foo num={42} />);
assert(wrapper.find('span').props().children === 42);
```
It doesn't really seem like there's a good reason that shouldn't work - the resulting way it renders is host-specific, and shouldn't really be the concern of the shallow renderer imo.

setupAdapters.js
milesj: Ah thanks, missed that sentence.

docs/future/compatibility.md
```diff
@@ -0,0 +1,184 @@
...
+
+The desired results of this proposal are the following:
+
+1. Cleaner code, easier maintenance, less bug prone.
+
+By standardizing on a single tree specification, the implementation of Enzyme would no longer have
+to take into account the matrix of supported structures and nuanced differences between different
+versions of React, as well as to some extent the differences between `mount` and `shallow`.
+
+2. Additional libraries can provide compatible adapters
```
lelandrichardson: true. i usually avoid this because when reading it in plain text it gets confusing

docs/future/migration.md
```diff
@@ -0,0 +1,94 @@
...
+two APIs are now symmetrical, starting off
+
+```js
+const x = 'x';
+const Foo = props => <div inner={props.outer} />
+const wrapper = mount(<Foo outer={x} />);
+```
+
+```js
+expect(wrapper.props()).to.deep.equal({ outer: x });
```
lelandrichardson: yes. this document isn't done yet

docs/future/migration.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+## Refs
+
+Refs no longer return a "wrapper". They return what the ref would actually be.
+
+
+## Keys
+
+keys no longer work? we should maybe fix this in the spec...
```
lelandrichardson: there already is one in the PR description

docs/future/migration.md
```diff
@@ -0,0 +1,94 @@
...
+
+# Migration Guide (for React 16)
+
+## Stateless Functional Components
+
+SFCs actually go down a different code path in react 16, which means that they
+dont have "instances" associated with them, which means there are a couple of things
+that we used to be able to do with enzyme + SFCs that will just no longer work.
+
+We could fix a lot of this if there was a reliable way to get from an SFC "fiber" to
```
lelandrichardson: it returns `null`. I prefer this over throwing i think. You might want to test to see if the instance is null rather than try/catch.

docs/future/compatibility.md
```diff
@@ -0,0 +1,184 @@
...
+}
+
+type EnzymeAdapter = {
+  // This is a method that will return a semver version string for the _react_ version that
+  // it expects enzyme to target. This will allow enzyme to know what to expect in the `instance`
+  // that it finds on an RSTNode, as well as intelligently toggle behavior across react versions
+  // etc. For react adapters, this will likely just be `() => React.Version`, but for other
+  // adapters for libraries like inferno or preact, it will allow those libraries to specify
+  // a version of the API that they are committing to.
+  getTargetApiVersion(): string;
```
aweary: If we just required that adapters be explicitly configured, would we need this `getTargetApiVersion` API at all? What happens if a React-like adapter (Preact, Inferno) exists for a version that collides with a React version? e.g., Preact has a 15.5 release.

src/ShallowWrapper.js
```diff
@@ -98,15 +91,20 @@ function validateOptions(options) {
...
-    // React 15.5+ exposes batching on shallow renderer itself
-    return renderer.unstable_batchedUpdates(fn);
+function getAdapter(options) {
+  if (options.adapter) {
+    return options.adapter;
   }
-  // React <15.5: Fallback to ReactDOM
-  return batchedUpdates(fn);
+  const adapter = configuration.get().adapter;
+  // TODO(lmr): warn about no adapter being configured
```
aweary: Should this throw instead of warn?

src/ShallowWrapper.js
```diff
@@ -119,27 +117,25 @@ class ShallowWrapper {
...
-          this.renderer.render(nodes, options.context);
-          const instance = this.instance();
-          if (
-            options.lifecycleExperimental &&
-            instance &&
-            typeof instance.componentDidMount === 'function'
-          ) {
-            instance.componentDidMount();
-          }
+      this.renderer = getAdapter(options).createRenderer({ mode: 'shallow', ...options });
```
aweary: If `mode` is being used by Enzyme internals should it be added to the `RendererOptions` type definition?

src/Utils.js
```diff
@@ -35,33 +17,14 @@ export function isCustomComponentElement(inst) {
...
-  if (node && node._currentElement) {
-    return (node._currentElement.props) || {};
-  }
-  if (REACT15 && node) {
-    if (internalInstance(node) && internalInstance(node)._currentElement) {
-      return (internalInstance(node)._currentElement.props) || {};
-    }
-  }
-
+function propsOfNode(node) {
```
aweary: `RSTTraversal` exports a `propsOfNode` utility, should we unify these?

src/adapters/EnzymeAdapter.js
aweary: Whats the strategy for exposing this class for adapter authors? Will it just be accessible from a `lib` folder like

```js
import EnzymeAdapter from 'enzyme/lib/EnzymeAdapter'
```

withDom.js
lelandrichardson: i'm not sure. this could cause unexpected behavior elsewhere in someone's codebase. That said, so could adding this polyfill... but without it, *nothing* works.... so we kinda need it.

src/adapters/ReactThirteenAdapter.js
lelandrichardson: i considered this but i'm not sure. I think throwing on unsupported modes might be enough. Pros/cons either way? Don't want to unnecessarily increase the surface area of this API

src/adapters/ReactSixteenAdapter.js
lelandrichardson: i'm not aware of any circumstance where it can be

docs/future/compatibility.md
```diff
@@ -0,0 +1,184 @@
...
+### Using different adapters with Enzyme
+
+At the top level, Enzyme would expose a `configure` method, which would allow for an `adapter`
+option to be specified and globally configure Enzyme's adapter preference:
+
+```js
+import Enzyme from 'enzyme';
+import ThirdPartyEnzymeAdapter from 'third-party-enzyme-adapter';
+
+Enzyme.configure({ adapter: ThirdPartyEnzymeAdapter });
```
lelandrichardson: Yeah, my thought is that this might get extended later on and configure is a broader term.

docs/future/compatibility.md
lelandrichardson: yeah. i think you're right. I think I will get rid of this, and let node semver declarations take care of the rest.

src/Utils.js
lelandrichardson: perhaps, yes. I've been trying to separate things mentally between adapters and the core enzyme lib so that they are easy to separate, and it's easier for me to find when things are no longer used

src/adapters/ReactThirteenAdapter.js
ljharb: It'd create a single source of truth for these strings; whereas otherwise, the string "mount" might need to appear more than once in a codebase. I'm a strong proponent of a string never appearing in a codebase more than once, if possible.

withDom.js
ljharb: In that case, can we use `rAF` without polyfilling?

docs/future/migration.md
ljharb: I think try/catch is more semantically correct.

aweary:
What do you all think about adding pointers for `nextSibling` and `prevSibling`? I found that having sibling pointers made it much easier when doing later traversal. For example, checking [adjacent](https://github.com/aweary/react-tester/blob/2163f45c623a252d357db252a1f5e5c61df3a123/src/selector-parser.js#L167-L176) or [general](https://github.com/aweary/react-tester/blob/2163f45c623a252d357db252a1f5e5c61df3a123/src/selector-parser.js#L181-L214) siblings. Since we're already traversing the whole tree, it should just require a constant amount of extra time to attach these pointers.

ljharb:
@aweary doesn't that open up the possibility that `nextSibling` or `prevSibling` could be wrong? In other words, it gives the adapter the ability to hijack the traversal algorithm (or am I misunderstanding)

aweary:
@ljharb we already trust that the adapter is returning an accurate `RSTNode`, so that doesn't sound like a problem to me. If we update the `RSTNode` type to include sibling pointers we're just extending the contract we're already requiring of adapters.

docs/future/migration.md
jwbay: Hoping this doesn't fall through the cracks -- is there something about the new architecture that prevents an adapter from always grabbing from the latest output, or node, whenever a traversal starts? (i.e. `myRootWrapper.find(...)`)

withDom.js
TiraO: Yeah, it seems like the `require('raf/polyfill')` should move inside the `if (!global.document)` since there is most likely a `requestAnimationFrame` already available when `global.document` exists.

For someone running tests in a browser the current code might make it look like react16 is more compatible with old browsers than it is.

ljharb: "compatible with shims" is still compatible; but I think it'd be better to avoid mutating the global in any situation inside `withDom`

TiraO:
@lelandrichardson I tried this out on my tests and I ran into a couple of issues: 

1. I had to call `Enzyme.configure`. I don't think users should be forced to call `Enzyme.configure` by default. Enzyme should work out-of-the-box. Also, `require('enzyme/setupAdapters')` in my test helper file didn't work for some reason.

1. Explicitly calling `update` looks like it isn't updating child props correctly (haven't really dug into this)

1. `simulate` (on a non-root node) is throwing an error `Error: ReactWrapper::update() can only be called on the root`

ljharb:
@TiraO "working out of the box by default" is the current state, and it's why we have issues with dependencies. It's important that users explicitly choose the version of React (or "not React") they're using.

TiraO:
@ljharb Fair enough. Would there be both a react 16 native adapter and a react 16 dom adapter?

ljharb:
@TiraO it would be possible to create one for each; i'm not sure what we're planning on releasing with, but the whole concept is that it's extensible.

withDom.js
TiraO: Isn't that kind of the point of `withDom` though? It's responsible for setting up global `document`, `window`, etc. I don't see how it would work if it didn't modify globals.
```javascript
    global.document = jsdom('');
    global.window = document.defaultView;
```

ljharb: lol yes, sorry. i meant mutating the global outside of DOM-related things, and out of the scope of a withDOM-wrapped test.

koba04:
Is `enzyme@3.0.0-alpha.1` based on this PR?

What about merging this PR into a branch like `v3` and releasing alpha or beta versions from the branch?
It's helpful to track the progress of supporting React v16 and I can work on supporting React v16.

koba04:
Is fragments support in TODOs?
The code including fragments seems not to work.

src/ShallowWrapper.js
lelandrichardson: (TODO)

src/adapters/elementToTree.js
lelandrichardson: (TODO)

docs/future/compatibility.md
```diff
@@ -0,0 +1,184 @@
...
+### React Standard Tree (RST)
+
+This proposal hinges on a standard tree specification. Keep in mind that this tree needs to account for more
+than what is currently satisfied by the output of something like `react-test-renderer`, which is currently
+only outputting the "host" nodes (ie, HTML elements). We need a tree format that allows for expressing a full
+react component tree, including composite components.
+
+```js
+// Strings and Numbers are rendered as literals.
+type LiteralValue = string | number
```
lelandrichardson: this is react 16 specific behavior that i don't think we should put in the spec...

docs/future/compatibility.md
```diff
@@ -0,0 +1,184 @@
...
+  // adapters for libraries like inferno or preact, it will allow those libraries to specify
+  // a version of the API that they are committing to.
+  getTargetApiVersion(): string;
+
+  // Provided a bag of options, return an `EnzymeRenderer`. Some options can be implementation
+  // specific, like `attach` etc. for React, but not part of this interface explicitly.
+  createRenderer(?options: RendererOptions): EnzymeRenderer;
+
+  // converts an RSTNode to the corresponding JSX Pragma Element. This will be needed
+  // in order to implement the `Wrapper.mount()` and `Wrapper.shallow()` methods, but should
```
lelandrichardson: no, i'm not sure... but when someone tries, hopefully they will file a bug :)

docs/future/migration.md
lelandrichardson: yes, the new architecture makes this fundamentally not possible. I don't see any way around this.

package.json
lelandrichardson: (TODO)

docs/future/migration.md
jwbay: I was able to get this working as a POC: https://github.com/jwbay/enzyme/commit/df609cf18b3263bfff5d95bf119cd924928027fe

Since the wrapper .getNode functions are already the starting access point for all traversals, we can just get the latest root node there.

The only downside is we lose reference equality for nodes during traversal in some cases now that there's always a new object literal created in the adapter's getNode. I tried caching based on render output but it didn't make much of a difference because most traversals end up going through .wrap, which creates a new renderer.

It's probably also worth noting that `mount` now also requires `.update` in places it didn't before. This new test fails in this branch but passes in master: https://github.com/jwbay/enzyme/commit/c14e87b22dc248ce4e4ff1d0bf1ab2881e691fed

install-relevant-react.sh
```diff
@@ -19,3 +19,7 @@ fi
 if [ "$REACT" = "15" ]; then
     npm run react:15
 fi
+
+if [ "$REACT" = "15" ]; then
```
ljharb: this needs to be 16

src/adapters/elementToTree.js
```diff
@@ -4,7 +4,7 @@ function nodeTypeFromType(type) {
   if (typeof type === 'string') {
     return 'host';
   }
-  if (type && type.prototype && typeof type.prototype.render === 'function') {
+  if (type && type.prototype && type.prototype.isReactComponent) {
```
ljharb: why not just `type.isReactComponent`? (checking `type.prototype` is good because it excludes arrow functions and `extends null`)

lelandrichardson: @ljharb i tried that, but it `isReactComponent` got introduced after react 13.  I'm thinking that when i refactor this code into separate code modules per adapter, i'll use render for react 13 and just `isReactComponent` after that.

ljharb: gotcha. for now, should this just use `render` then, to be maximally compatible?

lelandrichardson: it's now using either for miximal compatibility. this is an outdated commit i think

src/RSTTraversal.js
```diff
@@ -110,7 +105,7 @@ export function buildPredicate(selector) {
           return node => nodeHasId(node, selector.slice(1));
 
         case SELECTOR.PROP_TYPE: {
-          const propKey = selector.split(/\[([a-zA-Z-]*?)(=|])/)[1];
+          const propKey = selector.split(/\[([a-zA-Z][a-zA-Z_\d\-:]*?)(=|])/)[1];
```
ljharb: can you elaborate on what this fixes?

src/ReactWrapper.jsx
```diff
@@ -74,22 +75,29 @@ class ReactWrapper {
     }
 
     if (!root) {
+      this.renderer = getAdapter(options).createRenderer({ mode: 'mount', ...options });
       const ReactWrapperComponent = createWrapperComponent(nodes, options);
-      this.component = renderWithOptions(
+      this.renderer.render(
         <ReactWrapperComponent
```
ljharb: if there's jsx in the file, doesn't it have to have `React` in scope?

src/ReactWrapper.jsx
ljharb: since we should check that it's a function (rather than letting an error be thrown async), should this still default to null or undefined?

src/ReactWrapper.jsx
```diff
@@ -264,11 +292,17 @@ class ReactWrapper {
...
     if (this.root !== this) {
       throw new Error('ReactWrapper::setState() can only be called on the root');
     }
-    this.instance().setState(state, callback);
+    if (typeof callback !== 'function') {
+      throw new Error('ReactWrapper::setState() expects a function as its second argument');
+    }
+    this.instance().setState(state, () => {
+      this.update();
+      callback();
```
ljharb: same question here

src/ShallowWrapper.js
ljharb: bump

src/Utils.js
```diff
@@ -167,6 +143,17 @@ function arraysEqual(match, left, right) {
   return left.length === right.length && left.every((el, i) => match(el, right[i]));
 }
 
+function childrenToArray(children) {
+  // NOTE(lmr): we currently use this instead of Children.toArray(...) because
+  // toArray(...) didn't exist in React 0.13
+  const result = [];
+  React.Children.forEach(children, (el) => {
```
ljharb: should this React-specific stuff be moved to the adapters?

src/Utils.js
```diff
@@ -202,13 +189,7 @@ export function isReactElementAlike(arg) {
   return React.isValidElement(arg) || isTextualNode(arg) || Array.isArray(arg);
```
ljharb: here's another React usage outside of an adapter

src/adapters/.eslintrc
```diff
@@ -0,0 +1,12 @@
+{
+  "rules": {
+    "import/no-extraneous-dependencies": 0,
+    "import/no-unresolved": 0,
+    "import/extensions": 0,
+    "react/no-deprecated": 0,
+    "react/no-find-dom-node": 0,
+    "react/no-multi-comp": 0,
+    "no-underscore-dangle": 0,
+    "class-methods-use-this": 0
```
ljharb: presumably we can work to remove these overrides when we restructure the adapters?

src/adapters/ReactFifteenAdapter.js
```diff
@@ -0,0 +1,205 @@
...
+  mapNativeEventNames,
+  propFromEvent,
+  withSetStateAllowed,
+  assertDomAvailable,
+} from './Utils';
+
+function compositeTypeToNodeType(type) {
+  switch (type) {
+    case 0:
+    case 1: return 'class';
```
ljharb: why both 0 and 1? what's the difference?

src/adapters/ReactFifteenAdapter.js
```diff
@@ -0,0 +1,205 @@
...
+      ref: el.ref,
+      instance: inst._instance || inst._hostNode || null,
+      rendered: values(inst._renderedChildren).map(instanceToTree),
+    };
+  }
+  if (inst._hostNode) {
+    if (typeof el !== 'object') {
+      return el;
+    }
+    const children = inst._renderedChildren || { '.0': el.props.children };
```
ljharb: `.0`?

src/adapters/ReactFifteenAdapter.js
```diff
@@ -0,0 +1,205 @@
...
+      },
+      batchedUpdates(fn) {
+        return withSetStateAllowed(() => renderer.unstable_batchedUpdates(fn));
+      },
+    };
+  }
+
+  createStringRenderer(/* options */) {
+    return {
+      render(el /* , context */) {
```
ljharb: shouldn't the context here be passed down to ReactDOMServer?

src/adapters/ReactThirteenAdapter.js
ljharb: bump

src/adapters/Utils.js
```diff
@@ -0,0 +1,83 @@
+import { REACT013 } from '../version';
+
+export function mapNativeEventNames(event) {
```
ljharb: this seems like something that has to live in each adapter, and not something that can be shared in "utils"

src/validateAdapter.js
```diff
@@ -0,0 +1,22 @@
...
+
+      import Adapter from 'enzyme-adapter-react-15';
+
+      To find out more about this, see http://airbnb.io/enzyme/docs/installation/index.html
+    `);
+  }
+  if (!(adapter instanceof EnzymeAdapter)) {
+    throw new Error(
+      'Enzyme Internal Error: configured enzyme adapter did not inherit from the ' +
+      'EnzymeAdapter base class',
```
ljharb: this can be all on one line; long strings don't count against max-len.

docs/future/migration.md
```diff
@@ -0,0 +1,190 @@
...
+import Adapter from 'enzyme-adapter-react-16';
+
+Enzyme.configure({ adapter: new Adapter() });
+```
+
+The list of adapter npm packages for React semver ranges are as follows:
+
+- `enzyme-adapter-react-16` for `^16.0.0-0`
+- `enzyme-adapter-react-15` for `^15.5.0`
+- `enzyme-adapter-react-15.4` for `>= 15.0.0 && <15.5.0`
```
lelandrichardson: yes. this is needed for other reasons. the dependencies between these packages need to differ

src/RSTTraversal.js
lelandrichardson: I believe without this things like `wrapper.find('[data-foo]')` don't work

src/adapters/ReactFifteenAdapter.js
lelandrichardson: one is `PureComponent` and the other is `Component`

src/adapters/ReactFifteenAdapter.js
lelandrichardson: implementation detail of react 15. This is what children looks like here

src/adapters/Utils.js
lelandrichardson: in the followup PR where these become their own modules, i'll split this up

src/validateAdapter.js
lelandrichardson: but they count against readability

.eslintrc
```diff
@@ -12,6 +12,7 @@
     "id-length": 0,
     "new-cap": [2, { "capIsNewExceptions": ["AND"] }],
     "react/jsx-pascal-case": [2, { "allowAllCaps": true }],
+    "import/first": 0,
```
lelandrichardson: it could, but it seems stupid to do it for every file. The reason I did this was because i couldn't figure out how to get Karma to run `setupAdapters.js` before running the test suite, and that is now required.

src/adapters/ReactFifteenAdapter.js
lelandrichardson: It doesn't actually matter here, since the context is lifted in `render.jsx`. `context` will never have a value here with the way it is currently written. This could be refactored later if we wanted, but it wouldn't actually affect the public API at all.

src/ReactWrapper.jsx
lelandrichardson: it is in scope in this file

docs/future/migration.md
lelandrichardson: IMO this discussion should be moved to an issue in react if it's something we care enough about.

src/ReactWrapper.jsx
lelandrichardson: no it should default to a noop because the previous behavior was that it wasn't required.

src/ShallowWrapper.js
lelandrichardson: I'm just going to change the terminology to "Host Components". This PR moves enzyme to an adapter architecture, but removing concepts of react entirely from enzyme and making it possible to use enzyme for react-like libraries is *close*, but will require some additional refactoring i'm sure. For things like terminology of "DOM Components" etc, I think these ideas are fairly baked in in other parts of the code as well. I think that that's okay at this point.

src/Utils.js
lelandrichardson: yes, but I was having trouble doing this properly. There are still several parts of enzyme that will depend on react for now. Removing that dependency is still probably something we want to do, but I'm considering out of the scope of this PR.

src/adapters/.eslintrc
lelandrichardson: yeah I think so. or each adapter can have its own eslintrc that makes sense for it

src/ReactWrapper.jsx
ljharb: right, but isn't the point of adapters that the main enzyme package doesn't import React?

src/adapters/ReactFifteenAdapter.js
ljharb: gotcha, makes sense

src/validateAdapter.js
ljharb: lol only if your screen is set to be super narrow :-p either way our style guide allows for 100-char lines, so this string should be fine all on one line

src/ReactWrapper.jsx
ljharb: right - it's not required here tho either. defaulting it to undefined means we'd throw if a non-undefined non-function was provided.

src/Utils.js
ljharb: Can the adapter not export a `childrenToArray`?

src/ReactWrapper.jsx
lelandrichardson: i'm adding a typeof check before the `setChildProps` call. But keeping it defaulting to `noop`. Seem reasonable?

src/validateAdapter.js
lelandrichardson: it doesn't all fit in 100 characters, which is what i have my editor set to. I wrap earlier than i need to here (which i've now corrected), but i'm still going to keep strings inside of 100 char for readability

src/ReactWrapper.jsx
ljharb: Sure, I guess - it just adds function call overhead in the case when nothing is provided, so i'm not sure why this is better than defaulting to `undefined`.

src/version.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import semver from 'semver';
```
ljharb: let's move this inside the test dir?

test/ReactWrapper-spec.jsx
```diff
@@ -1408,7 +1435,7 @@ describeWithDOM('mount', () => {
       expect(() => wrapper.setState({ id: 'bar' }, 1)).to.throw(Error);
     });
 
-    itIf(REACT15, 'should throw error when cb is not a function', () => {
+    itIf(is('>=15 || ^16.0.0-alpha'), 'should throw error when cb is not a function', () => {
```
ljharb: open question: should this throw in all react versions?

test/ReactWrapper-spec.jsx
```diff
@@ -1785,6 +1808,7 @@ describeWithDOM('mount', () => {
     });
 
     describeIf(!REACT013, 'stateless function components', () => {
+      // TODO(lmr): this is broken now
```
ljharb: should this todo be removed?

test/ReactWrapper-spec.jsx
```diff
@@ -1892,10 +1916,10 @@ describeWithDOM('mount', () => {
           ]}
         />,
       );
-      expect(wrapper.children().length).to.equal(3);
-      expect(wrapper.children().at(0).hasClass('foo')).to.equal(true);
-      expect(wrapper.children().at(1).hasClass('bar')).to.equal(true);
-      expect(wrapper.children().at(2).hasClass('baz')).to.equal(true);
+      expect(wrapper.rendered().children().length).to.equal(3);
```
ljharb: open question: why is `rendered` necessary?

src/version.js
lelandrichardson: done

src/ReactWrapper.jsx
```diff
@@ -531,13 +589,8 @@ class ReactWrapper {
...
     this.single('simulate', (n) => {
-      const mappedEvent = mapNativeEventNames(event);
-      const eventFn = Simulate[mappedEvent];
-      if (!eventFn) {
-        throw new TypeError(`ReactWrapper::simulate() event '${event}' does not exist`);
-      }
-
-      eventFn(findDOMNode(n), mock);
+      this.renderer.simulateEvent(n, event, mock);
+      this.root.update();
```
lelandrichardson: nothing really is guaranteed other than `simulatedEvent()` finishing here... which will call the event I believe (at least our tests indicate), but of course the event could schedule something async

docs/future/migration.md
```diff
@@ -0,0 +1,372 @@
...
+
+At the time of writing this, Enzyme publishes "officially supported" adapters for React 0.13.x,
+0.14.x, 15.x, and 16.x. These adapters are npm packages of the form `enzyme-adapter-react-{{version}}`.
+
+You will want to configure Enzyme with the adapter you'd like to use before using enzyme in your
+tests. The way to do this is whith `Enzyme.configure(...)`. For example, if your project depends
+on React 16, you would want to configure Enzyme this way:
+
+```js
+import Enzyme from 'enzyme';
```
lelandrichardson: aesthetically i kinda prefer this way for some reason, but i don't think i can justify it in any way :P

zachlendon:
Any sense when a release might be made available with this change?

ljharb:
None yet.

midastouchprd:
any update on release with these changes?

PR #1014, opened 2017-06-27
Support npm@5 scripts behavior

ryanve:
enzyme isn't currently working npm@5 because `prepublish` no longers runs after local `npm install` like it does in previous versions. However the new `prepare` hook reproduces this behavior. My thinking is that aliasing these 2 commands should enable enzyme to work in any npm version. On publish via npm@5 both will run but this seems tolerable because one will replace the other.

- [npm scripts documentation](https://docs.npmjs.com/misc/scripts)
- [prepublish transition strategy](https://github.com/npm/npm/issues/10074)

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "homepage": "http://airbnb.io/enzyme/",
   "main": "build",
   "scripts": {
-    "prepublish": "not-in-publish || (npm run clean && npm run build && safe-publish-latest)",
+    "prepare": "not-in-publish || (npm run clean && npm run build && safe-publish-latest)",
+    "prepublish": "npm run prepare",
     "preversion": "npm run clean && npm run check",
     "postversion": "git push && git push --tags && npm run clean && npm run docs:publish",
     "version": "npm run build",


ryanve:
Hmm well using `npm@5` we are get errors for modules `require`'d by `enzyme@2.9.1`'s build folder.

```
PhantomJS 2.1.1 (Mac OS X 0.0.0) ERROR
  Error: Cannot find module "uuid/v4"
  at webpack:///~/enzyme/build/Utils.js:53:0 <- tests.webpack.js:26739~
```

Whereas with npm@4 it works fine.  Any idea then what would cause such errors? Can anyone verify that enzyme works with npm@5? Thanks for looking at this @ljharb

ljharb:
It definitely works with npm 5.

I'm not sure why you get that error, but `uuid` is required in the implementation code - any chance you're using a package-lock.json with npm 5, but that `npm ls` exits nonzero for? In other words, your lockfile is invalid?

ryanve:
Good idea to check for that. I tried deleting `package-lock.json` and reinstalling everything but still got the same error. I'll close this PR because the `in-publish` logic does seem sound. I'll post here when I find a solution. Thanks for the help :)

Issue #1020, opened 2017-07-05
Cannot find module 'uuid/v4' from 'Utils.js'


ryanve:
With `enzyme@2.9.1`and `uuid@3.1.0`  I get this `require` error. It seems likely that this relates to #1001 and #998. We are getting this error in multiple projects. cc @lencioni

```
Cannot find module 'uuid/v4' from 'Utils.js'
```

### `npm ls uuid`

```
├─┬ babel-cli@6.24.1
│ └─┬ chokidar@1.7.0
│   └─┬ fsevents@1.1.1
│     └─┬ node-pre-gyp@0.6.33
│       └─┬ request@2.79.0
│         └── uuid@3.0.1
├─┬ enzyme@2.9.1
│ └── uuid@3.1.0
├─┬ gulp-imagemin@2.4.0
│ └─┬ imagemin@4.0.0
│   ├─┬ buffer-to-vinyl@1.1.0
│   │ └── uuid@2.0.3
│   └─┬ imagemin-optipng@4.3.0
│     └─┬ exec-buffer@2.0.1
│       └─┬ tempfile@1.1.1
│         └── uuid@2.0.3  deduped
└─┬ node-sass@3.13.1
  └─┬ request@2.81.0
    └── uuid@3.0.1
```

If I manually change the Utils.js file from `require('uuid/v4')` to `require('uuid').v4` then it works.

ljharb:
I'm not sure why https://unpkg.com/uuid@3.1.0/v4.js wouldn't work if it's truly at v3.1.0. Can you verify that `node_modules/enzyme/node_modules/uuid/v4.js` exists, and that `node_modules/enzyme/node_modules/uuid/package.json` is at v3.1.0?

ryanve:
I'm not sure why either. Yes the v4 file exists and that package path is v3.1.0.

ryanve:
I confirmed this same error on <b>node</b> 7.10.0 and 8.1.3 and on <b>npm</b> 4.2.0 and npm 5.0.3

kouhin:
Same error!
Old version of uuid may be pinned in node_modules and will be used by webpack or other bundle tools, although new version is installed in enzyme/node_modules. 
I fixed this error by `npm install uuid --save` to replace the old version in node_modules.

ryanve:
[...]

PR #1021, opened 2017-07-05
Lock uuid version to 3.0.1

ryanve:
Testing this out in attempt to fix https://github.com/airbnb/enzyme/issues/1020

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "object.entries": "^1.0.4",
     "object.values": "^1.0.4",
     "prop-types": "^15.5.10",
-    "uuid": "^3.0.1"
+    "uuid": "3.0.1"
   },
   "devDependencies": {
     "babel-cli": "^6.24.1",


ryanve:
I was testing this out on [our fork](https://github.com/plangrid/enzyme). Did not mean to PR here. 

With this change we didn't get the uuid error but instead get this one. 

```
Cannot find module 'enzyme/build/ShallowWrapper' from 'utils.js'
```

ljharb:
On the fork, you may need to manually run the build process, as those files aren't in git.

ryanve:
Right—thanks :)

ryanve:
Ok after that it goes back to giving the original uuid error. So this is not the fix.

ljharb:
can you console.log what `require.resolve('uuid')` is, inside enzyme? I still suspect it's something about your node_modules layout.

ryanve:
`require.resolve('uuid')` is `/Users/me/project/node_modules/uuid/index.js`

`require.resolve('uuid/v4')` is `/Users/me/project/node_modules/uuid/v4.js`

ljharb:
OK, it's using `node_modules/uuid`, not `node_modules/enzyme/node_modules/uuid` - what version  does *that* package.json have in it? I'm going to guess not v3.1.0 - which should mean that `npm ls` should be exiting nonzero.

Perhaps you have a lock file with an invalid dependency graph?

ryanve:
I deleted `node_modules` and `package-lock.json`. I think the graph is ok. Could it relate to the `build` step? [uuid/v4](https://github.com/kelektiv/node-uuid/blob/v3.1.0/v4.js) exports a commonJS module not an ES6 one. [Importing as a member fixes it.](https://github.com/plangrid/enzyme/pull/3/files)

```
import { v4 as uuidv4 } from 'uuid';
```

ljharb:
That's only because `require('uuid').v4` is working for you, which shouldn't be needed in v3.1. iow, `import uuidv4 from 'uuid/v4'` should still be working.

Perhaps your build process is indeed breaking it?

PR #1030, opened 2017-07-12
[docs] Ensure `react-addons-test-utils` is set to `react-dom`

juanca:
Perhaps related issues:

- https://github.com/airbnb/enzyme/issues/940
- https://github.com/airbnb/enzyme/issues/968

Issue #940, opened 2017-05-12
can not find module 'react-addon-test-utils' from 'node_modules/enzyme/build'


mlding:
Our project is using :
`React: ^15.5.4`
`react-dom: ^15.5.4`
`react-test-renderer: ^15.5.4`

according to airbnb/enzyme website, If you are using React >=15.5, in addition to enzyme, you will have to ensure that you also have the `react-dom` and `react-test-renderer` npm modules installed, we already installed these npm modules, only that in React 0.14 or React <15.5 need to install the `react-addons-test-utils`.
when we use enzyme to test our react component, there is a problem:
`can not find module 'react-addon-test-utils' from 'node_modules/enzyme/build'`.
If we install the module 'react-addon-test-utils', then the test is OK, but our React >=15.5, Why we need to install the `react-addon-test-util`?

Anybody has solution to this problem? Thanks~

ljharb:
What version of enzyme are you using?

jenyeeiam:
I have the same set up, and have tried enzyme 2.8.0, 2.8.1, 2.8.2 to no avail :(

ljharb:
react-addons-test-utils is only attempted to be required when [either of these two requires](https://github.com/airbnb/enzyme/blob/master/src/react-compat.js#L100-L102) fail.

With v2.8.2 installed, if you try to require those things in the `node` repl, what happens?

zettadam:
This still fails with React 15.5.4, ReactDom 15.5.4 and Enzyme 2.8.2 on my Mac and on Linux:

```
WARNING in ./~/encoding/lib/iconv-loader.js
Critical dependencies:
9:12-34 the request of a dependency is an expression
 @ ./~/encoding/lib/iconv-loader.js 9:12-34

ERROR in ./~/enzyme/build/react-compat.js
Module not found: Error: Cannot resolve module 'react-addons-test-utils' in /Users/../.node_modules/enzyme/build
 @ ./~/enzyme/build/react-compat.js 143:20-54
```

ljharb:
And in the node repl, if you type `require('react-dom/test-utils')` and `require('react-test-renderer/shallow')`, what happens?

(also, just to confirm, `npm ls` is existing zero?)

zettadam:
`npm ls` returns no errors. I have also installed `react-addons-test-utils@15.5.1` and still see the same error.

ljharb:
And what about those repl requires?

rafaelkallis:
[...]

PR #1031, opened 2017-07-12
[Docs] Update karma docs for react >15.5

naganowl:
Follows the suggestion https://github.com/airbnb/enzyme/issues/940#issuecomment-310690420 and complements https://github.com/airbnb/enzyme/pull/1030


ljharb:

PR #1033, opened 2017-07-13
[Docs] Update Common Gotchas for ShallowWrapper setProps

paulsturgess:
As discussed here: https://github.com/airbnb/enzyme/issues/34 `setProps` does not trigger the `componentDidUpdate` lifecycle method by default.

I've updated the Common Gotchas section of the `setProps` documentation with the solution to set `lifecycleExperimental` to true.


kumar303:
Hi. The bit about `lifecycleExperimental: true` would have been really helpful to me if it were documented. It would be nice to see this in the documentation sometime soon. Thanks.

PR #1039, opened 2017-07-19
Update to GitBook 3

### LINKED ISSUE #1038: Update GitBook
Currently, this repo is using v2 of GitBook and the latest version is v3 with v4 just around the corner. To prevent getting left behind we should update to v3. Previous discussion in #1036.

samit4me:
Fixes #1038

Update GitBook to latest version and update configuration so the current documentation works without any changes.


SUMMARY.md
```diff
@@ -0,0 +1 @@
+docs/README.md
```
ljharb: is this a symlink? if so, does it also work on windows?

samit4me: Didn't appear to work, good pickup :) I moved the docs/README.md into the root and renamed to SUMMARY.md (the GitBook default name for Table of Contents).

samit4me:
Found unrelated issues on Windows when building docs, see error below:
```JS
Template render error: (D:\Development\enzyme\docs\api\ShallowWrapper\containsAllMatchingElements.md) [Line 30, Column 25]
  expected variable end
```

This has something to do with spaces and variables, but was resolved by escaping the JSX blocks using inline object literals as per solution on https://github.com/GitbookIO/gitbook/issues/1198.

Hope that's okay to include in this PR?

ljharb:
I think it's OK to include, but it'd be ideal to figure out exactly which characters are causing issues - i'd prefer to change those rather than adding the raw blocks.

samit4me:
It's the `{{ fontSize:13 }}`. They could be removed as I don't think they add any value. Thoughts?

ljharb:
I'm confused, that's perfectly valid jsx. why is that causing trouble?

samit4me:
It looks like jsx isn't well supported by GitBook (see [templating](https://toolchain.gitbook.com/templating/)) as it uses nunjunks which uses `{{ varable }}` so I guess that's why.

It looks like the prism plugin was used for that exact reason, I removed it as it was throwing errors on both macOS and Windows and assumed it was only for highlighting. Turns out I was wrong (no big surprise there :)

I'll have to look into why it doesn't work I guess, wish me luck.

samit4me:
It looks like JSX isn't well supported by GitBook (see [templating](https://toolchain.gitbook.com/templating/)) as it uses nunjunks which uses `{{ varable }}` so I guess that's why.

Seems to be a Windows only issue and has something to do with line endings "maybe", I've logged an issue upstream to try and get answers https://github.com/GitbookIO/gitbook/issues/1827.

I think the path of least resistance would be to remove these Object literal props as they don't appear to be adding any value, then we won't need to escape anything?

ljharb:
Yeah, i guess for now let's move them to `const style = { fontSize: 13 };` and `{style}` to see if that works?

samit4me:
Great idea and it works 🎉, just pushed up the changes.

ljharb:
k, this LGTM, but before approval i want to test it out locally

ljharb:
One thing I notice in npm 5+ - `gitbook install` seems to add deps to `package.json`:
```diff
+    "gitbook-plugin-anchors": "^0.7.1",
+    "gitbook-plugin-codeblock-disable-glossary": "0.0.1",
+    "gitbook-plugin-collapsible-menu": "^1.0.3",
+    "gitbook-plugin-edit-link": "^2.0.2",
+    "gitbook-plugin-github": "^2.0.0",
```

@samit4me do you think these should be explicit devDeps instead of letting gitbook implicitly install them?

samit4me:
Just switched over to Node 8/NPM 5 and got the same thing, the plugins were added as "dependencies" which doesn't seem correct, they really are "devDependencies". I've put them into "devDependencies" and removed node_modules, did a `npm i && npm run react:15 && npm run docs:watch` and everything appears to be working, so I will push the update.

docs/api/ShallowWrapper/debug.md
```diff
@@ -40,14 +40,11 @@ Outputs to console:
...
+const handleClick = () => {};
 const wrapper = shallow((
   <Book
     title="Huckleberry Finn"
-    cover={{
-      url: 'http://some.url/to/img.png',
-      width: 40,
-      height: 80,
-    }}
+    cover={handleClick}
```
nfcampos: I don't think the change to this line makes sense, why would a prop named `cover` be a function?

samit4me:
Yeah 💯 agree, we could make it anything other than an object due to the template issues on windows. Do you have any suggestions?

docs/api/ShallowWrapper/debug.md
samit4me: Yeah 💯 agree, we could make it anything other than an object due to the template issues on windows. Do you have any suggestions?

samit4me: What do you think of the updated one?

samit4me:
Just checking in to see if this update is still wanted? I've been keeping it up-to-date with `master` for a month now and was wondering if I should keep doing so?

ljharb:
@samit4me it's definitely still desired, sorry for the delay - with enzyme v3 incoming, there's been lots of changes in master. hopefully it's not too laborious to update it yet again?

samit4me:
Awesome @ljharb excited to see V3. I've updated again.

Issue #911, opened 2017-04-21
Enzyme not unmounting by itslef


hakunin:
Hi, I noticed that Enzyme doesn't unmount the components I mounted after each test.
In my case, they will listen and react to stuff, thus the test gets slower and slower.

Are we supposed to do it manually? 
I'd expect a high level library like Enzyme to unmount everything at the end of each test.

**Without unmounting:**
![screenshot 2017-04-21 14 47 56](https://cloud.githubusercontent.com/assets/65846/25278090/a02cc2c4-26a1-11e7-9c00-07efb294df9b.png)

**With unmounting its plenty fast but I get errors from the unmount code**
![screenshot 2017-04-21 14 53 36](https://cloud.githubusercontent.com/assets/65846/25278233/5bc4561e-26a2-11e7-83de-de77dcc46fe2.png)

I tried clearing the timers in the `afterEach()` function where I try to unmount but it didn't help. Any ideas?

ljharb:
Yes, you're supposed to do it manually.

It would be impossible for enzyme to know when your test is over, or even that you're writing a test at all.

Your component probably needs to clean up after itself in `componentWillUnmount`.

hakunin:
> It would be impossible for enzyme to know when your test is over
How about `afterEach`? Does that not mean the test is over?

ljharb:
It is impossible for enzyme to know that any of its code is running in an `afterEach`, for example. Obviously your test is over inside the afterEach, but only *you* know that - not enzyme.

hakunin:
I don't follow. Are you saying Enzyme can't add `afterEach` callback that would cleanup the mounted components?

ljharb:
Yes, because enzyme doesn't know you're using mocha, or any other test framework. `afterEach` doesn't exist in every test framework, it exists in the one you chose to use.

hakunin:
then a simple `enzyme.cleanup()` in a global `afterAll` hook would still work, or even something like `enzyme.init(jest)`, but I will simply create a wrapper myself :) thanks for explaining it

ljharb:
Indeed, something that would be useful for you here would be a wrapper that kept in memory every enzyme wrapper that was ever created, and provided an unmount/cleanup method to clean them all up, so you could call that in `afterEach`.

That belongs in a separate enzyme wrapper package, but would be very useful.

silvenon:
[...]

PR #1043, opened 2017-07-24
Add note that mount() requires cleanup

silvenon:
I was following the discussion at #911 and thought that this might be a useful note to add. I can also incorporate it in the example below if you want.

### DIFF
--- a/docs/api/mount.md
+++ b/docs/api/mount.md
@@ -10,6 +10,8 @@ want to run your tests inside of a browser, the recommended approach to using `m
 on a library called [jsdom](https://github.com/tmpvar/jsdom) which is essentially a headless browser
 implemented completely in JS.
 
+**Note**: unlike shallow or static rendering, full rendering actually mounts the component in the DOM, which means that tests can affect each other if they are all using the same DOM. Keep that in mind while writing your tests and, if necessary, use [`.unmount()`](ReactWrapper/unmount.md) or something similar as cleanup.
+
 ```jsx
 import { mount } from 'enzyme';
 import sinon from 'sinon';


lelandrichardson:
Love it. Thanks!

Hillshum:
Should this be extended to the docs for `shallow` as well? I've found that even using shallow rendering I had tests affecting each other when doing things like `window.setInterval`

silvenon:
Example?

Hillshum:
Here's what I ran into
```javascript
//component
componentDidMount() {
  this.intervalId = window.setInterval(()=> {this.setState({...})}, 100)
}

componentWillUnmount() {
  window.clearInterval(this.intervalId)
}

// test (this is using lolex for timer control)
it('should clear the interval on timeout', ()=> {
   const updateSpy = jest.spyOn(Timer.prototype, 'updateTimer')
      wrapper = enzyme.shallow(<Timer />)
      clock.tick(200)
      wrapper.unmount()
      updateSpy.mockClear()

      clock.tick(350)

      expect(updateSpy).toHaveBeenCalledTimes(0) // if other shallow wraps are unmounted this fails
      updateSpy.mockRestore()
}
```

silvenon:
More context might be more useful, but even looking at this leads me to believe that this test is flaky because you're spying on the **prototype**, which is shared in all tests using this component. You should be spying on the **instance**, i.e. the wrapper.

Shallow rendering doesn't output anything to the DOM (it doesn't require DOM at all), therefore it doesn't require a cleanup.

Hillshum:
My thinking was that spying on the instance was better, but I was unable to find a way to access the instance to attach my spy on the proper method before `componentDidMount` fired, therefore the function registered with `setInterval` was not spied. Maybe I'm missing something obvious however.

Regardless, even with shallow rendering this test suite registers a function with the `setInterval` API once for each test that never gets cleaned up without a call to `.unmount`.

silvenon:
I haven't been using Enzyme for quite a while so I completely forgot that lifecycle methods work with shallow rendering. This is a good reason to mention unmounting in the docs, you should submit a PR. 😉

Issue #1061, opened 2017-08-07
React 15.5 errors for `react/addons` & `react-addons-test-utils`


ericgio:
I'm using Webpack 3.4 and React 15.6. I was able to configure everything to get tests running, but enzyme still complains about the lack of `react/addons` and `react-addons-test-utils`. The former is gone and the latter is deprecated, both as of React 15.5.

```
ERROR in ./node_modules/enzyme/build/react-compat.js
Module not found: Error: Can't resolve 'react/addons' in '/enzyme-example-karma-webpack/node_modules/enzyme/build'
 @ ./node_modules/enzyme/build/react-compat.js 40:14-37 41:44-67
 @ ./node_modules/enzyme/build/ReactWrapper.js
 @ ./node_modules/enzyme/build/index.js
 @ ./test/Foo-test.js

ERROR in ./node_modules/enzyme/build/react-compat.js
Module not found: Error: Can't resolve 'react-addons-test-utils' in '/enzyme-example-karma-webpack/node_modules/enzyme/build'
 @ ./node_modules/enzyme/build/react-compat.js 137:18-52
 @ ./node_modules/enzyme/build/ReactWrapper.js
 @ ./node_modules/enzyme/build/index.js
 @ ./test/Foo-test.js
```

package.json:
```
"dependencies": {
  "react": "^15.6.1",
  "react-dom": "^15.6.1",
  "react-test-renderer": "^15.6.1"
},
```

karma.conf.js:
```
externals: {
  'react/lib/ExecutionEnvironment': 'react',
  'react/lib/ReactContext': 'react',
  'react-test-renderer': 'react-test-renderer',
}
```

Repro case using the Karma/Webpack example:
https://github.com/ericgio/enzyme-example-karma-webpack

chansuke:
Same error occurs in my environment too.

davecarlson:
Same here

naganowl:
https://github.com/airbnb/enzyme/issues/940#issuecomment-310690420 might be relevant

ericgio:
Adding the following to `externals` in my karma.conf.js got rid of the errors:

```
'react/addons': 'react',
'react-addons-test-utils': 'react-dom',
```

However, it seems like we're just getting rid of the symptoms here. Those lines shouldn't be necessary.

ljharb:
I don't think this is relevant any more now that we're on v3.

PR #1062, opened 2017-08-07
Update Webpack guide for React 15.5

ericgio:
I was having trouble configuring my setup correctly for Webpack + React 15.5 based on the current guide. This PR adds documentation for those using React 15.5+, where `react/addons` is gone and `react-addons-test-utils` is deprecated in favor of `react-test-renderer`.


ericgio:
Note that Enzyme still raises errors about missing `react/addons` and `react-addons-test-utils` but tests will run. I've opened a [separate issue](https://github.com/airbnb/enzyme/issues/1061) for this.

naganowl:
https://github.com/airbnb/enzyme/pull/1031 is relevant to this as well.

ericgio:
As noted in #1061, I was able to silence the errors by adding the following to `externals`:

```
'react/addons': 'react',
'react-addons-test-utils': 'react-dom',
```

I'm hesitant to add this to the docs, however, since the errors seem like a bug and these lines shouldn't be necessary.

ljharb:

PR #1068, opened 2017-08-11
Updated `withDom` to support jsdom@9.x.x and jsdom@10.x.x and greater.

fritzmonkey:


fritzmonkey:
Seems to. I manually tested with 6.x.x, 7.x.x, 8.x.x, 9.x.x and 11.x.x. Are there automated tests that I can run to verify this, or a guide for manual testing?

ljharb:
Automated tests across jsdom versions aren't something we have set up, unfortunately.

If you wanted to add an automated test, even though it'd only run on the latest jsdom, that's a start :-)

fritzmonkey:
I thought about this for a bit, and I'm not really sure what the best way to go about testing this would be. Any advice?

ljharb:
I suppose a test that uses withDOM, that would fail if jsdom wasn't installed correctly, would be a decent start?

indexzero:
@ljharb spent some time working with @fritzmonkey on this today. We talked a lot about how to accomplish this without expanding the build matrix too much. `node@8` tests are currently failing for some reason, but we'll get that sorted.

Wanted input from you (and/or @lelandrichardson & other core contributors) about [the new build matrix](https://travis-ci.org/airbnb/enzyme/builds/267011187) itself before we spent too much more time in this direction. So, thoughts?

ljharb:
I think @lelandrichardson is the best one to answer this.

indexzero:
@lelandrichardson 7 day bump. Thoughts on the changes to CI? Appreciate your input!

ljharb:
@fritzmonkey this has clearly slipped through the cracks; if you have any interest in updating it, i'd be happy to review.

fritzmonkey:
It definitely slipped through the cracks on my end as well. I closed it because I'm not sure if it is even relevant at this point. If you think there is still value here, I can get this updated.
I still have some questions around the test matrix though.`jsdom` is now on v16. It doesn't seem sustainable to me to continue to add versions to the test matrix, especially if we we need to run these tests against multiple version of react.

ljharb:
I don’t mind having a test run for every major of jsdom from 6-16, personally. I’d also be fine to land it with a few of them failing (as allowed failures) if that set us up for proper testing moving forwards.

PR #1072, opened 2017-08-16
[Fix] SetProps - Children objects should be new instances

Tzook:
Each time we update the props of a component with an array of children or an element child, React sends a new instance and not the same instance.
That means that pure components (or components do shallow equal the children) should re-render.

See the following discussion in React, stating that this is indeed the wanted behavior: [link](https://github.com/facebook/react/issues/6515).

I've added unit tests that cover the case - they used to fail before and now they pass.

Thanks.

### DIFF
--- a/src/ReactWrapper.jsx
+++ b/src/ReactWrapper.jsx
@@ -14,6 +14,8 @@ import {
   nodeEqual,
   nodeMatches,
   getAdapter,
+  shouldCloneChildren,
+  cloneChildren,
 } from './Utils';
 import {
   debugNodes,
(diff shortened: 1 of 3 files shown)


ljharb:
Did they fail on enzyme 2? master is now on its way to enzyme 3, so it's not in a reliable state for bugfixes.

ljharb:
@Tzook did you figure this out? which enzyme version did this pertain to?

Tzook:
It's for enzyme 3, @ljharb.

ljharb:
Thanks; which version of React?

Tzook:
Tested on 13, 14 and 15. It's relevant for 16 as well.

PR #1083, opened 2017-08-20
Make private properties more private and harder to use

lelandrichardson:
to: @ljharb @aweary 

This is in preparation for Enzyme v3.0. We are moving properties that are considered private and implementation details to use symbols and harder to rely on property names in an effort to make future migrations easier.

### DIFF
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -7,7 +7,7 @@ import sinon from 'sinon';
 import { createClass } from './_helpers/react-compat';
 import { shallow, render, ShallowWrapper } from 'enzyme';
 import { describeIf, itIf, itWithData, generateEmptyRenderData } from './_helpers';
-import { ITERATOR_SYMBOL, withSetStateAllowed } from 'enzyme/build/Utils';
+import { ITERATOR_SYMBOL, withSetStateAllowed, sym } from 'enzyme/build/Utils';
 import { REACT013, REACT014, REACT16, is } from './_helpers/version';
 
 // The shallow renderer in react 16 does not yet support batched updates. When it does,
@@ -4300,14 +4300,14 @@ describe('shallow', () => {
     it('works with a name', () => {
       const wrapper = shallow(<div />);
       wrapper.single('foo', (node) => {
-        expect(node).to.equal(wrapper.node);
+        expect(node).to.equal(wrapper[sym('__node__')]);
       });
     });
 
     it('works without a name', () => {
       const wrapper = shallow(<div />);
       wrapper.single((node) => {
-        expect(node).to.equal(wrapper.node);
+        expect(node).to.equal(wrapper[sym('__node__')]);
       });
     });
   });
--- a/packages/enzyme/src/ReactWrapper.jsx
+++ b/packages/enzyme/src/ReactWrapper.jsx
@@ -14,6 +14,8 @@ import {
   nodeEqual,
   nodeMatches,
   getAdapter,
+  sym,
+  privateSet,
 } from './Utils';
 import {
   debugNodes,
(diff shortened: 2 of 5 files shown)


packages/enzyme/src/ReactWrapper.jsx
```diff
@@ -1060,4 +1071,25 @@ if (ITERATOR_SYMBOL) {
   });
 }
 
+function privateWarning(prop, extraMessage) {
+  Object.defineProperty(ReactWrapper.prototype, prop, {
+    get() {
+      throw new Error(`
+        Attempted to access ShallowWrapper::${prop}, which was previously a private property on
+        Enzyme ShallowWrapper instances, but is no longer and should not be relied upon.
```
aweary: Need to change `ShallowWrapper` to `ReactWrapper`.

ljharb:
It'd be cool if we could go a bit slower on merging these things; it's Sunday and cell reception is limited on my drive home from out of town.

lelandrichardson:
@ljharb i'll slow down... i'm trying to take advantage of time when i have it though, and some of these PRs are hard to do in parallel.

ljharb:
PRs don't have to block proceeding on multiple local branches, and rebasing/PRing as things are merged :-)

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -4300,14 +4300,14 @@ describe('shallow', () => {
     it('works with a name', () => {
       const wrapper = shallow(<div />);
       wrapper.single('foo', (node) => {
-        expect(node).to.equal(wrapper.node);
+        expect(node).to.equal(wrapper[sym('__node__')]);
```
ljharb: Can these underscores be added by the `sym` function, instead of hardcoded in the text?

packages/enzyme/src/ReactWrapper.jsx
```diff
@@ -1060,4 +1071,25 @@ if (ITERATOR_SYMBOL) {
   });
 }
 
+function privateWarning(prop, extraMessage) {
```
ljharb: this function seems like it could live in a separate file and be shared, and take the proto as an argument?

packages/enzyme/src/Utils.js
```diff
@@ -362,3 +362,15 @@ export function displayNameOfNode(node) {
 
   return type.displayName || (typeof type === 'function' ? functionName(type) : type.name || type);
 }
+
+export function sym(s) {
+  return typeof Symbol === 'function' ? Symbol.for(`enzyme.${s}`) : s;
```
ljharb: the presence of `Symbol` doesn't ensure `Symbol.for` is present - this should maybe be a fallback from `Symbol.for` to a memoized `Symbol` to ``` `___${s}___` ```?

packages/enzyme/src/Utils.js
```diff
@@ -362,3 +362,15 @@ export function displayNameOfNode(node) {
...
+}
+
+export function privateSet(obj, prop, value) {
+  Object.defineProperty(obj, prop, {
```
ljharb: any reason not to use a `privateGet` as well, so it could later be refactored to use `WeakMap`s?

packages/enzyme/src/Utils.js
lelandrichardson: can you explain what you're proposing a little bit more? i'm not sure I understand

packages/enzyme/src/Utils.js
lelandrichardson: no real reason, no. It would add some syntactic overhead relative to `obj[prop]`, but i'm not opposed to it if we want to. I figured we'd just keep this setup until private properties are at whatever stage makes you comfortable with us using them :)

ljharb: that's exactly why i want them to use `WeakMap`, because that's how private properties will be transpiled :-)

packages/enzyme/src/Utils.js
ljharb:
```js
const syms = {};
function sym(s) {
  const k = `enzyme.___${s}___`;
  if (typeof Symbol === 'function') {
    if (typeof Symbol.for === 'function') return Symbol.for(k);
    const sym = Symbol(k);
    syms[s] = sym;
    return sym;
  }
  return k;
}
```

PR #1084, opened 2017-08-20
Remove all React dependencies from enzyme

lelandrichardson:
to: @ljharb @aweary 

This PR removes all react dependencies from the core enzyme library. In order to do this, the component "wrappers" were moved from enzyme into the adapters. This ends up cleaning up a number of things, and I think worked out rather nicely.

### DIFF
--- a/packages/enzyme/src/ReactWrapper.js
+++ b/packages/enzyme/src/ReactWrapper.js
@@ -1,11 +1,9 @@
-import React from 'react';
 import cheerio from 'cheerio';
 import flatten from 'lodash/flatten';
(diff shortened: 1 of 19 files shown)


packages/enzyme/src/ReactWrapper.js
```diff
@@ -117,7 +99,7 @@ class ReactWrapper {
         privateSet(this, NODES, nodes);
       }
       this.length = this[NODES].length;
-      privateSet(this, COMPONENT, null);
+      // privateSet(this, COMPONENT, null);
```
ljharb: uncomment, or delete?

PR #1088, opened 2017-08-22
props not merged when shallow rendering in lifecycleExperimental

LarsHassler:
When shallow rendering a component with `lifecycleExperimental: true` the lifecyle methods `componentWillReceiveProps` and `shouldComponentUpdate` will only be called with the value given to `setProps`, but not with old and new props merged.

It works for shallow rendering without lifecycleExperimental.

Due to the "[dirty hacks](https://github.com/airbnb/enzyme/blob/da9140713a946ef3bef23952e693bc917166a0b3/packages/enzyme/src/ShallowWrapper.js#L272-L293)" i'm not sure if there is a better way to do this, but this at least fixes the dirty hacks.

### DIFF
--- a/packages/enzyme/src/ShallowWrapper.js
+++ b/packages/enzyme/src/ShallowWrapper.js
@@ -262,7 +262,7 @@ class ShallowWrapper {
         const state = instance.state;
         const prevProps = instance.props || this[UNRENDERED].props;
         const prevContext = instance.context || this[OPTIONS].context;
-        const nextProps = props || prevProps;
+        const nextProps = { ...prevProps, ...props };
         const nextContext = context || prevContext;
         if (context) {
           this[OPTIONS] = { ...this[OPTIONS], context: nextContext };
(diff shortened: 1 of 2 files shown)


packages/enzyme/src/ShallowWrapper.js
```diff
@@ -262,7 +262,7 @@ class ShallowWrapper {
         const state = instance.state;
         const prevProps = instance.props || this[UNRENDERED].props;
         const prevContext = instance.context || this[OPTIONS].context;
-        const nextProps = props || prevProps;
+        const nextProps = Object.assign({}, prevProps || {}, props || {});
```
ljharb: The fallback objects are not required here, and this should use object spread.

LarsHassler: Never tried without the fallback, good to know ;)

Updated the syntax

LarsHassler:
Commits are rebased now

PR #1093, opened 2017-08-28
[Deps] [breaking] update `cheerio` to v1

ljharb:
See https://github.com/cheeriojs/cheerio/issues/1047#issuecomment-325221409 for context.

Hoping to merge this ASAP so it can be part of v3.

cc @jugglinmike

This semver range also will include `^1.0.0` once it's released, verified on https://semver.npmjs.com/

### DIFF
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -2817,7 +2817,7 @@ describe('shallow', () => {
   });
 
   describe('.render()', () => {
-    it('should return a cheerio wrapper around the current node', () => {
+    it('returns a cheerio wrapper around the current node', () => {
       class Foo extends React.Component {
         render() {
           return (<div className="in-foo" />);
@@ -2832,15 +2832,21 @@ describe('shallow', () => {
           );
         }
       }
+
       const wrapper = shallow(<Bar />);
-      expect(wrapper.render().find('.in-bar')).to.have.length(1);
+
+      const rendered = wrapper.render();
+      expect(rendered.is('.in-bar')).to.equal(true);
+      expect(rendered).to.have.lengthOf(1);
+
       const renderedFoo = wrapper.find(Foo).render();
-      expect(renderedFoo.find('.in-foo')).to.have.length(1);
+      expect(renderedFoo.is('.in-foo')).to.equal(true);
+      expect(renderedFoo.is('.in-bar')).to.equal(false);
       expect(renderedFoo.find('.in-bar')).to.have.length(0);
     });
 
-    describeIf(!REACT013, 'stateless function components', () => {
-      it('should return a cheerio wrapper around the current node', () => {
+    describeIf(!REACT013, 'stateless functional components', () => {
+      it('returns a cheerio wrapper around the current node', () => {
         const Foo = () => (
           <div className="in-foo" />
         );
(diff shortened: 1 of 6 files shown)


aweary:
@ljharb test suite seems to be failing

```
TypeError: _cheerio2.default.load(...)(...).root is not a function
```

ljharb:
Yes, I can't get them running locally yet to fix :-/

jugglinmike:
`root` is a static method on the jQuery-like object produced by `load`. I guess
the recommendation we made is a little inaccurate. I'll think about how to make
a better explanation for our migration guide.

For your purposes here, could you try exposing simply `cheerio.load('')(html)`?

ljharb:
What's the functional difference? Is the fragment not jquery-like?

What about `cheerio(cheerio.load('')(html))`?

jugglinmike:
Here again, my explanation is more confusing than helpful. "jQuery-like" can
mean a couple things, depending on your interpretation of "jQuery".

The return value of `cheerio.load('')` could be called "jQuery-like" because it
returns a *function* that is like the global function defined by jQuery (i.e.
`window.$` or `window.jQuery`). This functions has the static methods defined
by jQuery such as `parseHTML`, but it also has a Cheerio-specific static method
named `root`.

The return value of `cheerio.load('')('<div>')` is "jQuery-like" because it
returns an array-like *object* whose members are elements (sometimes called a
"jQuery collection").

My apologies for making this more confusing than it needs to be. If it's any
consolation, this conversation is going to help me improve Cheerio's migration
guide, and that will hopefully make the transition easier for others when we do
publish version 1.0. I'd be happy to contribute to this branch and get things
passing, but I may not have any availability until this weekend. Please let me
know!

ljharb:
That sounds great; I'll add you to my fork today so you can add to this PR.

.travis.yml
```diff
@@ -3,12 +3,12 @@ node_js:
   - "8"
   - "6"
   - "4"
+before_install:
+  - 'if [ "${TRAVIS_NODE_VERSION}" = "0.6" ]; then npm install -g npm@1.3 ; elif [ "${TRAVIS_NODE_VERSION}" != "0.9" ]; then case "$(npm --version)" in 1.*) npm install -g npm@1.4.28 ;; 2.*) npm install -g npm@2 ;; esac ; fi'
```
nfcampos: why do we need this and the `install` script?

.travis.yml
```diff
@@ -19,7 +19,7 @@ sudo: false
 matrix:
   fast_finish: true
   include:
-   - node_js: "6"
+   - node_js: "node"
```
nfcampos: why this change?

ljharb: There's no benefit to pinning the linter on node 6.

nfcampos: is there a benefit to pinning the karma tests, seeing as they run in a browser?

.travis.yml
ljharb: It ensures npm doesn't break, and deps can be installed on any node version. This will be rebased out tho; only the latest commit on the PR matters.

nfcampos: but we dont use versions of node in travis other than 8, 6 and 4, so why test for 0.6 and 0.4, etc?

ljharb: Because then I can copy paste my known-correct Travis config lines from all my other repos :-)

.travis.yml
ljharb: Only if the karma tools are c deps. I don't know much about karma.

ljharb:
OK, got this working :-)

The effective breaking change is that now `render` returns a wrapper around the *rendered thing*, just like shallow and mount will (once v3 is done) - instead of a “root” that requires you dive into `.children()`. This will make all our APIs more consistent.

jugglinmike:
I discovered the same thing last night, but I didn't have the time/imagination
to think of a workaround to preserve Enzyme's current behavior. Even now, I'm
not sure it's possible--that old root "element" was a little like a document
fragment in that it had children but no representation in the rendered DOM.

...so I'm glad that not only is the breaking change acceptable, but that you
find it preferable for consistency's sake!

On the Cheerio side, I plan on updating the migration guide with information
about the `root` method. We didn't modify its implementation, but now that
we're working with more realistic documents, users should be aware that what
they get back will be different than prior releases (though more in-line with
the browser overall).

nfcampos:
@ljharb are we compiling a list of breaking changes for v3?

.travis.yml
nfcampos: i dont think it makesany difference, but no point in changing it now

ljharb:
@nfcampos i believe leland has one in progress.

PR #1099, opened 2017-08-30
[Fix] `mount`: do not mutate `Component.contextTypes`

### LINKED ISSUE #1091: ReactWrapperComponent assumes that node.type.contextTypes isExtensible
Problem: When creating a wrapper using a component that uses Mobx-react inject and providing options.childContextTypes I get an error:

`Cannot add property ..., Object is not extensible`

I think this is because:

In reactWrapperComponent.jsx there is this line
`Object.assign(childContextTypes, options.childContextTypes);`

which assumes childContextTypes is extensible, i.e. ` Object.isExtensible(childContextTypes) === true` -> see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isExtensible

and as I'm using mobx-react `inject` which is calling Object.seal on the injected context Type Object this blows up with the error mentioned above.

I'd suggest the simple fix of changing line 73 in ReactWrappeComponent.jsx from:
```Object.assign(childContextTypes, options.childContextTypes);```
to:
``` const allChildContextTypes = Object.assign({}, childContextTypes, options.childContextTypes)```

and line 78 to: 
```WrapperComponent.childContextTypes = allChildContextTypes```

Currently it mutates the childContextTypes, and as long as nothing else is dependent on it doing that then it should be fine,

thoughts?

### PR DESCRIPTION
**Pull request by ljharb**
Fixes #1091.

Turns out that this was only an issue on class-based components (possibly SFCs as well), presumably because createClass clones `contextTypes` for us.

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -94,6 +94,46 @@ describeWithDOM('mount', () => {
       expect(wrapper.find(SimpleComponent)).to.have.length(1);
     });
 
+    describe('does not attempt to mutate Component.childContextTypes', () => {
+      const SimpleComponent = createClass({
+        displayName: 'Simple',
+        render() {
+          return <div />;
+        },
+      });
+
+      class ClassComponent extends React.Component {
+        render() {
+          return (
+            <div>
+              {this.context.a}
+              <SimpleComponent />
+            </div>
+          );
+        }
+      }
+      ClassComponent.contextTypes = Object.freeze({ a: PropTypes.string });
+
+      const CreateClassComponent = createClass({
+        contextTypes: ClassComponent.contextTypes,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/enzyme-test-suite/test/ReactWrapper-spec.jsx [hunk 1]
```diff
@@ -94,6 +94,47 @@ describeWithDOM('mount', () => {
       expect(wrapper.find(SimpleComponent)).to.have.length(1);
     });
 
+    describe('does not attempt to mutate Component.childContextTypes', () => {
+      const SimpleComponent = createClass({
+        displayName: 'Simple',
+        // contextTypes: Object.freeze({ b: PropTypes.string }),
```
aweary: Should this be deleted or uncommented?

ljharb: deleted, thanks

ljharb:
(this bug originated in #171)

PR #1113, opened 2017-09-15
[BREAKING] Add .rendered() and make shallow() and mount() semantics consistent

lelandrichardson:
to: @ljharb

This PR updates shallow() to have similar semantics to mount() with respect to what the wrapper that is returned by them.

The idea is that mount(<Foo />) and shallow(<Foo />) both return a wrapper that's around the Foo itself, and not what it rendered. This means that calling .type() and .instance() on the resulting wrapper will return Foo and an instance of Foo respectively.

This is essentially the alternative to #1107 

This is a huge breaking change, and i'm not convinced that this is the right API. But putting up this PR for comments. Theoretically, I find it slightly more appealing than #1107, but there might be some things in this PR that are done slightly wrong. Not sure how I feel TBH.

### DIFF
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -37,15 +37,15 @@ describe('shallow', () => {
 
       const wrapper = shallow(<Foo bar />);
 
-      expect(wrapper.type()).to.equal(Box);
-      expect(wrapper.props().bam).to.equal(true);
+      expect(wrapper.type()).to.equal(Foo);
+      expect(wrapper.rendered().type()).to.equal(Box);
+      expect(wrapper.props().bam).to.equal(undefined);
+      expect(wrapper.rendered().props().bam).to.equal(true);
       expect(wrapper.instance()).to.be.instanceOf(Foo);
-      expect(wrapper.children().at(0).type()).to.equal('div');
+      expect(wrapper.rendered().children().type()).to.equal('div');
       expect(wrapper.find(Box).children().props().className).to.equal('div');
       expect(wrapper.find(Box).children().at(0).props().className).to.equal('div');
       expect(wrapper.find(Box).children().props().className).to.equal('div');
-      expect(wrapper.children().type()).to.equal('div');
-      expect(wrapper.children().props().bam).to.equal(undefined);
     });
   });
 
@@ -155,7 +155,7 @@ describe('shallow', () => {
         </div>
       );
       const wrapper = shallow(<Foo foo="qux" />);
-      expect(wrapper.type()).to.equal('div');
+      expect(wrapper.rendered().type()).to.equal('div');
       expect(wrapper.find('.bar')).to.have.length(1);
       expect(wrapper.find('.qoo').text()).to.equal('qux');
     });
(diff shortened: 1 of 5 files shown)


packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -3780,7 +3792,7 @@ describe('shallow', () => {
     const wrapper = shallow(<Foo />);
     expect(wrapper).to.have.length(1);
     expect(wrapper.html()).to.equal(null);
-    expect(wrapper.type()).to.equal(null);
+    expect(wrapper.rendered().length).to.equal(0);
```
ljharb: Why is this not `expect(wrapper.rendered()).to.eql([null])`?

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -2558,7 +2570,7 @@ describe('shallow', () => {
 
           const context = { name: 'foo' };
           const wrapper = shallow(<Foo />);
-          expect(wrapper.find(Bar).shallow({ context }).text()).to.equal('foo');
+          expect(wrapper.find(Bar).shallow({ context }).rendered().text()).to.equal('foo');
```
ljharb: I think this is the same issue as `.first('div')` above, let's fix

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1045,10 +1045,10 @@ describe('shallow', () => {
 
         const context = { x: 'yolo' };
         const wrapper = shallow(<Foo x={5} />, { context });
-        expect(wrapper.first('div').text()).to.equal('yolo');
+        expect(wrapper.text()).to.equal('yolo');
 
         wrapper.setProps({ x: 5 }); // Just force a re-render
-        expect(wrapper.first('div').text()).to.equal('yolo');
```
ljharb: this shouldn't need to be changed

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -4178,7 +4190,7 @@ describe('shallow', () => {
         Foo.displayName = 'CustomWrapper';
 
         const wrapper = shallow(<Wrapper />);
-        expect(wrapper.name()).to.equal('CustomWrapper');
+        expect(wrapper.rendered().name()).to.equal('CustomWrapper');
```
ljharb: add `expect(wrapper.name()).to.equal('Wrapper')`

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -4319,10 +4331,10 @@ describe('shallow', () => {
 
     it('dives + shallow-renders when there is one component child', () => {
       const wrapper = shallow(<DoubleWrapsRendersDOM />);
-      expect(wrapper.is(WrapsRendersDOM)).to.equal(true);
+      expect(wrapper.rendered().is(WrapsRendersDOM)).to.equal(true);
```
ljharb: add `expect(wrapper.is(DoubleWrapsRendersDOM)).to.equal(true)`

packages/enzyme-test-suite/test/_helpers/index.jsx
```diff
@@ -96,7 +96,7 @@ export function generateEmptyRenderData() {
     { message: 'false', value: false, expectResponse: true },
     { message: 'null', value: null, expectResponse: true },
 
-    // Returns false for empty, valid returns
+    // // Returns false for empty, valid returns
```
ljharb: ?

PR #1096, opened 2017-08-29
Remove travis allowed failure

nfcampos:
It passes now. When was this failing?

PR #1119, opened 2017-09-16
Remove Karma flakiness by providing browser shims

lelandrichardson:
This fixes the travis karma flakiness, and removes the allowed failure. This does this by adding in airbnb-browser-shims to the karma tests so that we don't have to worry about which version of chrome they are running on.

### DIFF
--- a/packages/enzyme-test-suite/test/_helpers/setupShims.js
+++ b/packages/enzyme-test-suite/test/_helpers/setupShims.js
@@ -1,0 +1,15 @@
+/* eslint no-undef: 0, global-require: 0 */
+/**
+ * This file is needed only when karma runs the test suite. We can't guarantee
+ * what browser travis will actually run karma with, so we need to load in
+ * browser shims to make sure everything works that we expect. I'd love to
+ * put this somewhere else (ie, karma.conf.js), but I can't figure out how
+ * to tell karma to run a file before everything else. This is the next best
+ * thing I guess...
+ */
+const isBrowser = typeof window !== 'undefined' &&
+  Object.prototype.toString.call(window) === '[object Window]';
+
+if (isBrowser) {
+  require('airbnb-browser-shims');
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,9 +34,6 @@ matrix:
      env: KARMA=true REACT=15
    - node_js: "6"
      env: KARMA=true REACT=16
-  allow_failures:
-    - node_js: "6"
-      env: KARMA=true REACT=16
 env:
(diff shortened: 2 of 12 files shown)


packages/enzyme-test-suite/test/_helpers/setupShims.js
```diff
@@ -0,0 +1,15 @@
+/* eslint no-undef: 0, global-require: 0 */
+/**
+ * This file is needed only when karma runs the test suite. We can't guarantee
+ * what browser travis will actually run karma with, so we need to load in
+ * browser shims to make sure everything works that we expect. I'd love to
+ * put this somewhere else (ie, karma.conf.js), but I can't figure out how
+ * to tell karma to run a file before everything else. This is the next best
+ * thing I guess...
+ */
+const isBrowser = typeof window !== 'undefined' &&
```
ljharb: Why is a typeof window check not sufficient?

packages/enzyme-test-suite/test/_helpers/setupShims.js
```diff
@@ -0,0 +1,15 @@
...
+  Object.prototype.toString.call(window) === '[object Window]';
+
+if (isBrowser) {
+  require('airbnb-browser-shims');
```
ljharb: This is actually both safe to require in node, and recommended to do so.

lelandrichardson: hmm. doing so caused many issues. perhaps it's related to jsdom?

packages/enzyme-test-suite/test/_helpers/setupShims.js
lelandrichardson: because we screw that up w/ jsdom.

lelandrichardson:
This appears to have not actually fixed the flakiness.... so it seems like this is not the reason for the karma flakiness :/

packages/enzyme-test-suite/test/_helpers/setupShims.js
ljharb: Ah, yeah true, that makes sense.

nfcampos:
@lelandrichardson it did fix flakiness caused by lack of polyfills,it just didn't fix the other cause of flakiness,some tests appear to cause a full page reload. This shoudl still be merged

lelandrichardson:
@nfcampos any thoughts on how to fix the page reload issue? It'd be nice to fix karma in a single PR if possible.

ljharb:
Rebased this on top of #1096.

PR #1120, opened 2017-09-16
Add karma tests to allowed failures

lelandrichardson:
Adding this in temporarily since karma is sporadically failing. I'm investigating the fix, but for now I don't think we should mark these as required

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,16 @@ matrix:
    - node_js: "6"
      env: KARMA=true REACT=16
   allow_failures:
-    - node_js: "6"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -35,8 +35,16 @@ matrix:
    - node_js: "6"
      env: KARMA=true REACT=16
   allow_failures:
-    - node_js: "6"
-      env: KARMA=true REACT=16
+   - node_js: "6"
```
ljharb: these all need another space of indentation

lelandrichardson: lol WAT. how did that happen

PR #1127, opened 2017-09-18
Add public root method

lelandrichardson:
This adds a public `root()` method to the shallow and mount wrappers. This is a reasonable public API to have, and chai-enzyme needs this for a couple of things.


ljharb:
@lelandrichardson Can we add some tests for this method?

PR #1133, opened 2017-09-19
Fix componentDidUpdate when updating by setState on v16

koba04:
In v16, ShallowRenderer doesn't call `componentDidUpdate` even updating by `setState`. 
https://github.com/facebook/react/pull/10372
This PR is to fix this.


koba04:
l know ShallowRenderer v16 calls cDU without a context argument but I didn't include that because I think it should be fixed separately.

packages/enzyme-adapter-react-13/src/ReactThirteenAdapter.js
```diff
@@ -143,6 +143,7 @@ class ReactThirteenAdapter extends EnzymeAdapter {
     let isDOM = false;
     let cachedNode = null;
     return {
+      disableComponentDidUpdateOnSetState: true,
```
ljharb: This seems like a really awkward way to extend the adapter interface.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -3403,8 +3403,7 @@ describe('shallow', () => {
 
     context('updating state', () => {
       // NOTE: There is a bug in react 16 shallow renderer where prevContext is not passed
-      // into componentDidUpdate. Skip this test for react 16 only. add back in if it gets fixed.
```
ljharb: prevContext *should* be passed into componentDidUpdate; is react not going to fix this?

packages/enzyme-adapter-react-13/src/ReactThirteenAdapter.js
koba04: What about this?

```js
return {
  options: {
    disableComponentDidUpdateOnSetState: true
  }
```

I can do this without to extend the adapter interface but it requires to access the React version directly from ShallowWrapper. I think ShallowWrapper should depend on adapters only.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
koba04: Yes. #1139 is a PR for that.
You can see the change at [Unreleased] section in CHANGELOG.md 

https://github.com/facebook/react/blob/master/CHANGELOG.md

ljharb: Then if that's fixed, why do we need an option in the adapter?

koba04: The PR is to support that componentDidUpdate no longer receives prevContext argument, which is not only ShallowRenderer.
This PR is to support that ShallowRenderer no longer call componentDidUpdate.
These are different fixes so I added the options each other.

ljharb: Why does componentDidUpdate not receive a prevContext argument anymore? It seems like it should.

koba04: The reason seems to be the following. cc: @bvaughn

> Even with that parameter present- if context isn't managed in the same way as props and state (PR #8655) then it could be inconsistent between the instance and the fiber.

> Update: Chatted in meatspace with @acdlite about this. He pointed out another existing stack limitation with fiber WRT shouldComponentUpdate. (Seems like context is already flaky but fiber will make it more so.) The only safe way to use context is to pass singleton/static props (eg the way react-redux Provider passes down store) that components subscribe to for changes.

https://github.com/facebook/react/pull/8631#issuecomment-270450606

ljharb: Wow, that seems like a pretty major change to the way context works then.

I've not found context to be flaky at all in React <= 15.

@lelandrichardson any suggestions for a better name for this part of the adapter interface?

bvaughn: > Why does componentDidUpdate not receive a prevContext argument anymore? It seems like it should.

@ljharb Check out facebook/react/pull/8631 for a description of the options we considered.

> The only safe way to use context is to pass singleton/static props...that components subscribe to for changes.

Yes. That sounds accurate.

packages/enzyme-adapter-react-13/src/ReactThirteenAdapter.js
lelandrichardson: I wonder if the option should actually be the opposite. `enableComponentDidUpdateOnSetState`, making it an opt-in flag for adapters, and one that only react 16 will be using?

koba04: @lelandrichardson I think ShallowRenderer won't support componentDidUpdate in future releases so the flag is required by only versions before v16. Using `enableComponentDidUpdateOnSetState` means that enzyme has to set the flag for future versions of React permanently. So I used `disableComponentDidUpdateOnSetState` for this.
But it might make sense to use `enableComponentDidUpdateOnSetState` and set the flag until dropping v15 support.

lelandrichardson: that makes sense.  I think there might be more use cases for adding "options" for the adapter. I think i like giving them their own namespace though: `options.enableComponentDidUpdateOnSetState` or something.

Also, I wonder if it's better to put this on the adapter instance itself instead of the object it returns from `createRenderer`?  This way, we can set an empty `this.options = { ... }` in the `EnzymeAdapter` base constructor, which would mean people wouldn't have to set defaults.

Thoughts?

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -145,6 +145,12 @@ function nodeToHostNode(_node) {
 }
 
 class ReactSixteenAdapter extends EnzymeAdapter {
+  constructor() {
+    super();
+    this.options = { ...this.optins,
```
lelandrichardson: typo

packages/enzyme-adapter-react-13/src/ReactThirteenAdapter.js
koba04: That makes sense! I've updated this PR.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
koba04: Fixed

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -145,6 +145,12 @@ function nodeToHostNode(_node) {
 }
 
 class ReactSixteenAdapter extends EnzymeAdapter {
+  constructor() {
+    super();
+    this.options = { ...this.options,
```
ljharb: `...this.options` needs to go on its own line

packages/enzyme-test-suite/package.json
```diff
@@ -40,4 +40,4 @@
     "react": "^15.5.0",
     "react-dom": "^15.5.0"
   }
-}
+}
```
ljharb: all files need to have a trailing newline

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -369,7 +369,39 @@ class ShallowWrapper {
...
     this.single('setState', () => {
       withSetStateAllowed(() => {
-        this.instance().setState(state, callback);
+        const adapter = getAdapter(this[OPTIONS]);
+        const instance = this.instance();
+        const prevProps = instance.props;
+        const prevState = instance.state;
+        const prevContext = instance.context;
+        let shouldRender = true;
+        // dirty hack: avoid calling shouldComponentUpdate twice
```
ljharb: why is this needed? can you elaborate in the comment?

koba04: Added a comment. Does this make sense?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -369,7 +369,42 @@ class ShallowWrapper {
...
+        let originalShouldComponentUpdate;
+        if (
+          this[OPTIONS].lifecycleExperimental &&
+          adapter.options.enableComponentDidUpdateOnSetState &&
+          instance &&
+          typeof instance.shouldComponentUpdate === 'function'
+        ) {
+          originalShouldComponentUpdate = instance.shouldComponentUpdate;
+          instance.shouldComponentUpdate = (...args) => {
+            shouldRender = originalShouldComponentUpdate.apply(instance, args);
```
ljharb: should this function also `instance.shouldComponentUpdate = originalShouldComponentUpdate`?

I wonder if this should use `sinon` as a dependency rather than overwriting; what about things like enumerability, setters, etc?

lelandrichardson:
made small tweak based on @ljharb's comment. merging now in interest of time. Thanks!

packages/enzyme/src/ShallowWrapper.js
koba04: > I wonder if this should use sinon as a dependency rather than overwriting; what about things like enumerability, setters, etc?

Yeah, I can write this like the following if I use `sinon`.

```js
        let shouldRender = true;
        let spy;
        if (instance.shouldComponentUpdate) {
          spy = sinon.spy(instance, 'shouldComponentUpdate');
        }
        if (spy) {
          shouldRender = spy.getCall(0).returnValue;
          spy.restore();
        }
```

Also `sinon` seems to have a better support about enumerability so I think it's worth to consider this after releasing v3. ES2015 Proxy might be an option for that.

koba04:
Thanks!

packages/enzyme/src/ShallowWrapper.js
ljharb: I'd prefer to avoid Proxy; I'm not sure it would work here anyways, and it would very tightly constrict the engines enzyme can run on.

koba04: True. To support `< Node v6`, `sinon.spy` would be nice.

ljharb: A followup PR is always appreciated :-D

koba04: I'll work on it 👍

austinh:
Has this been merged into v3.1.0? Because I cannot get componentDidUpdate to run when using setState

ljharb:
Yes, it's in v3.1; please file a new issue if you have trouble still.

PR #1139, opened 2017-09-23
Fix componentDidUpdate no longer receives prevContext on React v16

koba04:
This PR is to fix https://github.com/facebook/react/pull/8631.
Currently, a test is failing because of #1133.

This PR depends on #1133 so I should apply this patch to the changes of #1133 when the PR is merged.


lelandrichardson:
We will also want to make this on `adapter.options` as well instead of the renderer

koba04:
I'll work on this after #1133 is merged.

koba04:
Thanks!

PR #1140, opened 2017-09-23
Breaking: lifecycleExperimental by default

koba04:
There is a discussion in #678 about this.
If enzyme chooses this option, the time has come.

I think this PR needs more works, which are tests and documentation.
Does enzyme enable lifecyleExperimental by default at v3?

### DIFF
--- a/packages/enzyme/src/ShallowWrapper.js
+++ b/packages/enzyme/src/ShallowWrapper.js
@@ -14,6 +14,7 @@ import {
   isCustomComponentElement,
   ITERATOR_SYMBOL,
   getAdapter,
+  makeOptions,
   sym,
   privateSet,
   cloneElement,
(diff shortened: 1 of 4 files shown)


koba04:
I'm working on rebasing and fixing a test.

koba04:
Thanks!

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -108,8 +109,9 @@ function getRootNode(node) {
  * @class ShallowWrapper
  */
 class ShallowWrapper {
-  constructor(nodes, root, options = {}) {
-    validateOptions(options);
+  constructor(nodes, root, passedOptions = {}) {
+    validateOptions(passedOptions);
```
ljharb: How come ReactWrapper doesn't validate its options?

Issue #1163, opened 2017-09-27
wrapper is not updated when using conditional rendering in v3


yesmeck:
I found that when using conditional rendering, after setProps, wrapper is not updated properly.

```javascript
// src/Hello.js
import React from 'react';

export default ({ show }) => (
  <div>
    {show ? <span>hello</span> : null}
  </div>
);
```

```javascript
// tests/Hello.test.js
import { mount } from 'enzyme';
import React from 'react';
import Hello from '../src/Hello';

test('Hello', () => {
  const wrapper = mount(<Hello show />)

  expect(wrapper.find('span').length).toBe(1)

  wrapper.setProps({ show: false });
  expect(wrapper.find('span').length).toBe(0)

  wrapper.setProps({ show: true });
  wrapper.update(); // even though calling wrapper.update explicitly
  console.log(wrapper.debug()); // no span
  console.log(wrapper.html()); // has span
  expect(wrapper.find('span').length).toBe(1) // failed
});
```

Here is a reproduction repo https://github.com/yesmeck/enzyme-set-props-bug

ljharb:
What version of React are you using?

yesmeck:
```
"react": "^16.0.0",
"react-dom": "^16.0.0"
```

Here is the full dependencies https://github.com/yesmeck/enzyme-set-props-bug/blob/master/package.json

tricoder42:
I can confirm this behavior with `enzyme@3.0.0` and `enzyme-adapter-react-16`. Upgrading to `enzyme@3.0.0`, but keeping `react@15` works fine.

neoziro:
Same issue for me, update is not working for React 16.

neoziro:
I investigated it and found the source of the bug:

https://github.com/airbnb/enzyme/blob/master/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js#L66

At the first render, vnode.memoizedProps is the source of truth.
At the second, vnode.alternate.memoizedProps is the source of truth.
At the third, vnode.memoizedProps is the source of truth.

That was it is named `alternate` I think. We should confirm this from the React team and ask them the correct way to handle it.

Should we alternate the node at each render? @gaearon @sebmarkbage @acdlite

robin-drexler:
Having the same issue with react adapter ^15.5.0.

Axbon:
[...]

Issue #1153, opened 2017-09-26
enzyme@3: wrapper is not updated even if there was a new render


deepsweet:
Hi. Just faced this after updating to v3.

Given the following deps:

```
├─ enzyme-adapter-react-16@1.0.0
├─ enzyme@3.0.0
├─ react-dom@16.0.0
├─ react-test-renderer@16.0.0
└─ react@16.0.0
```

And the HOC:

```js
import React, { Component } from 'react'

const testHOC = (Target) => class TestClass extends Component {
  constructor (...args) {
    super(...args)

    this.state = {}
    this.onTest = this.onTest.bind(this)
  }

  onTest () {
    this.setState({
      foo: true
    })
  }

  render () {
    console.log(this.state)

    return (
      <Target {...this.props} {...this.state} onTest={this.onTest} />
    )
  }
}

export default testHOC
```

And the test:

```js
import React from 'react'
import { mount } from 'enzyme'
import testHOC from './testHOC'

describe('testHOC', () => {
  it('should work', () => {
    const Target = () => null
    const EnhancedTarget = testHOC(Target)
    const wrapper = mount(
      <EnhancedTarget />
    )

    wrapper.find(Target).prop('onTest')()

    console.log(
      wrapper.debug()
    )
  })
})

```

I see:

```
  console.log testHOC.js:18
    { foo: true }

  console.log testHOC.spec.js:15
    <TestClass>
      <Target onTest={[Function]} />
    </TestClass>
```

So the was a new render, but wrapper just didn't react to it. Works fine with v2.

Please correct me if I was wrong all this time and test should be written in another way.

ljharb:
Wrappers are immutable in enzyme 3; this is in the migration docs.

Try `wrapper.update()` after invoking your function.

deepsweet:
Already tried and now re-tried, same result.

j4k:
I'm also experiencing the same behaviour with the same deps on some of my component tests (non-HOC) where the state updates, but the wrapper does not, even after calling `.update()`. 

Placing console.log's in the component under tests render method show's things rendering correctly during the test, I just can't seem to assert properly after an `.update()`, `.debug()` outputs the initial component. `.state()` shows the correct state after an `.update()`

yesmeck:
Im also experiencing the same behavior when using conditional rendering, here is a reproduction repo https://github.com/yesmeck/enzyme-set-props-bug.

yesmeck:
[...]

PR #1169, opened 2017-09-27
Fix node resolving in React 16 adapter

neoziro:
I added a test relative to https://github.com/airbnb/enzyme/issues/1163#issuecomment-332438576

I tried to fix it but I couldn't figure out when to use alternate node.

### DIFF
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -19,6 +19,7 @@ import {
   createMountWrapper,
   propsWithKeysAndRef,
 } from 'enzyme-adapter-utils';
+import findCurrentFiberUsingSlowPath from './findCurrentFiberUsingSlowPath';
 
 const HostRoot = 3;
 const ClassComponent = 2;
@@ -63,7 +64,7 @@ function toTree(vnode) {
   // TODO(lmr): I'm not really sure I understand whether or not this is what
   // i should be doing, or if this is a hack for something i'm doing wrong
   // somewhere else. Should talk to sebastian about this perhaps
-  const node = vnode.alternate !== null ? vnode.alternate : vnode;
+  const node = findCurrentFiberUsingSlowPath(vnode);
   switch (node.tag) {
     case HostRoot: // 3
       return toTree(node.child);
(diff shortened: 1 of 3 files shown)


neoziro:
@lelandrichardson It is probably not the cleaner way to do it but it fixes a huge bug. I think merging it and publishing a patch would be great!

gaearon:
Do you have a plan for what we need to expose on React side to remove this hack?

marvinpinto:
I was running into a an issue similar to https://github.com/airbnb/enzyme/issues/1153 so I followed @gaearon's advice and tried this patch out. It seemed to do the trick for me :+1:

I had to sprinkle `wrapper.update()` after dom-manipulating events and it worked fine for the most part.

Here is what I did to test this locally:

- Cloned https://github.com/neoziro/enzyme.git
- Switched to the `react-16-updated-props` branch.
- Inside the cloned `enzyme` repo: `lerna bootstrap && lerna run build`
- Then replaced the `enzyme`, `enzyme-adapter-react-16`, and `enzyme-adapter-utils` packages (in my `node_modules` directory) with the newly built lerna alternatives (in the `packages` directory).

I ran into an oddity where enzyme was throwing the following error:
``` text
Enzyme Internal Error: configured enzyme adapter did not inherit from the EnzymeAdapter base class
```
I assumed it was something weird related to how I built this package and so I just commented out the relevant lines in `validateAdapter.js` - https://github.com/neoziro/enzyme/blob/react-16-updated-props/packages/enzyme/src/validateAdapter.js#L16-L21

aweary:
> Do you have a plan for what we need to expose on React side to remove this hack?

@gaearon ideally React would expose a way to call `toTree` on an instance. This logic is already duplicated from the test renderer, so unifying and exporting it could mean that future adapters don't need to implement `toTree` themselves.

lelandrichardson:
@neoziro thanks for hitting this! I think this is good and am going to merge. It puts us in a strictly better state than where we are now.

That said, I'm definitely interested in hashing out more details with @aweary and @gaearon around getting the react 16 adapter to a state where it doesn't depend on any internals. I agree with Brandon that this likely doesn't need to be much more than just exposing a `ReactTestUtils.toTree(instance)` or something like this. The function itself is kind of already built for the test renderer, but isn't publicly exposed.

Perhaps this wednesday (at react wednesday) could be a good time to hash out some of these details.

In the mean time, I'm going to merge this to get some people unblocked.

Oh, also, @marvinpinto the error you got is indicative of multiple copies of `enzyme` being installed in your `node_modules` directory, thus `instanceof` doesn't match between them.

neoziro:
@lelandrichardson you are welcome! It was a pleasure to help!

PR #1179, opened 2017-09-27
feat: Added hostNodes method to ReactWrapper

### LINKED ISSUE #1174: Enzyme 3.0.0 "find()" queries not just DOM elements
I just switched to Enzyme 3.0.0 and React 16 and I noticed that some of my tests are failing because Enzyme `.find` is finding more than one element with the given ID.

```
// App.js
function Test(props) {
  return <div {...props} />
}

export default function App() {
  return <Test id="foobar" />;
}
```

```
// App.test.js

it('finds a single DOM node with the given ID', () => {
  expect(mount(<App />).find('#foobar').length).toBe(1); // fails saying that it found 2 elements
});
```

Repro repository here:
https://github.com/FezVrasta/enzyme-id-bug-report

Just run `yarn && yarn test` to reproduce.

I don't think this is the expected behavior because:

1. It didn't work this way previously
2. The `.find` method is now looking for DOM Elements **AND** React elements with the given ID instead of just looking at the DOM elements.

If this is the expected behavior, how can I obtain the previous result? Is there a `findDOMNode` or something similar?
Or maybe an helper function that strips from the selection all the "non DOM" elements?

Thanks!

FezVrasta:
fixes #1174


FezVrasta:
Going to add some tests tomorrow!

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -393,6 +393,17 @@ describeWithDOM('mount', () => {
...
+    
+    it('should strip out any non-hostNode', () => {
+      const Foo = props => <div {...props} />;
+      const wrapper = mount(
+        <div>
+          <Foo className="foo" />
+        </div>,
+      );
+      expect(wrapper.find('.foo').length).to.equal(2);
+      expect(wrapper.find('.foo').hostNodes().length).to.equal(1);
```
ljharb: could we assert on the contents too, not just the length? something like:
```js
const hostNodes = wrapper.find('.foo').hostNodes();
expect(hostNodes).to.have.lengthOf(1);
expect(hostNodes.is('div')).to.equal(true);
```

FezVrasta: 😩 I was trying to perform the PR trough the GitHub UI but I guess I'll have to clone the project locally 😅

FezVrasta: done!

FezVrasta:
Sorry, it slipped in accidentally. Done.

ljharb:
(if you need help rebasing, i'm happy to take care of it)

FezVrasta:
I'm not sure if the rebase ended up well 🤔

FezVrasta:
Ok now it looks much better 😅

ljharb:
@FezVrasta still looks like there's a merge commit; would you like me to rebase it for you?

FezVrasta:
I don't see merge commits?
image

ljharb:
aha, now it's good :-)

FezVrasta:
Still, Travis doesn't look that happy... 🙄

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -399,6 +399,20 @@ describe('shallow', () => {
...
+    it('should strip out any non-hostNode', () => {
+      const Foo = props => <div {...props} />;
+      const wrapper = shallow(
+        <div>
+          <Foo className="foo" />
+          <div className="foo" />
+        </div>,
+      );
+      const hostNodes = wrapper.find('.foo').hostNodes();
+      expect(wrapper.find('.foo')).to.have.length(2);
```
ljharb: maybe try `lengthOf` instead of `length` here?

FezVrasta: Other tests in the same page use `.length`, why so?

ljharb: They should all be using lengthOf; it was just a thought about fixing the travis tests.

FezVrasta: It doesn't seem to help

ljharb:
k, rebased, added some docs, and fixed the issue (it needs `.filterWhere`; `.filter` takes a string, not a function).

ljharb:
and finally, React 13 doesn't support SFCs; so the component in the tests needed to be class-based.

docs/api/ShallowWrapper/hostNodes.md
```diff
@@ -0,0 +1,18 @@
...
+#### Returns
+
+`ShallowWrapper`: A new wrapper that wraps the filtered nodes.
+
+
+
+#### Examples
+
+```jsx
+const wrapper = mount(<div><MyComponent className="foo" /><div className="foo" /></div>);
```
pfhayes: should this be `shallow`?

ljharb: Whoops, good catch! Want to submit a PR?

TheSharpieOne:
This is kinda critical for migration from v2 to v3. Not trying to rush, but I don't want to [monkey patch](https://github.com/airbnb/enzyme/issues/1174#issuecomment-332603355) this... so any plans on releasing this sooner than later?

ljharb:
I think we're hoping to cut a release this week.

lelandrichardson:
@TheSharpieOne enzyme v3.1.0 was just published. enjoy.

docs/api/ShallowWrapper/hostNodes.md
pfhayes: https://github.com/airbnb/enzyme/pull/1215

PR #1186, opened 2017-09-28
auto update wrappers behind a flag

### LINKED ISSUE #1175: Restore wrapper auto-update behavior behind a feature flag?
The breaking change in v3 around needing `.update()` calls is responsible for the vast majority of our broken tests after upgrading:
```
right after react@16, enzyme@3 upgrades:
Tests:       1232 failed, 4816 passed, 6048 total

after patching auto-update behavior back into enzyme by editing node_modules:
Tests:       325 failed, 5723 passed, 6048 total
```

I did see the mention in the migration guide that it was a conscious choice, but I'd disagree with it being an uncommon need. There are two things that drive a high level of need for .update for us -- leveraging component state heavily and making AJAX calls directly in components. The former leads to lots of lifting state up and letting child components change state in parents (`<ChildComponent prop={() => this.setState({ x: 1 })} />`). The latter leads to async setState calls.

If a project leverages external data stores like Redux, or uses some kind of library that externalizes data fetching like Apollo, everything comes in through props and I would imagine the need for .update likely goes down drastically.

Could we restore the auto-update behavior behind a flag (ideally global 😅)? The implementation is basically just the guts of .update put into .getNodeInternal functions:
```js
if (this[ROOT] === this) {
  this[NODE] = getRootNode(this[RENDERER].getNode());
  this[NODES] = [this[NODE]];
}
```

jwbay:
This restores the auto update feature introduced by #490 for v3, but behind a feature flag. `mount` now needs the same treatment to keep the tree in sync that shallow used to. Fixes #1175.

For what it's worth, since pre-v3 both shallow and mount essentially auto-updated by default, the only Enzyme tests that break *without* the flag checks are these two, which are fixable by asserting with deep equality if that's on the table.
https://github.com/airbnb/enzyme/blob/106b5d14c79a1dfc4d3ee13f87b13ffb609c5c4a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx#L3696-L3710

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -3683,6 +3683,55 @@ describeWithDOM('mount', () => {
     });
   });
 
+  describe('out-of-band state updates with autoUpdate', () => {
+    class Child extends React.Component {
+      render() {
+        return <span />;
+      }
+    }
+
+    class Test extends React.Component {
+      componentWillMount() {
+        this.state = {};
+      }
+
+      asyncSetState() {
+        setImmediate(() => {
+          this.setState({ showSpan: true });
+        });
+      }
+
+      callbackSetState() {
+        this.setState({ showSpan: true });
+      }
+
+      render() {
+        return (
+          <div>
+            {this.state && this.state.showSpan && <span className="show-me" />}
+            <button className="async-btn" onClick={() => this.asyncSetState()} />
+            <Child callback={() => this.callbackSetState()} />
+          </div>
+        );
+      }
+    }
+
+    it('should have updated output after an asynchronous setState', (done) => {
+      const wrapper = mount(<Test />, { autoUpdate: true });
+      wrapper.find('.async-btn').simulate('click');
+      setImmediate(() => {
+        expect(wrapper.find('.show-me').length).to.equal(1);
+        done();
+      });
+    });
+
+    it('should have updated output after child prop callback invokes setState', () => {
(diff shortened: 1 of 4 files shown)


koba04:
This PR seems to be reasonable to me.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -3683,6 +3683,55 @@ describeWithDOM('mount', () => {
...
+            <button className="async-btn" onClick={() => this.asyncSetState()} />
+            <Child callback={() => this.callbackSetState()} />
+          </div>
+        );
+      }
+    }
+
+    it('should have updated output after an asynchronous setState', (done) => {
+      const wrapper = mount(<Test />, { autoUpdate: true });
+      wrapper.find('.async-btn').simulate('click');
```
ljharb: i'd prefer to avoid `simulate`; can this do `.prop('onClick')()` instead?

packages/enzyme/src/ReactWrapper.js
```diff
@@ -675,7 +683,7 @@ class ReactWrapper {
    */
   parents(selector) {
     const allParents = this.wrap(
-      this.single('parents', n => parentsOfNode(n, this[ROOT].getNodeInternal())),
+      this.single('parents', n => parentsOfNode(n, this[ROOT][NODE])),
```
ljharb: This seems like it might be a bugfix that should go in separately from an `autoUpdate` option?

ljharb:
@jwbay are you no longer interested in continuing this?

Issue #1181, opened 2017-09-27
Immutable.js iterable children not rendering in v3


kmcq:
Hello!

Congrats on the v3 release! It must feel awesome to have this rewrite released.

I have noticed that in v3 Immutable JS children are not "rendering":

```js

// App.js

import Immutable from 'immutable';

class App extends Component {
  render() {
    return (
      <div className="App">
        {Immutable.List(['first', 'second', 'third']).map(item => <span key={item}>{item}</span>)}
      </div>
    );
  }
}

// App.test.js

import { shallow as enzyme2shallow } from 'enzyme2';
import { shallow as enzyme3shallow } from 'enzyme';

it('renders immutable children in enzyme 2', () => {
  expect(enzyme2shallow(<App />).debug()).toMatchSnapshot();
});

it('does not render immutable children in enzyme 3', () => {
  expect(enzyme3shallow(<App />).debug()).toMatchSnapshot();
});

// App.test.js.snap

exports[`does not render immutable children in enzyme 3 1`] = `
"<div className=\\"App\\">
  <undefined />
</div>"
`;

exports[`renders immutable children in enzyme 2 1`] = `
"<div className=\\"App\\">
  <span>
    first
  </span>
  <span>
    second
  </span>
  <span>
    third
  </span>
</div>"
`;
```

If I add a `toArray()` call to that `List` then this renders as expected, but since React accepts this code, I think enzyme should as well.

I made a repo that shows this bug as simply as I could think: https://github.com/kmcq/immutable-children-enzyme. That's where those code snippets come from.

Thank you so much! Let me know if I can help fix this. Cheers!

kmcq:
hopefully also relevant: I used the `react-16` adapter for this example, but at work we're using the `react-15.4` adapter and I had the same issue.

stevenmusumeche:
@kmcq that you for that PR, this is the last step before we can upgrade to 16 in production.

Apercu:
@ljharb Hey 😃 Do you mind publishing a patch release for this? Not sure if there is a process from the contributing doc, thanks!

ljharb:
It’s on my list to get to soon.

PR #1187, opened 2017-09-28
Support array-like iterable children in v3

kmcq:
## What this is

Fixes https://github.com/airbnb/enzyme/issues/1181.

This adds support for non-Array, but still iterable, children, such as Immutable-js Lists, Sets, etc., which is supported by React:
```js
const Component = () => {
  const listOfThings = Immutable.List([1, 2, 3]);
  return <div>{listOfThings.map(thing => <span>{thing}</span>)}</div>;
};
```

The React PR that originally added this functionality is here: https://github.com/facebook/react/pull/2376.

## The code

- Check to see if `children` is iterable, and if so, use `Array.from` to convert it to an array.

I like this approach because it compartmentalizes the knowledge that children are sometimes things other than arrays.


packages/enzyme-adapter-utils/src/elementToTree.js
```diff
@@ -0,0 +1,65 @@
+const ITERATOR_SYMBOL = typeof Symbol === 'function' && Symbol.iterator;
```
kmcq: I saw that enzyme has this elsewhere, but in a different package. If there's a better way to share this across lerna packages, please let me know 😄

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -81,43 +88,6 @@ export function assertDomAvailable(feature) {
   }
 }
 
-export function nodeTypeFromType(type) {
```
kmcq: Moved this to its own file so that `elementToTree` could still use it

kmcq:
after more searching around the codebase some more, I see that the `elementToTree` that I patched is not the only `elementToTree` (v13 has its own) and also there are potentially others places that would have to change for this to work. I'm also not convinced that the tests I wrote are the best way to test that this works (I left them failing on purpose).

I'm going to pause on this for now to see if any maintainers have ideas about how best to tackle this.

lelandrichardson:
Interesting! I didn't know React had this capability.

kmcq:
fyi I plan on coming back to this next week!

kmcq:
@lelandrichardson I have changed this to now be about adding Immutable.js support. Please see the updated description and let me know if you have any questions!

kmcq:
Hi @ljharb thanks for the review. I have made this more generic -- I did some more research and found out that all iterables support `Array.from`. This seems more in line with the language spec. Please read again and let me know :)

packages/enzyme-adapter-react-13/src/ReactThirteenElementToTree.js
```diff
@@ -23,6 +24,8 @@ export default function elementToTree(el) {
   let rendered = null;
   if (Array.isArray(children)) {
     rendered = flatten(children, true).map(elementToTree);
+  } else if (isIterable(children) && typeof children !== 'string') {
+    rendered = flatten(Array.from(children), true).map(elementToTree);
```
ljharb: since we use babel, this could also be `flatten([...children], true)`

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -95,6 +95,15 @@ export function nodeTypeFromType(type) {
   return 'function';
 }
 
+export function isIterable(obj) {
+  return (
+    typeof Symbol === 'function' &&
+    typeof Symbol.iterator === 'symbol' &&
+    obj != null &&
```
ljharb: i'd probably move the `obj` check higher, since that's faster then checking Symbols

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -104,6 +113,8 @@ export function elementToTree(el) {
   let rendered = null;
   if (Array.isArray(children)) {
     rendered = flatten(children, true).map(elementToTree);
+  } else if (isIterable(children) && typeof children !== 'string') {
+    rendered = flatten(Array.from(children), true).map(elementToTree);
```
ljharb: similarly, `flatten([...children], true)`

packages/enzyme-test-suite/test/RSTTraversal-spec.jsx
```diff
@@ -183,6 +183,36 @@ describe('RSTTraversal', () => {
...
+        <div key="a" />,
+        <div key="b" />,
+      ]);
+      const node = $(
+        <div>
+          {twoDivIterable}
+        </div>,
+      );
+      treeForEach(node, spy);
+      expect(spy.callCount).to.equal(3);
```
ljharb: Could we also use `expect(spy.calls).to.deep.equal` in both places, and assert that it's called with the right items?

It'd also be great to test with a Map as well as a Set

kmcq: Yes for sure, I like that. The tree objects are pretty complex, how do you feel about just checking part of it, say keys? Like this:

```js
it('should handle Map children', () => {
  const spy = sinon.spy();
  const twoDivMap = new Map([
    [<div key="a" />],
    [<div key="b" />],
  ]);
  const node = $(
    <div key="root">
      {twoDivMap}
    </div>,
  );
  treeForEach(node, spy);
  expect(spy.callCount).to.equal(3);
  const keys = spy.args.map(arg => arg[0].key);
  expect(keys).to.deep.equal(['root', 'a', 'b']);
});

```

packages/enzyme-adapter-utils/src/Utils.js
kmcq: Awesome, I wasn't aware of that! Thanks!

packages/enzyme-test-suite/test/RSTTraversal-spec.jsx
kmcq: ah nevermind, even better I can just use `$` and check the whole thing:
```js
it('should handle Map children', () => {
  const spy = sinon.spy();
  const twoDivMap = new Map([
    [<div key="a" />],
    [<div key="b" />],
  ]);
  const divA = $(<div key="a" />);
  const divB = $(<div key="b" />);
  const node = $(
    <div key="root">
      {twoDivMap}
    </div>,
  );
  treeForEach(node, spy);
  expect(spy.callCount).to.equal(3);
  const nodes = spy.args.map(arg => arg[0]);
  expect(nodes).to.deep.equal([node, divA, divB]);
});
```

I'll get this committed soon!

kmcq:
Alrighty! @ljharb travis is happy and I think I answered your comments. Can you please review again?

kmcq:
Thanks! Actually the v13 adapter is special-cased, the others all use `enzyme-adapter-utils`' version of `elementToTree`, which is part of this!

stevenmusumeche:
Just wanted to chime in that I am also having this issue with ImmutableJS iterables.  Any idea on when this PR would be merged and released?

ljharb:
Waiting on a couple more reviews first.

hellosmithy:
Also having the same issue, and know of others waiting on this too. I would argue this is not a minor issue, it's breaking functionality that worked before and causes tests to fail on any projects making heavy use of ImmutableJS. Is there any chance of getting the extra reviews done to push this through? I'm happy to help if I can.

Is it worth adding a more generic iterables test e.g.

```
it('should handle iterable children', () => {
  const spy = sinon.spy();
  const iterable = {
    [Symbol.iterator]: function* () {
      yield <div key="a" />;
      yield <div key="b" />;
    }
  };
  const divA = $(<div key="a" />);
  const divB = $(<div key="b" />);
  const node = $(
    <div>
      {iterable}
    </div>,
  );
  treeForEach(node, spy);
  expect(spy.callCount).to.equal(3);
  const nodes = spy.args.map(arg => arg[0]);
  expect(nodes).to.deep.equal([node, divA, divB]);
});
```

ljharb:
@hellosmithy wait, rendering iterables worked in v2?

stevenmusumeche:
Yes, it works in v2, and has for over a year.

kmcq:
Yea it currently works in v2. I put a small example app in #1181 that runs v2 and v3 to show the difference

kevin940726:
This issue is also blocking our team to upgrade react and enzyme 🙁 . Would be nice to merge it soon.

packages/enzyme-adapter-react-13/src/ReactThirteenElementToTree.js
```diff
@@ -23,6 +24,8 @@ export default function elementToTree(el) {
   let rendered = null;
   if (Array.isArray(children)) {
     rendered = flatten(children, true).map(elementToTree);
+  } else if (isIterable(children) && typeof children !== 'string') {
+    rendered = flatten([...children], true).map(elementToTree);
```
hellosmithy: This is not currently working for nested iterables. E.g.

```
<div>
  { iterable1 }
  { iterable2 }
</div>
```

I think we might need a recursive flatten function that caters for either an array or an iterable:

```
export function flatten(arrs) {
  return arrs.reduce((flattened, item) =>
    flattened.concat(Array.isArray(item) || isIterable(item) ? flatten([...item]) : [item]), []);
}
```

packages/enzyme-test-suite/test/RSTTraversal-spec.jsx
```diff
@@ -183,6 +183,63 @@ describe('RSTTraversal', () => {
...
+      const spy = sinon.spy();
+      const twoDivArray = [
+        <div key="a" />,
+        <div key="b" />,
+      ];
+      const divA = $(<div key="a" />);
+      const divB = $(<div key="b" />);
+      const node = $(
+        <div>
+          {twoDivArray}
```
hellosmithy: Suggest we add tests for sibling children (nested arrays and also nested iterables) e.g.

```
<div>
  {array1}
  {array2}
</div>
```

```
<div>
  {iterable1}
  {iterable2}
</div>
```

packages/enzyme-adapter-react-13/src/ReactThirteenElementToTree.js
ljharb: good call; recursion may be required here

kmcq:
@hellosmithy I appreciate the review! Thanks for catching that this didn't support sibling iterables. I added code to support them (thanks for helping to write that!) and tests for sibling arrays, Maps, and Sets. Lemme know what you think :)

stevenmusumeche:
Any idea when this might be released?

ljharb:
@stevenmusumeche when a few more collaborators have reviewed it.

kmcq:
would it be helpful if I periodically rebase this? otherwise I can just do it once we're ready

stevenmusumeche:
I hate to be a pest, but is there a way to move the review up in priority for the other collaborators?  The code and tests look great to me, but I am not intimately familiar with the codebase.

It is something that worked before in v2 and doesn't in v3 and is preventing quite a few companies from upgrading to React 16.

ljharb:
@kmcq please do rebase it, yet :-)

kmcq:
rebased!

stevenmusumeche:
Thank you all so much for what you do.

kmcq:
Thanks @ljharb!

I appreciate everyone's reviews. This was my first time contributing to a library like this and you all made it fun and I learned some things too. Cheers!

kamui:
Thanks @kmcq!

Any ETA on when a new version will be cut? This was the only thing blocking us from upgrading to react 16. Thanks!

Inucoder:
Looking forward for this to be released! Great work guys!

ljharb:
@kamui no ETA yet. I'm shooting for this week, ASAP.

ljharb:
All adapters, as well as enzyme-adapter-utils, have been bumped.

kmcq:
Hey @ljharb, sorry to bug you but it looks like the React 15 adapter's release didn't make it to NPM: https://www.npmjs.com/package/enzyme-adapter-react-15. I checked and the rest look good 👍 Is that something that you need to handle or is that one just lagging behind? Thanks!

Inucoder:
Hey guys, I'm running enzyme `3.1.0`, enzyme-adapter-react-15 `1.0.4`, enzyme-adapter-utils `1.1.0` along with Karma.

```
'use strict';

import Enzyme          from 'enzyme';
import Immutable     from 'immutable';
import Adapter         from 'enzyme-adapter-react-15';
import {isArrayLike} from 'enzyme-adapter-utils';

Enzyme.configure({adapter: new Adapter()});

describe('Enzyme 3', () => {
  it('should play nice with Immutable.List children', () => {
    const children = Immutable.fromJS([1, 2, 3]);

    expect(isArrayLike(children))
      .toBe(true);
  });
});
```

Above unit test doesn't pass, since `isArrayLike` doesn't think `Immutable.List` is array-like, to be more specific `isIterable` from enzyme-adapter-utils returns false for an Immutable List instance. I also can see you are testing against ES6 Map and Set in your test suite, but support for Immutable children is missing.

I could add Immutable as a dependency of enzyme-adapter-utils to extend `isArrayLike` for this use case. Or if you think about a better way to achieve it, I could give it a shot, add some tests and submit a PR.

I think having support for Immutable children makes sense because:

1. Was supported in v2.
2. It's supported by React itself.

I'm looking forward to collaborate, keep the great work!

ljharb:
@kcmq thanks, your comment reminded me to publish it :-) all set.

@Inucoder `Immutable.List` should be an iterable, even if not array-like - so I'd expect it should work. Are you using the latest version of immutable?

Inucoder:
@ljharb, the test I shared was with Immutable `3.8.2`, which is the latest stable release. Just pulled `4.0.0-rc.9` and test still not passing. I can add support for it based on how React checks for it https://github.com/facebook/react/pull/2376/commits/c07ea0ba34df57fc54b218d6b4edcd090ed6411c. Let me know your thoughts.

Inucoder:
@ljharb, I found that the issue was because we are running our unit tests on PhantomJS, and ES6 Symbol API is missing there, and might be the case for old browsers. Tried to run the test on recent Chrome version and works just fine.

We could change

```
function isIterable(obj) {
  return (
    obj != null &&
    typeof Symbol === 'function' &&
    typeof Symbol.iterator === 'symbol' &&
    typeof obj[Symbol.iterator] === 'function'
  );
}
```

to 

```
function isIterable(obj) {
  return (
    obj != null && (
      (
        typeof Symbol === 'function' &&
        typeof Symbol.iterator === 'symbol' &&
        typeof obj[Symbol.iterator] === 'function'
      ) ||
      obj['@@iterator']
    )
  );
}
```

that's exactly what React does.

ljharb:
Hmm - `@@iterator` is a fallback that only a few versions of Firefox officially supported; I'm hesitant to support it in enzyme; although since React unfortunately chose to support it, we might have to.

Can you confirm whether React supported this behavior down to 0.13? Also, can you file a separate issue about this?

Inucoder:
For sure @ljharb, I'll take a look into that.

ljharb:
Linking to #1334 as well.

PR #1192, opened 2017-09-28
Implement spyMethod and use it to inspect the result of shouldComponentUpdate

koba04:
This PR is to follow up https://github.com/airbnb/enzyme/pull/1133#discussion_r140950790

### DIFF
--- a/packages/enzyme/src/ShallowWrapper.js
+++ b/packages/enzyme/src/ShallowWrapper.js
@@ -18,6 +18,7 @@ import {
   sym,
   privateSet,
   cloneElement,
+  spyMethod,
 } from './Utils';
 import { debugNodes } from './Debug';
 import {
@@ -310,78 +311,61 @@ class ShallowWrapper {
         const { state } = instance;
         const prevProps = instance.props || this[UNRENDERED].props;
         const prevContext = instance.context || this[OPTIONS].context;
-        const nextProps = { ...prevProps, ...props };
         const nextContext = context || prevContext;
         if (context) {
           this[OPTIONS] = { ...this[OPTIONS], context: nextContext };
         }
         this[RENDERER].batchedUpdates(() => {
+          // When shouldComponentUpdate returns false we shouldn't call componentDidUpdate.
+          // so we spy shouldComponentUpdate to get the result.
           let shouldRender = true;
-          // dirty hack:
-          // make sure that componentWillReceiveProps is called before shouldComponentUpdate
-          let originalComponentWillReceiveProps;
-          if (
-            !this[OPTIONS].disableLifecycleMethods &&
-            instance &&
-            typeof instance.componentWillReceiveProps === 'function'
-          ) {
-            instance.componentWillReceiveProps(nextProps, nextContext);
-            originalComponentWillReceiveProps = instance.componentWillReceiveProps;
-            instance.componentWillReceiveProps = () => {};
-          }
-          // dirty hack: avoid calling shouldComponentUpdate twice
-          let originalShouldComponentUpdate;
+          let spy;
           if (
             !this[OPTIONS].disableLifecycleMethods &&
             instance &&
             typeof instance.shouldComponentUpdate === 'function'
           ) {
-            shouldRender = instance.shouldComponentUpdate(nextProps, state, nextContext);
-            originalShouldComponentUpdate = instance.shouldComponentUpdate;
+            spy = spyMethod(instance, 'shouldComponentUpdate');
           }
-          if (shouldRender) {
-            if (props) this[UNRENDERED] = cloneElement(adapter, this[UNRENDERED], props);
-            if (originalShouldComponentUpdate) {
-              instance.shouldComponentUpdate = () => true;
-            }
-
-            this[RENDERER].render(this[UNRENDERED], nextContext);
-
-            if (originalShouldComponentUpdate) {
-              instance.shouldComponentUpdate = originalShouldComponentUpdate;
-            }
+          if (props) this[UNRENDERED] = cloneElement(adapter, this[UNRENDERED], props);
+          this[RENDERER].render(this[UNRENDERED], nextContext);
+          if (spy) {
+            shouldRender = spy.getLastReturnValue();
+            spy.restore();
+          }
+          if (
+            shouldRender &&
+            !this[OPTIONS].disableLifecycleMethods &&
+            instance
+          ) {
             const lifecycles = getAdapterLifecycles(adapter);
-            if (
-              !this[OPTIONS].disableLifecycleMethods &&
-              instance
-            ) {
+
+            if (lifecycles.getSnapshotBeforeUpdate) {
+              let snapshot;
+              if (typeof instance.getSnapshotBeforeUpdate === 'function') {
+                snapshot = instance.getSnapshotBeforeUpdate(prevProps, state);
+              }
               if (
-                lifecycles.getSnapshotBeforeUpdate
-                && typeof instance.getSnapshotBeforeUpdate === 'function'
+                lifecycles.componentDidUpdate &&
+                typeof instance.componentDidUpdate === 'function'
               ) {
-                const snapshot = instance.getSnapshotBeforeUpdate(prevProps, state);
-                if (typeof instance.componentDidUpdate === 'function') {
-                  instance.componentDidUpdate(prevProps, state, snapshot);
-                }
-              } else if (typeof instance.componentDidUpdate === 'function') {
-                if (
-                  lifecycles.componentDidUpdate &&
-                  lifecycles.componentDidUpdate.prevContext
-                ) {
-                  instance.componentDidUpdate(prevProps, state, prevContext);
-                } else {
-                  instance.componentDidUpdate(prevProps, state);
-                }
+                instance.componentDidUpdate(prevProps, state, snapshot);
+              }
+            } else if (
+              lifecycles.componentDidUpdate &&
+              typeof instance.componentDidUpdate === 'function'
+            ) {
+              if (lifecycles.componentDidUpdate.prevContext) {
+                instance.componentDidUpdate(prevProps, state, prevContext);
+              } else {
+                instance.componentDidUpdate(prevProps, state);
               }
             }
-            this.update();
           // If it doesn't need to rerender, update only its props.
           } else if (props) {
             instance.props = props;
           }
-          if (originalComponentWillReceiveProps) {
-            instance.componentWillReceiveProps = originalComponentWillReceiveProps;
-          }
+          this.update();
         });
       });
     });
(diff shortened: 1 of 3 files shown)


packages/enzyme/src/ShallowWrapper.js
```diff
@@ -268,65 +269,41 @@ class ShallowWrapper {
...
           ) {
-            instance.componentWillReceiveProps(nextProps, nextContext);
-            originalComponentWillReceiveProps = instance.componentWillReceiveProps;
-            instance.componentWillReceiveProps = () => {};
+            spy = sinon.spy(instance, 'shouldComponentUpdate');
+          }
+          if (props) this[UNRENDERED] = cloneElement(adapter, this[UNRENDERED], props);
+          this[RENDERER].render(this[UNRENDERED], nextContext);
+          if (spy) {
+            shouldRender = spy.getCall(0).returnValue;
```
koba04: Should we care about calling `shouldComponentUpdate` multiple times?

ljharb: Maybe it should get the last call, not the first?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -380,26 +357,23 @@ class ShallowWrapper {
...
-          instance.shouldComponentUpdate = (...args) => {
-            shouldRender = originalShouldComponentUpdate.apply(instance, args);
-            instance.shouldComponentUpdate = originalShouldComponentUpdate;
-            return shouldRender;
-          };
+          spy = sinon.spy(instance, 'shouldComponentUpdate');
         }
         instance.setState(state, callback);
+        if (spy) {
+          shouldRender = spy.getCall(0).returnValue;
```
ljharb: Also here

koba04: I've fixed it!

koba04:
I'm not sure what the oldest JS engine sinon supports is.
But according to the CI results of sinon, sinon supports Node v4 or later.
In browser environment, according to its sourcelabs results, sinon supports IE11.

lelandrichardson:
I'm not sure I agree with this change. This seems like a huge (and often problematic) dependency to take on for very little benefit.

ljharb:
Fair point (not sure about problematic tho; sinon's always worked fine in my experience) the benefit I see is ensuring that the property descriptor remains the same after restoring, which is tedious to do correctly.

lelandrichardson:
Honestly I would prefer that we do our own thing here instead of pulling in sinon. If we need to do this in a more robust way and build a util function, or depend on a package that does *just* that, then that's fine. I'm also wondering if this would have strange conflicts with people using sinon on their own with `shouldComponentUpdate` and sinon handling sinon-spied functions in a different way in that case.

lelandrichardson:
@ljharb when i say "problematic" i mean that I know that in the past webpack and browserify both had problems bundling sinon. I think that might have been fixed in a more recent version though.

ljharb:
Those are all valid concerns. Let's definitely hold off on merging this for now.

koba04:
I'm working on to implement own spyMethod instead of `sinon.spy`.

koba04:
I've implemented a custom spy function, which is a minimum implementation needed by enzyme.

packages/enzyme/src/Utils.js
```diff
@@ -266,3 +266,34 @@ export function cloneElement(adapter, el, props) {
     { ...el.props, ...props },
   );
 }
+
+export function spyMethod(instance, methodName) {
+  let lastReturnValue;
+  const originalMethod = instance[methodName];
+  const hasOwn = Object.hasOwnProperty.call(instance, methodName);
```
ljharb: this should use the `has` package

packages/enzyme/src/Utils.js
```diff
@@ -266,3 +266,34 @@ export function cloneElement(adapter, el, props) {
...
+      const result = originalMethod.apply(this, args);
+      lastReturnValue = result;
+      return result;
+    },
+  });
+  return {
+    restore() {
+      if (hasOwn) {
+        /* eslint-disable no-param-reassign */
+        instance[methodName] = originalMethod;
```
ljharb: this won't actually restore the original property descriptor; to do this properly, it needs to use `Object.getOwnPropertyDescriptor` and `Object.defineProperty` when available, and fall back to normal reading and assignment.

packages/enzyme/src/Utils.js
```diff
@@ -266,3 +266,34 @@ export function cloneElement(adapter, el, props) {
...
+        /* eslint-enable no-param-reassign */
+      } else {
+        /* eslint-disable no-param-reassign */
+        delete instance[methodName];
+        /* eslint-enable no-param-reassign */
+      }
+    },
+    get lastReturnValue() {
```
ljharb: I don't think this needs to be, or should be, a getter - can it just be a function?

koba04:
@ljharb 

React has plans for v16.3. https://github.com/facebook/react/issues/12152
They include API changes for lifecycle methods.

* Add `UNSAFE_` prefix lifecycle methods
    * `UNSAFE_componentWillMount`, `UNSAFE_compnentWillReceiveProps`, `UNSAFE_componentWillUpdate`
* Add `static getDerivedStateFromProps`

Currently, ShallowWrapper overrides `componentWillReceiveProps` and `shouldComponentUpdate` to get the result of `shouldComponentUpdate`.
It means enzyme have to deal with these changes at ShallowWrapper.

This PR will solve that.
This PR avoid to override these lifecycle methods by using a spy function.
So if this PR is merged, we don't have to care about the changes.

koba04:
React v16.3 has been released.
The current implementation of `shallow` doesn't cover the case that `UNSAFE_componentWillReceiveProps` should be called when `shouldComponentUpdate` returns `false` because enzyme overrides the implementation.
This PR will resolve the case.

https://github.com/airbnb/enzyme/pull/1192/files#diff-fa27c82ee4fde075bae6173848e17c82L263

koba04:
I'm not sure why CI is failing...

https://travis-ci.org/airbnb/enzyme/jobs/399026371

koba04:
All checks have passed 🎉

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -310,78 +311,60 @@ class ShallowWrapper {
...
-            if (originalShouldComponentUpdate) {
-              instance.shouldComponentUpdate = originalShouldComponentUpdate;
-            }
             const lifecycles = getAdapterLifecycles(adapter);
+            // getSnapshotBeforeUpdate
+            let snapshot;
             if (
-              !this[OPTIONS].disableLifecycleMethods &&
-              instance
+              lifecycles.getSnapshotBeforeUpdate
```
ljharb: this no longer seems to be checking `disableLifecycleMethods`?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -310,78 +311,60 @@ class ShallowWrapper {
...
+              && typeof instance.getSnapshotBeforeUpdate === 'function'
             ) {
-              if (
-                lifecycles.getSnapshotBeforeUpdate
-                && typeof instance.getSnapshotBeforeUpdate === 'function'
-              ) {
-                const snapshot = instance.getSnapshotBeforeUpdate(prevProps, state);
```
ljharb: i have a mild preference for splitting the "snapshot" and "context" paths entirely, instead of trying to interleave them with `let snapshot`

packages/enzyme/src/Utils.js
```diff
@@ -242,3 +243,42 @@ export function cloneElement(adapter, el, props) {
...
+  let lastReturnValue;
+  const originalMethod = instance[methodName];
+  const hasOwn = has(instance, methodName);
+  let descriptor;
+  if (hasOwn) {
+    descriptor = Object.getOwnPropertyDescriptor(instance, methodName);
+  }
+  Object.defineProperty(instance, methodName, {
+    configurable: true,
+    enumerable: false,
```
ljharb: shouldn't this be `!descriptor || !!descriptor.enumerable`?

packages/enzyme/src/ShallowWrapper.js
koba04: Checking `disableLifecycleMethods` is here.

https://github.com/airbnb/enzyme/pull/1192/files#diff-fa27c82ee4fde075bae6173848e17c82R338

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -310,78 +311,58 @@ class ShallowWrapper {
...
-            if (
-              !this[OPTIONS].disableLifecycleMethods &&
-              instance
+
+            if (lifecycles.getSnapshotBeforeUpdate) {
+              let snapshot;
+              if (typeof instance.getSnapshotBeforeUpdate === 'function') {
+                snapshot = instance.getSnapshotBeforeUpdate(prevProps, state);
+              }
+              if (typeof instance.componentDidUpdate === 'function') {
```
koba04: Should I check `lifecycles.componentDidUpdate`?

ljharb: yes, that'd be great

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -310,78 +311,58 @@ class ShallowWrapper {
...
         const nextContext = context || prevContext;
         if (context) {
           this[OPTIONS] = { ...this[OPTIONS], context: nextContext };
         }
         this[RENDERER].batchedUpdates(() => {
+          // When shouldComponentUpdate returns false we shouldn't call componentDidUpdate.
+          // so we spy shouldComponentUpdate to get the result.
           let shouldRender = true;
-          // dirty hack:
-          // make sure that componentWillReceiveProps is called before shouldComponentUpdate
```
ljharb: do we no longer need this hack?

packages/enzyme/src/Utils.js
koba04: But shouldn't an instance method be enumerable?

ljharb: the replacement should be enumerable, or not, to match the one it's spying on

packages/enzyme/src/ShallowWrapper.js
ljharb: Can you elaborate on why?

koba04: The current version calls `shouldComponentUpdate` manually to get the result.
But `componentWillReceiveProps` should be called before `shouldComponentUpdate` so we should call `componentWillReceiveProps` manually before `shouldComponentUpdate`.

In this PR, we no longer call `shouldComponentUpdate` manually because we can spy the method.
As the result, we can remove the dirty hack because we don't need to call `componentWillReceiveProps` and `shouldComponentUpdate` manually.

ljharb: Perfect, thanks.

koba04:
Thanks 🎉

PR #1196, opened 2017-09-29
Fix typo from readme

lalnuo:


dubbha:
This contradicts my previous PR, where I fix ES5 version:
https://github.com/airbnb/enzyme/pull/1194

Besides actually fixing the ES5 version, I also changes "Enzyme" to "enzyme" there. Because Adapter is a constructor here (used with "new"), but Enzyme is not.

Moreover, it is not needed at all in the ES6 example. In the basically the same example in https://github.com/airbnb/enzyme/blob/master/docs/installation/react-16.md it is omitted, and it makes perfect sense:

```js
import { configure } from 'enzyme';

```

ljharb:
Closing since #1194 superseded it.

PR #1204, opened 2017-09-30
Add an npmignore file to all packages

lelandrichardson:
Adds an npmignore file to all of the packages since the root one does not apply to npm publishing anymore


lelandrichardson:
@ljharb i considered this, but thought that they might need to diverge over time. Thoughts?

ljharb:
That's fair; but it'd be easy to copy it when a need arose. Not a huge deal either way.

Issue #1173, opened 2017-09-27
Add better dev/test instructions in CONTRIBUTING.md


jack-lewin:
Hi,

I'm hoping to contribute to enzyme, but I'm having trouble running the test suite. Not sure if I've missed anything with the setup, but I'd appreciate some help getting up and running! Thanks a lot.

#### Steps taken:
1. `git clone https://github.com/airbnb/enzyme`
2. `cd enzyme`
3. `npm install`
4. `npm run react:15` - this waits for ~10 secs, and then exits without feedback. Not sure whether that's the issue?
5. `npm test` fails, with a linter error.

Error message:

```
/path/to/enzyme/packages/enzyme-adapter-react-15/src/ReactFifteenAdapter.js

lerna ERR! execute   10:31  error  Unable to resolve path to module 'enzyme'                import/no-unresolved
lerna ERR! execute   10:31  error  Missing file extension for "enzyme"                      import/extensions
lerna ERR! execute   20:8   error  Unable to resolve path to module 'enzyme-adapter-utils'  import/no-unresolved
lerna ERR! execute   20:8   error  Missing file extension for "enzyme-adapter-utils"        import/extensions
```

#### List of globally-installed packages:
```
├── create-react-app@1.4.1
├── eslint@4.7.2
├── eslint-plugin-import@2.7.0
├── jest@21.2.0
├── lerna@2.2.0
├── npm@5.4.2
├── serve@6.1.0
├── standard@10.0.3
└── yarn@1.1.0
```

lelandrichardson:
Hey @jack-lewin, you're right - we should add some steps to the contributing guide.

Due to our repo structure and hard to explain things, our test suite currently runs on *built* enzyme. This means you need to run `npm run build` before running the test suite.

If you are actively developing, the workflow I recommend is to run `npm run build:watch` in one terminal, and `npm run test:watch` in another terminal. This should provide you with ~realtime feedback.

jack-lewin:
@lelandrichardson thanks for clearing that up - I've got it built & running now. I'll put in a PR for the docs over the weekend!

ljharb:
Merged in #1207.

PR #1207, opened 2017-10-01
Clarify dev workflow in CONTRIBUTING.md

jack-lewin:
* Explains that the test suite & linter run on built Enzyme
* Adds guidance on dev workflow (as per recommendation in #1173)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,7 +46,12 @@ npm run react:16
 
 ### Running Tests
 
+The test suite runs on *built* Enzyme.
+
 ```bash
+# build Enzyme locally before testing
+npm run build
+
 # run tests on whatever version of React is currently installed
 npm test
 ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -62,10 +74,15 @@ npm run test:watch
 This codebase adheres to the [Airbnb Styleguide](https://github.com/airbnb/javascript) and is
 enforced using [ESLint](http://eslint.org/).
 
+As with the test suite, linting rules apply to *built* Enzyme.
```
lelandrichardson: This is maybe only partially correct. I believe the `import/*` eslint rules rely on built enzyme because they actually look at the filesystem and require that the built files be there, but linting doesn't actually run on the built files, they run on the src files... but they require the built files to be there. Kind of annoying that it works that way but hey, this is the babel-compiled world we live in...

jack-lewin: Good point - I've added some clarification to that.

PR #1234, opened 2017-10-05
move react-addons-test-utils to dependencies

brucewpaul:
making these explicit dependencies will make it easier for other
projects and packages that allow for multiple versions of react


brucewpaul:
@ljharb

ljharb:
Let's combine this and #1252? I can do that on the command line.

PR #1236, opened 2017-10-06
Update mount.md

withintheruins14:
It's not clear you're saying this is a recommended alternative to using mount


withintheruins14:
Ah, rereading it - it makes sense to me now. Thanks 👌🏼

PR #1241, opened 2017-10-08
make param name consistant

jstoebel:
suggest makeing param name same as what's in the side bar


jstoebel:
good point: see #1244

ljharb:
I'm going to combine these two PRs; they should be done in the same PR, not in a separate one.

PR #1242, opened 2017-10-09
Update docs of ref method to be consistent with version 3

adam-golab:


docs/api/ReactWrapper/ref.md
```diff
@@ -12,7 +12,7 @@ NOTE: can only be called on a wrapper instance that is also the root instance.
 
 #### Returns
 
-`ReactWrapper`: A wrapper of the node that matches the provided reference name.
+`ReactComponent`: The node that matches the provided reference name.
```
yesmeck: It can be a HTMLElement.

Issue #1161, opened 2017-09-27
[v3] Consider adding react-test-renderer to the deps for ReactSixteenAdapter


mattkrick:
Moving to v3, on React v16, I get the following error:
```
 Cannot find module 'react-test-renderer/shallow' from 'ReactSixteenAdapter.js'
```

I think it'd be nice to explicitly depend on this so folks don't need to install it on their own.

lelandrichardson:
Like react and react-dom, this is a peer dependency: https://github.com/airbnb/enzyme/blob/master/packages/enzyme-adapter-react-16/package.json#L44

We could make it an explicit dep, but then I think we'd have to also add react as a dep?  I'm concerned about this causing other errors from people having duplicate react versions in their node_modules.

ljharb:
React *should* be a peer dep of each react adapter; as should all the react things the adapter needs. Is React not already a peer dep of each adapter?

ljharb:
@mattkrick what does `npm ls` output? If it exits nonzero, then nothing can be expected to work.

Every React thing must always be a peer dep, and always must be explicitly depended on in the top level - that's how singletons work.

lelandrichardson:
@ljharb react and react-test-renderer are both peer deps already. i'm saying i think if we moved react-test-renderer to a dep, then react would need to move to be a dep as well? i'm not quite sure.  Anyway, I feel like react-test-renderer should stay a peer dep unless anyone has a convincing argument for it not to be?

ljharb:
Absolutely it should; react things should never be anything but peer deps except in the top-level app.

mattkrick:
i agree with keeping react & react-dom as a peerDep.
With React 16, facebook has removed all "add-ons" from react (prop-types, react-transition-group, react-test-renderer, etc). the dependencies for react-test-renderer are:
```
"dependencies": {
    "fbjs": "^0.8.16",
    "object-assign": "^4.1.1"
  },
  "peerDependencies": {
    "react": "^16.0.0-beta.5"
  },
```

since everyone will need this to use enzyme, **_and it doesn't have a hard dep on react_**, i think it's a good candidate for a dependency in enzyme.

ljharb:
Absolutely not; since enzyme is used in an app, having enzyme-adapter-react-16 hard-depend on react would risk duplicates occuring.

NOTHING EVER can safely directly depend on `react` except the top-level app, period.

mattkrick:
[...]

PR #1252, opened 2017-10-12
Make react-test-renderer a dependency instead of a peer dependency

bdwain:
This addresses #1161 . It makes react test renderer a regular dependency instead of a peer dependency so that users do not have to install it themselves. They should not have to because they don't use it directly.

This is safe because the test renderer does not have a dependency on react, so there will be no second copy. It has a peer dependency, and as long the version of test renderer that is installed has its peer dependency satisfied, everything should be fine.

### DIFF
--- a/docs/installation/react-014.md
+++ b/docs/installation/react-014.md
@@ -7,12 +7,6 @@ installed, you should do so:
 npm i --save react@0.14 react-dom@0.14
 ```
 
-Further, enzyme with React 0.14 requires the test utilities addon be installed:
-
-```bash
-npm i --save-dev react-addons-test-utils@0.14
-```
-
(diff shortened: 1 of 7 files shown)


ljharb:
@bdwain what about the react 15.4 adapter?

bdwain:
ah yea i guess that can be a peer dependency too. I was just looking for react test renderer so i didn't think about it. I can add it.

bdwain:
i just saw #1234 . Looks like it has overlap with this. I didn't realize before because i was only looking for react-test-renderer.

ljharb:
I'll combine the two and merge them together.

bdwain:
rebased this off the latest master

PR #1254, opened 2017-10-12
stop adding null string key when there is a ref

### LINKED ISSUE #1240: Components with refs getting "null" (string) key even if no key is set
I have a component that returns something like
```js
return <Dropdown ref={this.setRef}>...</Dropdown>;
```

I am trying to shallow render it, but when i try to verify the element returned from .get(0), it has a key of "null", the string not the value null, even though there should be no key in the return value.

I've been trying to dig through the code and I think the problem is [here](https://github.com/airbnb/enzyme/blob/master/packages/enzyme-adapter-utils/src/Utils.js#L121). The function `propsWithKeysAndRef` is adding BOTH key and ref to props, even if only one of them has a value. React [stringifies all valid keys in createElement](https://github.com/facebook/react/blob/master/src/isomorphic/classic/element/ReactElement.js#L196), and a key is valid if it is not undefined. [This seems like it was an intentional decision](https://github.com/facebook/react/issues/5519). 

The problem seems to be that in enzyme 3, it reruns the node through react.createElement, which now thinks you explicitly set the key to null, and so it sets it to "null". I guess an element with no key gets a "key" property of null, not undefined.

Unfortunately, I don't think the answer is to just not return key as one of the props from `propsWithKeysAndRef` if it is null, because if you explicitly set it to null, then that would be the correct thing to do. 

Is there a way to store the original key value on the node somehow, so that you can look for undefined explicitly?

bdwain:
This fixes #1240 . Since react turns all valid (aka not undefined) keys into strings in react.createElement, it made sense to just remove falsey keys from nodes, so that when they were passed back through createElement, they didn't get turned into strings.

The tests I added fail with errors like 

`AssertionError: expected 'null' to equal null` 

without the changes I made.


bdwain:
added #1254 to fix this.

bdwain:
is there anything else needed before this is looked at? It'd be great to get it merged. Idk how many people it's affecting but I know my project can't upgrade to react 16 until it's done.

bdwain:
cool thanks. i rebased off of master

bdwain:
I rebased this. Does it need more approvals?

ljharb:
Yes, I'd like a couple more reviewers to take a look first.

bdwain:
Hi @ljharb is there anything i can do to get this reviewed and merged? I know it's blocking me from updating to react 16, and I'm sure it's blocking others as wll.

ljharb:
I'll give it a try locally.

bdwain:
any luck with this? It's getting hard to keep it up to date because merge conflicts keep happening as master updates.

ljharb:
Still working on it :-) I'll keep it up to date for you tho!

bdwain:
thanks!

tryggvigy:
I believe https://github.com/airbnb/enzyme/issues/1511 was introduced in this PR

PR #1261, opened 2017-10-13
Fix to call componentDidUpdate on setState of React v16

### LINKED ISSUE #1247: Lifecycle methods aren't called as expected when state changes
I've upgraded from Enzyme 2.x to 3.x and noticed that tests that were previously passing now fail. I've limited these down to tests that depend on lifecycle methods like `componentDidUpdate` to be called. 

In https://github.com/airbnb/enzyme/pull/1140 `lifecycleExperimental` was enabled by default, which to my understanding means lifecycle methods like `componentDidUpdate` should be called when `setState` is called on a shallow-rendered component. However, I am not seeing this behavior. 

Even more weird: the [migration guide makes it sound](http://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html#lifecycle-methods) as if 2.x wouldn't have been calling these lifecycle methods, but I assume they were since my tests were passing before upgrading 🤔 

**I've created a simplified example of what I'm trying to describe here: https://github.com/sawyerh/enzyme-bug**

The lifecycle method `componentDidUpdate` doesn't appear to be called when the component is rendered like this:

```jsx
...
wrapper.setState({...})
```

However `componentDidUpdate` does appear to be called when the component is rendered either of these ways:

1. `const wrapper = shallow(<Foo />, { lifecycleExperimental: true });`
2. `const wrapper = mount(<Foo />);`

----

Maybe related: #1201

koba04:
@ljharb @lelandrichardson This is to fix #1247 🙏 
Currently, componentDidUpdate is not called when calling setState with React v16.

### DIFF
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -1028,7 +1028,7 @@ describe('shallow', () => {
       expect(wrapper.first('div').text()).to.equal('yolo');
     });
 
-    it('should call componentWillReceiveProps, shouldComponentUpdate and componentWillUpdate with merged newProps', () => {
+    it('should call componentWillReceiveProps, shouldComponentUpdate, componentWillUpdate, and componentDidUpdate with merged newProps', () => {
       const spy = sinon.spy();
 
       class Foo extends React.Component {
@@ -1041,6 +1041,9 @@ describe('shallow', () => {
         }
         componentWillUpdate(nextProps) {
           spy('componentWillUpdate', this.props, nextProps);
+        }
+        componentDidUpdate(prevProps) {
+          spy('componentDidUpdate', prevProps, this.props);
         }
         render() {
           return (
@@ -1066,6 +1069,11 @@ describe('shallow', () => {
         ],
         [
           'componentWillUpdate',
+          { a: 'a', b: 'b' },
+          { a: 'a', b: 'c', d: 'e' },
+        ],
+        [
+          'componentDidUpdate',
           { a: 'a', b: 'b' },
           { a: 'a', b: 'c', d: 'e' },
         ],
@@ -3098,7 +3106,7 @@ describe('shallow', () => {
         ]);
       });
 
-      // componentDidUpdate does not seem to get called in react 16 beta.
+      // componentDidUpdate is not called in react 16
       itIf(REACT16, 'calls expected methods for setState', () => {
         wrapper.setState({ bar: 'bar' });
         expect(spy.args).to.deep.equal([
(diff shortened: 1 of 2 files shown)


packages/enzyme/src/ShallowWrapper.js
```diff
@@ -387,7 +387,7 @@ class ShallowWrapper {
         // so we replace shouldComponentUpdate to know the result and restore it later.
         let originalShouldComponentUpdate;
         if (
-          this[OPTIONS].lifecycleExperimental &&
```
ljharb: hm, i wonder if this is causing a number of other bugs related to lifecycle methods?

koba04: I can't find issues causing by this in enzyme repository, but there might be in outside.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1026,7 +1026,7 @@ describe('shallow', () => {
       expect(wrapper.first('div').text()).to.equal('yolo');
     });
 
-    it('should call componentWillReceiveProps, shouldComponentUpdate and componentWillUpdate with merged newProps', () => {
+    it('should call componentWillReceiveProps, shouldComponentUpdate and componentWillUpdate and componentDidUpdate with merged newProps', () => {
```
ljharb: You have a duplicate "and" here (before both the ultimate and penultimate items); also please use an Oxford comma :-)

ljharb:
@koba04 this needs to be rebased, so as to remove all merge commits; i have some commits i need to add to master first, so i'll rebase this branch for you after that, and then merge it.

koba04:
@ljharb Rebased!

But I couldn't run unit tests on my local because `lerna bootstap` never finish at the following point... Is it just me?

```
lerna info version 2.5.1
lerna info versioning independent
lerna info Bootstrapping 9 packages
lerna info lifecycle preinstall
lerna info Installing external dependencies
⸨         ░░░░░░░░░⸩ ⠙ install dependencies: verb actions 9 actions, concurrency 4
```

I'm using macOS 10.12.6 and Node v8.9.0 and npm v5.5.1.

ljharb:
@koba04 hm, that should definitely work. i'll give it a shot.

ljharb:
(altho you don't need to run lerna yourself; `npm install` should do all the necessary things)

koba04:
Thanks! To be accurate, that is a result through `npm install`. If anyone doesn't have the trouble, my environment would have a problem.

koba04:
Thanks!

koba04:
https://github.com/airbnb/enzyme/pull/1261#issuecomment-341760426

(I avoided the trouble by `lerna bootstrap --concurrency 1` 🤔)

luissmg:
Is this fixed and already in the latest version? Because I am using 3.3.0 and this problem persists. The workaround for me is to use mount instead of shallow.

PR #1262, opened 2017-10-13
Update Mount docs

joemcloE:
getNode and getNodes are no longer supported so removing the bad link


ljharb:
Replaced by #1714; thanks for the contribution!

PR #1263, opened 2017-10-14
Add Portal support

jquense:
I’m happy to add tests, but i don’t really understand the test suite nor where the appropriate place to add them is for the v16 only feature. 

@aweary from what I can tell the Portal fiber is similar to a host component in that it’s essentially a transparent wrapper over it’s rendered children. It does hold some host container info which may be required for actual DOM traversal but in terms of the RST tree this should give an accurate representation as React sees it.

### DIFF
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -69,7 +69,7 @@ function toTree(vnode) {
   switch (node.tag) {
     case HostRoot: // 3
       return toTree(node.child);
-    case HostPortal:
+    case HostPortal: // 4
       return toTree(node.child);
     case ClassComponent:
       return {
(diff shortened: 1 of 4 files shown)


alansouzati:
Hey folks, any news here? I really want to enable tests for my React portal components 👍

louisscruz:
Are any efforts being made to provide portal support? This seems to be the final bit in migrating our code base (and its Enzyme tests) to React 16.

marudor:
I've merged master into this PR and tests seem to work.
Could you rebase to master and see if it passes?

ljharb:
@jquense please rebase PRs from the command line when updating them, so as to avoid merge commits :-)

jquense:
@ljharb Sorry, it's not that i'm unsure how/where to write a test, but how to write ones that sufficiently test the change to ya'lls satisfaction. A single test that does a `find` through a portal will exercise the code path, but not particularly aggressively. If ya'll ok with none of the other traversal methods hitting this that seems fine, if not i'm not sure you want me to rewrite every test to include a portal either.

thanks for the review!

ljharb:
I think that the simple test you describe would be the minimum of what we'd want; ideally, we'd write tests for all the traversal methods, to ensure that none of them throw.

louisscruz:
@jquense Are you planning on writing the test? If not, I could take a crack at it.

jquense:
I'm planning on it but a bit busy, happy to get some help :)

PR #1264, opened 2017-10-15
Typescript setup documentation

camurai:
add additional instructions to install types for enzyme and enzyme-adapter-react-16 for typescript as well as the change to the setup file to include
import * as Adapter from 'enzyme-adapter-react-16'; as just "import Adapter..." from does not work.

### DIFF
--- a/docs/installation/react-16.md
+++ b/docs/installation/react-16.md
@@ -23,12 +23,6 @@ And then you're ready to go!  In your test files you can simply `require` or `im
 
 Typescript:
 
-You will need to additionally install the types files for enzyme and enzyme-adapter-react-16:
-```bash
-npm i --save-dev @types/enzyme
-npm i --save-dev @types/enzyme-adapter-react-16
-```
-
 ```ts
 // setup file
 import { configure } from 'enzyme';
--- a/docs/installation/react-16.md
+++ b/docs/installation/react-16.md
@@ -21,6 +21,29 @@ npm i --save-dev enzyme enzyme-adapter-react-16
 
 And then you're ready to go!  In your test files you can simply `require` or `import` enzyme:
 
+Typescript:
+
+You will need to additionally install the types files for enzyme and enzyme-adapter-react-16:
+```bash
+npm i --save-dev @types/enzyme
+npm i --save-dev @types/enzyme-adapter-react-16
+```
+
+```ts
+// setup file
+import { configure } from 'enzyme';
+import * as Adapter from 'enzyme-adapter-react-16';
+
+configure({ adapter: new Adapter() });
+```
+
+```ts
+// test file
+import { shallow, mount, render } from 'enzyme';
+
+const wrapper = shallow(<Foo />);
+```
+
 ES6:
 ```js
 // setup file


docs/installation/react-16.md
```diff
@@ -21,6 +21,29 @@ npm i --save-dev enzyme enzyme-adapter-react-16
...
+You will need to additionally install the types files for enzyme and enzyme-adapter-react-16:
+```bash
+npm i --save-dev @types/enzyme
+npm i --save-dev @types/enzyme-adapter-react-16
+```
+
+```ts
+// setup file
+import { configure } from 'enzyme';
+import * as Adapter from 'enzyme-adapter-react-16';
```
ljharb: If Typescript requires this, it's either a massive bug, or a huge UX roadbump, in TypeScript - I would expect TypeScript to loudly and clearly document this such that any TS user should automatically be aware of it.

Separately, do you *need* to install the types files? Either way, we don't maintain the type devs in DefinitelyTyped, so referencing them here seems to be committing to more support than we're offering.

camurai: I did some additional experimentation and you are correct the types files are not necessary and can be left out.

But i can verify that the change to import * as Adapter is necessary for typescript, otherwise the following error is generated:
```bash
TypeError: enzyme_adapter_react_16_1.default is not a constructor
```

I did notice that there is a difference between how enzyme_adapter_react_16 structures exports and how Enzyme does (and enzyme's import works as expected) 
Enzyme explicitly sets the default like this:
```js
Object.defineProperty(exports, "__esModule", {
  value: true
});

exports['default'] = shallow;
```
where as the ReactSixteenAdapter does this:
```js
module.exports = ReactSixteenAdapter;
```
perhaps changing ReactSixteenAdapter to match Enzyme's export structure would correct the issue? Unfortunately that is getting outside of my knowledge base. All I know is that currently to use the ReactSixteenAdapter with typescript you have to import it as:
```ts
import * as Adapter from 'enzyme-adapter-react-16';
```

ljharb: That's because the React 16 adapter uses the standard node pattern, where the export is `module.exports` - enzyme uses a babel-transpiled ES Module pattern, where `__esModule` is `true` and the `default` property has the real export.

TypeScript should be able to import CJS modules without `import *`; if not; that's the massive UX roadbump I was referring to in my first comment.

jwbay: TypeScript doesn't require `import *` here, strictly speaking. `import x = require(` also works, as does `const x = require(`. However, the first is incompatible with an ES2015 module target, and the second has no type information.

`import *` is definitely not valid ES2015 for pulling in a constructor, but it's the best we've got until https://github.com/Microsoft/TypeScript/issues/16093 is done.

mikebridge:
I couldn't get any variant of `import` working in TypeScript 2.6 & React 16 so I switched to `require` :

```typescript
import { configure } from "enzyme";
const Adapter = require("enzyme-adapter-react-16");
```

docs/installation/react-16.md
```diff
@@ -21,6 +21,23 @@ npm i --save-dev enzyme enzyme-adapter-react-16
 
 And then you're ready to go!  In your test files you can simply `require` or `import` enzyme:
 
+Typescript:
```
vsiao: nit: TypeScript

retailSteveDave:
> 
> 
> I couldn't get any variant of `import` working in TypeScript 2.6 & React 16 so I switched to `require` :
> 
> ```ts
> import { configure } from "enzyme";
> const Adapter = require("enzyme-adapter-react-16");
> configure({ adapter: new Adapter() });
> ```

This is what ended up working for me after browsing not only this thread, but [this one](https://github.com/airbnb/enzyme/issues/1284) as well. Now I'm on to the next error...

PR #1274, opened 2017-10-17
increase clarity of Selector documentation

SavePointSam:
Updates Selector documentation to explicitly explain `key` and `ref`, use proper CSS selector terminology, and re-organize to better differentiate from CSS-based selectors and React Prop Selectors.

### DIFF
--- a/docs/api/selector.md
+++ b/docs/api/selector.md
@@ -1,7 +1,8 @@
 # enzyme Selectors
 
-Many methods in enzyme's API accept a *selector* as an argument. Selectors in enzyme can fall into
-one of the following four categories:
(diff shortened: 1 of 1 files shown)


docs/api/selector.md
```diff
@@ -1,66 +1,71 @@
...
+));
+
+wrapper.find('[foo=3]');
+wrapper.find('[bar=false]');
+wrapper.find('[title="baz"]');
+```
+
+**The Key and Ref Prop**
+
+While in most cases, any React prop can be used, there are exceptions. The `key` and `ref` props will never work. This decision comes from how React uses these props internally and should not be relied upon.
```
ljharb: "internally, which means they should not be relied upon"

SavePointSam: updated!

Issue #1285, opened 2017-10-20
Supported Node Version


samit4me:
I recently encountered the following error after updating to v3.

```
error rst-selector-parser@2.2.2: The engine "node" is incompatible with this module. Expected version ">=5".
```

This error was thrown in a node 4 environment and it looks like #1086 switched to `rst-selector-parser` which only supports node 5+ according to [this file](https://github.com/aweary/rst-selector-parser/blame/master/package.json#L33). Given that node 4 is no longer active ([node release schedule](https://github.com/nodejs/Release)) I am curious if `enzyme` has officially dropped support for node 4??? If so, would it make sense to update `package.json` to reflect this?

ljharb:
No, it hasn't. rst-selector-parser needs to fix its engines field. @aweary?

aweary:
I removed the engines field in https://github.com/aweary/rst-selector-parser/pull/10 and just published the 2.2.3 patch release, which should fix this.

samit4me:
That is fantastic, this should fix the issue once `enzyme` is using the new version, thank you ✨ 

I wondering if the [`engines`](https://docs.npmjs.com/files/package.json#engines) field should still be specified to document the version of node that your stuff works on. That was my original intent of this ticket as I was unable to find out what versions of node is supported by `enzyme`.

If the `engines` field is specified and the node version does not match `yarn` will throw an error and stop whereas `npm` will throw a warning and continue. I guess which one you prefer is a personal preference but I think documenting the supported versions makes life easier when stuff isn't working.

For example, @aweary [determined](https://github.com/aweary/rst-selector-parser/pull/10) that `rst-selector-parser` test suite passes all the way back to 0.12 but that information is not easily available for new users of the library. They would have to either a test it themselves or did through all the PRs to find what versions of node are supported.

ljharb:
@samit4me fwiw enzyme doesn't need to update for you to get the new version; run `npm update` or `rm -rf node_modules && npm install`.

samit4me:
You are correct @ljharb removing node_modules and re-installing does work. What are your thoughts on documenting the supported node versions?

ljharb:
If yarn didn't error out, I'd do it - but *because* yarn errors out, I'd rather just not specify "engines" at all, and let users figure out if it works or not on their own.

samit4me:
Yeah, that's cool I totally get that. I think I on the other side of the fence and like how it errors out on node mismatch as it doesn't waste any more time than necessary.

[...]

PR #1286, opened 2017-10-20
Add supported node version

samit4me:
Specify supported node versions as per https://docs.npmjs.com/files/package.json#engines.

#1285

### DIFF
--- a/packages/enzyme/package.json
+++ b/packages/enzyme/package.json
@@ -33,6 +33,9 @@
   ],
   "author": "Leland Richardson <leland.richardson@airbnb.com>",
   "license": "MIT",
+  "engines": {
+    "node": ">=5"
+  },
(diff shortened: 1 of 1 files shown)


packages/enzyme/package.json
```diff
@@ -33,6 +33,9 @@
   ],
   "author": "Leland Richardson <leland.richardson@airbnb.com>",
   "license": "MIT",
+  "engines": {
+    "node": ">=5"
```
ljharb: I don't think we should specify this (and yarn shouldn't be enforcing it anyway - npm does not).

I think @aweary should remove or update "engines" on rst-selector-parser instead.

ljharb:
Closed in https://github.com/airbnb/enzyme/issues/1285#issuecomment-338257975

PR #1288, opened 2017-10-20
Update rst-selector-parse to prevent issues in node 4

samit4me:
#1285


aweary:
@samit4me can you verify that doing that ^ fixes the issue for you before we merge this?

PR #1296, opened 2017-10-24
Update contains.md

the2nhi:
we don't need a '1.' since it's not a step

### DIFF
--- a/docs/api/ReactWrapper/contains.md
+++ b/docs/api/ReactWrapper/contains.md
@@ -6,7 +6,7 @@ It will determine if an element in the wrapper matches the expected element by c
 
 #### Arguments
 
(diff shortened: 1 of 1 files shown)


docs/api/ReactWrapper/contains.md
```diff
@@ -6,7 +6,7 @@ It will determine if an element in the wrapper matches the expected element by c
 
 #### Arguments
 
-1. `nodeOrNodes` (`ReactElement|Array<ReactElement>`): The node or array of nodes whose presence you are detecting in the current instance's
```
ljharb: Even though it's a list of 1, it's still an ordered list. This is consistent with the rest of the docs as well.

ljharb:
Thanks for the contribution though!

Issue #1299, opened 2017-10-24
State update not affecting .props()


peter-mouland:
I have come across a test that worked in v2, but not in v3. But as far as I can tell, it should still work.

The component state is updated via a `scroll` event.  This in-turn should update the component className.  I can verify that the state has changed, but when i test the className, `.props()` doesn't know it has updated.  I can even confirm, using `.html()`, that the className has updated.

i.e.
This fails: 
`expect(backToTopButton.props().className).toMatch('back-to-top--visible')`
but this is the backToTopButton.html() :
`    <div id="back-to-top" class="back-to-top back-to-top--visible"><span class="back-to-top__chevron">test-file-stub</span></div>`

I have added a new test to compensate, which is written using `.find(className)` rather than `.props().className` and this seems to work.  This makes me think it's worth raising as a potential issue. 

Is this an issue, or have I missed something?

Thanks

## component
```jsx
// BackToTop.jsx
/* global window, document */
import React from 'react'
import PropTypes from 'prop-types'
import Svg from 'argos-svg'
import debounce from 'argos-utils/esnext/debounce'
import SmoothScroll from 'argos-utils/esnext/smooth-scroll'
import Chevron from 'argos-bolt/src/icons/chevron.svg'

import './back-to-top.scss'

function scrollToTop() {
  SmoothScroll.scrollHere(document.body, 0, 300)
}

class BackToTop extends React.Component {
  static propTypes = {
    triggerOffset: PropTypes.number
  }

  static defaultProps = {
    triggerOffset: 600
  }

  constructor(props) {
    super(props)
    this.state = { visible: false }
    this.handleScroll = debounce(this.handleScroll, 50)
  }

  componentDidMount() {
    window.addEventListener('scroll', this.handleScroll)
    this.handleScroll()
  }

  componentWillUnmount() {
    window.removeEventListener('scroll', this.handleScroll)
  }

  handleScroll = () => {
    if (window.pageYOffset > this.props.triggerOffset && !this.state.visible) {
      this.setState({ visible: true })
    } else if (window.pageYOffset < this.props.triggerOffset && this.state.visible) {
      this.setState({ visible: false })
    }
  }

  render() {
    return (
      <div id='back-to-top' className={`back-to-top${this.state.visible ? ' back-to-top--visible' : ''}`} onClick={scrollToTop}>
        <Svg className={'back-to-top__chevron'} height={'13px'} width={'20px'}>{Chevron}</Svg>
      </div>
    )
  }
}

[...]

PR #1300, opened 2017-10-24
Calling `props()` after a state change should be updated to `find()`

peter-mouland:
This PR has come about from this issue: https://github.com/airbnb/enzyme/issues/1299

If I have misunderstood anything please let me know.

Thanks

### DIFF
--- a/docs/guides/migration-from-2-to-3.md
+++ b/docs/guides/migration-from-2-to-3.md
@@ -102,6 +102,49 @@ Although this is a breaking change, I believe the new behavior is closer to what
 actually expect and want. Having enzyme wrappers be immutable results in more deterministic tests
 that are less prone to flakiness from external factors.
 
+### Calling `props()` after a state change
+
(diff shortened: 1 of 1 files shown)


bregenspan:
As far as I understand it, this is a side effect of https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#element-referential-identity-is-no-longer-preserved, so maybe it makes sense to put this as a sub-head under that?

ljharb:
@peter-mouland any interest in updating this PR?

peter-mouland:
hey, yeah - can do, do you have any stats on how many people look at the migrate docs now?

ljharb:
I'm afraid github doesn't offer any of those stats :-/ but i did get a bug or three filed about its content in the last 6 months, so it's not nobody!

peter-mouland:
cool - consider it updated.  I've also updated the example to make it more consistent / readable (it was a while ago i wrote this up, old code never looks good 🤦‍♂️ !!)

PR #1301, opened 2017-10-25
revert peer dependency change for react-test-renderer 14, 15, and 15.4

### LINKED ISSUE #1277: enzyme-adapter-react-15.4 incompatible with react/react-dom 15.3
Hi!

The Enzyme docs ([here](http://airbnb.io/enzyme/#installation)) say to use `enzyme-adapter-react-15.4` when using React 15.0-15.4.

As of #1234 / #1252, `react-addons-test-utils` is no longer a peerDependency of `enzyme-adapter-react-15.4`, but a normal dependency. This means that instead of the user being able to specify a suitable version, the adapter uses the version range `15.0.0-0 - 15.4.x`, which resolves to 15.4.2, and so is incompatible with react/react-dom 15.3.x.

For example when using:
```js
  "dependencies": {
    "react": "15.3.2",
    "react-dom": "15.3.2",
    ...
  },
  "devDependencies": {
    "enzyme": "3.1.0",
    "enzyme-adapter-react-15.4": "1.0.2",
    ...
  }
```

...`enzyme-adapter-react-15.4` pulls in `react-addons-test-utils` 15.4.2, which contains:
```js
module.exports = require('react-dom/lib/ReactTestUtils');
```
...which doesn't exist in react-dom 15.3.2, and so results in:
```
Module not found: Error: Can't resolve 'react-dom/lib/ReactTestUtils' in '....\node_modules\react-addons-test-utils'
```

Compare this to `react-addons-test-utils` 15.3.2, which contains:
```js
module.exports = require('react/lib/ReactTestUtils');
```

I would presume the solutions to be either:
1) Revert the peer dependency change for #1234 / #1252 for older versions of the adapter.
2) Create a new `enzyme-adapter-react-15.3` that has a reduced version range for `react-addons-test-utils`.

Many thanks :-)

CC @brucewpaul, @ljharb, @bdwain.

bdwain:
Fix #1277 by reverting the portions of #1252 that affected react 14, 15, and 15.4. React-test-renderer/react-addons-test-utils will again be a peer dependency for those versions.

Leaving react 16's test renderer as a regular dependency because they changed their versioning scheme for 16 so it should be safe (see https://github.com/facebook/react/issues/11278).

### DIFF
--- a/docs/installation/react-014.md
+++ b/docs/installation/react-014.md
@@ -7,6 +7,12 @@ installed, you should do so:
 npm i --save react@0.14 react-dom@0.14
 ```
 
+Further, enzyme with React 0.14 requires the test utilities addon be installed:
+
+```bash
+npm i --save-dev react-addons-test-utils@0.14
+```
+
 Next, to get started with enzyme, you can simply install it with npm:
 
 ```bash
--- a/docs/installation/react-15.md
+++ b/docs/installation/react-15.md
@@ -7,6 +7,12 @@ installed, you should do so:
 npm i --save react@15 react-dom@15
 ```
 
+Further, enzyme requires the test utilities addon be installed:
+
+```bash
+npm i --save-dev react-test-renderer@15
+```
+
 Next, to get started with enzyme, you can simply install it with npm:
 
 ```bash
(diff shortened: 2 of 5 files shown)


bdwain:
i know :(. At least they changed it for v16.

PR #1311, opened 2017-10-31
fix: move react-test-renderer to hard deps for ^15.5.x adapter

evan-scott-zocdoc:
The ^16.0.0 adapter includes `react-test-renderer` by default, so it was surprising that the appropriate version wasn't included for ^15.5.0 as well.

I left the 15.0-15.4 adapter alone since its dependency structure is different, but the change should perhaps be made for test utils as well.

### DIFF
--- a/packages/enzyme-adapter-react-15/package.json
+++ b/packages/enzyme-adapter-react-15/package.json
@@ -38,13 +38,13 @@
     "lodash": "^4.17.4",
     "object.assign": "^4.0.4",
     "object.values": "^1.0.4",
-    "prop-types": "^15.5.10"
+    "prop-types": "^15.5.10",
+    "react-test-renderer": "^15.5.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


packages/enzyme-adapter-react-15/package.json
```diff
@@ -38,13 +38,13 @@
...
     "object.values": "^1.0.4",
-    "prop-types": "^15.5.10"
+    "prop-types": "^15.5.10",
+    "react-test-renderer": "^15.5.0"
   },
   "peerDependencies": {
     "enzyme": "^3.0.0",
     "react": "^15.5.0",
-    "react-dom": "^15.5.0",
-    "react-test-renderer": "^15.5.0"
```
ljharb: This is intentional; it needs to be a peer dep, because in react < 16, the minor version of react-test-renderer needs to match the minor version of react as well.

See #1301 and #1252 and #1277 for more info.

Issue #1316, opened 2017-11-01
support react's hydrate for enzyme-adapter-react-16


brucewpaul:
It would be great to support testing react's hydrate method as opposed to just the render method. This could maybe be done via an options argument?

cc @ljharb

ljharb:
With an empty DOM node, won't `hydrate` and `render` be identical?

brucewpaul:
Ya I think so, but this would be able to specifically test server side rendering of a component.

cc @goatslacker

ljharb:
Certainly `mount` invokes `ReactDOM.render`; an adapter option could be used in the v16+ adapter to specify `ReactDOM.hydrate` be used.

I'm just not sure what the point would be - all we'd be doing is testing ReactDOM itself, since `hydrate` and `render` theoretically do the same thing on an empty container.

goatslacker:
There's currently a warning when you try to `render()` SSR'd markup, so it would at the very least silence that warning.

ljharb:
In that case, maybe it'd be better to always use `hydrate` in the React 16 adapter here?

goatslacker:
We should have both, for non server rendered content I think you want to use `render`

PR #1317, opened 2017-11-01
add hydrate method that uses reactDOM.hydrate

brucewpaul:
this pr solves the issue in https://github.com/airbnb/enzyme/issues/1316

react16 has a hydrate function, so it would be great to support that in this adapter.


ljharb:
Replaced by #1707.

Issue #1303, opened 2017-10-25
Update docs to remove `lifecycleExperimental` option


maurocarrero:
When `shallow` is used with `lifecycleExperimental` option as false, `componentDidMount` is called.

From [shallow docs](http://airbnb.io/enzyme/docs/api/shallow.html): 
`
options.lifecycleExperimental: (Boolean [optional]): If set to true, the entire lifecycle (componentDidMount and componentDidUpdate) of the React component is called. The current default value is false with enzyme v2, but the next major version will flip the default value to true.
`

You can check the issue in this [repo with example](https://github.com/maurocarrero/react16-enzyme-shallow-did-mount)

Let me know if additional details are needed.
Thanks.

blainekasten:
What version are you using? I think we would want to remove that flag more than fix any old compatibility.

maurocarrero:
Hi @blainekasten, is detailed in the repo's README:
`"enzyme": "^3.1.0",
"enzyme-adapter-react-16": "^1.0.2",`

I would expect that shallow's default behavior avoids running those methods that belongs to a mounted component, but per docs, that would be already changed:
`but the next major version will flip the default value to true`

ljharb:
`lifecycleExperimental` is gone; the flag you want to set is `disableLifecycleMethods: true`.

We may need to update the docs.

maurocarrero:
That option is the one I found to fix the unit tests after upgrading. Everything is in the repo I added at the beginning.

Agree, you may need to update the docs. Thanks.

maurocarrero:
@ljharb I can help with this.

maurocarrero:
https://github.com/airbnb/enzyme/pull/1319

PR #1319, opened 2017-11-03
[lifecycleExperimental] Issue #1303

maurocarrero:
Removes lifecycleExperimental from ShallowWrapper docs.
~~Updates code throwing an error when this option is used.~~
~~Updates related unit tests.~~
Updates CONTRIBUTING doc.

### DIFF
--- a/docs/api/shallow.md
+++ b/docs/api/shallow.md
@@ -46,10 +46,7 @@ describe('<MyComponent />', () => {
   - `options.context`: (`Object` [optional]): Context to be passed into the component
   - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, `componentDidMount`
 is not called on the component, and `componentDidUpdate` is not called after
-[`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md). Default to `false`. 
-  - `options.lifecycleExperimental`: (`Boolean` [optional]): If set to true, the entire lifecycle
-(`componentDidMount` and `componentDidUpdate`) of the React component is called. The current default value
-is `false` with enzyme v2, but the next major version will flip the default value to `true`.
+[`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md). Default to `false`.
 
 #### Returns
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,10 @@ npm run react:14
 npm run react:15
 ```
 
+```bash
+# switch to React 16
+npm run react:16
+```
 
 ### Running Tests
 


docs/api/shallow.md
```diff
@@ -46,10 +46,7 @@ describe('<MyComponent />', () => {
   - `options.context`: (`Object` [optional]): Context to be passed into the component
   - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, `componentDidMount`
 is not called on the component, and `componentDidUpdate` is not called after
-[`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md). Default to `false`. 
-  - `options.lifecycleExperimental`: (`Boolean` [optional]): If set to true, the entire lifecycle
-(`componentDidMount` and `componentDidUpdate`) of the React component is called. The current default value
-is `false` with enzyme v2, but the next major version will flip the default value to `true`.
+[`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md). Default to `false`.
```
maurocarrero: I removed lifecycleExperimental mentions but perhaps you want to add some info related to the fat that now shallow has this behaviour by default.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -3117,74 +3108,15 @@ describe('shallow', () => {
...
-        [true, false, undefined].forEach((value) => {
-          expect(() => shallow(<div />, {
-            lifecycleExperimental: value,
-          })).not.to.throw();
+          })).to.throw(/not longer supported/);
         });
       });
     });
 
-    context('mounting phase', () => {
```
maurocarrero: I guess this is not needed anymore.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -3527,6 +3459,13 @@ describe('shallow', () => {
             { foo: 'bar' }, { foo: 'baz' },
             { foo: 'context' },
           ]);
+        } else {
+          expected.push([
+            'componentDidUpdate',
+            { foo: 'props' }, { foo: 'props' },
+            { foo: 'bar' }, { foo: 'baz' },
+            undefined,
```
maurocarrero: I do not have enough insight into enzyme to evaluate whether this is acceptable, it smells bad. I wanted to create this PR anyway so you can guide me on how this should behave and fix it (or not), the code provoking this is in ShallowWrapper.js:413. If this is as weird to you as to me, I can give it more time and research how the adapter works, that option is undefined for React16.

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -402,7 +391,7 @@ class ShallowWrapper {
         instance.setState(state, callback);
         if (
           shouldRender &&
-          this[OPTIONS].lifecycleExperimental &&
+          !this[OPTIONS].disableLifecycleMethods &&
```
maurocarrero: This change could be the cause for the undefined in the context, my comment above. But it makes sense at first sight.

koba04: This will be fixed at #1261

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
koba04: This will be fixed at #1261

maurocarrero: Will revert this change then.

ljharb:
I'm actually not sure about the code change; it might make this a semver-major.

Perhaps it'd be better to remove it from the docs, but leave it in the API?

maurocarrero:
Agree, it'd be a breaking change, updating only the docs now. 
As it is now those who try to use it (because they already knew the flag), could receive misleading errors and no-op behaviour.
I'd be glad to help further.

PR #1321, opened 2017-11-03
New Package: enzyme-adapter-react-helper

brucewpaul:
This new package aims to help make testing with enzyme on multiple versions of React much easier by abstracting what needs to be installed depending on the version of React.

The idea would be that you would import this package and call `setupEnzymeAdapter(enzymeOptions, adapterOptions)` which will set up enzyme with the correct adapter.

When using it on ci, you will either need to supply the `enzyme-adapter-react-install` script an argument with the correct semver (`enzyme-adapter-react-install 16.0.0`) or as an environment variable (`REACT=16.0.0 enzyme-adapter-react-install`). The environment variable takes precedent (if present) over the supplied argument.


brucewpaul:
Possible Ideas:
- We could make the `enzyme-adapter-react-install` script without any arguments and assume the most recent React version is being used.

brucewpaul:
cc @ljharb

ljharb:
The reason I think a default version should be supplied - although it could totally be `*` or `latest` - is because otherwise it seems like it won't be explicit what's being used.

packages/enzyme-adapter-react-helper/src/enzyme-adapter-react-install.js
```diff
@@ -0,0 +1,56 @@
...
+
+console.log('Cleaning up React and related packages...');
+if (shell.exec('npm uninstall --no-save react react-dom react-test-renderer react-addons-test-utils enzyme-adapter-react-14 enzyme-adapter-react-15.4 enzyme-adapter-react-15 enzyme-adapter-react-16 && rimraf node_modules/react-test-renderer node_modules/react && npm prune').code !== 0) {
+  console.error('An uninstallation failed');
+  shell.exit(1);
+}
+
+console.log(`Installing React@${reactVersion} and related packages...`);
+if (semver.satisfies(reactVersion, '^16.0.0-0')) {
+  if (shell.exec('npm run install-peerdeps -S enzyme-adapter-react-16').code !== 0) {
```
ljharb: i think this needs to be `npm-run`; otherwise it'll look for an "install-peerdeps" script? (same throughout the file)

ljharb: altho actually, i think we need to `require('npm-run')` and use it programmatically?

brucewpaul:
cc @ljharb

ljharb:
Released as v1.0.2 (unpublished the first 2; there were some bugs)

Fauntleroy:
Forgive me if this is the wrong place to ask about this—this is one of the few places I know of that even mentions this package—but are there any instructions for how to use this module? I can't find anything in the docs or readmes. Or is this not intended for external use?

ljharb:
It’s intended for external use, but hasn’t been properly documented yet.

Check out react-dates for an example of if in use; a PR to add docs would be great.

PR #1322, opened 2017-11-03
Adds failing test for cached prop in wrapper

tzvipm:
It seems that when `root.update()` is called, any child wrappers that have already been constructed maintain a cached version of their props.


ljharb:
This is intended behavior in v3.

Issue #1328, opened 2017-11-06
Iterable children not working on environments where Symbol API is not available


Inucoder:
Support for iterable children was recently released https://github.com/airbnb/enzyme/pull/1187. It enables to write Enzyme tests for components that return array-like children, for example ES6 Map and Set, and Immutable List.

We run our unit tests on PhantomJS, and iterable children support was not working for us. But works just fine on latest version of Chrome.

After some discussion with @ljharb in the comments of https://github.com/airbnb/enzyme/pull/1187, I found that the problem is that the iterable support that was implemented for Enzyme relies on the ES6 Symbol API. A polyfill for Symbol makes sense to make it work on our environment, but there are a few reasons why I think a fallback should be added as part of Enzyme instead.

1. It used to work on v2. Enzyme was able to deal with iterable children even though Symbol API was not available.
2. React implements a fallback to `@@iterator` when Symbol API is not available since v 13. Here is a fiddle that shows iterable children support on React 13 https://jsfiddle.net/inucoder/n11gvjgz/ and the source code using `@@iterator` can be found here https://github.com/facebook/react/blob/v0.13.0/src/utils/getIteratorFn.js
3. It's not a hard to fix issue that can help make users of Enzyme life easier.

The purposed fix is as follows. In https://github.com/airbnb/enzyme/blob/master/packages/enzyme-adapter-utils/src/Utils.js from `enzyme-adapter-utils`, we could change the code for `isIterable` function from

```
function isIterable(obj) {
  return (
    obj != null &&
    typeof Symbol === 'function' &&
    typeof Symbol.iterator === 'symbol' &&
    typeof obj[Symbol.iterator] === 'function'
  );
}
```

to

```
function isIterable(obj) {
  return (
    obj != null && (
      (
        typeof Symbol === 'function' &&
        typeof Symbol.iterator === 'symbol' &&
        typeof obj[Symbol.iterator] === 'function'
      ) ||
      obj['@@iterator']
    )
  );
}
```

If you guys think it makes sense to include such change, I'm open to submit a PR during this week (our team is looking forward to bump to v3). Thanks so much and keep the great work!

Inucoder:
After testing above change, I can make below unit test pass when Symbol API is unavailable.

```
'use strict';

import Enzyme        from 'enzyme';
import Immutable     from 'immutable';
import Adapter       from 'enzyme-adapter-react-15';
import {isArrayLike} from 'enzyme-adapter-utils';

[...]

PR #1334, opened 2017-11-07
Better support for iterable children

Inucoder:
…e Object.

Update isIterable function to leverage getIteratorFn.
Rewrite flatten function to leverage getIteratorFn.
Add unit test for flatten function.
Add unit tests for arbitrary iterable children.
Support for Symbol.iterator and @@iterator iterables.

### DIFF
--- a/packages/enzyme-adapter-utils/src/Utils.js
+++ b/packages/enzyme-adapter-utils/src/Utils.js
@@ -110,11 +110,11 @@ function getIteratorFn(obj) {
 }
 
 function isIterable(obj) {
-  return Boolean(getIteratorFn(obj));
+  return !!getIteratorFn(obj);
 }
 
 export function isArrayLike(obj) {
-  return Array.isArray(obj) || (isIterable(obj) && typeof obj !== 'string');
+  return Array.isArray(obj) || (typeof obj !== 'string' && isIterable(obj));
 }
 
 export function flatten(arrs) {
--- a/packages/enzyme-adapter-utils/src/Utils.js
+++ b/packages/enzyme-adapter-utils/src/Utils.js
@@ -92,13 +92,25 @@ export function nodeTypeFromType(type) {
   return 'function';
 }
 
-function isIterable(obj) {
-  return (
-    obj != null &&
-    typeof Symbol === 'function' &&
-    typeof Symbol.iterator === 'symbol' &&
-    typeof obj[Symbol.iterator] === 'function'
+function getIteratorFn(obj) {
+  const iteratorFn = obj && (
+    (
+      typeof Symbol === 'function' &&
+      typeof Symbol.iterator === 'symbol' &&
+      obj[Symbol.iterator]
+    ) ||
+    obj['@@iterator']
   );
+
+  if (typeof iteratorFn === 'function') {
+    return iteratorFn;
+  }
+
+  return undefined;
+}
+
+function isIterable(obj) {
+  return Boolean(getIteratorFn(obj));
 }
 
 export function isArrayLike(obj) {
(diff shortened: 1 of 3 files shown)


Inucoder:
Fixes https://github.com/airbnb/enzyme/issues/1328 by implementing closer to React support for iterable children. @ljharb

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -92,24 +92,55 @@ export function nodeTypeFromType(type) {
...
+
+  if (typeof iteratorFn === 'function') {
+    return iteratorFn;
+  }
+
+  return undefined;
+}
+
+function isIterable(obj) {
+  return Boolean(getIteratorFn(obj));
```
ljharb: this should use `!!` rather than `Boolean()`.

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -92,24 +92,55 @@ export function nodeTypeFromType(type) {
...
 }
 
 export function isArrayLike(obj) {
   return Array.isArray(obj) || (isIterable(obj) && typeof obj !== 'string');
```
ljharb: (predates your PR, but still) let's flip the typeof check here; `isIterable` shouldn't be called at all in the string case.

packages/enzyme-test-suite/test/RSTTraversal-spec.jsx
```diff
@@ -316,6 +316,115 @@ describe('RSTTraversal', () => {
       expect(nodes).to.deep.equal([node, divA, divB, divC, divD]);
     });
 
+    describe('support for arbitrary iterable children', () => {
+      const makeDivIterator = (lowerBound, upperBound) => {
+        let counter = lowerBound;
```
ljharb: `let counter = lowerBound + 'a'.charCodeAt(0);` would avoid needing to look up "a"'s charCode every time

Inucoder:
@ljharb, working on the feedback from your review. Thanks so much.

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -92,24 +92,55 @@ export function nodeTypeFromType(type) {
...
+    } else {
+      flatItem = item;
+    }
+
+    flatArrs = flatArrs.concat(flatItem);
+
+    step = iterator.next();
+  }
+
+  return flatArrs;
```
aweary: This feels like a lot of added complexity when `flatten` should almost always be called with arrays.  It just seems like we're doing the same work that `[...iterable]` is doing, except that this approach won't benefit from any internal optimizations that the spread operator may have.

Inucoder: Our use case is running our unit test suite on PhantomJS, where some of our components return Immutable List instance of React elements. As described in https://github.com/airbnb/enzyme/issues/1328, that feature was recently added but was not working on PhantomJS due the lack of support for Symbol.iterator. I agree the increased complexity is evident, and that's why I aimed for good test coverage. The main reason to make this change is to catch up with React, which is very flexible with the children is able to handle, and switching from v2 to v3 may break some middle to large projects using a pattern like us and results in additional effort to make the bump. Really appreciate your comments.

Inucoder: @aweary, what if I add below code at the beginning of `flatten` function?

```
if (Array.isArray(arrs)) {
  return arrs.reduce(
    (flattened, item) => flattened.concat(isArrayLike(item) ? flatten(item) : item),
    [],
  );
}
```

That way `flatten` will be as performant as it used to before my change, for the most common case where `arrs` is a regular JS `Array`. But gracefully fallbacks to support any kind of iterable children React can also handle.

Let me know your thoughts. cc @ljharb

ljharb: Seems reasonable to me.

Inucoder: Cool, I'm going to make that change and push.

Inucoder:
Made the optimization for the most common case and pushed @ljharb cc @aweary. Thanks for your feedback.

ljharb:
@Inucoder would you please rebase this on the command line, so as to remove all merge commits?

Inucoder:
It's done @ljharb, a coworker gave me a helping hand to do the rebase.

ljharb:
@Inucoder thanks! it looks like it's still out of date with master tho; you may need to fetch and redo the rebase? (if this gets onerous, I'm happy to take care of it for you)

Inucoder:
Should be good now @ljharb, not a big deal.

lelandrichardson:
Seems good to me. Thanks for working on this.

PR #1350, opened 2017-11-12
Change ShallowWrapper.text() trim spaces with same behavior as ReactWrapper.text()

idanilt:
ReactWrapper.text() return string with the spaces as is, ShallowWrapper.text() removing double spaces and replacing &nbsp; with regular space (removing doubles as well).

### DIFF
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -1648,11 +1648,12 @@ describe('shallow', () => {
           <div>&nbsp;</div>
           <div>&nbsp;&nbsp;</div>
         </div>
-      )
+      );
+
       const wrapper = shallow(Space);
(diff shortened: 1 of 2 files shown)


packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1633,6 +1633,29 @@ describe('shallow', () => {
       matchesRender(<div>&gt;</div>);
     });
 
+    it('should handle spaces with same behavior as ReactWarpper.text()', () => {
```
ljharb: `ReactWarpper.text` → `ReactWrapper.text`

idanilt:
In the end ReactWrapper.text() don't do anything to spaces, so the solution was to remove the `replace` at all.
I added a fix for `npm run react:n` for Windows machines

ljharb:
@idanilt would you mind rebasing this on latest master?

PR #1360, opened 2017-11-14
Fix code example in .exists() documentation

mkarajohn:
Corrects `.to.be(false)` to `to.be.false`

### DIFF
--- a/docs/api/ShallowWrapper/exists.md
+++ b/docs/api/ShallowWrapper/exists.md
@@ -14,5 +14,5 @@ Returns whether or not the current node exists.
 
 ```jsx
 const wrapper = shallow(<div className="some-class" />);
-expect(wrapper.find('.other-class').exists()).to.be(false);
+expect(wrapper.find('.other-class').exists()).to.equal(false);
 ```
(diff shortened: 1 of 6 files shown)


docs/api/ShallowWrapper/exists.md
```diff
@@ -14,5 +14,5 @@ Returns whether or not the current node exists.
 
 ```jsx
 const wrapper = shallow(<div className="some-class" />);
-expect(wrapper.find('.other-class').exists()).to.be(false);
```
ljharb: This is actually correct as-is; we use chai's expect, not jest's; separately, no-op matchers (ie, any matcher that isn't a function call) are extremely dangerous, because `expect(value).to.be.fals` or `expect(value).to.be.yogurt` will both silently fail even though those matchers don't exist.

mkarajohn: I am using chai's `expect` as well, not using jest, and `.to.be()` throws with `.to.be is not a function` and as much as I searched I could not find any part of the documentation where chai's `.be` is used as a function. 

I think something from your reply must be going over my head :/

ljharb: hmmmm good call, you're right. however, `.to.be.false` isn't the right fix :-)

ljharb: ok so, you're totally right :-) let's change this to `.to.equal(false)`

mkarajohn: Done!

PR #1361, opened 2017-11-14
[DOCS] Add callback arg to setProps header in mount.md

DannyDelott:
Match the signature as expressed in`setProps.md`

### DIFF
--- a/docs/api/mount.md
+++ b/docs/api/mount.md
@@ -162,7 +162,7 @@ Simulates an event on the current node.
 #### [`.setState(nextState) => ReactWrapper`](ReactWrapper/setState.md)
 Manually sets state of the root component.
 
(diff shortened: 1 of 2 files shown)


docs/api/mount.md
```diff
@@ -170,7 +170,7 @@ Simulates an event on the current node.
 #### [`.setState(nextState) => ReactWrapper`](ReactWrapper/setState.md)
 Manually sets state of the root component.
 
-#### [`.setProps(nextProps) => ReactWrapper`](ReactWrapper/setProps.md)
+#### [`.setProps(nextProps[, callback]) => ReactWrapper`](ReactWrapper/setProps.md)
```
ljharb: Thanks, does a similar change need to be made in the shallow docs?

DannyDelott: Without looking at any of the code, the documentation for shallow's `setProps` does not describe a callback argument.

Issue #1366, opened 2017-11-15
setState callback is not called (React 16?)


mabounassif:
The following issue is still an issue: https://github.com/airbnb/enzyme/issues/953

It was blocked because of `react-test-renderer`, which fixed the issue in the following PR: https://github.com/facebook/react/pull/10106. 

Via this commit: https://github.com/facebook/react/commit/5ac62095996cce7b5441d9fad73fbcda1e3d2fca

Basically we must update `react-test-renderer` to `^16.1.1`

mabounassif:
Update here: https://github.com/airbnb/enzyme/pull/1367

PR #1367, opened 2017-11-15
Update react test renderer to v16.1.1 for React@16

mabounassif:
This addressed the following issue: https://github.com/airbnb/enzyme/issues/1366

### DIFF
--- a/packages/enzyme-adapter-react-16/package.json
+++ b/packages/enzyme-adapter-react-16/package.json
@@ -39,7 +39,7 @@
     "object.assign": "^4.0.4",
     "object.values": "^1.0.4",
     "prop-types": "^15.5.10",
-    "react-test-renderer": "^16.0.0-0"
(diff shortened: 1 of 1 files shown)


bdwain:
I don't think this change is necessary. ^16.0.0-0 is satisfied by 16.1.1. The user just needs to say `npm update` or `yarn upgrade` and it should update test renderer for them. No need to change the semver every time they release a patch.

ljharb:
Since this fixes a bug, it might be a good change to make - but does this react-test-renderer work with all versions of react v16?

bdwain:
it does work. they intend to support all of v16 as they update test renderer. the peer dependency on react is only ^16.0.0

mabounassif:
I agree with what has been said above. I should have realized it was a caret - I'm a bit embarrassed :)

Closing this.

PR #1382, opened 2017-11-24
Update of text.md

jorrit-g:
JSX element are stripped from returned string as far as I understand

### DIFF
--- a/docs/api/ReactWrapper/text.md
+++ b/docs/api/ReactWrapper/text.md
@@ -21,8 +21,12 @@ expect(wrapper.text()).to.equal('important');
 ```
 
 ```jsx
-const wrapper = mount(<div><Foo /><b>important</b></div>);
-expect(wrapper.text()).to.equal('<Foo />important');
+function Foo() {
+  return <div>This is</div>;
+}
+
+const wrapper = mount(<div><Foo /> <b>really</b> important</div>);
+expect(wrapper.text()).to.equal('This is really important');
 ```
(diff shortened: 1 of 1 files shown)


docs/api/ReactWrapper/text.md
```diff
@@ -22,7 +22,7 @@ expect(wrapper.text()).to.equal('important');
 
 ```jsx
 const wrapper = mount(<div><Foo /><b>important</b></div>);
-expect(wrapper.text()).to.equal('<Foo />important');
+expect(wrapper.text()).to.equal('important');
```
ljharb: Good catch! The text would actually also include whatever text `Foo` rendered; perhaps this should also include a definition for a `Foo`?

docs/api/ReactWrapper/text.md
```diff
@@ -21,8 +21,8 @@ expect(wrapper.text()).to.equal('important');
 ```
 
 ```jsx
-const wrapper = mount(<div><Foo /><b>important</b></div>);
-expect(wrapper.text()).to.equal('<Foo />important');
+const wrapper = mount(<div><Foo>This is <b>really</b> important</Foo></div>);
```
ljharb: sorry for the confusion; the children you pass into `Foo` are irrelevant; it matters what the `Foo` component *itself* renders.

I was envisioning something like
```jsx
function Foo() {
  return <div>this is a foo</div>
}
```

PR #1385, opened 2017-11-26
Fix Sort Order of Full DOM Rendering Methods

dhurlburtusa:
I noticed that two methods were not in the apparent alphabetical order.


ljharb:
Thanks!

PR #1404, opened 2017-11-29
patch semver range due to bug in semver.intersects

brucewpaul:
this pr fixes an issue where if you supplied 15 as `reactVersion` to `semver.intersects(reactVersion, '^16.0.0-0')`, it would return true. This is a bug in semver.intersects and the `-0` will have to removed until it is fixed

cc @ljharb


ljharb:
(linking to https://github.com/npm/node-semver/issues/223)

PR #1423, opened 2017-12-08
[Fix] Convert nodes to RST nodes before comparing

### LINKED ISSUE #1422: Add RST mapping to contains
I'm making an enzyme-inferno-adapter, but I've hit some problems getting the last few tests to pass. 

Enzyme requires `className` to be inside `props`. This is fine normally, because we can convert the inferno VNode tree to an enzyme RST tree. 

But when you call `contains`, enzyme compares the RST node to the newly created VNode. The RST node has `className` inside `props`, the inferno VNode has className on the vNode.

Example:

RST Node:

```js
{
type: 'div',
  props: {
    className: 'bar'
  }
}
```

Inferno VNode
```js
type: 'div',
  props: {
  },
  className: 'bar'
}
```

That means we can't call a method like this:
```js
const a = <div className="bar" />;
expect(mount(Component).contains(a)).to.equal(true);
```

Can we convert the node that `contains` is called with to an RST node before comparing? 

I'm happy to make the PR

eddyerburgh:
- Convert nodes in `contains` functions to RST before comparing 

- Fixes #1422

### DIFF
--- a/packages/enzyme/src/ReactWrapper.js
+++ b/packages/enzyme/src/ReactWrapper.js
@@ -374,9 +374,13 @@ class ReactWrapper {
    * @returns {Boolean}
    */
   contains(nodeOrNodes) {
+    const isArray = Array.isArray(nodeOrNodes);
+    const rstNodeOrNodes = isArray
+      ? nodeOrNodes.map(getAdapter().elementToNode)
+      : getAdapter().elementToNode(nodeOrNodes);
     const predicate = Array.isArray(nodeOrNodes)
-      ? other => containsChildrenSubArray(nodeEqual, other, nodeOrNodes)
-      : other => nodeEqual(nodeOrNodes, other);
+      ? other => containsChildrenSubArray(nodeEqual, other, rstNodeOrNodes)
+      : other => nodeEqual(rstNodeOrNodes, other);
     return findWhereUnwrapped(this, predicate).length > 0;
   }
 
@@ -397,7 +401,8 @@ class ReactWrapper {
    * @returns {Boolean}
    */
   containsMatchingElement(node) {
-    const predicate = other => nodeMatches(node, other, (a, b) => a <= b);
+    const rstNode = getAdapter().elementToNode(node);
+    const predicate = other => nodeMatches(rstNode, other, (a, b) => a <= b);
     return findWhereUnwrapped(this, predicate).length > 0;
   }
 
@@ -424,7 +429,9 @@ class ReactWrapper {
       throw new TypeError('nodes should be an Array');
     }
 
-    return nodes.every(node => this.containsMatchingElement(node));
+    const rstNodes = nodes.map(getAdapter().elementToNode);
+
+    return rstNodes.every(rstNode => this.containsMatchingElement(rstNode));
   }
 
   /**
@@ -446,7 +453,10 @@ class ReactWrapper {
    * @returns {Boolean}
    */
   containsAnyMatchingElements(nodes) {
-    return Array.isArray(nodes) && nodes.some(node => this.containsMatchingElement(node));
+    const rstNodes = nodes.map(getAdapter().elementToNode);
+
+    return Array.isArray(rstNodes)
+    && rstNodes.some(rstNode => this.containsMatchingElement(rstNode));
   }
 
   /**
--- a/packages/enzyme/src/ReactWrapper.js
+++ b/packages/enzyme/src/ReactWrapper.js
@@ -358,8 +358,10 @@ class ReactWrapper {
    * @returns {Boolean}
    */
   matchesElement(node) {
-    const rstNode = getAdapter().elementToNode(node);
-    return this.single('matchesElement', () => nodeMatches(rstNode, this.getNodeInternal(), (a, b) => a <= b));
+    return this.single('matchesElement', () => {
+      const rstNode = getAdapter().elementToNode(node);
+      return nodeMatches(rstNode, this.getNodeInternal(), (a, b) => a <= b);
+    });
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/enzyme/src/ReactWrapper.js
```diff
@@ -446,7 +454,10 @@ class ReactWrapper {
    * @returns {Boolean}
    */
   containsAnyMatchingElements(nodes) {
-    return Array.isArray(nodes) && nodes.some(node => this.containsMatchingElement(node));
+    const rstNodes = nodes.map(getAdapter().elementToNode);
```
ljharb: this change makes `.map` be called before `Array.isArray` is called to ensure it's an array.

Let's change this to `if (!Array.isArray(nodes)) { return false; }` at the top of the function, and then we can do the entire check in a single `some`

packages/enzyme/src/ReactWrapper.js
```diff
@@ -374,9 +375,13 @@ class ReactWrapper {
    * @returns {Boolean}
    */
   contains(nodeOrNodes) {
+    const isArray = Array.isArray(nodeOrNodes);
+    const rstNodeOrNodes = isArray
+      ? nodeOrNodes.map(getAdapter().elementToNode)
+      : getAdapter().elementToNode(nodeOrNodes);
     const predicate = Array.isArray(nodeOrNodes)
```
ljharb: this line can use `isArray` now, I think?

packages/enzyme/src/ReactWrapper.js
```diff
@@ -374,9 +375,13 @@ class ReactWrapper {
...
-      ? other => containsChildrenSubArray(nodeEqual, other, nodeOrNodes)
-      : other => nodeEqual(nodeOrNodes, other);
+      ? other => containsChildrenSubArray(nodeEqual, other, rstNodeOrNodes)
```
ljharb: probably at this point, rather than two paired multiline ternaries, this would be better as an `if`/`else` branch.

packages/enzyme/src/ReactWrapper.js
```diff
@@ -358,7 +358,8 @@ class ReactWrapper {
    * @returns {Boolean}
    */
   matchesElement(node) {
-    return this.single('matchesElement', () => nodeMatches(node, this.getNodeInternal(), (a, b) => a <= b));
+    const rstNode = getAdapter().elementToNode(node);
+    return this.single('matchesElement', () => nodeMatches(rstNode, this.getNodeInternal(), (a, b) => a <= b));
```
ljharb: let's do the elementToNode conversion here inside the arrow function, so that it happens after `this.single` succeeds.

ljharb:
@eddyerburgh is this change not needed in `shallow` as well?

eddyerburgh:
@ljharb Yes, you're right. I've only been implementing mount so haven't looked into shallow. I'll add the changes now

packages/enzyme/src/ReactWrapper.js
```diff
@@ -374,9 +377,16 @@ class ReactWrapper {
...
   contains(nodeOrNodes) {
+    const adapter = getAdapter(this[OPTIONS]);
+
     const predicate = Array.isArray(nodeOrNodes)
-      ? other => containsChildrenSubArray(nodeEqual, other, nodeOrNodes)
-      : other => nodeEqual(nodeOrNodes, other);
+      ? other => containsChildrenSubArray(
+        nodeEqual,
+        other,
+        nodeOrNodes.map(adapter.elementToNode),
```
ljharb: Let’s make this use an arrow function, so adapters only get the single argument

eddyerburgh:
It was already implemented for contains in ShallowWrapper, so I refactored `contains` in ReactWrapper. 

I also realized `containsAllMatchingElements` and `containsAnyMatchingElements` call `containsMatchingElement`, so I removed the unncessary calls to `containsMatchingElement`.

Finally, while I'm here, in `ReactWrapper.contains` there's no check that the node/nodes is a React element. Should I add a test and write a check, like there is [in `ShallowWrapper.contains`](https://github.com/airbnb/enzyme/blob/master/packages/enzyme/src/ShallowWrapper.js#L441)?

packages/enzyme/src/ReactWrapper.js
ljharb: I also see this is how it already is in https://github.com/airbnb/enzyme/pull/1423/files#diff-fa27c82ee4fde075bae6173848e17c82R448; let’s change both

ljharb:
> I also realized containsAllMatchingElements and containsAnyMatchingElements call containsMatchingElement, so I removed the unncessary calls to containsMatchingElement.

This was done intentionally, so there couldn’t be any drift between how these three methods worked - please keep them linked.

ljharb:
In `mount`, someone might pass a DOM node; I’m not sure what the best solution is there.

eddyerburgh:
>  so I removed the unncessary calls to `containsMatchingElement`

Sorry, I meant that I removed the unnecessary calls to `elementToNode`. They are all still linked.

ljharb:
gotcha, sounds good then :-)

packages/enzyme/src/ReactWrapper.js
eddyerburgh: > Let’s make this use an arrow function, so adapters only get the single argument
 
Do you mean the `nodeOrNodes` `map` callback should be an arrow function?

ljharb: yep! like `.map(x => adapter.elementToNode(x))`

eddyerburgh: Done 🙂

eddyerburgh:
Thanks @ljharb . Is there anything else you need me to do to get this merged?

Also, do you know when this will be released?

ljharb:
@eddyerburgh at the moment, you're good; i'm waiting for other collaborators to review it first.

After it's merged, it will be released whenever we're ready to publish the next one :-) hopefully soon.

eddyerburgh:
@collaborators, could we get this reviewed and merged?

ljharb:

PR #1431, opened 2017-12-13
Add example to contains() docs with an embedded expression

willj:
This may be obvious to most, but it wasn't for me, that contains() is comparing JSX (rather than the resulting html) and therefore curly braces are required with embedded expressions.

### DIFF
--- a/docs/api/ShallowWrapper/contains.md
+++ b/docs/api/ShallowWrapper/contains.md
@@ -56,6 +56,17 @@ expect(wrapper.contains([
 ])).to.equal(false);
 ```
 
+```jsx
+const calculatedValue = 2 + 2;
+
+const wrapper = shallow((
+  <div>
+    <div data-foo="foo" data-bar="bar">{calculatedValue}</div>
+  </div>
+));
+
+expect(wrapper.contains(<div data-foo="foo" data-bar="bar">{4}</div>)).to.equal(true);
+```
 
 #### Common Gotchas
 


docs/api/ShallowWrapper/contains.md
```diff
@@ -56,10 +56,22 @@ expect(wrapper.contains([
...
+expect(wrapper.contains(<div data-foo="foo" data-bar="bar">{4}</div>)).to.equal(true);
+```
 
 #### Common Gotchas
 
 - `.contains()` expects a ReactElement, not a selector (like many other methods). Make sure that
 when you are calling it you are calling it with a ReactElement or a JSX expression.
 - Keep in mind that this method determines equality based on the equality of the node's children as
 well.
+- Embedded expressions must be wrapped with curly braces
```
ljharb: This part, I'm not so sure about - this is just how jsx works, and isn't required if you're using `React.createElement`.

docs/api/ShallowWrapper/contains.md
```diff
@@ -56,10 +56,22 @@ expect(wrapper.contains([
...
+```jsx
+let calculatedValue = 2 + 2;
+
+const wrapper = shallow((
+  <div>
+    <div data-foo="foo" data-bar="bar">{calculatedValue}</div>
+  </div>
+));
+
+expect(wrapper.contains(<div data-foo="foo" data-bar="bar">{4}</div>)).to.equal(true);
```
ljharb: This is a great addition.

docs/api/ShallowWrapper/contains.md
willj: I certainly would've picked up my error much earlier with just the example added above, so I'll remove this line.

I think it was just the fact all the sample test JSX happened to be showing valid HTML which reinforced my newbie assumptions :)

docs/api/ShallowWrapper/contains.md
```diff
@@ -56,6 +56,17 @@ expect(wrapper.contains([
 ])).to.equal(false);
 ```
 
+```jsx
+let calculatedValue = 2 + 2;
```
ljharb: this needs to be a `const`

willj: Sorry I missed this, I've fixed this, and rebased to 1 commit on the latest master

PR #1450, opened 2017-12-22
Printing String objects as strings instead of objects

joeldenning:
This pull request updates the way that nodes are printed when being debugged.

#### Use case in question

```jsx
// Sorta weird to provide a String object as a prop instead of a string, but React supports it.
<div className={new String('foo')} />
```

#### Previous behavior
Debugging the above node previously logged the following:
```jsx
<div className={{...}} />
```

#### Proposed behavior
I propose it would make more sense to print props that are String objects as strings:

```jsx
<div className="foo" />
```

#### Notes
Providing String objects for element props like className is something that is supported by the DOM, and thus supported by React.

- See [this react fiddle](https://jsfiddle.net/69z2wepo/96039/) for an example of String objects being accepted as the className prop in React.
- Run the following code in a browser console to see that this behavior is something supported by browsers themselves:

```js
document.body.className = new String('foo');
document.body.className; // this will log "foo"
```


packages/enzyme/src/Debug.js
```diff
@@ -32,7 +32,11 @@ function propString(prop) {
     case 'boolean':
       return `{${prop}}`;
     case 'object':
-      return '{{...}}';
+      if (prop instanceof String) {
```
ljharb: `instanceof` is never reliable, and should not be used.

If we want to detect boxed strings, we need to use the `is-string` library.

packages/enzyme/src/Debug.js
```diff
@@ -32,7 +32,11 @@ function propString(prop) {
...
+        return `"${prop.toString()}"`;
```
ljharb: `String(x)` is always strictly better than `x.toString()`, and inside a template literal, both are unnecessary - this can just be ``` `"${prop}"` ```

joeldenning:
@ljharb thanks for the feedback - just pushed some changes.

packages/enzyme/src/Debug.js
```diff
@@ -32,7 +33,11 @@ function propString(prop) {
     case 'boolean':
       return `{${prop}}`;
     case 'object':
-      return '{{...}}';
+      if (isString(prop)) {
+        return `"${prop}"`;
+      } else {
```
ljharb: The else here isn’t necessary; the following return can be de-indented

joeldenning: 👍 updated

PR #1453, opened 2017-12-26
Fix to call a setState callback after finishing the the render

### LINKED ISSUE #1451: setState on shallowWrapper needs explicit update
After upgrading to Enzyme V3, we started observing that now we need explicit update after setState. I read the migration guide and somehow i didn't get sense this is expected and hence raising issue here, please correct me if my understanding is wrong.

Here is the sample component

```
export class ComponentA { 

    public render(): JSX.Element {

        return (
            <div className={"myClass"}>
                {
                    this.state.myState ?
                        this._getComponentB() :
                        this.props.children
                }
            </div>
        );
    }

    private _getComponentB(): JSX.Element {
        return (
            <ComponentB
                text={this.props.text} />
        );
    }
}
```

In the test I am trying to do following, see in the setState callback I have to explicit update now to make test pass. Is this expected? I feel this is common scenario and enzyme should handle this.

```
let shallowWrapper = Enzyme.shallow(<ComponentA text="text" />);

shallowWrapper.setState({
                    isLoading: false
                }, () => {
			shallowWrapper.update();
			// Assertion here
		});
```

koba04:
This PR is to fix #1451.
Currently, a setState callback is called before finishing the render because enzyme is passing the callback into `instance.setState`.
So I've fixed it to call the callback after `this.update()`


packages/enzyme/src/ShallowWrapper.js
```diff
@@ -400,6 +402,10 @@ class ShallowWrapper {
           }
         }
         this.update();
+        // call the setState callback
+        if (callback) {
+          callback.call(instance);
```
ljharb: are we sure that React calls the callback with zero arguments, and the instance as the receiver?

koba04: > are we sure that React calls the callback with zero arguments, and the instance as the receiver?

I think so.

* Implementation
    * https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberUpdateQueue.js#L352
* Demo
    * https://codepen.io/koba04/pen/XVpEEy?editors=0011

ljharb: That implementation doesn't show where `update.callback` is assigned; but your demo is convincing (for the latest version of React).

However, in React ^15.4, it seems to get one argument set to `undefined` in your demo (whereas in <= 15.3, and >= 16, it seems to get zero arguments). We should maintain that behavior via adapters.

koba04: Thanks! I'll work on to support < 15.3

koba04: I've tested for arguments of setState callback.

```js
setState({foo: 'bar'}, function(...args) {
  console.log(args);
});
```

* 0.13.3 ... `[]`
* 0.14.9 ... `[]`
* 15.4.2 ... `[ undefined ]`
* 15.6.2 ... `[ undefined ]`
* 16.2.0 ... `[]`

The arguments seem to be `[undefined]` only in v15.

ljharb: In my testing, it's `^15.4` - ie, >= 15.4 and < 16. If you check `15.3`, you should get `[]`.

koba04: Thanks, you are right. I've updated the list.

* 0.13.3 ... `[]`
* 0.14.9 ... `[]`
* 15.3.2 ... `[]`
* 15.4.2 ... `[ undefined ]`
* 15.4.2 ... `[ undefined ]`
* 15.5.4 ... `[ undefined ]`
* 15.6.2 ... `[ undefined ]`
* 16.2.0 ... `[]`

koba04: Changing the argument by depending on React's version seems to be painful.
So I'll try to update the implementation to use a spy method like  https://github.com/airbnb/enzyme/pull/1192 to know the argument.

ljharb: It's more than just painful; enzyme itself must not know about React at all, given that non-React adapters need to be able to work.

In other words, this functionality needs to be something that every adapter implements, and enzyme merely calls into.

koba04: > enzyme itself must not know about React at all, given that non-React adapters need to be able to work.

I agree with you. But calling setState callback without argument is depending on current React implementation.
Spying setState callback and using the result later is to reproduce this functionality so I think it can be applied any adapters which are including non-React adapters.

```js
instance.setState(state, spy);
:
if (callback) {
  // replay setState callbacks
  spy.getCalls().forEach((spyCall) => {
    const context = spyCall.thisValue;
    const args = spyCall.args;
    callback.apply(context, args);
  });
}
```

ljharb: Your spy suggestion is interesting, but I was suggesting something like `adapter.invokeSetStateCallback(instance, callback)` - I'm not sure which is better.

koba04: Thanks, I got it.
Are you ok to implement a default behavior into `EnzymeAdapter`? If not, it will be a breaking change.

ljharb: Yes, I think that would make sense.

packages/enzyme-test-suite/package.json
```diff
@@ -49,4 +49,4 @@
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-react": "^7.5.1"
   }
-}
+}
```
ljharb: let's revert this file; every file should always have a trailing newline

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -385,7 +385,9 @@ class ShallowWrapper {
             return shouldRender;
           };
         }
-        instance.setState(state, callback);
+        // We don't pass the setState callback here
+        // to guarantee to call the callback after finishing the render
```
ljharb: can we add a test case for this? or does `expect(wrapper.find('div').prop('className')).to.eql('bar');` cover that

koba04: Yes, `expect(wrapper.find('div').prop('className')).to.eql('bar');` is a test for this.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -8,12 +8,16 @@ import { ITERATOR_SYMBOL, withSetStateAllowed, sym } from 'enzyme/build/Utils';
...
-import { REACT013, REACT014, REACT16, is } from './_helpers/version';
+import { REACT013, REACT014, REACT16, REACT154, REACT155, REACT156, is } from './_helpers/version';
 
 // The shallow renderer in react 16 does not yet support batched updates. When it does,
 // we should be able to go un-skip all of the tests that are skipped with this flag.
 const BATCHING = !REACT16;
 
+
+// some React versions pass undefined as an argument of setState callback.
+const CALLING_SETSTATE_CALLBACK_WITH_UNDEFINED = REACT154 || REACT155 || REACT156;
```
ljharb: could this be made a bit more robust by changing it to `!REACT15 && !REACT150 && !REACT151 && !REACT152 && !REACT153`, since those are the only 4 versions of 15 that will ever not have this behavior?

packages/enzyme-adapter-react-15.4/src/ReactFifteenFourAdapter.js
```diff
@@ -252,6 +254,15 @@ class ReactFifteenFourAdapter extends EnzymeAdapter {
   createElement(...args) {
     return React.createElement(...args);
   }
+
+  invokeSetStateCallback(instance, callback) {
+    // React in <= 15.3, and >= 16 pass undefined to a setState callback
+    if (MINOR_VERSION >= 4) {
+      callback.call(instance, undefined);
+    } else {
+      callback.call(instance);
```
ljharb: maybe this could do `super.invokeSetStateCallback(instance, callback)`?

koba04:
Thank you for your review!

ljharb:
Tests seem to be failing in React 15, both locally and in travis. I'm going to see what I can do to figure it out now.

packages/enzyme-test-suite/package.json
koba04: The diff is added again😉 I'm not digging into that but I guess it is added by a script to run tests.

packages/enzyme-adapter-react-15.4/src/ReactFifteenFourAdapter.js
```diff
@@ -252,6 +253,15 @@ class ReactFifteenFourAdapter extends EnzymeAdapter {
...
     return React.createElement(...args);
   }
+
+  invokeSetStateCallback(instance, callback) {
+    // React in >= 15.4, and < 16 pass undefined to a setState callback
+    ifReact(
+      '^15.4',
+      () => { callback.call(instance, undefined); },
+      () => { super.invokeSetStateCallback(instance, callback); },
+    );
```
koba04: `ifReact` returns a function, not calling it so you have to call a function which is returned from `ifReact`.

https://github.com/airbnb/enzyme/blob/master/packages/enzyme-adapter-react-helper/src/ifReact.js#L7

packages/enzyme-test-suite/package.json
ljharb: ah, weird. maybe npm itself. not a big deal.

packages/enzyme-adapter-react-15.4/src/ReactFifteenFourAdapter.js
ljharb: lol, oops. shouldn't tests have failed?

ljharb: Will fix in a followup.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -8,12 +8,15 @@ import { ITERATOR_SYMBOL, withSetStateAllowed, sym } from 'enzyme/build/Utils';
...
 import { describeIf, itIf, itWithData, generateEmptyRenderData } from './_helpers';
-import { REACT013, REACT014, REACT16, is } from './_helpers/version';
+import { REACT013, REACT014, REACT15, REACT150_4, REACT16, is } from './_helpers/version';
 
 // The shallow renderer in react 16 does not yet support batched updates. When it does,
 // we should be able to go un-skip all of the tests that are skipped with this flag.
 const BATCHING = !REACT16;
 
+// some React versions pass undefined as an argument of setState callback.
+const CALLING_SETSTATE_CALLBACK_WITH_UNDEFINED = REACT15 && !REACT150_4;
```
koba04: I guess this condition doesn't cover a case that React v15.4 calls setState callback with `undefined`.

But all tests are passing in my environment. The reason is `ifReact` is checking with React v16.2 because `enzyme-adapter-react-helper` has React v16.2 into its `node_modules`.
I'm not sure why `enzyme-adapter-react-helper` has React v16.2 into its `node_modules`.

packages/enzyme-adapter-react-15.4/src/ReactFifteenFourAdapter.js
koba04: I think so we should add a test to guarantee to call a setState callback.
Currently, we don't have a test for it.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
koba04: Thanks! I'll dig into it later.

koba04: `enzyme-adapter-react-helper` shouldn't have `react` as a devDependencies?

https://github.com/airbnb/enzyme/blob/master/packages/enzyme-adapter-react-helper/package.json#L50

ljharb: since it's a peer dep, it also needs to be a dev dep.

koba04: Other packages don't have react as a devDeps. Why does only this package have it?

ljharb: That's a bug leftover from the v3 migration; they all should - every peer dep in every npm package anywhere must always have the identical peer dep string as a dep or a dev dep.

ljharb: I've almost got a fix PR with a regression test; i'll put it up shortly.

ljharb: https://github.com/airbnb/enzyme/pull/1465

Issue #1475, opened 2018-01-12
wrapper.find(y).at(x).exists() always returns true


crymis:
### Current behavior
`expect(wrapper.find('DOESNOTEXIST').at(0).exists()).toBe(true);` 
^ always works

### Expected behavior
`expect(wrapper.find('DOESNOTEXIST').at(0).exists()).toBe(false);`

#### API

- [x] shallow
- [ ] mount
- [ ] render

#### Version

| library         | version
| ---------------- | -------
| Enzyme           | ^3.1.1
| React            |  ^16.0.0

#### Adapter

- [x] enzyme-adapter-react-16 (^1.0.4)
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-15.4
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] enzyme-adapter-react-helper
- [ ] others ( )

ljharb:
We have explicit tests for this:
 - https://github.com/airbnb/enzyme/blob/master/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx#L2810-L2818
 - https://github.com/airbnb/enzyme/blob/master/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx#L2826-L2832

Can you provide the code for whatever constructs `wrapper`, and also `wrapper.find('DOESNOTEXIST').at(0).debug()`?

crymis:
Thanks for the answer.
The issue is related with the combination of the `at` and the `exists` methods.
I can provide a code example next week.

KordonDev:
Hi
I have added a pull request, which adds a failing test in ShallowWrapper just after the one mentioned by you. Hope this helps for understanding.

The problem is in short form, that `at(0)` returns undefined, which will be a ShallowWrapper with 1 child, which is `undefined`. And exists only checks if the length is greater than 0.

PR #1478, opened 2018-01-13
Add failing .at(0).exists() test

KordonDev:
This pull request adds a test for the issue https://github.com/airbnb/enzyme/issues/1475 but doesn't fix it.


KordonDev:
I moved it to `.at()`

ljharb:
I'm tweaking this a bit. `wrapper.find('.foo')` is *supposed* to find nothing, because `.find` does not consider the wrapper itself - and the wrapper *is* a `.foo`.

PR #1492, opened 2018-01-25
Add null option to shallow type() comment & docs

ialexryan:
Also fixes a docs typo I ran into that was replicated in a few places.


ialexryan:
done!

Issue #1149, opened 2017-09-26
[Improved Fiber Support] Support Array-rendering components


lelandrichardson:
I believe we still have a little bit of work to do on this front.

nickmccurdy:
I noticed that all my `.find()` calls fail in components that I've migrated to Array returns. It seems like I need to avoid it entirely until this is supported in Enzyme.

aamirafridi:
👍🏼
I have to wrap the array of jsx into a div just to pass the tests :(

udivankin:
FYI wrapping code with <Fragment /> that was implemented in React v16.2 seems to do the trick.

zerubeus:
@udivankin Nice but we have to migrate to React 16.2 to be able to use this new API :/

FezVrasta:
And `Fragment` is not supported by Enzyme, anyway

tkrotoff:
@udivankin 
> `<Fragment />` seems to do the trick

Does not work

jshado1:
`<React.Fragment>` as a wrapper is working for me (React 16.2.0 & Enzyme 3.3.0):

```jsx
const wrapper = shallow(<MyComponent />);

console.log(
    wrapper.debug(),
    wrapper.children().debug(),
);
```
log output:
```jsx
<Symbol(react.fragment)>
  <PaymentIcon amount="$10.00" displayName="John Jingleheimer" type="card" size="huge" />
  <Header as="h4" content={[undefined]} />
</Symbol(react.fragment)>

<PaymentIcon amount="$10.00" displayName="John Jingleheimer" type="card" size="huge" />
<Header as="h4" content={[undefined]} />
```

Which is exactly what I expect.

brandonduff:
@jshado1 In your example, `wrapper.text()` and `wrapper.html()` throw this error:

```
Error: Trying to get host node of an array
```
This issue is also relevant https://github.com/airbnb/enzyme/issues/1178

LandonSchropp:
@lelandrichardson This has turned into a blocker for my team, and I'd love to help resolve it. 

* Would you be open to accepting a PR to fix this?
* Is this an issue that somebody who's not familiar with the codebase can solve?
* Could you point me in the right direction for what needs to be done? 
* Is there any documentation covering the adapter interface? 
* Is the current API for the adapters (particularly `nodeToHostNode`) compatible with some of the new paradigms in React? 
* Does the core library need to be updated to support this?

ljharb:
[...]

PR #1498, opened 2018-01-27
Fix ShallowWrapper for array-rendering components

vsiao:
This PR is in support of https://github.com/airbnb/enzyme/issues/1149.
There are two main assumptions that I've fixed:

- in `ReactSixteenAdapter`, flatten and map over the render output if possible, instead of assuming that it is always an element
- in `ShallowWrapper`, reuse existing logic to set `NODE` and `NODES` appropriately when the root node is an array

### DIFF
--- a/packages/enzyme/src/ShallowWrapper.js
+++ b/packages/enzyme/src/ShallowWrapper.js
@@ -99,6 +99,17 @@ function getRootNode(node) {
   return node.rendered;
 }
 
+function privateSetNodes(wrapper, nodes) {
+  if (!Array.isArray(nodes)) {
+    privateSet(wrapper, NODE, nodes);
+    privateSet(wrapper, NODES, [nodes]);
+  } else {
+    privateSet(wrapper, NODE, nodes[0]);
+    privateSet(wrapper, NODES, nodes);
+  }
+  privateSet(wrapper, 'length', wrapper[NODES].length);
+}
+
 /**
  * @class ShallowWrapper
  */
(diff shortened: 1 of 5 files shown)


packages/enzyme/src/ShallowWrapper.js
```diff
@@ -122,25 +122,26 @@ class ShallowWrapper {
           instance.componentDidMount();
         });
       }
-      privateSet(this, NODE, getRootNode(this[RENDERER].getNode()));
-      privateSet(this, NODES, [this[NODE]]);
-      this.length = 1;
+      this.setNodesInternal(getRootNode(this[RENDERER].getNode()));
```
ljharb: this makes `setNodesInternal` a public function; it's important to minimize the publicly observable API. Can this be moved to a module-scoped function?

vsiao:
I applied the same module-scoped `NODES`-setter helper to `ReactWrapper` and added a basic compatibility test to `ReactWrapper-spec`, but I don't know if that counts as "taking care" of `mount`. The test I wrote would've passed even before my changes.

There are deeper problems associated with achieving full support for array-rendering that I'd rather not get into right now—primarily in that `ReactDOM.findDOMNode(instance)` will return the first DOM node rendered by `instance` and not a full array. So I imagine one of two things has to happen in order to move forward: a) React exposes an API to get a list of DOM nodes rendered by an instance, or b) Enzyme stops using `findDOMNode`. Until then, any `ReactWrapper` method that uses `adapter.nodeToHostNode()` will be broken when applied to a component that returns an array.

ljharb:
Hmm. Is there any way inside the React 16 adapter, in findDOMNode, to determine when this bug is happening?

Separately, could the adapter keep track of which components returned an array or fragment, and then inside findDOMNode, know to return an array of nodes?

vsiao:
Sorry, I can't commit time to exploring that. I don't think it's a great idea to do it in this pull request anyway, given that this is nicely self-contained and a strict improvement over what was there before.

vsiao:
@ljharb is there anything else I can do to get this merged?

louisscruz:
@ljharb Is there a plan to publish this soon? I currently have `enzyme-adapter-react-16` version 1.1.1 installed, but this change doesn't come along with it. I'm still seeing this in my `node_modules/enzyme-adapter-react-16/src/ReactSixteenAdapter.js`:

```
 rendered: elementToTree(output),
```

ljharb:
I'm hoping to get a number of changes in the next release related to fragments/arrays; I'll take a look at publishing soon.

ljharb:

PR #1499, opened 2018-01-27
[Shallow]Call update() automatically through get querys

koba04:
In Enzyme v3, I saw many issues about changes that we have to call `update()` after `instance.setState()`.

Actually, `wrapper.update()` doesn't rerender even though the name implies to update something.
`wrapper.update()` only does to get a latest ReactElementTree from ReactShallowRenderer.
I think it would be nice if developers don't care about calling `wrapper.update()`.

* https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#for-mount-updates-are-sometimes-required-when-they-werent-before

In this case, enzyme can't handle all cases that need to call `wrapper.update()`.

```js
expect(wrapper.find('p').text()).to.equal('0');
wrapper.find('.button').prop('onClick'); // Update the value to '1'
// enzyme should call wrapper.update() but enzyme can't handle that
// because the above calls ReactComponent instance's setState directly 
expect(wrapper.find('p').text()).to.equal('0');
wrapper.update(); // Get the latest nodes from ShallowRenderer
expect(wrapper.find('p').text()).to.equal('1');
```

In order to resolve this, this PR adopts a pull-based approach, which calls `wrapper.update()` in getter for internal nodes.
So we can get the latest node through the enzyme APIs (`find`, `text`, ...).

```js
expect(wrapper.find('p').text()).to.equal('0');
wrapper.find('.button').prop('onClick'); // Update the value to '1'
// wrapper has nodes that are before updating.
expect(wrapper.find('p')/* Get the latest nodes from ShallowRenderer */.text()).to.equal('1');
```

This PR is for ShallowWrapper, but I could apply this to ReactWrapper as well.

But there is a limitation.
This PR doesn't cover that a wrapper isn't RootShallowWrapper.

```js
let child = wrapper.find('p');
expect(child.text()).to.equal('0');
child.prop('onClick'); // Update the value to '1'
expect(child).to.equal('0');
// we have to get the child from the wrapper again
let child = wrapper.find('p');
expect(child).to.equal('1');
```

Covering the case might be difficult because enzyme has to get `child` from the updated RootShallowWrapper with the same condition.
But enzyme has to care about a case that can't get `child` due to the update.
I think it seems to be good that wrappers that are not RootShallowWrapper treat as immutable.

BTW. ShallowWrapper has the condition(`this[ROOT] === this`) in many methods, I think it becomes clearer if ShallowWrapper can be separated `RootShallowWrapper` and `ChildShallowWrapper`.

### DIFF
--- a/packages/enzyme/src/ShallowWrapper.js
+++ b/packages/enzyme/src/ShallowWrapper.js
@@ -97,6 +97,10 @@ function getRootNode(node) {
   return node.rendered;
 }
 
+function getRootNodeInternal(wrapper) {
+  return wrapper[ROOT][NODE];
+}
+
 function privateSetNodes(wrapper, nodes) {
   if (!Array.isArray(nodes)) {
     privateSet(wrapper, NODE, nodes);
@@ -150,10 +154,12 @@ class ShallowWrapper {
     return this[ROOT];
   }
 
-
   getNodeInternal() {
     if (this.length !== 1) {
       throw new Error('ShallowWrapper::getNode() can only be called when wrapping one node');
+    }
+    if (this[ROOT] === this) {
+      this.update();
     }
     return this[NODE];
   }
@@ -167,7 +173,7 @@ class ShallowWrapper {
     if (this.length !== 1) {
       throw new Error('ShallowWrapper::getElement() can only be called when wrapping one node');
     }
-    return getAdapter(this[OPTIONS]).nodeToElement(this[NODE]);
+    return getAdapter(this[OPTIONS]).nodeToElement(this.getNodeInternal());
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/enzyme/src/ShallowWrapper.js
```diff
@@ -152,11 +152,17 @@ class ShallowWrapper {
     return this[ROOT];
   }
 
+  getRootNodeInternal() {
```
ljharb: we tried pretty hard in v3 to minimize the public API. Can this be a normal function instead of a class method?

koba04: I've fixed it.

koba04:
Thanks!

PR #1500, opened 2018-01-28
add options arg to API docs for debug()

finnigantime:


finnigantime:
@ljharb Looks to me like mount API docs already show options:
screen shot 2018-01-29 at 2 13 14 pm

ljharb:
@finnigantime that's on `mount` itself; i mean on ReactWrapper's `.debug` method.

ljharb:
I took care of the rebase and the mount docs for you.

finnigantime:
Ah, I was confused what you meant but now all makes sense. Thanks!

PR #1506, opened 2018-02-02
Updating docs for shallow disableLifecycleMethods

joeldenning:
As of enzyme v3, disableLifecycleMethods defaults to true instead of false. Even though this is explained in a variety of places, the actual documentation for `shallow()` still says that it defaults to false, which is no longer true.

### DIFF
--- a/docs/api/shallow.md
+++ b/docs/api/shallow.md
@@ -48,7 +48,7 @@ describe('<MyComponent />', () => {
   - `options.context`: (`Object` [optional]): Context to be passed into the component
   - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, `componentDidMount`
 is not called on the component, and `componentDidUpdate` is not called after
-[`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md). Default to `false`.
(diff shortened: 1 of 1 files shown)


docs/api/shallow.md
```diff
@@ -48,7 +48,7 @@ describe('<MyComponent />', () => {
   - `options.context`: (`Object` [optional]): Context to be passed into the component
   - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, `componentDidMount`
 is not called on the component, and `componentDidUpdate` is not called after
-[`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md). Default to `false`.
+[`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md). Defaults to `true`.
```
joeldenning: Am happy to change this to explain that it defaults to `true` for v3 but `false` for v2, if that is more helpful for the readers. I thought I would just start with what is true for `enzyme@latest`, though.

joeldenning:
Oops, I got confused on this. The docs are actually correct, as is -- the double negative of "disable defaults to false" confused me, but it is correct. Closing.

PR #1513, opened 2018-02-05
Add support for react context element types, fixes #1509

### LINKED ISSUE #1509: New React context API adds new tag types
The new API for context, publicly released just a few hours ago in the React 16.3 alpha, adds new tag types for the `Provider` and `Consumer` components that are returned from the `createContext` function.

When updating my module to use the new API, I get an unknown node error from the `ReactSixteenAdapter.js`:

```
Enzyme Internal Error: unknown node with tag 13
```

I went in and took a stab at updating the adapter, and found that the Consumer prop has another new tag type, 12.

I added the following lines:

```diff
+ var ContextProvider = 13
+ var ContextConsumer = 12
```

and tried adding cases for both of those in the module's switch statement, but none of the existing case handling options worked to get my tests passing again.

#### API

- [ ] shallow
- [x] mount
- [ ] render

#### Version

| library         | version
| ---------------- | -------
| Enzyme           | 3.3.0
| React            |  16.3.0-alpha.0

#### Adapter

- [x] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13

jquense:
There are some outstanding support concerns here with shallow rendered per https://github.com/facebook/react/issues/12152 but i wasn't attempting to fix those here. This just adds support for traversal over said elements. I was not sure how to structure the peer deps in a way that would exercise this test as well...

cc @aweary


aweary:
This might be better as a new adapter like `ReactSixteenThreeAdapter`, since it's a new API and 16.2 and earlier don't have it. It will possibly depend on how context is supported in the new shallow renderer as well.

jquense:
my thought was that it'd reduce some churn for folks since it's not a breaking change but that might go out the window for handling context only using hte new API. In which case it may be easier to make a new one...Idk

aweary:
Yeah, I think reducing churn is good concern. There is precedence for creating a new adapter for a minor release ([`ReactFifteenFourAdapter`](https://github.com/airbnb/enzyme/blob/master/packages/enzyme-adapter-react-15.4/src/ReactFifteenFourAdapter.js)), but to be fair that split was around before we required people to explicitly configure an adapter. I just think if we're going to traversing the new context elements, we should also support actually using the new context API, so it might be best to hold off until we figure out what that'll look like.

jquense:
> we should also support actually using the new context API, so it might be best to hold off until we figure out what that'll look like.

definately agree on supporting hte new context, but most of that change is gonna be internal ReactWrapper stuff. It'd be nice to at least not have tests break for folks who start using the context api in their own code. I mean if we can do it at the same time great, but i'm not confident in _my_ ability to tackle it so idk

aweary:
I guess my concern is that if context doesn't actually work at all, maybe it should break? I don't want anyone to assume because it renders the providers/consumers without failing that that means the adapter supports them. I also think it should fail if you try to use it with React <=16.2, which is why I'm leaning towards a separate adapter. We didn't do that with fragments though AFAIK, so maybe it's not a big deal.

I haven't been paying a lot of attention to what's been happening with Enzyme, @ljharb can you chime in here?

jquense:
I think there are two concerns here. The first, being enzyme should work in react trees that use `createContext`, and the second is that internally Enzyme `mount()` and `shallow` should handle their `context` options with the new APi since it'll break when the old on is removed.

> I guess my concern is that if context doesn't actually work at all

So this isn't the case, if a user renders a Provider/Consumer it'll work, and internally if passed `context` it'll also work but use the old api. 

Ultimately tho I think Enzyme even needs a `context` option (depending on shallow renderer) since its straightforward to do in a test manually now, e.g. `<Provider><Root></Provider>` that and there is probably no way to make the enzyme api work without a breaking change since you can't arbitrarily pass context anymore, you _must_ have the paired Consumer to access it.

aweary:
>I think there are two concerns here. The first, being enzyme should work in react trees that use createContext, and the second is that internally Enzyme mount() and shallow should handle their context options with the new API since it'll break when the old on is removed.

Right, I'm suggesting that if one works the other should probably work as well. Mostly because people will probably be rendering a lot of consumers without providers.

> So this isn't the case, if a user renders a Provider/Consumer it'll work, and internally if passed context it'll also work but use the old api.

So my worry is related to the second concern you mentioned, specifically users thinking that Enzyme's API for providing context to a tree will work rendering standalone consumers. I'm assuming this would not work:

```jsx
const Button = ({ label }) => (
  <ThemeConsumer>
    {theme => <button style={theme.button}>{label}</button>}
  </ThemeConsumer>
)
```

and they try to pass it context via `mount`'s option like:

```jsx
const theme = {
  button: { fontSize: 15 }
}
const root = mount(<Button label="foo">, { context:  { theme }  })
```

> Ultimately tho I think Enzyme even needs a context option (depending on shallow renderer) since its straightforward to do in a test manually now, e.g. <Provider><Root></Provider> that and there is probably no way to make the enzyme api work without a breaking change since you can't arbitrarily pass context anymore, you must have the paired Consumer to access it.

The one thing that sucks about requiring that consumers always exist inside a provider is that with `shallow` you'll always have to use `dive` to get to the actual rendered component(s). It's the same problem we have now with shallow rendering trees with `react-redux` providers and other similar APIs, but at least with the existing context API you can side-step the issue and pass the store or whatever into the `context` option.

jquense:
> The one thing that sucks about requiring that consumers always exist inside a provider

is this the case? That will break so much of my context code...

aweary:
> is this the case? That will break so much of my context code...

Not in the general case. It will just use the default value provided to `createContext`. I mean in the context (no pun intended 😄) of testing components in isolation; you'll probably want to test that `Button` with something other than the default value in some cases, and requiring a `Provider` to do that might not be the best solution.

That's why `mount` and `shallow` have the `context` option, so you don't have to wrap your components in actual context providers.

jquense:
Phew! :P 

> That's why mount and shallow have the context option, so you don't have to wrap your components in actual context providers.

definitely, I was thinking though that the new API can't work like this, the user will at the least have to provide Enzyme with the provider since it can't take a plain object and send to the correct consumers. AT the very least the API will change to passing a Provider to `context:` in which case i might be simpler to just have the user use the provider.

packages/enzyme-test-suite/package.json
```diff
@@ -49,4 +49,4 @@
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-react": "^7.5.1"
   }
-}
+}
```
ljharb: please ensure all files always have a trailing newline

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -71,6 +73,10 @@ function toTree(vnode) {
       return toTree(node.child);
     case HostPortal: // 4
       return toTree(node.child);
+    case ContextProvider:
+      return toTree(node.child);
+    case ContextConsumer:
+      return toTree(node.child);
```
aweary: Looks like the test renderer treats them as fragments https://github.com/facebook/react/pull/12237/  Should we do the same? It's really confusing how both files use the same APIs (`toTree`, `childrenToTree`) but don't use them consistently.

packages/enzyme-adapter-react-16/package.json
```diff
@@ -40,12 +40,12 @@
     "object.values": "^1.0.4",
     "prop-types": "^15.6.0",
     "react-reconciler": "^0.7.0",
-    "react-test-renderer": "^16.0.0-0"
+    "react-test-renderer": "^16.0.0-0  || ^16.3.0-0"
```
aweary: Why do we need to do this?

stepankuzmin:
Hi there! Any news on that?

packages/enzyme-adapter-react-16/package.json
jquense: `^16.0.0-0` only handles prerelease ranges the exact version `16.0.0` not any prerelease of any version of `16`

jquense:
I updated, let me know if there is anything i can do here :) I'd love to see this a release, we're already playing with react 16.3.0 and the context api and it's killing our tests :P

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -111,6 +112,11 @@ function toTree(vnode) {
     }
     case HostText: // 6
       return node.memoizedProps;
+    case Fragment: // 10
+    case Mode: // 11
+    case ContextProvider: // 13
```
ljharb: O.o ContextProvider and ContextConsumer are new node types, they're not just components?

jquense:
oops sorry shallow tests are wrong, one sec

packages/enzyme-test-suite/package.json
```diff
@@ -49,4 +49,4 @@
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-react": "^7.6.1"
   }
-}
+}
```
ljharb: please revert this; all files should have a trailing newline

jquense: sorry i'm not sure why my editor keeps doing this, it otherwise is set to add trailing nls

diegohaz:
Hey, guys. Is there any reason to not merge this?

aweary:
Nope

jquense:
Thanks y'all

lucasconstantino:
Any news on when this will get released?

atsu85:
Hi, is smth blocking the release?

lucasconstantino:
This is serious! I'm updating a bunch of libs and my tests are going nuts :scream: Could we at least have an RC with this?

ljharb:
There’s a few more issues i want to iron out before cutting a release; hopefully, I’ll have time to do it soon.

benf1977:
When I run the test suite I'm seeing a warning:
```
Warning: Failed prop type: Invalid prop `Component` supplied to `WrapperComponent`.
    in WrapperComponent
        ✓ should find elements through Context elements
```
If I copy the test into my environment I'm actually seeing an exception thrown. Is this an issue to be addressed separately? Can I prevent it from failing in my environment? 
```
    Expected the function not to throw an error.
    Instead, it threw:
      Failed prop type: Invalid prop `Component` supplied to `WrapperComponent`.
        in WrapperComponent
```
Thanks!

baldwmic:
@ljharb any update on when a release for this will happen? we've migrated our codebase to the new Context API in React 16.3 and this bug is causing our tests to fail so just waiting on a release. Thanks 😄

Dremora:
@baldwmic to overcome this, we are using a simplified version of [jamiebuilds/create-react-context](https://github.com/jamiebuilds/create-react-context) polyfill. Copy [this file](https://gist.github.com/Dremora/e02fbbfb1eb5009abac92ccf71c4eb5f) into your project and and a reference to it in package.json under `jest.setupFiles`. Note that you'll need to provide a 3rd argument to all calls to `createContext`. If you're using flow, you'll also have to modify the type signature of `createContext`, type definitions are included in the gist.

baldwmic:
@Dremora thanks so much for the tip! My team doesn't want to change our configuration and dependencies for a short amount of time only to change it back once the official release occurs, but it's definitely nice to know a workaround exists.

kkwiatkowski:
@ljharb do you have any update on a release cut? Another poor soul is struggling hard to work around 😢

atsu85:
@ljharb
> There’s a few more issues i want to iron out before cutting a release; hopefully

Why can't you just release current master (or merge this onto previous release if master contains bugs)?

gziolo:
Same here, we had to disable tests for parts of the code that uses new Context API :(

sergeysolovev:
Guys, release this pls

theKashey:
😢 😿 😭

KrustyC:
Any update on when this is gonna be released?

KrustyC:
While you are waiting for the release you can install the `enzyme-adapter-react-16` package that we forked [here](https://github.com/School-Explained/enzyme-adapter-react-16). 
We tried to install `enzyme` directly from github but it didn't work, so we just took the react 16 adapter package with @jquense changes, made a couple of small changes (babel was out of date and not able to build). Further there was an issue importing a function from `enzyme-adapter-utlis`so we just add that directly into the adapter code. 

The build is available on github as well, so you can just clone it from our repo.

KevinHewson:
Do I have to beg?  I'm not above begging.  PLEASE release this soon.  We have a whole suite of tests that we have disabled now because of this issue.

lucasconstantino:
This is truly incredible. The single most important tool for testing React, does not keep up to date with the latest React version - for a matter of **months** not days.

ramesaliyev:
Well... Since no one cares i've ended up publishing latest code changes for myself. I'll use my own packages until official release. I've also made them public in case anyone else wants to use.

Install
```bash
npm i @pisano/enzyme @pisano/enzyme-adapter-react-16
```

Use
```javascript
import Enzyme, { shallow, mount } from '@pisano/enzyme';

import React16Adapter from '@pisano/enzyme-adapter-react-16';

Enzyme.configure({
  adapter: new React16Adapter()
});
```

**Note:** I didnt change any code. Only pulled and published master. The repo is [here](https://github.com/Pisano/enzyme).

davidfloegel:
@ramesaliyev see above, @KrustyC already did

ramesaliyev:
@davidfloegel yeah i saw that but i dont like using git repos as npm modules, maybe this is okay today but in the past it causes trouble because of the npm bugs.

baldwmic:
@ljharb @aweary @jquense can you please release a new version of Enzyme with this fix? It's becoming difficult for many users of enzyme (as you can see above) to upgrade to the new version of React since we are waiting on this fix for our tests to pass.

Or at the very least, it would be nice to know a timeline on when a new release can be expected. Of course, thank you for all your work giving to the open source community 😄

jquense:
i cannot release anything i'm not a maintainer

aweary:
I also don't have publish rights on npm. Someone from Airbnb will need to publish a release.

jaredpalmer:
This is blocking Formik from going to 16.3. Would love to see a release tomorrow.

KevinGrandon:
This was taking a long time to land, so for the meantime I've published a package which patches enzyme in the postinstall hook, using the patch from @lucasconstantino above. 

Repo: https://github.com/KevinGrandon/enzyme-react-16-context-patch
NPM package: https://www.npmjs.com/package/enzyme-context-patch

We wanted a small temporary and encapsulated package that we could install, simply uninstall once this code lands. Hopefully it won't take too much longer 😅

KrustyC:
Any news guys?

JonathanPicques:
Last publish 5 months ago, would love to see this fix released

baldwmic:
Thank you @lucasconstantino and @KevinGrandon for sharing the alternative approach of patching the fix. I am having success with this approach and it means we don't have to switch to a fork! 😄 

Unfortunately, I couldn't use your package @KevinGrandon because we use yarn workspaces, and the postinstall script doesn't run for us, but I've figured out how to do it. Thanks for sharing.

koba04:
Do we have to wait for the next release until all issues related React v16 have been resolved?

* https://gitter.im/airbnb/enzyme?at=5aec95d540f24c4304524d83
* https://github.com/airbnb/enzyme/issues/1553

afenton90:
Whats the policy on releasing updates to enzyme? 
It seems strange that `master` is so widely out of date with what is published on `npm`.

baldwmic:
@ljharb Thank you so much for taking the time to explain! 😄 

Strategically, that makes sense a lot of sense why you have delayed in publishing. It's too bad others can't take some of the work of enzyme off your plate to allow progress to continue for the benefit of the community. I am sure there's a lot of context I'm missing, since I don't work at Airbnb so don't worry about my pondering...

I want to be clear that I am extremely grateful for the work you do. I think being an open source maintainer is an often thankless job, and I hope I didn't contribute in some way to your personal stress. If anything, the number of comments requesting a release on this pull request show how impactful Enzyme has been on the React testing community! Great work and thanks again! ❤️

nemoDreamer:
Cool, thanks for the update, @ljharb ! ❤️

joepuzzo:
@ljharb when you get this release in can you also add support for `React.forwardRef` ? I used the patch described above but still have the issue referenced here https://github.com/airbnb/enzyme/issues/1604

KevinGrandon:
Any chance for a pre-release? I think the expectation there is that pre-releases are unstable and folks should be willing to see bugs. Hopefully that should address the concern of new bug reports? Could even update the bug report template to reflect this as well, hopefully resulting in less noise.

dantman:
I'd happily try out an alpha pre-release with the React 16.3 handling implemented so far. Like the Babel 7 beta.

I really think it would be a good idea to officially release what has been done so far as an alpha because unlike other packages, the monorepo pattern used in enzyme means that it's not possible to use a git+ dependency to get the latest in-development version.

rodush:
now seems the same problem with the React-16.4

jcready:
@baldwmic you said:

> Unfortunately, I couldn't use your package KevinGrandon because we use yarn workspaces, and the postinstall script doesn't run for us, **but I've figured out how to do it.** Thanks for sharing.

Would you be kind enough to explain how you got things working with yarn workspaces?

baldwmic:
@jcready I tried looking into why post install script wasn't working, but it was becoming a time sink

Instead, since my team used a Makefile, I just wrote a command to patch the file from KevinGrandon's repo. So I didn't really solve it, but got around it.

You might be able to get around it by writing your own npm script to patch the file if you don't use Make. Maybe do your own postinstall. 

I don't have access to the private repo anymore but hope that helps!

stereobooster:
There is an issue with current implementation of the context in enzyme

```
  console.error node_modules/fbjs/lib/warning.js:33
    Warning: Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.
```

https://github.com/stereobooster/react-context-issue

and 

https://github.com/stereobooster/react-context-issue/tree/enzyme-react-adapter-future

ljharb:
@stereobooster would you mind filing a new issue about that?

draperd:
@KevinGrandon Thanks for building that package. Could you confirm if your package actually should support the new context features or if it just allows components using it to be mounted? (It definitely lets my components mount where they wouldn't before, but context doesn't seem to be passed 😕 )

KevinGrandon:
> @KevinGrandon Thanks for building that package. Could you confirm if your package actually should support the new context features or if it just allows components using it to be mounted? (It definitely lets my components mount where they wouldn't before, but context doesn't seem to be passed 😕 )

Weird, I would think that it should work, but I didn't do the initial implementation. If you have a PR/patch I'm happy to try to include it in the package fork. Unfortunately not much time to investigate this further at the moment.

draperd:
@ljharb The test case I can be found in this test for a project I've been working on: https://github.com/draperd/react-forms-processor/blob/master/packages/core/src/renderer.test.js#L61 (the commented line should pass but is failing with undefined rather than an empty object which should be provided from the context)

I'm not entirely sure how I'd create a PR for a test without pulling in my entire library?

ljharb:
@draperd thanks - the hope would be that you could construct a minimal test case that doesn't require your whole library :-) https://github.com/draperd/react-forms-processor/blob/master/packages/core/src/components/Button.js is relatively small, so perhaps you could condense that along with a test into a single post? If you don't have time, or aren't willing, to make a PR, then filing a new issue with that minimal test case would also be super helpful <3

ljharb:

donaldpipowitch:
Thank you so much! Could you add this to the changelog? I couldn't found this issue. Thanks

donaldpipowitch:
Ah, I see. Thank you!

MartinDawson:
https://github.com/airbnb/enzyme/issues/1647 is still broken it seems.

ljharb:
@MartinDawson yes, that's why it's still open.

belfz:
How am I supposed to use this? I upgraded both `enzyme` and `enzyme-adapter-react-16` to latest versions, I try to `shallow` with `context` in the options and it still doesn't work for me. Couldn't find any instructions either.

jquense:
@belfz  this change adds support for rendering trees with the new Context elements, it doesn't change the API for setting context on mount, that is still old style contex. If you want to mount a component with some context, render it with the ContextProvider and a value. That said the new Context is not supported for shallow rendering, the React team has not decided to support that use case, it will work for `mount` tho

ljharb:
enzyme, however, will support that use case - it just doesn’t yet.

belfz:
Thank you!

James-Ga11agher:
@jquense - Enzyme's old style context isn't working for me with React 16's new Context... This is the code I have...

package.json
```
    "react": "^16.5.0",
    "enzyme": "^3.6.0",
    "enzyme-adapter-react-16": "^1.5.0",
```

Error.js
```
export default class Error extends Component {
  render() {
    return (
      <AppConsumer>
        {context => {
          return (
            <div>
              <h1>Error</h1>
              <h4>{context.error}</h4>
            </div>
          );
        }}
      </AppConsumer>
    );
  }
}
```

Error.test.js
```
const mountComponent = () => {
  const context = { error: 'foo' };
  element = mount(<Error />, { context });
  console.log(element.html());
  console.log(element.text());
};

describe('<Error />', () => {
  describe("AppContext's Error is displayed", () => {
    it('Given Error component is rendered', () => {
      mountComponent();
      element.setContext({ error: 'bar' });
      console.log(element.html());
      console.log(element.text());
    });
    it('Expect error display to be error passed in from context', () => {
      expect(element.text()).to.contain('bar');
    });
  });
});
```

The context never gets set for me. Is the above correct? 

@belfz - did the old style context work for you?

jquense:
@James-Ga11agher that won't work you can use old style context with new context. they aren't compatible apis.

James-Ga11agher:
@jquense - sorry, not following, "you can use old style context with new context. they aren't compatible apis."

Does Enzyme support React 16's new Context? Is there an example of this? On my search for this, this was the most promising posting.

jquense:
The Enzyme API's for setting context are only for **old style context**, they don't work for the new context api. if you want to set context for a tree using the new context API, render the Provider as part of the test. e.g. 

```
mount(<AppProvider><Error/></AppProvider>)
```

James-Ga11agher:
Ah I get you now, thank you so much 👍 Really appreciate the help 👍  In case anyone else has the same issue as me, working code below
```
const mountComponent = errorMessage => {
  element = mount(
    <AppProvider value={{ error: errorMessage }}>
      <Error />
    </AppProvider>,
  );
};

describe('<Error />', () => {
  it('renders without crashing', () => {
    mountComponent();
  });

  describe("AppContext's Error is displayed", () => {
    it('Given Error component is rendered', () => {
      mountComponent(errorMessage);
    });
    it('Expect error display to be error passed in from context', () => {
      expect(element.html()).toContain(errorMessage);
    });
  });
});
```

ljharb:
@James-Ga11agher no, enzyme does not yet support new-style context, see #1553. This PR is only one piece of it.

mbrowne:
Following up on the above comments about the existing `context` option in `mount()` and `shallow()`, I'm wondering if it will ever be possible to achieve a similar result using the new context API. I am currently in the process of setting up a theme for the `styled-components` in the app I'm working on. This is requiring updates to almost every test case that uses `mount()`, because either the component directly uses styled-components that need theme properties, or it uses the theme indirectly via a separate npm package with our shared UI components (also styled-components). So for every test where we could previously do something like this:

    const component = mount(<CreatorSummary creator={mockCreator} />)
    ...
    expect(component.state().open).toBe(true)

We would now have to do this:

    const component = mount(<ThemeProvider theme={theme}><CreatorSummary creator={mockCreator} /></ThemeProvider>)
    ...
    expect(component.find(CreatorSummary).instance().state.open).toBe(true)

...and this is just the tip of the iceburg -- the tests are already written so there is a lot of such code that would need to be updated.

With the current API, I was able to set up helper functions, e.g. `mountWithTheme()`, that use enzyme's `context` and `childContextTypes` options, as explained [here])(https://github.com/styled-components/jest-styled-components/issues/61#issuecomment-421043968) - which is obviously a much more convenient solution, and IMO a more elegant one too.

But soon we will be upgrading to styled-components 4 (currently in beta), which uses the new context API. So far the best solution I've been able to come up with for the new version is this hack that overrides one of styled-component's internal methods, which obviously isn't ideal:

```js
...
import * as theme from './theme'

function setupMockTheme() {
    const result = enzyme.mount(React.createElement(styled.div``))
    const BaseStyledComponent = result.childAt(0).type()

    BaseStyledComponent.prototype.renderOuter = function renderOuter(styleSheet) {
        this.styleSheet = styleSheet
        return this.renderInner(theme)
    }
}
```

Would it somehow be possible to add an option to enzyme so context properties could still be passed as an option to `mount()` and `shallow()` in a way that would work with the new context API? I realize the new context API works quite differently...any thoughts or suggestions are welcome. As an alternative, it might be helpful if it were possible to call methods like `state()` on non-root nodes (so tests wrapped with Provider components could still be written in a similar way), but that seems to be an inherent and deliberate limitation of enzyme...

ljharb:
See #1802, for setting state.

It may indeed be possible; i can’t know until i build support for new context.

My advice would be not to update to something that used new context until you’re able to test it properly :-)

mbrowne:
@ljharb Oh, I thought the new context support was mostly done already (I was looking at #1553, where only the issue about `dive()` not working is still open). Good to know there will be more robust support for it in the future.

Regarding `state()` and `setState()` and #1802, this is great. Are there any plans to allow other methods like `props()`, `prop()`, and `setProps()` to work on child nodes as well?

ljharb:
`props` and `prop` should already work on anything; `setProps` is a bit different since the child's parent controls what props it gets - i'm not sure it makes sense to allow that. Either way, let's discuss this in a new issue, since it's all off topic for this PR.

ljharb:
@c10b10 yes. if you're still seeing an issue, please file one.

Issue #1511, opened 2018-02-04
ReactWrapper.key returns undefined for empty string keys


tryggvigy:
### Current behavior
```js
  it('asd', () => {
    const foo = [{ a: '' }, { a: '2' }, { a: '3' }];
    const wrapper = mount(<div>{foo.map(i => <span key={i.a} />)}</div>);
    expect(wrapper.children().first().key()).toBe('');
  });
```
results in:
![image](https://user-images.githubusercontent.com/2373958/35782828-9f9c2778-09fe-11e8-94ac-8865e979b173.png)

### Expected behavior
I expected the test to pass.

#### API

- [x] shallow
- [x] mount
- [x] render

#### Version

| library         | version
| ---------------- | -------
| Enzyme           | 3.1.0
| React            |  16.2.0

#### Adapter

- [x] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-15.4
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] enzyme-adapter-react-helper
- [ ] others ( )

ljharb:
Does the same thing happen with `shallow`?

What about with other versions of React?

tryggvigy:
@ljharb This happens with `shallow` as well yes. I'm not sure if this happens with other versions of React. This issues popped up in our repo when I did a minor version bump of our prod deps:
<details><summary>Diff</summary><p>

```diff
diff --git a/ui/package.json b/ui/package.json
index a47590a10e..9ae4e2b1e9 100644
--- a/ui/package.json
+++ b/ui/package.json
@@ -102,48 +102,48 @@
     "classnames": "^2.1.3",
     "debug": "2.6.9",
     "dom-delegate": "~2.0.3",
-    "domify": "~1.3.1",
+    "domify": "~1.4.0",
     "domready": "^1.0.8",
-    "enzyme-adapter-react-16": "^1.0.1",
+    "enzyme-adapter-react-16": "^1.1.1",
     "escape-html": "~1.0.1",
     "events": "^1.0.2",
     "extend": "^2.0.1",
-    "findandreplacedomtext": "0.4.4",
+    "findandreplacedomtext": "0.4.6",
     "glob": "^5.0.10",
-    "handlebars": "^4.0.6",
-    "hbsfy": "^2.7.0",
+    "handlebars": "^4.0.11",
+    "hbsfy": "^2.8.1",
[...]

PR #1524, opened 2018-02-11
Allow empty strings as key props

tryggvigy:
## This PR will 
- Fix https://github.com/airbnb/enzyme/issues/1511.

## Why?
React allows empty string key props. We should too:

```js
const wrapper = shallow(
  <div>
    {[''].map(s => <span key={s}>{s}</span>)}
  </div>
);
wrapper.at(0).key() // === `undefined` on master,  === `''` on this PR
```


packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -153,6 +153,10 @@ export function flatten(arrs) {
   return flatArrs;
 }
 
+export function ensureKeyOrUndefined(key) {
+  return key || (key === '' ? '' : undefined);
```
tryggvigy: This check could be more restrictive.
```js
return typeof key === 'string' ? key : undefined;
```
Are there situations where a key can be anything other than a string or a falsey value?

bdwain: react stringifies all valid keys before storing them, and a key is valid if it is not undefined. So i think this workes either way. I think it's probably best to leave it as you have now

ljharb: maybe `typeof key === 'string' ? key : undefined`?

PR #1529, opened 2018-02-19
[Docs] Update list of enzyme-adapter-react-16 peer dependencies

edmorley:
Since as of #1252, `enzyme-adapter-react-16` has `react-test-renderer` as a dependency rather than a peer dependency, so users of enzyme no longer need to manually install it.


ljharb:
Thanks!

PR #1536, opened 2018-02-23
return null for missing keys

### LINKED ISSUE #1488: .key() isn't consistent with unwrapped element .key
### Current behavior

If a `key` is not specified on an element, Enzyme returns `undefined` from `.key()`

### Expected behavior

I would expect `.key()` to return `null` the same as React does for `.key` on an (unwrapped) element.

Contrived example: https://codesandbox.io/s/xpj8pvpj2o

My use case is that I'm writing a test for a library that ensures that an element generated from a shorthand prop (on a parent) matches a specific factory-generated counterpart. Part of this process involves the optional generation of a key, so I need to check that the keys match (which isn't covered by `.contains()` or similar).

As you can see from the example, the assertion works fine when a key is provided/expected, but breaks when it isn't. There are easy workarounds:

- Use `.get()` to retrieve the unwrapped element
- Compare with loose equality
- Wrap the other element

...but I thought I'd bring it up anyway.

#### API

- [x] shallow
- [x] mount
- [ ] render

#### Version

| library         | version
| ---------------- | -------
| Enzyme           | 3.3.0
| React            | 16.0.0

#### Adapter

- [x] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] others ( )

bdwain:
Fixes #1488 by checking if a key is undefined, and if so, returning null.

Without the code changes, the new tests will fail with errors like this

`AssertionError: expected undefined to equal null`

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -3153,6 +3153,15 @@ describeWithDOM('mount', () => {
       expect(wrapper.at(1).key()).to.equal('bar');
       expect(wrapper.at(2).key()).to.equal('');
     });
+
+    it('should return null when no key is specified', () => {
+      const wrapper = mount((
+        <ul>
+          <li>foo</li>
+        </ul>
+      )).find('li');
+      expect(wrapper.key()).to.equal(null);
+    });
   });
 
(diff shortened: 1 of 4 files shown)


bdwain:
fixed lint errors

bdwain:
thanks!

PR #1537, opened 2018-02-23
Add support for some pseudo selectors

jquense:
Adds some initial support for pseudo selectors, There are more that could be added but before I did that I started with the few I miss the most in order to get some feedback

### DIFF
--- a/packages/enzyme/src/selectors.js
+++ b/packages/enzyme/src/selectors.js
@@ -138,13 +138,36 @@ function matchAttributeSelector(node, token) {
   }
 }
 
+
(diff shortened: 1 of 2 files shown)


packages/enzyme/src/selectors.js
```diff
@@ -138,13 +138,38 @@ function matchAttributeSelector(node, token) {
   }
 }
 
+
+function matchPseudoSelector(node, token, root) {
+  // console.log(token)
```
ljharb: remove?

jquense: Oops

PR #1540, opened 2018-02-24
Make clearer the docs for .mount()

### LINKED ISSUE #1523: mount() oneliner in docs should be clearer
The [`ReactWrapper` docs](https://github.com/airbnb/enzyme/blob/8f0b7cd042de7c40b3d2a571204bc6444fe89dfa/docs/api/ReactWrapper/mount.md) say:

> `.mount() => ReactWrapper`
>
> A method that re-mounts the component.

A developer using Enzyme who reads this one-liner without reading the full docs will incorrectly believe that a call to `mount()` will re-mount *any* component. In fact, this method will only mount an *unmounted* component. It will do nothing if the component is already mounted.

A more precise summary would be something like:

> `.mount() => ReactWrapper`
>
> A method that (re-)mounts the component, if it is not currently mounted.

[`mount.md`](https://github.com/airbnb/enzyme/blob/261737722100326790c93efd7a5eee352a00d85d/docs/api/mount.md) and the [comments in `ReactWrapper.js`](https://github.com/airbnb/enzyme/blob/6f7f4fbae19050a912c70e6fbfe42c47b0851fe3/packages/enzyme/src/ReactWrapper.js) would need to be updated as well.

koba04:
Fix #1523

### DIFF
--- a/docs/api/ReactWrapper/mount.md
+++ b/docs/api/ReactWrapper/mount.md
@@ -1,6 +1,6 @@
 # `.mount() => Self`
 
-A method that re-mounts the component. This can be used to simulate a component going through
+A method that re-mounts the component, if it is not currently mounted. This can be used to simulate a component going through
 an unmount/mount lifecycle.
 
 #### Returns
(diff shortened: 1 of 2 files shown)


koba04:
Thanks!

Issue #1476, opened 2018-01-12
Debug masks arrays and object contents


blainekasten:
[Repl for example](https://repl.it/repls/MediumInexperiencedAfricanrockpython)

This was just raised as an issue in [enzyme-matchers](https://github.com/blainekasten/enzyme-matchers/issues/186).

I expected debug to print everything. Should it? If not, any advice on a different way to implement a strict matching element?

ljharb:
In https://github.com/airbnb/enzyme/commit/4f4c233096b944fef1c847a0ba38879d27556656, i improved the output for boxed primitives, but did not change the behavior for non-function objects.

I think that would be reasonable, and it'd be pretty trivial to change the `debug` output. @lelandrichardson, any idea why it was this way in the first place?

One possibility is that it could be done via an option to debug, so that the default output is minimal but the more verbose output is achievable.

ljharb:
Fixed in #1547.

PR #1547, opened 2018-03-01
Implement verbose debug output

blainekasten:
This implements `.debug({verbose: true})` which prevents boxing primitive values when stringified.

Fix for #1476

### DIFF
--- a/packages/enzyme-test-suite/test/Debug-spec.jsx
+++ b/packages/enzyme-test-suite/test/Debug-spec.jsx
@@ -629,6 +629,36 @@ describe('debug', () => {
 </div>`
       ));
     });
+
+    it('options.verbose causes arrays and objects to be verbosely printed', () => {
+      class Foo extends React.Component {
+        render() {
+          const nestedData = {
+            a: [1, 3, { true: true }],
+            b: false,
+            c: { d: 'f' },
+          };
+          nestedData.d = nestedData.a;
+          const arry = [1, 2, { f: nestedData.c }];
+          return (
+            <div data-json={nestedData} data-arry={arry}>Test Component</div>
+          );
+        }
+      }
+
+      const wrapper = shallow(<Foo />);
+      expect(wrapper.debug({ verbose: true })).to.equal((
+        `<div data-json={{ a: [ 1, 3, { true: true } ], b: false, c: { d: 'f' }, d: [ 1, 3, { true: true } ] }} data-arry={[ 1, 2, { f: { d: 'f' } } ]}>
+  Test Component
+</div>`
+      ));
+
+      expect(wrapper.debug({ verbose: false })).to.equal((
+        `<div data-json={{...}} data-arry={{...}}>
+  Test Component
+</div>`
+      ));
+    });
   });
 
   describeWithDOM('debug React wrapper', () => {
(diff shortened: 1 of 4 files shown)


packages/enzyme-test-suite/package.json
```diff
@@ -49,4 +49,4 @@
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-react": "^7.6.1"
   }
-}
```
ljharb: Please make sure all files always have a trailing newline :-)

packages/enzyme-test-suite/test/Debug-spec.jsx
```diff
@@ -626,6 +626,30 @@ describe('debug', () => {
   <span>
     span text
   </span>
+</div>`
+      ));
+    });
+
+    it('options.verbose causes boxed primitives to be unboxed', () => {
```
ljharb: None of these are boxed primitives; a boxed primitive is like `Object(2)`. Do you mean objects and arrays?

packages/enzyme/src/Debug.js
```diff
@@ -43,15 +44,19 @@ function propString(prop) {
     return `{${inspect(prop)}}`;
   }
   if (typeof prop === 'object') {
+    if (options.verbose) {
+      return `{${CircularJSON.stringify(prop)}}`;
```
ljharb: Why not `inspect(prop)`?

packages/enzyme-test-suite/package.json
blainekasten: yeah, not sure how this one happened. I'll fix it.

packages/enzyme-test-suite/test/Debug-spec.jsx
blainekasten: Ah. tbh, i heard you say it once and I didn't really know what it meant, so like everyone, I just pretended it was the right thing. my bad. I'll fix up the wording to just mean objects and arrays.

packages/enzyme/src/Debug.js
blainekasten: let me test this. I assumed it wouldn't do what I wanted it to.

blainekasten: This is currently the difference between the two libs:

```jsx
{ 
  inspect:      <div data-json={{ a: [ 1, 3, { true: true } ], b: false, c: { d: 'f' }, d: [ 1, 3, { true: true } ] }} data-arry={[ 1, 2, { f: { d: 'f' } } ]}>
  circularJSON: <div data-json={{"a":[1,3,{"true":true}],"b":false,"c":{"d":"f"},"d":"~a"}} data-arry={[1,2,{"f":{"d":"f"}}]}>
}
```

Seems like the inspect one looks more like real code, rather than json. But It will definitely crash if used against circular structures since it just prints them out directly. If I pass in something like:

```js
a = {};
a.b = {};
a.c = a.b;
a.b.c = a.c;
```

Passing that object as a prop will then crash the method.

We are using this method in enzyme-matchers and I think it would be really confusing if anyone ever ran into this.

blainekasten: Actually. Just tested and it looks like inspect does stringify circular properly! So i'll use that.

blainekasten:
@ljharb addressed your concerns dawg

packages/enzyme/src/Debug.js
ljharb: Awesome :-) I'm not particularly worried about circular cases anyways, but glad it handles it!

packages/enzyme-test-suite/test/Debug-spec.jsx
```diff
@@ -684,5 +708,31 @@ describe('debug', () => {
 </Bar>`
       ));
     });
+
+    it('options.verbose causes boxed primitives to be unboxed', () => {
```
ljharb: let's update this description also

packages/enzyme/src/ReactWrapper.js
```diff
@@ -988,8 +988,9 @@ class ReactWrapper {
   /**
    * Returns an HTML-like string of the shallow render for debugging purposes.
    *
-   * @param {Object} options - (Optional) Property bag of additional options.
-   * options.ignoreProps - if true, props are omitted from the string.
+   * @param {Object} [options] - Property bag of additional options.
+   * @param {boolean} [options.ignoreProps] - if true, props are omitted from the string.
+   * @param {boolean} [options.verbose] - if true, boxed primitives are unboxed.
```
ljharb: also here

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -1132,8 +1132,9 @@ class ShallowWrapper {
   /**
    * Returns an HTML-like string of the shallow render for debugging purposes.
    *
-   * @param {Object} options - (Optional) Property bag of additional options.
-   * options.ignoreProps - if true, props are omitted from the string.
+   * @param {Object} [options] - Property bag of additional options.
+   * @param {boolean} [options.ignoreProps] - if true, props are omitted from the string.
+   * @param {boolean} [options.verbose] - if true, boxed primitives are unboxed.
```
ljharb: also here

nfcampos:
Should we add a test for verbose: false that has arrays or objects to test the difference in behaviour between verbose false and true?

LGTM otherwise

ljharb:
Yes, that's a great idea

blainekasten:
@nfcampos thanks for the call out. I added that and addressed the rest of @ljharb's comments

blainekasten:
ping @ljharb

ljharb:
@blainekasten I don't see any tests for verbose: false?

blainekasten:
@ljharb https://github.com/airbnb/enzyme/pull/1547/files#diff-3cc78530daa735c5d407c3f768d35c57R743

ljharb:
aha, thanks :-) i'll rebase it and rereview.

PR #1555, opened 2018-03-03
Use jsdoc Boolean over boolean

oliviertassinari:
JSDoc isn't case sensitive: https://github.com/jsdoc3/jsdoc/issues/1046, but it doesn't mean we shouldn't try to reduce variance.


ljharb:
We’re not really using jsdoc here, or else we’d be linting for it.

PR #1569, opened 2018-03-12
Add event name mapping for animation events

robbiemccorkell:
Adding support for css3 animation events:
- [animationstart](https://developer.mozilla.org/en-US/docs/Web/Events/animationstart)
- [animationiteration](https://developer.mozilla.org/en-US/docs/Web/Events/animationiteration)
- [animationend](https://developer.mozilla.org/en-US/docs/Web/Events/animationend)

The support for these events in React's documentation can be found [here](https://reactjs.org/docs/events.html#animation-events)


ljharb:
Looks like these were added in https://github.com/facebook/react/pull/6005, which is in v15+.

Issue #1566, opened 2018-03-09
"findWhere" by node.text() when mounted causes "Cannot read property 'textContent' of null"


angelikatyborska:
### Current behavior

I would like to write a test that checks if a certain phrase appears anywhere in my rendered component. When I use `shallow`, I have no issues:

```javascript
const container = shallow(
  <div>
    <div>apple</div>
    <div>banana</div>
    <div>orange</div>
  </div>
);

const nodesWithBanana = container.findWhere(node => /banana/.test(node.text()));
console.log(nodesWithBanana.length);
// 3 (outer div, inner div, the text node itself)
```

However, when I switch to `mount` instead, the same operation gives me an error that does not make a lot of sense to me:

```javascript
const container = mount(
  <div>
    <div>apple</div>
    <div>banana</div>
    <div>orange</div>
  </div>
);

const nodesWithBanana = container.findWhere(node => /banana/.test(node.text()));
// TypeError: Cannot read property 'textContent' of null
```
Full stacktrace:
```
TypeError: Cannot read property 'textContent' of null
    at ReactWrapper.<anonymous> (/Users/angelika/code/MyApp/node_modules/enzyme/build/ReactWrapper.js:799:43)
    at ReactWrapper.single (/Users/angelika/code/MyApp/node_modules/enzyme/build/ReactWrapper.js:1534:25)
    at ReactWrapper.text (/Users/angelika/code/MyApp/node_modules/enzyme/build/ReactWrapper.js:798:21)
    at /Users/angelika/code/MyApp/source/components/AnnotationComponent.spec.jsx:165:76
    at /Users/angelika/code/MyApp/node_modules/enzyme/build/ReactWrapper.js:1409:18
    at /Users/angelika/code/MyApp/node_modules/enzyme/build/RSTTraversal.js:75:9
    at treeForEach (/Users/angelika/code/MyApp/node_modules/enzyme/build/RSTTraversal.js:65:5)
    at /Users/angelika/code/MyApp/node_modules/enzyme/build/RSTTraversal.js:68:12
    at Array.forEach (<anonymous>)
    at treeForEach (/Users/angelika/code/MyApp/node_modules/enzyme/build/RSTTraversal.js:67:24)
    at /Users/angelika/code/MyApp/node_modules/enzyme/build/RSTTraversal.js:68:12
    at Array.forEach (<anonymous>)
[...]

PR #1582, opened 2018-03-16
[Tests] Extend findWhere null nodes test

angelikatyborska:
As requested by @ljharb, this is a failing test case for issue https://github.com/airbnb/enzyme/issues/1566.

`expect(spy.callCount).to.equal(3);` - I am not certain if this assertion should expect 3 calls or 4 calls. That depends if the `null` node that is currently passed will be skipped or fixed not to be `null`.


mhuggins:
+1 on merging this and getting a fix.

ljharb:
Hmm - I'm not sure I think there's a bug here. I think that it makes sense for `findWhere` to receive every node - specifically in case it wants to find text nodes - and that it makes sense for `.getDOMNode()` to return `null` on text nodes.

However, I think that `.text()` on a textNode should just return the text, so i'll update this PR with the updated tests, and then add a fix.

Thank you - this helped me understand the problem!

xixixao:
This changes the type of text() to be `?string`, so we should update Flow types (and presumably TS?).

ljharb:
@xixixao we don't use or maintain our own types, so please file issues in the appropriate places.

PR #1585, opened 2018-03-18
You need to import the component

anshu0612:
It's important for the users to know that they need to import the component. Since after rendering React component , they turn into html tags.


anshu0612:
@ljharb AFAIK none!

PR #1592, opened 2018-03-23
Forward ref

jquense:
This is built on top the Context PR, i'll rebase once that is merged

Adds support for forwardRef to accompany the new changes.

The dependencies for these to work are getting fairly specific at this point, I’d like to bump the rep range for react-test-renderer, but that effectively bumps the peer dips for react, maybe worth splitting to a new adapter at this point?

It would be really great if we could use toTree from react-test-renderer directly…


petegleeson:
Hi mate can we rebase these changes with master now #1513 is merged?

The lack of forwardRef support is a bit of a blocker for us at the moment. I have time to work on this so please let me know if there is anything I can do.

jquense:
CI failure looks like a travis stalled build

packages/enzyme/src/Utils.js
```diff
@@ -7,8 +7,47 @@ import configuration from './configuration';
 import validateAdapter from './validateAdapter';
 import { childrenOfNode } from './RSTTraversal';
 
+const hasSymbol = typeof Symbol === 'function' && Symbol.for;
+
 export const ITERATOR_SYMBOL = typeof Symbol === 'function' && Symbol.iterator;
 
+// TODO use react-is when a version is released with no peer dependency on React
```
Droeftoeter: The latest react-is release has your change in it and does not have react as peer dependency anymore. Could this be updated?

azmenak:
As @petegleeson mentioned, this PR seems to be a blocker for anyone using forwardRefs. Anything we can do to get this moving?

Seems the `react-is` changes are strait forward, is there any way to add those changes to this PR? Happy to make those edits if it helps.

sambokai:
Any news on this?

janv:
What's holding this back?

cyan33:
Does this PR fix the forwardRef issue and is good to merge? Much appreciated if this gets merged soon. 😄

nemoDreamer:
Yup... I'm about to switch my tests that rely on this back to Facebook's TestRenderer... 😢

jquense:
If anyone needs an interim solution i've released `@monastic.panic/enzyme-adapter-react-16` folks can use

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
```diff
@@ -55,7 +60,7 @@ export default function createMountWrapper(node, options = {}) {
     }
   }
   WrapperComponent.propTypes = {
-    Component: PropTypes.oneOfType([PropTypes.func, PropTypes.string]).isRequired,
+    Component: PropTypes.oneOfType([PropTypes.func, PropTypes.string, specialType]).isRequired,
```
ljharb: `PropTypes.element`?

packages/enzyme-test-suite/test/_helpers/react-compat.js
```diff
@@ -37,9 +38,16 @@ if (is('^16.3.0-0')) {
   createContext = null;
 }
 
+if (is('^16.3.0-0')) {
+  ({ forwardRef } = require('react'));
```
ljharb: let's combine createContext with forwardRef, since they're in the same versions.

packages/enzyme/src/Debug.js
```diff
@@ -66,6 +94,7 @@ function indentChildren(childrenStrs, indentLength) {
 
 export function debugNode(node, indentLength = 2, options = {}) {
   if (typeof node === 'string' || typeof node === 'number') return escape(node);
+  if (typeof node === 'function') return '[function child]';
```
ljharb: why is this necessarily a function child?

packages/enzyme/src/Utils.js
ljharb: a) `react-is` must never be used in new places outside of a react adapter, since non-react adapters exist. b) `react-is` does not work on react 0.13.

So, all of these things need to be exposed via the adapter interface - instead of being part of enzyme.

packages/enzyme/src/Utils.js
```diff
@@ -7,8 +7,47 @@ import configuration from './configuration';
...
+export const AsyncMode = hasSymbol ? Symbol.for('react.async_mode') : 0xeacf;
+export const ContextProvider = hasSymbol ? Symbol.for('react.provider') : 0xeacd;
+export const ContextConsumer = hasSymbol ? Symbol.for('react.context') : 0xeace;
+export const Element = hasSymbol ? Symbol.for('react.element') : 0xeac7;
+export const ForwardRef = hasSymbol ? Symbol.for('react.forward_ref') : 0xead0;
+export const Fragment = hasSymbol ? Symbol.for('react.fragment') : 0xeacb;
+export const Portal = hasSymbol ? Symbol.for('react.portal') : 0xeaca;
+export const StrictMode = hasSymbol ? Symbol.for('react.strict_mode') : 0xeacc;
+
+export function typeOf(node) {
```
ljharb: this needs unit tests.

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
jquense: its not an element right it's a component class, e..g `element.type`

ljharb: `$$typeof` isn't in react 13 (which is why `react-is` doesn't work on it), so i'm not sure that the propType is correct.

packages/enzyme/src/Debug.js
jquense: I'm not sure what else to call a `props.children` that's a function. Perhaps cases also hit here, but i'm not sure what they are.

ljharb: right but `debugNode` won't necessarily be called on a child.

packages/enzyme/src/Utils.js
jquense: I'm not sure what's actionable here. It' seems like you are saying we should update all adapters to provide a `typeOf` api?

ljharb: Yes, i think that might be what's required :-/

if so, let's do that in a separate PR, so that when this one is rebased, it can only include the diffs to typeOf?

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -112,10 +113,11 @@ function toTree(vnode) {
...
       return node.memoizedProps;
-    case Fragment: // 10
-    case Mode: // 11
-    case ContextProvider: // 13
-    case ContextConsumer: // 12
+    case Fragment:
+    case ContextProvider:
+    case ContextConsumer:
+    case Mode:
+    case ForwardRef:
```
cyan33: If we want to find the ForwardRef itself, we should add a node for it, not only its children here in this case.

cyan33:
To fix finding forwardRef itself, I had to change this function:

```
export function nodeMatchesObjectProps(node, props) {
  if (typeOf(node) === ForwardRef) {
    return isSubset(node.type, replaceUndefinedValues(props));
  }
  return isSubset(propsOfNode(node), replaceUndefinedValues(props));
}
```

and in the react-16-adapter, in the `toTree` method:
```
case ForwardRef:
    return {
        nodeType: 'forward_ref',
        type: node.type,
        props: { ...node.memoizedProps },
     };
```

Not sure if it's correct.

jquense:
Alright, as @cyan33 notes, the problem here is that we _can't_ add support for ForwardRef selecting without some sort of brand check.

```js
mount(<TreeWithForwardRef/>).find(ForwardRefComponent)
```
Doesn't work here because selectors are pretending to be agnostic, but are actually codifying the assumption that element types can only be functions or strings, which isn't true anymore for React. The simplest but dirty approach is to go with what I have here for `typeof` logic, which is honestly an extension of the current state of affairs, even if that SOA isn't the intended one. The alternative, as @ljharb notes, is to add a `isValidElementType` api to adapters and pass that through to the selectors, which is a much bigger refactor.

ljharb:
I think adding that API to the adapters is a fine idea; let's just do it in a separate PR and rebase this after it's merged.

wmonk:
Thanks for your work on this! Is there any capacity to get this moving forward? I've tried the patched packages but none of them seemed to work.

ljharb:
@wmonk once #1701 has tests and is merged, this can be rebased and re-evaluated.

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
```diff
@@ -3,6 +3,11 @@ import PropTypes from 'prop-types';
 
 /* eslint react/forbid-prop-types: 0 */
 
+// TODO: use react-is?
+const specialType = PropTypes.shape({
+  $$typeof: PropTypes.any.isRequired,
```
ljharb: React 0.13 doesn't have `$$typeof`, a div in 0.13 looks like:
```js
{ type: 'div',
  key: null,
  ref: null,
  _owner: null,
  _context: {},
  _store: { props: {}, originalProps: {} } }
```

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
ljharb: We could have multiple things here - but we might want to add a propType to each adapter for "a component" - that way react 0.13 can specify that the function has to be a class component, for example, and not an SFC.

ljharb:
@jquense the propType doesn't seem to be protecting against anything; no tests fail when I comment it out. Why is it needed?

jquense:
It was complaining if you try and use a special element for mounting, e.g. a forwardRef(). `mount(<ForwardedRef/>)` should warn.

ljharb:
@jquense oh interesting, why? Can we add a test that fails when that happens?

jquense:
rebased. I confirmed that the tests do fail if you remove the special propType, but i had to make sure i restarted the test server for whatever reason. 

>  oh interesting, why?

The special element types aren't functions or strings. We could do a custom prop-type using the calling adapters `isValidElementType` but it get s a bit messy passing it in.

ljharb:
I don't run the tests in watch mode, and I'm still unable to reproduce the test failure when I revert line 70 in the React 16 adapter.

ljharb:
@jquense by "why" i meant, why would you be able to mount `<div><ForwardRef /></div>` but not `<ForwardRef />`? It seems like both should work.

jquense:
They both work insofar as the mount and are queryable but the latter fails the prop type for the mount wrapper. In the second case the Component passed to createMountWrapper is the type off the element passed in, which is the ForwardRef.

I'm really not sure why it doesn't warn for you...is it clear that ForwardRef.type would fail this prop type while div.type wouldn't? If it doesn't fail I'd think there is something with the warnings being swallowed during a test run

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
jquense: isValidElementType that we added would be the the right check I think. We could pass that in and use it in a custom proptype

ljharb:
How are you testing it? I run `npm run build && npx mocha packages/enzyme-test-suite/test` from the root.

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
aldeed: @jquense This may be helpful since I just figured this same prop type check out: https://github.com/facebook/prop-types/issues/200#issuecomment-409391696

(Eagerly awaiting this PR merge)

aldeed:
@ljharb @jquense I made a fork with these changes and it generally works. The one issue I'm seeing is that `wrapper.instance()` always returns `null` when it wraps a forwardRef return object. It's possible that that's an issue only in my fork since I had to selectively build and publish the related packages, and maybe I don't have some newer code. But it also seems like it could be due to the `instance: null` part of this PR. Maybe need to add a test to ensure that `wrapper.instance()` works as expected?

jquense:
The null for instance makes sense these components, like stateless function components, don't have public instance s

mhuggins:
What can we do in the meantime?

jquense:
I just had a baby so my time for for this is a bit limited to say the least :P I do think I added a test that was at least meant to test this in here somewhere. Still not sure if it does actually test it tho

ljharb:
Congrats!

It's currently not tested, since if I remove the special propType handling everything will still pass :-/ if anyone else wants to post a link to a commit that adds/updates the tests, I'm happy to pull it in.

ljharb:
I've played with this a bit; I still can't get the propType to matter. Separately, I've added some debug tests that fail.

suchipi:
I'm gonna pick this up and try to finish it.

suchipi:
I rebased this, added a test for the special proptype handling, and fixed the failing tests; see https://github.com/airbnb/enzyme/pull/1738

ljharb:
@suchipi please do NOT ever create a duplicate PR - that forever clutters the git refs of the repo. I very explicitly said "post a link to a commit that adds/updates the tests" meaning "do NOT create a PR".

suchipi:
Oh. Sorry.

https://github.com/suchipi/enzyme/commit/b4cecf121a7ab44c0b4fa23cac4ff2ece68f0cdb

suchipi:
It seems the solution for the debug failures in my referenced commit doesn't work after all; the tests pass, but when running on my app at work, there are new failures related to prop values changing. I'm guessing it's because I'm using pendingProps instead of memoizedProps, so that solution isn't viable.

ehellman:
What's the suggested workaround for this now until it's fixed?

ljharb:
@ehellman same as any other react feature not supported by enzyme: either don't use the feature, or don't test components that use it, until enzyme has support for it.

wmonk:
🎉 is there an ETA for this getting published to npm?

ljharb:
It was merged 9 hours ago; you asked 6 hours ago. Have some patience.

jquense:
Thanks for finishing this up @ljharb!

probablyup:
@ljharb looks like there's been a few enzyme patches releases since this was merged. Could we get a react 16 adapter release also to start using this? 🙇

ljharb:
@probablyup those were backports that don't include this (or other semver-minor) changes. This feature remains unreleased, but will be released when things are ready.

kenips:
@ljharb I know this isn't typical release style but considering the huge volume here could you please please please consider cutting a `beta`, `next`, or `1.3.0.beta-1` for ease of testing this? Thank you <3

kenips:
Thank you @ljharb your great work is very much appreciated.

PR #1605, opened 2018-04-05
Update find.md (to.have.length doesn't works) 

anshu0612:
The current implementation doesn't seem to work.
`Returns: TypeError: Cannot read property 'have' of undefined. `


anshu0612:
Yup right, this also seems to works, in fact, a better way to do it.

Issue #1614, opened 2018-04-11
Typo in 2->3 migration guide


3lvcz:
### what is wrong

Mistyped enzyme version number in first sentence of the following text

### original text:

Prior enzyme v3, we would observe the following behavior:

```jsx
wrapper.find(Box).children().debug();
// => <div className="div" />
```
In enzyme v3, we now have .children() return the rendered children. In other words, it returns the element that is returned from that component's render function.

```jsx
wrapper.find(Box).children().debug();
// =>
// <div className="box">
//   <div className="div" />
// </div>
```

PR #1615, opened 2018-04-11
fix typo in from 2 to 3 migration guide

3lvcz:
Resolves issue #1614


docs/guides/migration-from-2-to-3.md
```diff
@@ -135,7 +135,7 @@ Although the `<Box ... />` element has a `children` prop of `<div className="div
 rendered children of the element that the box component renders is a `<div className="box">...</div>`
 element.
 
-Prior enzyme v3, we would observe the following behavior:
+Prior enzyme v2, we would observe the following behavior:
```
ljharb: In fact this is correct as-is (altho it should be "Prior to enzyme v3") - prior to v2 is v1. Prior to v3 is v1 and v2, which is what this sentence is describing.

PR #1659, opened 2018-05-19
Drop Node v4 and add v10 for TravisCI

koba04:
Node v4 is no longer maintained. We should support Node v10 instead of v4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 node_js:
+  - "10"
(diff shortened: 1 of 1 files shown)


koba04:
OK, thanks!

ljharb:
Note that adding v10 is great, if you wanted to reopen this PR and reduce it down to that?

koba04:
@ljharb I've created the PR as another PR #1660.

PR #1684, opened 2018-06-20
Fix finding forwardRef itself

cyan33:
This PR is built on top of @jquense 's https://github.com/airbnb/enzyme/pull/1592

It fixes when the `wrapper.find` method cannot find the forwardRef itself. The reason this is happening is that `React.forwardRef` returns a type that is unknown to enzyme. The return type looks like this:

```js
{$$typeof: Symbol.for('react.forward_ref'), render: [Function]}
```

What we really want to search is the thing wrapped inside of `forwardRef`, so one way to do this is to call the `render` method and use the return element to re-build a predicate based on that.

### DIFF
--- a/packages/enzyme-adapter-utils/src/createMountWrapper.jsx
+++ b/packages/enzyme-adapter-utils/src/createMountWrapper.jsx
@@ -3,6 +3,11 @@ import PropTypes from 'prop-types';
 
 /* eslint react/forbid-prop-types: 0 */
 
+// TODO: use react-is?
+const specialType = PropTypes.shape({
+  $$typeof: PropTypes.any.isRequired,
+}).isRequired;
+
 /**
  * This is a utility component to wrap around the nodes we are
  * passing in to `mount()`. Theoretically, you could do everything
@@ -55,7 +60,7 @@ export default function createMountWrapper(node, options = {}) {
     }
   }
   WrapperComponent.propTypes = {
-    Component: PropTypes.oneOfType([PropTypes.func, PropTypes.string]).isRequired,
+    Component: PropTypes.oneOfType([PropTypes.func, PropTypes.string, specialType]).isRequired,
     props: PropTypes.object.isRequired,
     context: PropTypes.object,
   };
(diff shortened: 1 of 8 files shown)


packages/enzyme-adapter-utils/src/createMountWrapper.jsx
```diff
@@ -3,6 +3,11 @@ import PropTypes from 'prop-types';
 
 /* eslint react/forbid-prop-types: 0 */
 
+// TODO: use react-is?
+const specialType = PropTypes.shape({
```
jquense: should use `react-is`

packages/enzyme/src/selectors.js
```diff
@@ -227,6 +227,12 @@ export function buildPredicate(selector) {
       if (hasUndefinedValues) {
         throw new TypeError('Enzyme::Props can’t have `undefined` values. Try using ‘findWhere()’ instead.');
       }
+      // the selector could also be a forwardRef
+      if (typeOf(selector) === ForwardRef) {
+        // re-build the predicate based on what is wrapped by forwardRef
+        // rather than the forwardRef itself
+        return buildPredicate(selector.render().props);
```
jquense: I don't think this is the right way to do this but maybe i'm misunderstanding. We should copy waht react-test-renderer does and this code should go in the react-16-adapter

aweary: Yeah the selector module is used for all adapters, so the node type should be opaque in this module.

ljharb:
@jquense are you able to pull this PR into yours, such that we don't have to have stacked PRs?

jquense:
I'll update my PR. I think someone actually already PR'd My PR with the correct fix

ljharb:
(I'll leave this open and sync the two of them myself when the first one's ready to merge)

cyan33:
Actually this PR's change is not correct as @aweary and @jquense pointed out that the change should not be in enzyme core. I've updated my fix in the comment of @jquense's PR. I'll close it for now.

PR #1696, opened 2018-06-29
[enzyme-adapter-react-*, shallow] [refactor]: add “lifecycles” adapter option

ljharb:
The intention here is to make a more scalable way to indicate which lifecycle methods specific adapters support.

The `getSnapshotBeforeUpdate` change here is breaking, but has not yet been published.

The other changes should be both backwards and forwards compatible - in other words, a before-PR enzyme should work with an after-PR adapter, and a before-PR adapter should work with an after-PR enzyme - which makes all the versions non-major. **Please check this part closely.**

This will create a nice clean place for the rest of the new React 16 lifecycle methods to go.


packages/enzyme/src/ShallowWrapper.js
```diff
@@ -405,22 +461,24 @@ class ShallowWrapper {
...
+            lifecycles.getSnapshotBeforeUpdate &&
             typeof instance.getSnapshotBeforeUpdate === 'function'
           ) {
             const snapshot = instance.getSnapshotBeforeUpdate(prevProps, prevState);
             if (typeof instance.componentDidUpdate === 'function') {
               instance.componentDidUpdate(prevProps, prevState, snapshot);
             }
           } else if (typeof instance.componentDidUpdate === 'function') {
-            if (adapter.options.supportPrevContextArgumentOfComponentDidUpdate) {
+            if (lifecycles.componentDidUpdate.prevContext) {
```
brucewpaul: is `lifecycles.componentDidUpdate` guaranteed in this context as opposed to `lifecycles.componentDidUpdate.onSetState`? Otherwise we should check to see if `lifecycles.componentDidUpdate` exists.

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -88,6 +94,48 @@ function validateOptions(options) {
...
+      }),
+      ...(hasLegacyPrevContextArg && {
+        prevContext: !!supportPrevContextArgumentOfComponentDidUpdate,
+      }),
+    }
+    : null;
+
+  return {
+    ...lifecycles,
+    ...(componentDidUpdate && { componentDidUpdate }),
```
brucewpaul: I could foresee a problem where there is a mismatch between 
```
supportPrevContextArgumentOfComponentDidUpdate: true
``` 
and 
```
lifecycles: {
  ...lifecycles,
  componentDidUpdate: {
    prevContext: false,
  },
},
```

this code above would mean that the legacy settings would the one that is honored. Would we instead want to honor the new lifecycles settings object over the legacy settings if both are present? This isn't a block, but more of a discussion. Also not sure if there is a clean way to do this...

brucewpaul: I suppose that this would be caught here:

```
typeof enableComponentDidUpdateOnSetState !== 'undefined' &&
lifecycles.componentDidUpdate &&
lifecycles.componentDidUpdate.onSetState !== enableComponentDidUpdateOnSetState
```
but then that makes the error message potentially misleading, as the new setting format is set, but is different.

packages/enzyme/src/ShallowWrapper.js
ljharb: `lifeCycles.componentDidUpdate` is ensured by the check on line 468

packages/enzyme/src/ShallowWrapper.js
ljharb: Good call; i'll improve the error message.

PR #1698, opened 2018-06-29
Bugfix for using general sibling selector on root

krawaller:
Right now `.find('#foo ~ #bar')` will throw an error if `#foo` matches the root element.

This PR adds test and the missing parent check that caused the problem.


ljharb:
@krawaller btw you should add the email address you used on these commits to your github account, so you can be linked to the commits.

krawaller:
Oh, right! Done, thank you @ljharb!

PR #1707, opened 2018-07-06
[enzyme-adapter-react-16] [New] `mount`: add `hydrateIn` option

### LINKED ISSUE #1316: support react's hydrate for enzyme-adapter-react-16

ljharb:
Fixes #1316

Reviewers: please help ensure that the following scenarios will work after this PR:
 - new enzyme + new adapter
 - new enzyme + old adapter
 - old enzyme + new adapter

### DIFF
--- a/packages/enzyme/src/Utils.js
+++ b/packages/enzyme/src/Utils.js
@@ -21,9 +21,25 @@ export function getAdapter(options = {}) {
 }
 
 export function makeOptions(options) {
+  const { attachTo, hydrateIn } = options;
+
+  if (attachTo && hydrateIn && attachTo !== hydrateIn) {
+    throw new TypeError('If both the `attachTo` and `hydrateIn` options are provided, they must be === (for backwards compatibility)');
+  }
+
+  // neither present: both undefined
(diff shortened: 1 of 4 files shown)


packages/enzyme/src/Utils.js
```diff
@@ -21,9 +21,25 @@ export function getAdapter(options = {}) {
 }
 
 export function makeOptions(options) {
+  const { attachTo, hydrateIn } = options;
+
+  if (attachTo && hydrateIn && attachTo !== hydrateIn) {
```
brucewpaul: would there ever be a case when both are present?  Do we just want to not allow this and then hydrate -> attach on pre16 versions?

ljharb: So, I’ve normalized the options here - which means it will end up having both the best time it hits this code path. So altho users really shouldn’t supply both, i cant think of how to differentiate between user options and things passed in via shallow calls or adapter options etc.

PR #1710, opened 2018-07-09
React adapter 16: portals and roots may render fragments

jquense:
This matches what react-test-renderer does as well.


packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -833,6 +833,17 @@ describeWithDOM('mount', () => {
...
 
+    itIf(REACT16, 'should mount fragments', () => {
+      const wrapper = mount(
+        <React.Fragment>
+          <p>hello</p>
+          <span>boo</span>
+        </React.Fragment>
+      );
+
+      expect(wrapper).to.have.lengthOf(2);
```
ljharb: can we add assertions that the wrapper has a p and a span?

jquense:
I removed the two shallow tests, i'm not sure why i added them in the first place. ShallowRenderer doesn't support mounting a Fragment, and a Portal doesn't make any sense, since it's a DOM specific element.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -251,11 +252,13 @@ class ReactSixteenAdapter extends EnzymeAdapter {
...
           isDOM = false;
           const { type: Component } = el;
+
           const isStateful = Component.prototype && (
             Component.prototype.isReactComponent
             || Array.isArray(Component.__reactAutoBindPairs) // fallback for createClass components
           );
-          if (!isStateful) {
+
+          if (!isStateful && typeof Component === 'function') {
```
jquense: This is just so we get a clearer error is a user tries to mount a Fragment or Portal, they will see the ShallowTestRenderer  informative error instead of the much less clear "Component is not a function"

ljharb:
@jquense I'd actually added them, for test parity. We *should* be able to shallow-render a fragment and a Portal.

jquense:
I'm not sure what you want to do in this case, it doesn't make much sense to me to intentionally hobble the mount cases for parities sake in my mind.

The main issue tho is that upstream shallow renderer does not allow it. Same is true for the new context types as well

ljharb:
Hmm. I certainly don't want to limit what `mount` can do.

However, I also don't want to limit what `shallow` can do just because the upstream renderer can't handle it. Is there *any* way that a Portal, Fragment, or Context component couldn't just be treated like a `div`, in that in shallow mode it simply renders its children?

jquense:
I think we can probably get clever and hack something together but maybe we can split that into a new PR? The goal here was really to support selectors through these types deep in element tree not at the root. at moment the adapter assumes they can only render single children. The tests have sort of accidentally brought to light this inconsistency, but it's one that already exists, not one introduced by this PR

PR #1721, opened 2018-07-22
[New] `shallow`: `setProps`: add callback argument

### LINKED ISSUE #1716: `setProps`: fix callback support
See https://github.com/airbnb/enzyme/compare/setProps and https://travis-ci.org/airbnb/enzyme/builds/406918917 for the failing tests, on basically all react versions.

A PR that makes these tests pass (or corrects them, if they're wrong) would be greatly appreciated!

emuraton:
Fixes #1716

### DIFF
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -1240,6 +1240,66 @@ describe('shallow', () => {
       ]);
     });
 
+    it('should throw if an exception occurs during render', () => {
+      class Trainwreck extends React.Component {
+        render() {
+          const { user } = this.props;
+          return (
+            <div>
+              {user.name.givenName}
+            </div>
+          );
+        }
+      }
+
+      const similarException = ((() => {
+        const user = {};
+        try {
+          return user.name.givenName;
+        } catch (e) {
+          return e;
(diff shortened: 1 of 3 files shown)


packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1272,10 +1272,8 @@ describe('shallow', () => {
       expect(wrapper.find('.foo')).to.have.lengthOf(1);
 
       wrapper[sym('__renderer__')].batchedUpdates(() => {
-        wrapper.setProps({ id: 'bar', foo: 'bla' }, () => {
-          expect(wrapper.find('.bar')).to.have.lengthOf(1);
-        });
```
emuraton: I think callback is not supported for ShallowRendering
https://github.com/airbnb/enzyme/blob/master/packages/enzyme/src/ShallowWrapper.js#L381

ljharb: no, but it probably should be.

emuraton: does it mean I should change `setProps` function in `ShallowWrapper`?
Because right now `setProps()` only has on parameter for the props to update, right?

PR #1730, opened 2018-08-02
Adds Mount tracking

### LINKED ISSUE #437: Proposal: Add a `mount` function that acts as a context manager for a component
Neither enzyme's `mount` nor TestUtil's `renderIntoDocument` will clean up a component when you're done testing it (i.e., unmount it and ensure proper life-cycle methods are invoked.)  This can be problematic when your component has side-effects like listening to a global store singleton for updates, in which case each component you ever mount in your test suite will forever listen to store broadcasts, which could prohibitively clog your runtime's event loop. 

While coupling global state into a component's unit test is surely an anti-pattern, I'm confident this subtle issue is lurking it many react test suites.

I think enzyme is in a great position to a) raise awareness so developers are at least conscious of their components never being unmounted and b) provide a clean interface for safely mounting a component and ensuring it is automatically cleaned up when any assertions have completed.  I have a demo of this problem and a proposed solution here: https://github.com/griffinmyers/react-test, but the tl;dr of it is something like this: 

``` javascript
it('safeMounts', function() {
  safeMount(<Baz />, function(wrapper) {
    expect(wrapper.find('.baz').length).to.equal(1);
  });
});

it('safeMounts async', function(done) {
  safeMount(<Baz />, function(wrapper, unmount) {
    expect(wrapper.state('mountedOneSecondAgo')).to.equal(false);
    setTimeout(() => {
      expect(wrapper.state('mountedOneSecondAgo')).to.equal(true);
      unmount();
    }, 1000);
  }, done);
});
```

I'd love to hear some thoughts from you folks and am happy to work up a PR if you think this is something worth pursuing.

### LINKED ISSUE #1689: Mount tracking
**Is your feature request related to a problem? Please describe.**

When using full DOM rendering, I'd like to be able to orchestrate calling `unmount` via an afterEach-style hook, rather than having to call `unmount` in each test.

*Note* When `unmount` was not called at all, I experienced memory leaks when working with large test suites.

**Describe the solution you'd like**

Obviously, enzyme does not assume a specific test environment. I think the right solution here is for enzyme to provide some form of mount-tracking with an `unmountAll` function.

**Describe alternatives you've considered**

Currently, we expose a library which wraps enzyme's mount function, tracking the `ReactWrapper` instances, and then calling `unmount` on them via our library.

This solution is problematic, as most developers are familiar with `import {mount} from 'ezyme';`, and we'd end up needing a lint rule to prevent that. Also, I feel that this solution makes sense to be included in enzyme itself so that others can benefit from better handling of memory leaks.

**Additional context**

Some implementation considerations:

* if someone calls `unmount` manually, we probably want to delete any global references tot he underlying wrapper, so that the JS engine can properly garbage-collect it. It also doesn't make sense to call `unmount` multiple times anyhow.

* Many applications will use something like mocked timers int heir testing suites. We'll need to consider the case where these utilities are referenced via a component's `componentWillUnmount` hook. In such a case, if the mock is removed after the test and then we try to run a global `unmountAll` hook on that component, we may encounter errors from the mocking utility. One way to handle this is via documentation, suggesting the package consumer to manually call `unmount` on those wrappers before restoring any mocked utilities.

TzviPM:
Closes #1689. Closes #437.


tzvipm:
looks like one of the jobs in my build timed out before starting the tests: https://travis-ci.org/airbnb/enzyme/jobs/411307270 cc @ljharb

packages/enzyme/src/mountTracking.js
```diff
@@ -0,0 +1,31 @@
...
+let mountedWrappers = [];
+
+/**
+ * Stores a reference to a testing wrapper for later unmounting
+ * via unmountAllWrappers()
+ *
+ * @param {ReactWrapper|ShallowWrapper} wrapper
+ */
+export function trackMountedWrapper(wrapper) {
+  const { enableMountTracking } = get();
```
ljharb: i think this probably needs `options` to be passed through to `get`?

packages/enzyme/src/mountTracking.js
```diff
@@ -0,0 +1,31 @@
...
+  if (enableMountTracking) {
+    mountedWrappers.push(wrapper);
```
ljharb: what if it's already in there? this should probably use a Set

packages/enzyme/src/mountTracking.js
tzvipm: I'm not sure I follow. `get` just returns a shallow copy of the configuration object, does it not?

docs/installation/mount-tracking.md
```diff
@@ -0,0 +1,60 @@
...
+
+ES6:
+```js
+// setup file
+import { configure } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+
+configure({
+  adapter: new Adapter(),
+  enableMountTracking: true,
```
ljharb: it seems kind of weird to call this option "mount tracking" when it tracks both shallow and mount. maybe "enableSandbox", since `sinon-sandbox` is actually the behavior it's emulating?

packages/enzyme/src/index.js
```diff
@@ -15,4 +16,6 @@ module.exports = {
   ReactWrapper,
   configure,
   EnzymeAdapter,
+  trackMountedWrapper,
```
ljharb: why does this need to be re-exported?

tzvipm: for custom implementation of `mount` that extend the enzyme wrapper. We use this pattern often, for example, to encapsulate logic around application context.

ljharb: I don't think it needs to be in the root export tho - consumers can deep-import whatever they need from inside `build`

tzvipm:
@ljharb can you have another look when you get a chance?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -79,6 +80,15 @@ describeWithDOM('mount', () => {
       mount(<div ref={spy} />);
       expect(spy).to.have.property('callCount', 1);
     });
+
+    it('should call trackWrapper', () => {
+      const spy = sinon.spy();
+      const originalTrackWrapper = wrapperSandbox.trackWrapper;
```
ljharb: eesh, this is a really brittle test that will definitely break in babel 7 and when node ships native ESM support. is this necessary? can we instead test the actual behavior?

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -75,6 +76,15 @@ describe('shallow', () => {
       expect(wrapper.children().type()).to.equal('div');
       expect(wrapper.children().props().bam).to.equal(undefined);
     });
+
+    it('should call trackWrapper', () => {
+      const spy = sinon.spy();
+      const originalTrackWrapper = wrapperSandbox.trackWrapper;
```
ljharb: same here

packages/enzyme-test-suite/test/wrapperSandbox-spec.jsx
```diff
@@ -0,0 +1,35 @@
...
+import sinon from 'sinon';
+import { ReactWrapper, configure } from 'enzyme';
+import { get } from 'enzyme/build/configuration';
+
+import {
+  trackWrapper,
+  unmountAllWrappers,
+} from 'enzyme/build/wrapperSandbox';
+
+const originalConfig = get();
```
ljharb: this should probably be populated `before` or `beforeEach` rather than at require time

packages/enzyme-test-suite/test/wrapperSandbox-spec.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+describe('wrapperSandbox', () => {
+  afterEach(() => {
+    configure(originalConfig);
+  });
+
+  it('does what i expect', () => {
+    const wrapper = new ReactWrapper(<p>foo</p>);
+    const spy = sinon.spy();
+    wrapper.unmount = spy;
```
ljharb: `const spy = sinon.spy(wrapper, 'unmount');`

packages/enzyme/src/wrapperSandbox.js
```diff
@@ -0,0 +1,31 @@
...
+ * Unmounts all sandboxed Enzyme wrappers.
+ *
+ * Usually, this can be run once for an entire test suite after all each test
+ * (and its nested hooks)have been run. However, in some cases this may need
+ * to be run this manually.This is most commonly needed when a component uses
+ * timeouts/ animation frames that are mocked for tests; in that case, waiting
+ * until after you have restored those globals will lead to their stored
+ * identifiers being invalid.
+ */
+export function unmountAllWrappers() {
```
ljharb: i kind of feel like if `enableSandbox` is false, this function should throw

TzviPM:
Cleaning up my PRs and this doesn't look like it's going anywhere after 6 years 🤪

ljharb:
@TzviPM itd be great if you could restore and reopen it; the elapsed time has no bearing on the likelihood of merging.

PR #1734, opened 2018-08-08
Fix an adapter table style in README.md

koba04:


README.md
```diff
@@ -43,7 +43,7 @@ React:
 | `enzyme-adapter-react-16` | `^16.4.0-0` |
 | `enzyme-adapter-react-16.3` | `~16.3.0-0` |
 | `enzyme-adapter-react-16.2` | `~16.2` |
-| `enzyme-adapter-react-16.1` | `~16.0.0-0 || ~16.1` |
+| `enzyme-adapter-react-16.1` | `16.0.0-0 - 16.1.x` |
```
ljharb: I see how the pipes could cause a problem, but it’s essential to keep the same semver range here that the packages use. Can we find a way to make the pipes work?

koba04:
@ljharb I use `\` to escape `||`, which seems to work fine on GitHub.
The markdown preview of VSCode show `\` though.

I've updated the table in the migration guide.

koba04:
Thanks!

PR #1738, opened 2018-08-08
Forward ref

### LINKED ISSUE #1604: Enzyme Internal Error: unknown node with tag 14
### Current behavior
After update `React` to 16.3.0 version and use `forwardRef`, my tests is broken.

Component that uses `forwardRef` now exports as:
```
export default forwardRef((props, ref) => (
  <Input {...props} forwardedRef={ref} />
));
```

and test is:
```
test('with `id` prop', () => {
    const output = shallow(<Input id={id} />);

    expect(renderToJson(output)).toMatchSnapshot();
  });
``` 

throw error is: `Invariant Violation: ReactShallowRenderer render(): Shallow rendering works only with custom components, but the provided element type was 'object'.`

But if I use `render` method this error doesn't happened. 

When I use `mount` method with test is:
```
test('value state is changed', () => {
    const value = 'New value';
    const wrapper = shallow(<Input />);

    wrapper.setProps({ value });
    expect(wrapper.state().value).toEqual(value);
  });
```

I catch error is: `Enzyme Internal Error: unknown node with tag 14`

#### API

- [x] shallow
- [x] mount
- [ ] render

#### Version

| library         | version
| ---------------- | -------
| Enzyme           | 3.3.0
| React            |  16.3.0

#### Adapter

- [x] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] others ( )

suchipi:
Rebased version of https://github.com/airbnb/enzyme/pull/1592 with test added for the special proptype handling and failing tests fixed.

Fixes #1604.


packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -133,6 +135,17 @@ function toTree(vnode) {
     case ContextProviderType: // 13
     case ContextConsumerType: // 12
       return childrenToTree(node.child);
+    case ForwardRefType: {
+      return {
+        nodeType: 'function',
+        type: node.type,
+        props: { ...node.pendingProps },
```
suchipi: This feels fundamentally wrong, but it works? I'm not sure how React implements the fiber nodes for forward refs under the hood but it's possible memoizedProps is never equal to pendingProps...

packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
```diff
@@ -133,6 +134,17 @@ function toTree(vnode) {
     case ContextProviderType: // 13
     case ContextConsumerType: // 12
       return childrenToTree(node.child);
+    case ForwardRefType: {
```
suchipi: I added this logic to the 16.3 adapter in addition to the 16 adapter because React.forwardRef was added in 16.3.

ljharb:
Duplicate of #1592.

suchipi:
@ljharb should this PR be closed?

Alternatively, I can force-push the branch to contain nothing?

I don't understand what workflow you're optimizing around.

ljharb:
@suchipi no worries, i know it's probably not a common one.

Continue pushing to this one; I'll rebase and force push it so that both PRs match.

PR #1741, opened 2018-08-09
fix to run all tests in ShallowWrapper-spec.jsx

koba04:
Currently, the other tests except `merging props` don't run by `npm run test:only`.
I guess that `describe.only` intend to avoid failing by `should call componentWillReceiveProps for new renders` so I've skiped the test.

https://travis-ci.org/airbnb/enzyme/jobs/413839140

### DIFF
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -1330,7 +1330,7 @@ describe('shallow', () => {
       expect(wrapper.find('.bar')).to.have.lengthOf(1);
     });
 
-    describe.only('merging props', () => {
+    describe('merging props', () => {
       it('merges, not replaces, props when rerendering', () => {
         class Foo extends React.Component {
           render() {


packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1330,7 +1330,7 @@ describe('shallow', () => {
       expect(wrapper.find('.bar')).to.have.lengthOf(1);
     });
 
-    describe.only('merging props', () => {
```
ljharb: whoops, thanks :-D

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1468,7 +1468,7 @@ describe('shallow', () => {
       });
     });
 
-    it('should call componentWillReceiveProps for new renders', () => {
+    it.skip('should call componentWillReceiveProps for new renders', () => {
```
ljharb: this should be passing :-/ i'll look into it

ljharb:
I was able to unskip that test, and everything still passes.

PR #1759, opened 2018-08-16
Throw when mount/shallow rendering invalid root elements

jgzuke:
Throw `TypeError('ReactWrapper can only wrap valid elements')` when `ShallowWrapper` or `ReactWrapper` are created wrapping an invalid element as root.


packages/enzyme/src/ReactWrapper.js
```diff
@@ -85,8 +85,13 @@ class ReactWrapper {
     const options = makeOptions(passedOptions);
 
     if (!root) {
+      const adapter = getAdapter(options);
+      if (!adapter.isValidElement(nodes)) {
```
ljharb: hmm, `nodes` could be an array of valid elements; i'm not sure we can check it directly.

also, i think we probably want to check this on `root` too if that's provided?

jgzuke: I think if we try to `mount` an array of elements as the root that should still throw.

`root` elements should always have already been through this path when they were created, ie `wrap`/`ReactWrapper` constructor only get called with `null` when mounting (or diving in the shallow wrapper) or `this[ROOT]` when traversing, which is only set here.

jgzuke:
The tests added failed previously with 
```
AssertionError: expected [Function] to throw 'TypeError' but 'Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it\'s defined in, or you might have mixed up default and named imports.\n\nCheck the render method of `WrapperComponent`.' was thrown
+ expected - actual

-Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
-
-Check the render method of `WrapperComponent`.
+TypeError

at Context.<anonymous> (packages/enzyme-test-suite/test/ReactWrapper-spec.jsx:94:7)
```
for the mount tests and
```
AssertionError: expected [Function] to throw 'TypeError' but 'Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it\'s defined in, or you might have mixed up default and named imports.\n\nCheck the render method of `WrapperComponent`.' was thrown
+ expected - actual

-Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
-
-Check the render method of `WrapperComponent`.
+TypeError

at Context.<anonymous> (packages/enzyme-test-suite/test/ReactWrapper-spec.jsx:101:7)
```
for the shallow tests

PR #1760, opened 2018-08-16
Mounted portals show in debug and can be found

jgzuke:
React 16+ adapters `toTree` given a `HostPortal` returns a portal wrapping the child tree, instead of returning the child tree directly. This lets Portals show up in debug output and be found. Added the portal wrapper with a `nodeType` of `'portal'` and a `type` of react-is `Portal`.

### DIFF
--- a/packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
+++ b/packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
@@ -111,16 +111,19 @@ function toTree(vnode) {
   switch (node.tag) {
     case HostRoot: // 3
       return childrenToTree(node.child);
-    case HostPortal: // 4
+    case HostPortal: { // 4
+      const { stateNode: { containerInfo } } = node;
+      const props = { containerInfo };
       return {
         nodeType: 'portal',
         type: Portal,
-        props: {},
+        props,
         key: ensureKeyOrUndefined(node.key),
         ref: node.ref,
         instance: null,
         rendered: childrenToTree(node.child),
       };
+    }
     case ClassComponent:
       return {
         nodeType: 'class',
--- a/packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
+++ b/packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
@@ -111,9 +111,16 @@ function toTree(vnode) {
   switch (node.tag) {
     case HostRoot: // 3
       return childrenToTree(node.child);
-    case HostPortal: { // 4
-      return childrenToTree(node.child);
-    }
+    case HostPortal: // 4
+      return {
+        nodeType: 'portal',
+        type: Portal,
+        props: {},
+        key: ensureKeyOrUndefined(node.key),
+        ref: node.ref,
+        instance: null,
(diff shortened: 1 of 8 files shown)


packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
```diff
@@ -83,9 +83,16 @@ function toTree(vnode) {
   switch (node.tag) {
     case HostRoot: // 3
       return childrenToTree(node.child);
-    case HostPortal: { // 4
-      return childrenToTree(node.child);
-    }
+    case HostPortal: // 4
+      return {
+        nodeType: 'portal',
+        type: Portal,
```
jgzuke: Im not sure if it makes more sense to use `Portal` here, it may be better to just use `'Portal'`?

ljharb: typically `type` is either a string (an html element) or a function (component). I think using `the `Portal` symbol here is perfect.

packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
```diff
@@ -83,9 +83,16 @@ function toTree(vnode) {
...
+        props: {},
```
ljharb: I wonder if this shouldn't be `props: { children }` - ie, have the actual children provided to createPortal?

also, perhaps include the `containerInfo` as a "prop"?

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -94,6 +97,10 @@ export function displayNameOfNode(node) {
 
   if (!type) return null;
 
+  if (type === Portal) {
+    return 'Portal';
```
ljharb: i can't decide if it's better to add this here, or to require that adapters implement `displayNameOfNode` in order to support portals?

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -104,6 +111,9 @@ export function nodeTypeFromType(type) {
   if (type && type.prototype && type.prototype.isReactComponent) {
     return 'class';
   }
+  if (type && type === Portal) {
+    return 'portal';
```
ljharb: same question here

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -381,6 +384,48 @@ describeWithDOM('mount', () => {
...
+            <div className="in-portal">InPortal</div>,
+            containerDiv,
+          )}
+        </div>
+      );
+
+      const wrapper = mount(<Foo />);
+      expect(wrapper.debug()).to.equal(`<Foo>
+  <div>
+    <Portal>
```
ljharb: perhaps this could be `<Portal container={…}>` - with the container display being something like `containerDiv.outerHTML`?

jgzuke:
The test for finding Portals descendants is at https://github.com/airbnb/enzyme/blob/master/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx#L957 (since that behavior was working previously for `mount`) but I will move the new test closer.

PR #1761, opened 2018-08-16
Shallow renders though portal child

jgzuke:
`elementToTree` when given a `Portal` element returns a wrapper node and renders through children. Previously this would just return the element itself which would show in debug as `<Portal />` but not render through any children.


packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -168,9 +178,20 @@ export function ensureKeyOrUndefined(key) {
...
 
 export function elementToTree(el) {
-  if (el === null || typeof el !== 'object' || !('type' in el)) {
+  if (el === null || typeof el !== 'object' || !('type' in el || el.$$typeof === Portal)) {
     return el;
   }
+  if (el.$$typeof === Portal) {
+    return {
+      nodeType: nodeTypeFromType(Portal),
+      type: Portal,
```
jgzuke: Same comment here as https://github.com/airbnb/enzyme/pull/1760/files/f00cac8aab51269101a0b951d2bab3742409e01a#r210763428

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -94,6 +97,10 @@ export function displayNameOfNode(node) {
 
   if (!type) return null;
 
+  if (type === Portal) {
+    return 'Portal';
```
ljharb: Same question as on the other PR; should we add this here, or only on the adapters?

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -168,9 +178,20 @@ export function ensureKeyOrUndefined(key) {
 }
 
 export function elementToTree(el) {
-  if (el === null || typeof el !== 'object' || !('type' in el)) {
+  if (el === null || typeof el !== 'object' || !('type' in el || el.$$typeof === Portal)) {
```
ljharb: the `$$typeof` part should definitely not be here; that should be adapter-specific.

packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
```diff
@@ -72,6 +72,33 @@ function flatten(arr) {
...
+function nodeTypeFromType(type) {
+  if (type && type === Portal) {
+    return 'portal';
+  }
+
+  return utilNodeTypeFromType(type);
+}
+
+function elementToTree(el) {
+  if (!(el && el.$$typeof === Portal)) {
```
ljharb: i think this line can be `if (!isPortal(el)) {`?

packages/enzyme-adapter-react-16.2/src/ReactSixteenTwoAdapter.js
```diff
@@ -73,6 +73,33 @@ function flatten(arr) {
...
+function nodeTypeFromType(type) {
+  if (type && type === Portal) {
+    return 'portal';
+  }
+
+  return utilNodeTypeFromType(type);
+}
+
+function elementToTree(el) {
+  if (!(el && el.$$typeof === Portal)) {
```
ljharb: also isPortal here?

packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
```diff
@@ -78,6 +78,33 @@ function flatten(arr) {
...
+function nodeTypeFromType(type) {
+  if (type && type === Portal) {
+    return 'portal';
+  }
+
+  return utilNodeTypeFromType(type);
+}
+
+function elementToTree(el) {
+  if (!(el && el.$$typeof === Portal)) {
```
ljharb: isPortal

packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
```diff
@@ -78,6 +78,33 @@ function flatten(arr) {
   return result;
 }
 
+function nodeTypeFromType(type) {
+  if (type && type === Portal) {
```
ljharb: if type is falsy, it won't === Portal, so i think this can be `type === Portal ? 'portal' : utilNodeTypeFromType(type)`

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -180,7 +180,9 @@ export function ensureKeyOrUndefined(key) {
   return key || (key === '' ? '' : undefined);
 }
 
-export function elementToTree(el) {
+export function elementToTree(el, wrapperElementToTree = null) {
```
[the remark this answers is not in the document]
ljharb: good point. i guess this could be like `recurse = elementToTree`, and then use `recurse` below?

PR #1763, opened 2018-08-17
[Fix] `shallow`: `.setState()`: stub out `setState` on non-root code paths as well.

### LINKED ISSUE #1756: `this` is undefined in bound functions when shallow rendering
**Describe the bug**
The following test passes in 3.4.1 but fails in 3.4.2. 
```js
class Comp extends Component {
  state = {
    key: "",
  }

  componentDidMount() {
    this.instanceFunction();
  }

  instanceFunction = () => this.setState(() => ({ key: "value" }));

  render() {
    const { key } = this.state;
    return null;
  }
}

it("should work", () => {
  shallow(<Comp />);
});
```

The error in 3.4.2 is:
```shell
FAIL src/test.js
  ✕ should work (16ms)

  ● should work

    Method “setState” is only meant to be run on a single node. undefined found instead.

      11 |   }
      12 |
    > 13 |   instanceFunction = () => this.setState(() => ({ key: "value" }));
         |                                 ^
      14 |
      15 |   render() {
      16 |     const { key } = this.state;

      at ShallowWrapper.single (node_modules/enzyme/build/ShallowWrapper.js:1718:17)
      at ShallowWrapper.setState (node_modules/enzyme/build/ShallowWrapper.js:499:14)
      at Comp.ShallowWrapper.instance.setState (node_modules/enzyme/build/ShallowWrapper.js:184:33)
      at Comp.setState [as instanceFunction] (src/test.js:13:33)
      at Comp.instanceFunction [as componentDidMount] (src/test.js:10:10)
      at node_modules/enzyme/build/ShallowWrapper.js:189:20
      at Object.batchedUpdates (node_modules/enzyme-adapter-react-16/build/ReactSixteenAdapter.js:392:22)
      at new ShallowWrapper (node_modules/enzyme/build/ShallowWrapper.js:188:24)
      at shallow (node_modules/enzyme/build/shallow.js:21:10)
      at Object.<anonymous> (src/test.js:22:3)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        2.147s
Ran all test suites matching /src\/test.js/i.
error Command failed with exit code 1.
```

**To Reproduce**
1. Run the above test with `enzyme@3.4.2`

**Expected behavior**
The test should pass.

 - OS: Mac OSX (but also fails in Windows)
 - Browser n/a
 - Version n/a

ljharb:
Fixes #1756

cc @koba04

### DIFF
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -4967,6 +4967,33 @@ describe('shallow', () => {
         expect(wrapper.state('foo')).to.equal('onChange update');
         expect(spy).to.have.property('callCount', 1);
       });
+
+      it('should call `componentDidUpdate` when component’s `setState` is called', () => {
+        class Foo extends React.Component {
+          constructor(props) {
+            super(props);
+            this.state = {
+              foo: 'init',
+            };
+            this.update = () => this.setState({ foo: 'update' });
+          }
+
+          componentDidMount() {
+            this.update();
+          }
+
+          componentDidUpdate() {}
+
+          render() {
+            return <div>{this.state.foo}</div>;
+          }
+        }
+        const spy = sinon.spy(Foo.prototype, 'componentDidUpdate');
+
+        const wrapper = shallow(<Foo />);
+        expect(spy).to.have.property('callCount', 1);
+        expect(wrapper.state('foo')).to.equal('update');
+      });
     });
 
     describeIf(is('>= 16'), 'support getSnapshotBeforeUpdate', () => {
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -2142,7 +2142,7 @@ describe('shallow', () => {
       expect(wrapper.find('.bar')).to.have.lengthOf(1);
     });
 
-    it.skip('allows setState inside of componentDidMount', () => {
+    it('allows setState inside of componentDidMount', () => {
       class MySharona extends React.Component {
         constructor(props) {
           super(props);
@@ -2232,6 +2232,31 @@ describe('shallow', () => {
       const wrapper = shallow(<Foo />);
       expect(wrapper.state()).to.eql({ id: 'foo' });
       expect(() => wrapper.setState({ id: 'bar' }, 1)).to.throw(Error);
+    });
+
+    it('should preserve the receiver', () => {
+      class Comp extends React.Component {
+        constructor(...args) {
+          super(...args);
+
+          this.state = {
+            key: '',
+          };
+
+          this.instanceFunction = () => this.setState(() => ({ key: 'value' }));
+        }
+
+        componentDidMount() {
+          this.instanceFunction();
+        }
+
+        render() {
+          const { key } = this.state;
+          return key ? null : null;
+        }
+      }
+
+      expect(shallow(<Comp />).debug()).to.equal('');
     });
   });
 
--- a/packages/enzyme/src/ShallowWrapper.js
+++ b/packages/enzyme/src/ShallowWrapper.js
@@ -163,21 +163,19 @@ function privateSetNodes(wrapper, nodes) {
 class ShallowWrapper {
   constructor(nodes, root, passedOptions = {}) {
     validateOptions(passedOptions);
+
     const options = makeOptions(passedOptions);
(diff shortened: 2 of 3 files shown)


packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -2233,6 +2233,31 @@ describe('shallow', () => {
...
+          this.instanceFunction();
+        }
+
+        render() {
+          const { key } = this.state;
+          return key ? null : null;
+        }
+      }
+
+      expect(shallow(<Comp />).debug()).to.equal('');
```
koba04: Does `setState` in `componentDidMount` invoke `componentDidUpdate`?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -195,6 +193,15 @@ class ShallowWrapper {
       privateSetNodes(this, nodes);
     }
     privateSet(this, OPTIONS, root ? root[OPTIONS] : options);
+
+    if (!instance) {
+      ({ instance } = this[RENDERER].getNode());
+    }
+    // Ensure to call componentDidUpdate when instance.setState is called
+    if (instance && lifecycles.componentDidUpdate.onSetState && !instance[SET_STATE]) {
```
koba04: Nits: I should have overridden `setState` only in `options.disableLifecycleMethods` is false.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
ljharb: it should; cDU is invoked whenever it updates.

packages/enzyme/src/ShallowWrapper.js
ljharb: i'll add a test for that.

ljharb: actually, can you write one? :-D i'm not sure where to put it.

or is this just an optimization?

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
koba04: In order to be called cDU, we should stub `setState` so I think we should move calling cDM after the stubbing.

packages/enzyme/src/ShallowWrapper.js
koba04: https://github.com/airbnb/enzyme/pull/1763/commits/a05d4cb5dcd84dfd0ba3c8f074641b173e85242f is the right place. It is unnecessary when `options. disableLifecycleMethods` is `true`.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
koba04: I'll write the tests and the patch.
https://github.com/airbnb/enzyme/compare/fix_shallow_setstate...koba04:fix_shallow_setstate?expand=1

ljharb:
Thanks, this looks great. Will merge once tests pass.

monken:
Not sure why, but the 3.4.3 release broke my tests. I receive `Error: ShallowWrapper::setState() can only be called on the root` errors all over the place.

The relevant test is attached. I was able to figure out that `componentDidMount` is called twice with the 3.4.3 release, which was not the case before.

```js
  it('<MyComponent />', async () => {
    fetch.mockResponse(JSON.stringify(jsonData));
    const wrapper = shallow(<MyComponent />, { disableLifecycleMethods: true });
    await wrapper.instance().componentDidMount();
    wrapper.update();
    expect(fetch.mock.calls.length).toEqual(1);
    expect(wrapper.find(Table).length).toEqual(1);
  });
```

ljharb:
@monken please file a new issue; commenting on a merged PR isn't going to get anything fixed.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -2142,7 +2142,7 @@ describe('shallow', () => {
       expect(wrapper.find('.bar')).to.have.lengthOf(1);
     });
 
-    it.skip('allows setState inside of componentDidMount', () => {
+    it('allows setState inside of componentDidMount', () => {
```
talco: should not allow setState inside of componentDidMount

ljharb: Why would it not?

PR #1768, opened 2018-08-17
Fix to call unexpected cDM

### LINKED ISSUE #1765: 3.4.3 breaks tests with "ShallowWrapper::setState() can only be called on root"
**Describe the bug**
Not sure why, but the 3.4.3 release broke my tests. I receive Error: ShallowWrapper::setState() can only be called on the root errors all over the place. Seems to be related to #1756.

The relevant test is attached. I was able to figure out that componentDidMount is called twice with the 3.4.3 release, which was not the case before.

**To Reproduce**

```js
  it('<MyComponent />', async () => {
    fetch.mockResponse(JSON.stringify(jsonData));
    const wrapper = shallow(<MyComponent />, { disableLifecycleMethods: true });
    await wrapper.instance().componentDidMount();
    wrapper.update();
    expect(fetch.mock.calls.length).toEqual(1);
    expect(wrapper.find(Table).length).toEqual(1);
  });
```
The second call to `componentDidMount` has the following stack trace:

```
  console.error console.js:253
        at MyComponent.componentDidMount (/src/components/MyComponent.jsx:42:13)
        at /node_modules/enzyme/build/ShallowWrapper.js:204:20
        at Object.batchedUpdates (/node_modules/enzyme-adapter-react-16/build/ReactSixteenAdapter.js:342:22)
        at new ShallowWrapper (/node_modules/enzyme/build/ShallowWrapper.js:203:24)
        at ShallowWrapper.wrap (/node_modules/enzyme/build/ShallowWrapper.js:1763:16)
        at ShallowWrapper.find (/node_modules/enzyme/build/ShallowWrapper.js:815:21)
        at ShallowWrapper.exists (/node_modules/enzyme/build/ShallowWrapper.js:1711:52)
        at Object._callee$ (/src/components/tests/MyComponent.spec.jsx:38:20)
        at tryCatch (/node_modules/regenerator-runtime/runtime.js:62:40)
        at Generator.invoke [as _invoke] (/node_modules/regenerator-runtime/runtime.js:296:22)
```

### LINKED ISSUE #1766: componentDidMount is called any time `this.setState` is called
**Describe the bug**
`componentDidMount` is called on any call to `this.setState`. The bug was introduced in 3.4.2.

**To Reproduce**
Here's a sample component
```js
import React from 'react';

class TestComponent extends React.Component {
    state = {}
    componentDidMount() {
        this.props.onMount();
    }
    render() {
        return <button onClick={() => this.setState({ someState: 'some new state' })} />
    }
}

describe('componentDidMount', () => {
    it('will call componentDidMount on every call to setState', () => {
        const spy = jest.fn();
        const wrapper = shallow(<TestComponent onMount={spy} />);

        expect(spy).toHaveBeenCalledTimes(1);
        wrapper.find('button').simulate('click'); // updating state
        expect(spy).toHaveBeenCalledTimes(1);
    });
})
```

**Expected behavior**
I'd expect `componentDidMount` only to be called when I mount the component for the first time, or re-mount it explicitly. Instead, it calls it every time `this.setState` is called.

 - Linux
 - Chrome / not applicable
 - 3.4.2 and 3.4.3 and latest?

Relates to https://github.com/airbnb/enzyme/issues/1756

koba04:
Fixes #1766. Fixes #1765.

I missed that ShallowWrapper's constructor is called when finding child components.
In this case, the `renderer` is a `renderer` of the root, which means its `instance` is the instance of the root component. So `cDM` of the instance is called when creating a ShallowWrapper instance for child components.

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -5196,5 +5196,35 @@ describeWithDOM('mount', () => {
       expect(spy).to.have.property('callCount', 1);
       expect(wrapper.state('foo')).to.equal('update');
     });
+
+    it('should not call `componentDidMount` twice when a child component is created', () => {
+      class Foo extends React.Component {
+        constructor(props) {
+          super(props);
+          this.state = {
+            foo: 'init',
+          };
+        }
+
+        componentDidMount() {}
+
+        render() {
+          return (
+            <div>
+              <button onClick={() => this.setState({ foo: 'update2' })}>
+                click
+              </button>
+              {this.state.foo}
+            </div>
+          );
+        }
+      }
+      const spy = sinon.spy(Foo.prototype, 'componentDidMount');
+
+      const wrapper = mount(<Foo />);
+      expect(spy).to.have.property('callCount', 1);
+      wrapper.find('button').prop('onClick')();
+      expect(spy).to.have.property('callCount', 1);
+    });
   });
 });
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -4994,6 +4994,36 @@ describe('shallow', () => {
         expect(spy).to.have.property('callCount', 1);
         expect(wrapper.state('foo')).to.equal('update');
       });
+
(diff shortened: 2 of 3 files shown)


xitter:
Looking forward to getting this released soon.

koba04:
Thank you for the release!!
If there are any problems, please open a new issue!
(A test case to reproduce it is very helpful)

Issue #1769, opened 2018-08-18
Shallow renderer "ReactWrapper::setState() expects a function as its second argument" when setState given a callback within the class


rydash:
**Describe the bug**
Shallow rendering components fails when `setState(updater, callback)` is called within `componentDidUpdate` and `callback` refers to a function within the current component class, accessed via `this`.

**To Reproduce**
1. Create a component like this:
```js
import React from 'react';

class MyComponent extends React.Component {
  constructor() {
    super();
    this.state = {
      isLifecycle: false
    };
  }

  componentDidMount() {
    this.setState({ isLifecycle: true }, this.myCallback('Hello!'));
  }

  myCallback(message) {
    console.log(message);
  }

  render() {
    return <div></div>;
  }
}

export default MyComponent;
```

2. Set up a test:
```js
import React from 'react';
import { shallow } from 'enzyme';
import MyComponent from 'components/MyComponent';

describe('MyComponent', function() {
  it('renders', function() {
    const wrapper = shallow(<MyComponent />);
    expect(wrapper).to.be.ok;
  });
});
```

3. Run the test. (I did so using enzyme@3.4.4, enzyme-adapter-react-16@1.2.0, chai@4.1.2, and mocha@5.2.0. If other packages are relevant, I'm happy to describe them!)

4. Receive the following error:
```
0 passing (22ms)
1 failing

  1) MyComponent
       renders:
     TypeError: ReactWrapper::setState() expects a function as its second argument
      at ShallowWrapper.setState (node_modules/enzyme/build/ShallowWrapper.js:508:17)
      at MyComponent.ShallowWrapper.instance.setState (node_modules/enzyme/build/ShallowWrapper.js:190:35)
      at MyComponent.componentDidMount (src/js/components/MyComponent.js:12:10)
      at /Users/ryanmcgill/repos/fake/node_modules/enzyme/build/ShallowWrapper.js:196:22
      at Object.batchedUpdates (node_modules/enzyme-adapter-react-16/build/ReactSixteenAdapter.js:392:22)
      at new ShallowWrapper (node_modules/enzyme/build/ShallowWrapper.js:195:26)
      at shallow (node_modules/enzyme/build/shallow.js:21:10)
      at Context.<anonymous> (test/components/MyComponent.spec.js:7:21)
```

**Expected behavior**
The callback is called with the provided message, and the test succeeds.

**Desktop (please complete the following information):**
 - OS: macOS High Sierra 10.13.6
[...]

PR #1771, opened 2018-08-18
Fix setState regression

oliviertassinari:
I have noticed the following regression after upgrade to enzyme 3.4.4.

> TypeError: ReactWrapper::setState() expects a function as its second argument

I think that we are closer to how React behaves this way: https://github.com/facebook/react/blob/5031ebf6beddf88cac15f4d2c9e91f8dbb91d59d/packages/react-reconciler/src/ReactFiberClassComponent.js#L77-L92

### DIFF
--- a/packages/enzyme/src/ShallowWrapper.js
+++ b/packages/enzyme/src/ShallowWrapper.js
@@ -399,11 +399,11 @@ class ShallowWrapper {
    * @param {Function} cb - callback function
    * @returns {ShallowWrapper}
    */
-  setProps(props, callback = noop) {
+  setProps(props, callback) {
     if (this[ROOT] !== this) {
       throw new Error('ShallowWrapper::setProps() can only be called on the root');
     }
-    if (typeof callback !== 'function') {
+    if (callback && typeof callback !== 'function') {
       throw new TypeError('ShallowWrapper::setProps() expects a function as its second argument');
     }
     this.rerender(props);
@@ -424,14 +424,14 @@ class ShallowWrapper {
    * @param {Function} cb - callback function
    * @returns {ShallowWrapper}
    */
-  setState(state, callback = undefined) {
+  setState(state, callback) {
     if (this[ROOT] !== this) {
       throw new Error('ShallowWrapper::setState() can only be called on the root');
     }
     if (this.instance() === null || this[RENDERER].getNode().nodeType === 'function') {
       throw new Error('ShallowWrapper::setState() can only be called on class components');
     }
-    if (arguments.length > 1 && typeof callback !== 'function') {
+    if (callback && typeof callback !== 'function') {
       throw new TypeError('ReactWrapper::setState() expects a function as its second argument');
     }
     this.single('setState', () => {


oliviertassinari:
Well, my fix is probably wrong, but at least you get the idea. This is fine with React, I don't think that enzyme should throw.
```jsx
this.setState({}, null);
this.setState({}, undefined);
```

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -399,11 +399,11 @@ class ShallowWrapper {
    * @param {Function} cb - callback function
    * @returns {ShallowWrapper}
    */
-  setProps(props, callback = noop) {
+  setProps(props, callback) {
     if (this[ROOT] !== this) {
       throw new Error('ShallowWrapper::setProps() can only be called on the root');
     }
-    if (typeof callback !== 'function') {
+    if (callback && typeof callback !== 'function') {
```
ljharb: if you pass `undefined`, the current code will replace it with `noop` - this would allow you to silently pass a falsy primitive (false, NaN, 0, empty string, null). What's the utility of this?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -399,11 +399,11 @@ class ShallowWrapper {
    * @param {Function} cb - callback function
    * @returns {ShallowWrapper}
    */
-  setProps(props, callback = noop) {
+  setProps(props, callback) {
```
ljharb: either way, the default is required, so the `.length` of the function is correct.

oliviertassinari:
> enzyme is a test framework - its job is to catch bugs

How is it a bug if it's working in production?

> Do you have a legitimate use case for passing a falsy non-function to either setState or setProps?

I believe I have, the callback parameter is optional in my case, it can be undefined:
https://github.com/mui-org/material-ui/blob/a29c7f1a7ee885b70159856212206e99b8e86e56/packages/material-ui/src/ButtonBase/TouchRipple.js#L248

ljharb:
If you pass `undefined` with the current code, it will not throw - the default argument will replace it.

ljharb:
It's a bug because it's unintentional - that something *happens to work* doesn't make it not a bug.

If you're passing "not a callback", and "not nothing", where a callback is expected, you made a mistake, plain and simple.

packages/enzyme/src/ShallowWrapper.js
oliviertassinari: What about `null`? The default won't kick in. React check both:
https://github.com/facebook/react/blob/5031ebf6beddf88cac15f4d2c9e91f8dbb91d59d/packages/react-reconciler/src/ReactFiberClassComponent.js#L179

```js
function b(b = 'b') {
  return b
}

b() // 'b'
b(null) // null
b(undefined) // 'b'
b(false) // false
```

oliviertassinari:
> If you pass undefined with the current code, it will not throw

If you pass `undefined` with the current, it will throw.

packages/enzyme/src/ShallowWrapper.js
ljharb: You're right that `null` won't trigger the default; but i'm confused why you'd ever pass in a null in the first place.

ljharb:
Ah, you're right, I added that to setState (but not setProps).

That stricter check, however, has already caught a bug in someone's code - see #1769. I think that in this case the benefits outweigh the deviation from React in production. (Separately, I'm going to try to pursue adding the same strict check in react itself in v17)

packages/enzyme/src/ShallowWrapper.js
oliviertassinari: > i'm confused why you'd ever pass in a null in the first place.

Good question, I have no idea. I was curious to see how React would believe, turns out, they explicitly support it for some reason I'm not aware of.

ljharb:
I do see, however, that because of recent changes, the wrapper's `setState` will be called *in your component code* because we stub the instance's setState. That does seem like something we could fix, in the ShallowWrapper constructor - which would allow your code to work - but it would prevent people from being able to catch bugs like in #1769.

oliviertassinari:
> Separately, I'm going to try to pursue adding the same strict check in react itself in v17

@ljharb My main purpose with this pull request was to illustrate the change of behavior. I have already patched the Material-UI code to always have a callback function. Personally, I would stick to React behavior as much as possible. But I can understand you have a different vision. Do what you think is best.

packages/enzyme/src/ShallowWrapper.js
ljharb: I've asked for the motivation [here](https://github.com/facebook/react/commit/d3aeca1ff89e2ecedcd9be5ebe0a08336def5048#commitcomment-30185172)

ljharb:
Per the discussion; I'm going to address this by leaving the wrapper methods strict, but by making the shallow `setState` stub match React. This way, your code won't error, but people directly calling the methods on the wrapper will get the desired error.

oliviertassinari:
> leaving the wrapper methods strict, but by making the shallow setState stub match React.

@ljharb I haven't checked the implementation, but the behavior description sounds great :)!

PR #1774, opened 2018-08-19
Add children to mounted portal props

jgzuke:
Add `children` to mounted portal `props`. This brings the mount results in line with the shallow results and the other element types (which have `props.children`).

Also added the `shallow renders react portals` test. Compared to the `renders react portals` test above, `node.rendered.ref` comes back `undefined` instead of `null` which may be an issue with https://github.com/airbnb/enzyme/blob/master/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js#L103, if a portal can never have refs this might make sense to hard code as `null`. The inner divs `rendered` also comes back ` 'Hello World!'` instead of `['Hello World!']` which seems like the shallow renderer `Utils.elementToTree` is missing the check at https://github.com/airbnb/enzyme/blob/master/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js#L157.

### DIFF
--- a/packages/enzyme-test-suite/test/Adapter-spec.jsx
+++ b/packages/enzyme-test-suite/test/Adapter-spec.jsx
@@ -208,9 +208,10 @@ describe('Adapter', () => {
       const document = jsdom.jsdom();
       const options = { mode: 'mount' };
       const renderer = adapter.createRenderer(options);
+      const innerDiv = <div className="Foo">Hello World!</div>;
       const Foo = () => (
         createPortal(
-          <div className="Foo">Hello World!</div>,
+          innerDiv,
           document.body,
         )
       );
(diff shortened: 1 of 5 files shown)


packages/enzyme-test-suite/test/Adapter-spec.jsx
```diff
@@ -250,6 +254,56 @@ describe('Adapter', () => {
...
+        ref: null,
+        instance: null,
+        rendered: {
+          nodeType: 'portal',
+          type: Portal,
+          props: {
+            containerInfo: containerDiv,
+          },
+          key: undefined,
+          ref: undefined,
```
ljharb: based on your comment, it seems like this should be `null`, and not `undefined`. Thoughts?

jgzuke: Yep, does it make sense that portal `rendered.ref` should always be null? Im not sure what refs are here. I will follow up on the other issue (array wrapping `rendered` in shallow) as well but separately.

ljharb: Does ref ever return `undefined` for anything else? If not, then it should probably be null here.

Issue #1746, opened 2018-08-09
Shallow rendering: setState after setProps calls componentWillReceiveProps


maxkostow:
**Describe the bug**
On a shallow wrapper, calling `setState` after calling `setProps` triggers `componenetWillReceiveProps` which it should not.

I believe this has something to do with https://github.com/airbnb/enzyme/blob/9b4d0276f57e54be06aca6c3636120b3c4053310/packages/enzyme/src/ShallowWrapper.js#L361 which clones the props causing
https://github.com/facebook/react/blob/be4533af7d29f020499d3c01931e55ef8666a419/packages/react-test-renderer/src/ReactShallowRenderer.js#L183
to be `true` on the subsequent setState call
https://github.com/airbnb/enzyme/blob/9b4d0276f57e54be06aca6c3636120b3c4053310/packages/enzyme/src/ShallowWrapper.js#L441

**To Reproduce**
```
const React = require('react')
const Enzyme = require('enzyme') // 3.4.1
const Adapter = require('enzyme-adapter-react-16.3') // 1.0.0

Enzyme.configure({ adapter: new Adapter(), disableLifecycleMethods: true })

class A extends React.Component {
    constructor(props) {
        super(props)
        this.state = { a: 0 }
    }

    componentWillReceiveProps() {
        this.setState({ a: 1 })
    }

    render() {
        return React.createElement('div', {}, this.state.a)
    }
}

const e = Enzyme.shallow(React.createElement(A))

e.setState({ a: 2 })
console.log(e.state('a')) // logs `2`

e.setProps({})
console.log(e.state('a')) // logs `1`

e.setState({ a: 3 })
console.log(e.state('a')) // also logs `1` but should log `3`
```

**Expected behavior**
`componenetWillReceiveProps` should not be called after calling `setState`

ljharb:
Interesting, thanks for the report.

ljharb:
(Possibly related to #1742)

peanutenthusiast:
I'd be happy to submit a PR regarding this issue! Can probably submit in a week or two.

ljharb:
Fixed in #1779.

PR #1779, opened 2018-08-20
Shallow rendering, setState after setProps calls componentWillReceiveProps

peanutenthusiast:
Fix for #1746

### DIFF
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -2406,6 +2406,86 @@ describe('shallow', () => {
       });
     });
 
+    describe('should not call componentWillReceiveProps after setState is called', () => {
+      it('should not call componentWillReceiveProps upon rerender', () => {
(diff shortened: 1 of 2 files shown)


packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -2406,6 +2406,86 @@ describe('shallow', () => {
       });
     });
 
+    describeIf(is('>= 16.3'), 'should not call componentWillReceiveProps after setState is called', () => {
```
ljharb: i changed this from `is('16.3.0-0')` because we want it to apply to v16.4 and later also.

Can you elaborate, though, on why this wouldn't be true for all versions of react?

peanutenthusiast:
@ljharb Thanks for rebasing as well as the additional changes! I initially used `describeIf` as the test case specified a particular React version, but it's true that the patch does need to apply to all version of React.

I pulled down your changes, fixed `describeIf` to `describe`, and I've verified all tests pass with the new changes.

PR #1781, opened 2018-08-21
[Fix] `shallow`: `.parents`: ensure that one `.find` call does not affect another

### LINKED ISSUE #1780: parents: cannot read property 'reverse' of null
**Describe the bug**
Calling `parents` sequentially on two different sibling nodes breaks. I'm on:
- enzyme 3.4.4
- enzyme-adapter-react-16 1.2.0

```
TypeError: Cannot read property 'reverse' of null
    at parentsOfNode (http://localhost:9876/base/frontend/react-specs.js:41255:32)
    at ShallowWrapper.<anonymous> (http://localhost:9876/base/frontend/react-specs.js:62256:50)
    at ShallowWrapper.single (http://localhost:9876/base/frontend/react-specs.js:62811:25)
    at ShallowWrapper.parents (http://localhost:9876/base/frontend/react-specs.js:62255:41)
    at UserContext.<anonymous> (http://localhost:9876/base/frontend/react-specs.js:248367:18)
    at attempt (http://localhost:9876/base/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:4289:46)
    at QueueRunner.run (http://localhost:9876/base/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:4217:20)
    at QueueRunner.execute (http://localhost:9876/base/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:4199:10)
    at Spec.queueRunnerFactory (http://localhost:9876/base/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:909:35)
    at Spec.execute (http://localhost:9876/base/node_modules/jasmine-core/lib/jasmine-core/jasmine.js:526:10)
```

**To Reproduce**
The following test will reproduce the issue:

```jsx
it('breaks', () => {
  class Test extends React.Component {
    render() {
      return (
        <div>
          <div className="a">
            <div>A child</div>
          </div>
          <div className="b">
            <div>B child</div>
          </div>
        </div>
      );
    }
  }

  const wrapper = shallow(<Test />);

  const aChild = wrapper.find({ children: 'A child' });
  const bChild = wrapper.find({ children: 'B child' });

  aChild.parents({ className: 'a' });
  bChild.parents({ className: 'b' });
});
```

**Expected behavior**
It should not throw an error.

 - OS: macOS 10.13.6
 - Browser Chrome
 - Version 68.0.3440.106 (Official Build) (64-bit)

**Additional context**
In the above example, changing the order of accessing `parents` works:

```jsx
const aChild = wrapper.find({ children: 'A child' });
aChild.parents({ className: 'a' });

const bChild = wrapper.find({ children: 'B child' });
bChild.parents({ className: 'b' });
```

ljharb:
Fixes #1780.

These tests are failing; anyone who can help by posting a link to their branch would be very appreciated.

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -3893,6 +3893,55 @@ describeWithDOM('mount', () => {
       const formUp = input.parents('form');
       expect(formUp).to.have.lengthOf(1);
     });
+
+    it('works when called sequentially on two sibling nodes', () => {
+      class Test extends React.Component {
+        render() {
+          return (
+            <div>
+              <div className="a">
+                <div>A child</div>
+              </div>
+              <div className="b">
+                <div>B child</div>
+              </div>
+            </div>
+          );
+        }
+      }
+
+      const wrapper = mount(<Test />);
+
+      const aChild = wrapper.find({ children: 'A child' });
+      expect(aChild.debug()).to.equal(`<div>
+  A child
+</div>`);
+      expect(aChild).to.have.lengthOf(1);
+
+      const bChild = wrapper.find({ children: 'B child' });
+      expect(bChild.debug()).to.equal(`<div>
+  B child
+</div>`);
+      expect(bChild).to.have.lengthOf(1);
+
+      /*
+      const bChildParents = bChild.parents('.b');
+      expect(bChildParents.debug()).to.equal(`<div className="b">
+  <div>
+    B child
+  </div>
+</div>`);
+      expect(bChildParents).to.have.lengthOf(1);
+      */
+
+      const aChildParents = aChild.parents('.a');
+      expect(aChildParents.debug()).to.equal(`<div className="a">
+  <div>
+    A child
+  </div>
+</div>`);
+      expect(aChildParents).to.have.lengthOf(1);
+    });
   });
 
   describe('.parent()', () => {
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -3556,6 +3556,55 @@ describe('shallow', () => {
       expect(parents.at(0).hasClass('foo')).to.equal(true);
       expect(parents.at(1).hasClass('bax')).to.equal(true);
     });
+
+    it('works when called sequentially on two sibling nodes', () => {
+      class Test extends React.Component {
+        render() {
+          return (
+            <div>
+              <div className="a">
+                <div>A child</div>
+              </div>
+              <div className="b">
+                <div>B child</div>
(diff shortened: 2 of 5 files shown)


packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -3529,6 +3529,51 @@ describeWithDOM('mount', () => {
...
+
+      /*
+      const bChildParents = bChild.parents('.b');
+      expect(bChildParents.debug(`<div className="b">
+  <div>B child</div>
+</div>`));
+      expect(bChildParents).to.have.lengthOf(1);
+      */
+
+      const aChildParents = aChild.parents('.b');
```
jgzuke: @ljharb I believe this should this be `const aChildParents = aChild.parents('.a');`? (which fixes this test)

ljharb: thanks, i'll fix that

jgzuke:
It looks like this issue is coming from the changes in https://github.com/airbnb/enzyme/pull/1499, the state is changing when `getNodeInternal` calls `update`. If we remove the `update` calls from `getNode(s)Internal` these tests pass.

jgzuke:
@ljharb when `update` gets called we get a new (but identical) root tree. The problem is when we call `parents` and go to get the `pathToNode` from our node to its root, the `this[ROOT]` references a different tree than `this[NODES]` (`this[NODES]` references the old tree) so we never find a match and always return `null` from `pathToNode`.

Im not really sure what the correct fix is here. Im also not sure why we are updating on all `getNode(s)Internal` calls. But it seems like given that the a nodes parent can always be updated, we need some way to keep our non root `[NODES]` references up to date?

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -3556,6 +3556,51 @@ describe('shallow', () => {
...
+      /*
+      const bChildParents = bChild.parents('.b');
+      expect(bChildParents.debug(`<div className="b">
+  <div>B child</div>
+</div>`));
+      expect(bChildParents).to.have.lengthOf(1);
+      */
+
+      const aChildParents = aChild.parents('.a');
+      expect(aChildParents.debug(`<div className="a">
```
jgzuke: This should be
```
expect(aChildParents.debug()).to.equal(`<div className="a">
  <div>A child</div>
</div>`);
```
Same as above and in the `ReactWrapper-spec`

ljharb:
I think you're exactly right, and it's the reason I had to do the weird hierarchy stuff for simulateError's componentStack.

I think that `wrapper.update()` should update the non-root `NODES` list as well, such that `===` holds.

jgzuke:
@ljharb 👍 I will try to make that change

PR #1785, opened 2018-08-22
Fix to skip updates when nextState is null or undefined

### LINKED ISSUE #1783: Even when setState returns null, it calls cDU lifecycle.
**Describe the bug**
In React, if `setState` returns `null`, cDU is not called and so following does not make infinite loop even after cDU is called. (You can check it on [this codesandbox](https://codesandbox.io/s/18poplkr4q) by typing text on the input. It prints only one console log at a time.)

```js
export class TestComponent extends Component {
  constructor(props) {
    super(props);

    this.state = {};
  }

  componentDidUpdate() {
    this.setState(() => {
      return null;
    });
  }

  render() {
    return null;
  }
}
```

However, in `enzyme`, when I update component, i.e., invoking cDU with above component, will gives me infinite loop like

```
      at withSetStateAllowed (node_modules/enzyme/build/Utils.js:296:21)
      at ShallowWrapper.<anonymous> (node_modules/enzyme/build/ShallowWrapper.js:511:42)
      at ShallowWrapper.single (node_modules/enzyme/build/ShallowWrapper.js:1735:25)
      at ShallowWrapper.setState (node_modules/enzyme/build/ShallowWrapper.js:510:14)
      at TestComponent.ShallowWrapper.instance.setState (node_modules/enzyme/build/ShallowWrapper.js:190:35)
      at TestComponent.setState [as componentDidUpdate] (TestComponent.js:11:10)
      at node_modules/enzyme/build/ShallowWrapper.js:548:28
      at withSetStateAllowed (node_modules/enzyme/build/Utils.js:300:3)
      at ShallowWrapper.<anonymous> (node_modules/enzyme/build/ShallowWrapper.js:511:42)
      at ShallowWrapper.single (node_modules/enzyme/build/ShallowWrapper.js:1735:25)
      at ShallowWrapper.setState (node_modules/enzyme/build/ShallowWrapper.js:510:14)
      at TestComponent.ShallowWrapper.instance.setState (node_modules/enzyme/build/ShallowWrapper.js:190:35)
      at TestComponent.setState [as componentDidUpdate] (TestComponent.js:11:10)
      at node_modules/enzyme/build/ShallowWrapper.js:548:28
      // and repeat from line 1 to line 7
``` 

**To Reproduce**

1. `git clone https://github.com/Ailrun/enzyme-issue1783`
2. `npm i`
3. `npm test`

**Expected behavior**
Does not give infinite loop.

 - OS: Manjaro
 - Browser: is not used
 - Version: >=3.4.2, specifically, after https://github.com/airbnb/enzyme/pull/1742.

koba04:
Fix #1783 
This PR makes to prevent updates when setState returns `null` or `undefined`.
But the callback of setState's 2nd arguments should always be called as well as a behavior of ReactDOM.

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -2460,6 +2460,68 @@ describeWithDOM('mount', () => {
       });
     });
 
+    it('prevents the update if nextState is null or undefined', () => {
+      class Foo extends React.Component {
+        constructor(props) {
+          super(props);
+          this.state = { id: 'foo' };
+        }
+
+        componentDidUpdate() {}
+
+        render() {
+          return (
+            <div className={this.state.id} />
+          );
+        }
+      }
+
+      const wrapper = mount(<Foo />);
+      const spy = sinon.spy(wrapper.instance(), 'componentDidUpdate');
+      const callback = sinon.spy();
+      wrapper.setState(() => ({ id: 'bar' }), callback);
+      expect(spy).to.have.property('callCount', 1);
+      expect(callback).to.have.property('callCount', 1);
+
+      wrapper.setState(() => null, callback);
+      expect(spy).to.have.property('callCount', is('>= 16') ? 1 : 2);
+      expect(callback).to.have.property('callCount', 2);
+
+      wrapper.setState(() => undefined, callback);
+      expect(spy).to.have.property('callCount', is('>= 16') ? 1 : 3);
+      expect(callback).to.have.property('callCount', 3);
+    });
+
+    itIf(is('>= 16'), 'prevents an infinite loop if nextState is null or undefined from setState in CDU', () => {
+      class Foo extends React.Component {
+        constructor(props) {
+          super(props);
+          this.state = { id: 'foo' };
+        }
+
+        componentDidUpdate() {}
+
+        render() {
+          return (
+            <div className={this.state.id} />
+          );
+        }
+      }
+
+      let payload;
+      const stub = sinon.stub(Foo.prototype, 'componentDidUpdate')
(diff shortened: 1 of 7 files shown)


packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -2460,6 +2460,64 @@ describeWithDOM('mount', () => {
...
+
+        render() {
+          return (
+            <div className={this.state.id} />
+          );
+        }
+      }
+
+      const wrapper = mount(<Foo />);
+      const spy = sinon.spy(wrapper.instance(), 'componentDidUpdate');
```
ljharb: i think we can check both null and undefined here by spying on componentDidUpdate?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -446,6 +447,14 @@ class ShallowWrapper {
         const prevProps = instance.props;
         const prevState = instance.state;
         const prevContext = instance.context;
+
+        if (typeof state === 'function') {
+          state = state.call(instance, prevState, prevProps);
```
ljharb: i'll tweak this to avoid reassignment

koba04:
Travis is failed.
Ah, This behavior seems to be `>= 16`, so we need an adapter flag for this.

ljharb:
@koba04 meaning, prior to react 16, it's an infinite loop?

koba04:
@ljharb Yeah, prior to react 16, This seems to cause an infinite loop.

ljharb:
actually i changed my mind; this should be in the adapter, so that react versions < 16 *always* get the infinite loop, even shallow-rendering when setState returns nullish.

Ailrun:
I missed so many details in my issue... :disappointed: Thank you for noticing the version issue and fixing it too! You guys are really wonderfully awesome!

Issue #1784, opened 2018-08-21
setState getting into infinite loop


artola:
A component that fires `setState` at `componentDidMount` and `componentDidUpdate` enters in an infinite loop. Here in a simplified example that produces this situation.
Notice:
  - with versions 3.4.0 and 3.4.1 is OK.
  - since 3.4.2 it breaks.
  - commenting out 1 of the lifecycle calls, it pass.
  - same code with `mount` also breaks.

Interesting:
  - wrapping the 1 of the `setState` with a`setTimeout` WORKS.

  ```js
  componentDidUpdate() {
    setTimeout(() => {
      this.setState({value: false});
    });
  }
  ```

Example test:
```js
import React, {Component} from 'react';
import {shallow} from 'enzyme';

class Example extends Component {
  constructor(props: any) {
    super(props);

    this.state = {
      value: false,
    };
  }

  componentDidMount() {
    this.setState({value: false});
  }

  componentDidUpdate() {
    this.setState({value: false});
  }

  render() {
    return React.createElement('div', null, this.state.value);
  }
}

it('should not go crazy', () => {
  expect(() => {
    shallow(React.createElement(Example));
  }).not.toThrow();
});
```

With `"enzyme": "3.4.0"` and `"enzyme": "3.4.1"`:
```sh
 PASS  src/components/example.test.ts

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
```

With `"enzyme": "3.4.2"`:
```sh
 FAIL  src/components/example.test.ts
  ● should not go crazy

    expect(function).not.toThrow()

    Expected the function not to throw an error.
    Instead, it threw:
      Method “setState” is only meant to be run on a single node. undefined found instead.
          24 | }
          25 |
        > 26 | it('should not go crazy', () => {
             |          ^
          27 |   expect(() => {
          28 |     shallow(React.createElement(Example));
          29 |   }).not.toThrow();

          at ShallowWrapper.single (node_modules/enzyme/build/ShallowWrapper.js:1718:17)
          at ShallowWrapper.setState (node_modules/enzyme/build/ShallowWrapper.js:499:14)
          at Example.ShallowWrapper.instance.setState (node_modules/enzyme/build/ShallowWrapper.js:184:33)
          at Example.componentDidMount (src/components/example.test.ts:26:10)
          at node_modules/enzyme/build/ShallowWrapper.js:189:20
[...]

PR #1786, opened 2018-08-22
Fix to prevent updates with PureComponent

koba04:
Fix https://github.com/airbnb/enzyme/issues/1784#issuecomment-414902941

### DIFF
--- a/packages/enzyme/src/ShallowWrapper.js
+++ b/packages/enzyme/src/ShallowWrapper.js
@@ -158,6 +158,14 @@ function privateSetNodes(wrapper, nodes) {
   privateSet(wrapper, 'length', wrapper[NODES].length);
 }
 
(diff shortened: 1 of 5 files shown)


packages/enzyme/src/ShallowWrapper.js
```diff
@@ -339,6 +339,9 @@ class ShallowWrapper {
             && typeof instance.shouldComponentUpdate === 'function'
           ) {
             spy = spyMethod(instance, 'shouldComponentUpdate');
+          // PureComponent
+          } else if (instance.isPureReactComponent) {
+            shouldRender = !isEqual(prevProps, props) || !isEqual(state, instance.state);
```
ljharb: let's put this in some sort of `pureComponentShouldComponentUpdate` function, so here and line 471 can share the implementation.

PR #1790, opened 2018-08-24
Detect Fiber types to avoid inlining constants

gaearon:
This removes Enzyme's reliance on internal React reconciler constants which can change between releases (and *are* going to change in 16.4.3 which we plan to cut on Monday).

Instead, we're feature testing to detect the Fiber tags. The fields I'm relying on are relatively stable, and are expected to stay significantly more stable than the constants. The detection is done lazily (because we can't assume the DOM is available until you call `mount`). I've checked that other code paths don't use `FiberTags`.

This should work with all 16.x versions. I did some manual checking with 16.1, 16.2, 16.3, and the master React version (which would break without this). I haven't verified 16.0 but I'd expect it to work as well.

I included the revert of the reconciler dependency. This is because the code in it *does* rely on the constants (which we don't want). The inlined version here doesn't (because those constants are used only for invariants).

Longer term, we should be able to get rid both of this inlined piece, and of the constant detection. Instead we should share the infrastructure with React DevTools. I expect that we'll get to it in the coming months, and when we do, I'll send a PR for that. In the meantime I'm happy to help with bugfixing the React adapter if you discover any problems with my approach.

### DIFF
--- a/packages/enzyme-adapter-react-16/src/detectFiberTags.js
+++ b/packages/enzyme-adapter-react-16/src/detectFiberTags.js
@@ -1,0 +1,63 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+function getFiber(element) {
+  const container = global.document.createElement('div');
+  let inst = null;
+  class Tester extends React.Component {
+    render() {
+      inst = this;
+      return element;
+    }
+  }
+  ReactDOM.render(React.createElement(Tester), container);
+  return inst._reactInternalFiber.child;
+}
+
+module.exports = function detectFiberTags() {
+  const supportsMode = typeof React.StrictMode !== 'undefined';
+  const supportsContext = typeof React.createContext !== 'undefined';
+  const supportsForwardRef = typeof React.forwardRef !== 'undefined';
+
+  function Fn() {
+    return null;
+  }
+  // eslint-disable-next-line react/prefer-stateless-function
+  class Cls extends React.Component {
+    render() {
+      return null;
+    }
+  }
+  let Ctx = null;
+  let FwdRef = null;
+  if (supportsContext) {
+    Ctx = React.createContext();
+  }
+  if (supportsForwardRef) {
+    // React will warn if we don't have both arguments.
+    // eslint-disable-next-line no-unused-vars
+    FwdRef = React.forwardRef((props, ref) => null);
+  }
+
+  return {
+    HostRoot: getFiber('test').return.return.tag, // Go two levels above to find the root
+    ClassComponent: getFiber(React.createElement(Cls)).tag,
+    Fragment: getFiber([['nested']]).tag,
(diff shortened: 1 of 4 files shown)


packages/enzyme-adapter-react-16/src/detectFiberTags.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+  ReactDOM.render(React.createElement(Tester), container);
+  return inst._reactInternalFiber.child;
+}
+
+function detectFiberTags() {
+  const supportsMode = typeof React.StrictMode !== 'undefined';
+  const supportsContext = typeof React.createContext !== 'undefined';
+  const supportsForwardRef = typeof React.forwardRef !== 'undefined';
```
gaearon: These assume you're not going to use a newer `react` minor with an older `react-dom` minor. If that's not an assumption we're comfortable making, I can add some try-catches and fall back.

ljharb: I think that it's fine if things break with a mismatched react/react-dom - those have always had to match minor versions to work properly.

gaearon:
Above sounds correct.

What do you think about my concern re: mismatching React and ReactDOM versions? Do you aim to support this?

ljharb:
I [replied](https://github.com/airbnb/enzyme/pull/1790#discussion_r212701469):
> I think that it's fine if things break with a mismatched react/react-dom - those have always had to match minor versions to work properly.

Although it's ideal if react and react-dom only need to match within minors, i think it's reasonable to require that they match exactly. I think that could be helped, however, by making react-dom 16.4.3 peerDepend on `^16.4.3` of react - that would solve everything except "react 16.4.3, older react-dom", but i'm not sure there's a solution for that.

gaearon:
Sorry, GH didn't show me your comment at first.

>I think that could be helped, however, by making react-dom 16.4.3 peerDepend on ^16.4.3 of react - that would solve everything except "react 16.4.3, older react-dom", but i'm not sure there's a solution for that.

FWIW the only case I anticipate breaking is when `react` has a *larger minor* than `react-dom` (e.g. `react@16.4` with `react-dom@16.3`). Because our feature checks would pass but then the detection would fail. Let's see if this turns out to be a problem in practice.

ljharb:
In the past, I've seen many bugs caused by mismatched minors between react and react-dom; I think it is both an actual problem in practice, but also one that's easily addressed by telling people to manually keep those versions in sync. If enzyme helps surface that sooner than production, that's a feature not a bug imo :-D

gaearon:
Not sure I understand why CI is failing.

ljharb:
nbd, just a linter error in a tweak i made. I’ll fix.

aweary:
Thanks for doing this Dan! ❤️

PR #1793, opened 2018-08-27
Show functions and not <Profiler/> with react 16.3

lukeapage:
We haven't quite yet upgraded to react 16.4 and with 16.3 Profiler is not defined, meaning that any react nodes without a type (like a function) in shallow get serialized as `<Profiler/>`

Its not that important but I thought it could happen in the future, so I did a small refactor to not match undefined and undefined.

### DIFF
--- a/packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
+++ b/packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
@@ -417,26 +417,26 @@ class ReactSixteenThreeAdapter extends EnzymeAdapter {
   displayNameOfNode(node) {
     if (!node) return null;
     const { type, $$typeof } = node;
-    
+
     const nodeType = type || $$typeof;
 
     // newer node types may be undefined, so only test if the nodeType exists
     if (nodeType) {
       switch (nodeType) {
-        case AsyncMode: return 'AsyncMode';
-        case Fragment: return 'Fragment';
-        case StrictMode: return 'StrictMode';
-        case Profiler: return 'Profiler';
-        case Portal: return 'Portal';
+        case AsyncMode || NaN: return 'AsyncMode';
+        case Fragment || NaN: return 'Fragment';
+        case StrictMode || NaN: return 'StrictMode';
+        case Profiler || NaN: return 'Profiler';
+        case Portal || NaN: return 'Portal';
       }
     }
 
     const $$typeofType = type && type.$$typeof;
 
     switch ($$typeofType) {
-      case ContextConsumer: return 'ContextConsumer';
-      case ContextProvider: return 'ContextProvider';
-      case ForwardRef: {
+      case ContextConsumer || NaN: return 'ContextConsumer';
+      case ContextProvider || NaN: return 'ContextProvider';
+      case ForwardRef || NaN: {
         const name = type.render.displayName || functionName(type.render);
         return name ? `ForwardRef(${name})` : 'ForwardRef';
       }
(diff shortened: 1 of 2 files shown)


packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
```diff
@@ -417,27 +417,30 @@ class ReactSixteenThreeAdapter extends EnzymeAdapter {
...
     if (!node) return null;
     const { type, $$typeof } = node;
+    
+    const nodeType = type || $$typeof;
 
-    switch (type || $$typeof) {
-      case AsyncMode: return 'AsyncMode';
-      case Fragment: return 'Fragment';
-      case StrictMode: return 'StrictMode';
-      case Profiler: return 'Profiler';
```
ljharb: i wonder if a simpler change, since this is the 16.3 adapter, would be to just delete the Profiler line?

ljharb:
We'll also need the same change in the 16 adapter, since that still supports 16.0 - 16.4.

ljharb:
hmm, we have an explicit test that covers function children showing up as `[function]`. Can you share your test case?

lukeapage:
I'm a bit confused now.
1. The Profiler line is valid for 16.4, which I thought the 16.3 adapter supported?
2. I can do || NaN as a simpler change as per your commit, but then I may as well keep the code as-is - should I simplify it?

I don't have a (unit) test case at the moment but I can have a go at writing one once I understand what changes you want.

lukeapage:
I followed the instructions and I get errors running anything in enzyme

```
$ npm run env -- 16

> enzyme@0.0.1 env C:\git\enzyme
> babel-node ./env.js "16"

events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: spawn npm ENOENT
    at _errnoException (util.js:992:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:190:19)
    at onErrorNT (internal/child_process.js:372:16)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickDomainCallback (internal/process/next_tick.js:218:9)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! enzyme@0.0.1 env: `babel-node ./env.js "16"`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the enzyme@0.0.1 env script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

on node 8.11.4

ljharb:
@lukeapage it's entirely possible that our dev scripts don't work on windows; could you share the code itself, and i can try to repro it?

lukeapage:
I did add a unit test - it will fail because I don't know exactly what the function output is.

its a class that passes a function callback, shallow rendered and passed to jest's toMatchSnapshot
```
class B extends React.Component {
   render() {
       this.props.children();
   }
}
class A extends React.Component {
    render() {
       return <B>{() => <div/>}</B>;
    }
}

const wrapper = shallow(<A></A>);
expect(wrapper).toMatchSnapshot();
```
and its the snapshot serializer which uses the display name during serialization.

ljharb:
Thanks, I'll give that a shot.

ljharb:
What's the "snapshot serializer" here? enzyme doesn't have snapshot support - only `.debug()`, which i'd recommend instead - and it seems like your test case is passing just fine on master in React 16.3.

ljharb:
I'm going to merge this, since i still think these are good fixes - but I'm pretty confident your issue is that you're relying on non-enzyme snapshotting. If you stick with `.debug()` output, you should be fine.

PR #1797, opened 2018-08-28
[New] add `componentDidCatch` support, and `simulateError`

### LINKED ISSUE #1255: Enzyme 3 couldn't test component when ComponentDidCatch happend
My Component:
```javascript
class Error extends React.Component {
  render() {
    throw new Error('ErrorMessage');
    return <div />;
  }
}

class App extends React.Component {
  constructor() {
    super();
    this.state = {
      hasError: false,
      error: null,
      errorInfo: null
    }
  }
  componentDidCatch(error, info) {
    this.setState({
      hasError: true,
      error,
      errorInfo: info
    });
  }
  render() {
    const { hasError } = this.state;
    if (this.state.hasError) {
      return <div className="error-view">error view</div>
    }
    return <Error />
  }
}
```

My test: 
```javascript
const app = mount(<App />);
const errorTip = app.setState({
  hasError: true,
});
expect(app.find('.error-view').length).to.equal(1);
```

When `app` is mount，`app` throws the error and I couldn't get the `app` to test the error view. How do I solve it？

ljharb:
This PR adds a `.simulateError` API to shallow and mount wrappers, along with support for `componentDidCatch`.

Fixes #1255.


package.json
```diff
@@ -90,6 +90,7 @@
     "gitbook-plugin-github": "^2.0.0",
     "in-publish": "^2.0.0",
     "istanbul": "^1.0.0-alpha.2",
+    "istanbul-api": "~1.2",
```
jgzuke: Where is this being used?

ljharb: it's a dep of istanbul-api; see https://github.com/istanbuljs/istanbuljs/issues/216

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -4511,6 +4511,111 @@ describeWithDOM('mount', () => {
       });
     });
 
+    describeIf(is('>= 16'), 'componentDidCatch', () => {
+      describe('errors inside an error boundary', () => {
+        const errorToThrow = new EvalError('threw an error!');
+        // in React 16.0 - 16.2, and sole older nodes, the actual error thrown isn't reported.
```
calinoracation: I don't fully understand this comment: `in React 16.0 - 16.2, and sole older nodes`.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -4878,6 +4878,95 @@ describe('shallow', () => {
...
+          componentDidCatch(error, info) {
+            const { spy } = this.props;
+            spy(error, info);
+            this.setState({ throws: false });
+          }
+
+          render() {
+            const { throws } = this.state;
+            return (
+              <div>
```
calinoracation: Why the `span` nested in the `div` here?

packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
```diff
@@ -262,6 +263,19 @@ class ReactSixteenOneAdapter extends EnzymeAdapter {
       getNode() {
         return instance ? toTree(instance._reactInternalFiber).rendered : null;
       },
+      simulateError(nodeHierarchy, rootNode, error) {
+        const { instance: catchingInstance } = nodeHierarchy
+          .find(x => x.instance && x.instance.componentDidCatch) || {};
```
jgzuke: If I have multiple error boundary components in my tree, will this throw on the closest one to the current node?

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -222,3 +222,49 @@ export function propsWithKeysAndRef(node) {
...
+    throw error;
+  }
+
+  const componentStack = getComponentStack(
+    hierarchy,
+    getNodeType,
+    getDisplayName,
+  );
+
+  componentDidCatch.call(catchingInstance, error, { componentStack });
```
calinoracation: Why is `componentStack` an object here?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -4511,6 +4511,111 @@ describeWithDOM('mount', () => {
...
+          componentDidCatch(error, info) {
+            const { spy } = this.props;
+            spy(error, info);
+            this.setState({ throws: false });
+          }
+
+          render() {
+            const { throws } = this.state;
+            return (
+              <div>
```
calinoracation: Are the `div` and `span` here needed / helpful?

calinoracation: I think the more common pattern that we use would be along the lines of:

```jsx
return (
  <React.Fragment>
    <Thrower throws={throws} />
  </React.Fragment>
);
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -4878,6 +4878,95 @@ describe('shallow', () => {
...
+
+          expect(spy).to.have.property('callCount', 0);
+
+          expect(() => wrapper.find(Thrower).simulateError(errorToThrow)).not.to.throw();
+
+          expect(spy).to.have.property('callCount', 1);
+
+          expect(spy.args).to.be.an('array').and.have.lengthOf(1);
+          const [[actualError, info]] = spy.args;
+          expect(() => { throw actualError; }).to.throw(errorToThrow);
```
calinoracation: Any reason you use a different syntax to test between this one and the `mount` test, whereas here you rethrow the `actualError` but don't in the mount one?

packages/enzyme-adapter-utils/src/Utils.js
ljharb: React passes an “info” object with a componentStack property

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
ljharb: oops, typo for “some”

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
ljharb: True, but by adding them, i was able to find bugs in the hierarchy logic.

I’ll add another test for fragments.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
ljharb: To ensure that the hierarchy logic works properly :-)

packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
ljharb: It should throw on the first one it finds as it traverses upwards.

ljharb:
@calinoracation thanks for the reminder on the docs; I've added some.

docs/api/ReactWrapper/simulateError.md
```diff
@@ -0,0 +1,66 @@
...
+
+class ErrorBoundary extends React.Component {
+  componentDidCatch(error, info) {
+    const { spy } = this.props;
+    spy(error, info);
+  }
+
+  render() {
+    const { children } = this.props;
+    const { throws } = this.state;
```
a15n: unused

docs/api/ShallowWrapper/simulateError.md
```diff
@@ -0,0 +1,66 @@
...
+
+class ErrorBoundary extends React.Component {
+  componentDidCatch(error, info) {
+    const { spy } = this.props;
+    spy(error, info);
+  }
+
+  render() {
+    const { children } = this.props;
+    const { throws } = this.state;
```
a15n: unused

GreenGremlin:
This is great! I was just trying to write a test for an error boundary and this made it possible. Thank you!

packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
```diff
@@ -327,6 +342,16 @@ class ReactSixteenOneAdapter extends EnzymeAdapter {
             : elementToTree(output),
         };
       },
+      simulateError(nodeHierarchy, rootNode, error) {
+        simulateError(
```
GreenGremlin: Why not check the `nodeHierarchy` for error boundaries in a shallow renderer? Should this scenario require a mount renderer?

```js
// ErrorBoundary is a component with a componentDidCatch method that renders null
// if an error is encountered.
function MyComponent({children}) {
    return (
        <div>
            <h3>My component is cool!</h3>
            <ErrorBoundary>
                {children}
            </ErrorBoundary>
        </div>
    );
}

const BadChild = () => null;

const wrapper = shallow(
    <MyComponent>
        <BadChild />
    </MyComponent>
);
wrapper.find(BadChild).simulateError(new Error('That was bad'));

expect(wrapper.find('h3')).toExist();
```

ljharb: In a shallow render, there can't ever be any other error boundaries, since only the root node is actually rendered.

GreenGremlin: Sure, but if `simulateError` is called on an element in shallow render tree, it seems reasonable to render any error boundaries found above that element. This actually seems to work:

```js
const wrapper = shallow(
    <MyComponent>
        <BadChild />
    </MyComponent>
);
const errorBoundary = wrapper.find('ErrorBoundary').shallow();
errorBoundary.find(BadChild).simulateError(new Error('That was bad'));
expect(wrapper.find('h3')).toExist();
```

...but it doesn't really feel intuitive. I can see people trying my previous example and being confused when it doesn't work, especially since I did just that :\

ljharb: It's not reasonable or possible, because of how shallow rendering works. In shallow rendering, everything that's not the root node is treated as if it's a div - ie, as if the component implementation is `({ children }) => children`. Thus, there's no `componentDidCatch` anywhere, except the top.

PR #1802, opened 2018-09-03
[New] `mount`: `.state()`/`.setState()`: allow calling on children

### LINKED ISSUE #635: Support for accessing state of child components?
Maybe this is not supported for a reason.
But I find it convenient in many places for my testing.
Is there any way for me to see the state of child components?

### LINKED ISSUE #1289: ReactWrapper::setState() can only be called on the root
## Preface
I had went through #814 and #361 , but I still don't get the answer I want.  

## Explanation
I know you may say that in practice you should only test a child component in isolation, but I'm in a situation where the child component cannot be rendered without the parent component.   
For example, because I'm using [React Material-UI](http://www.material-ui.com/#/), I got to mount the element-to-be-tested inside a theme provider, if not error will be thrown during mounting : 
```jsx
const wrapper = mount(
        <MuiThemeProvider>
            <SubjectListView subjects={subjects}/>
        </MuiThemeProvider>);
```

## Problem
I need to set the state of `SubjectListView` but then I got this `ReactWrapper::setState() can only be called on the root` error. 

## Conclusion 
I really hope I can call `setState()` on child element, if not I have to wrap all my components with a `<MuiThemeProvider>` which is obviously not a good thing.

ljharb:
Fixes #635. Fixes #1289.

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -3045,6 +3045,63 @@ describeWithDOM('mount', () => {
 
       expect(mount(<Comp />).debug()).to.equal('<Comp />');
     });
+
+    describe('child components', () => {
+      class Child extends React.Component {
+        constructor(...args) {
+          super(...args);
+          this.state = { state: 'a' };
+        }
+
+        render() {
+          const { prop } = this.props;
+          const { state } = this.state;
+          return (
+            <div>
+              {prop} - {state}
+            </div>
+          );
+        }
+      }
+
+      class Parent extends React.Component {
+        constructor(...args) {
+          super(...args);
+          this.state = { childProp: 1 };
+        }
+
+        render() {
+          const { childProp } = this.state;
+          return <Child prop={childProp} />;
+        }
+      }
+
+      it('sets the state of the parent', () => {
+        const wrapper = mount(<Parent />);
+
+        expect(wrapper.text().trim()).to.eql('1 - a');
+
+        return new Promise((resolve) => {
+          wrapper.setState({ childProp: 2 }, () => {
+            expect(wrapper.text().trim()).to.eql('2 - a');
+            resolve();
+          });
+        });
+      });
+
+      it('sets the state of the child', () => {
+        const wrapper = mount(<Parent />);
+
+        expect(wrapper.text().trim()).to.eql('1 - a');
+
+        return new Promise((resolve) => {
+          wrapper.find(Child).setState({ state: 'b' }, () => {
+            expect(wrapper.text().trim()).to.eql('1 - b');
+            resolve();
+          });
+        });
+      });
+    });
   });
 
   describe('.is(selector)', () => {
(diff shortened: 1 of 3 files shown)


packages/enzyme/src/ReactWrapper.js
```diff
@@ -213,13 +213,13 @@ class ReactWrapper {
...
    *
-   * NOTE: can only be called on a wrapper instance that is also the root instance.
+   * NOTE: no matter what instance this is called on, it will always update the root.
    *
    * @returns {ReactWrapper}
    */
   update() {
     if (this[ROOT] !== this) {
-      throw new Error('ReactWrapper::update() can only be called on the root');
+      return this[ROOT].update();
```
ljharb: this is the part i want review on, in particular - is this the best way to solve it? should i instead keep the throw, but explicitly `this[ROOT].update()` in most of the places that calls this?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -2965,6 +2965,64 @@ describeWithDOM('mount', () => {
...
+
+    describe('child components', () => {
+      class Child extends React.Component {
+        constructor(...args) {
+          super(...args);
+          this.state = { a: 'a' };
+        }
+
+        render() {
+          const { _ } = this.props;
```
jgzuke: I find using `_` here makes this harder to read, it took me a while to realize this was getting passed through as `Child` props. I think I might be used to ignoring `_`, but maybe others might be the same 🤷‍♂️

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -2965,6 +2965,64 @@ describeWithDOM('mount', () => {
...
+          const { a } = this.state;
+          return (
+            <div>
+              {_}
```
jgzuke: This might also be easier to understand as `{_} - {a}`, `expect(wrapper.text().trim()).to.eql('1a');` wasn't super obvious what was going on

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -2657,6 +2657,59 @@ describe('shallow', () => {
...
+
+        return new Promise((resolve) => {
+          wrapper.setState({ _: 2 }, () => {
+            expect(wrapper.debug()).to.eql('<Child _={2} />');
+            resolve();
+          });
+        });
+      });
+
+      it('sets the state of the child', () => {
```
jgzuke: This message should be updated

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -3210,6 +3263,50 @@ describe('shallow', () => {
...
+        }
+      }
+
+      it('gets the state of the parent', () => {
+        const wrapper = shallow(<Parent />);
+
+        expect(wrapper.state()).to.eql({ _: 1 });
+      });
+
+      it('gets the state of the child', () => {
```
jgzuke: update description

packages/enzyme/src/ReactWrapper.js
jgzuke: Personally I think it makes sense to always update root. I dont see why anyone adding an `update` call that might happen in a non root node not wanting it not to update the root. We do already have a call to `this[ROOT].update()` in `simulate`, if we decide to change this we should update there.

It might be more clear setting `const root = ...` and operating on that instead of calling `this[ROOT].update()` though? When reading a function that calls itself I tend to assume it has n levels of recursion, instead of just one fall through, ie I would assume `this[ROOT]` might itself have a different root.

As an aside, it seems like a lot of things are guarded with `if (this[ROOT] !== this)`, would it make sense to make a `rootOnly(` similar to `single(`?

ljharb: That's a great idea; i'll look into doing that separately.

pgangwani:
WOW lovely.

PR #1811, opened 2018-09-08
Freeze ROOT_NODES for child wrappers

jgzuke:
Fixes the tests in https://github.com/airbnb/enzyme/pull/1781.

There are a few alternative ways to do fix those tests (and some things wrong with this one)

Solution 1 - This one! Save the original `ROOT` node tree, and use in calls to `parents` etc
- The saved root node tree will not actually match the `ROOT` node tree after root `update`. `update`s will affect the `ROOT` node tree but not the saved tree. This is already sort of the case, currently updates to the `ROOT` (for example setting state) would not affect `childWrapper[NODES]`, but would modify `childWrapper[ROOT][NODES]`. Given that we want child wrappers to be immutable I think this might make the most sense.
- It is possible here for something to still accidentally compare between the old and new tree by using `.root().getNodeInternal()` instead of `getRootNodeInternal()`

Solution 2 - On `update` we swap all child wrappers `ROOT`s to a copy of the old root.
- Calling and methods that update the parent would no longer work after an update. Ie
```js
const child = wrapper.find(Child);
wrapper.find(...) // Or anything that calls update on the wrapper
child.simulate(...) // Would affect the copy, not the real wrapper
```

Solution 3 - Make child wrappers not able to cause `update`s
- Breaking change, makes simulating child events and setting child state etc not possible

Solution 4 - Make child wrappers not immutable and update with component updates
- 🤷‍♂️ 

@ljharb

### DIFF
--- a/packages/enzyme/src/ReactWrapper.js
+++ b/packages/enzyme/src/ReactWrapper.js
@@ -31,6 +31,7 @@ const RENDERER = sym('__renderer__');
 const UNRENDERED = sym('__unrendered__');
 const ROOT = sym('__root__');
 const OPTIONS = sym('__options__');
+const ROOT_NODES = sym('__rootNodes__');
 
 /**
  * Finds all nodes in the current wrapper nodes' render trees that match the provided predicate
@@ -57,8 +58,18 @@ function filterWhereUnwrapped(wrapper, predicate) {
   return wrapper.wrap(wrapper.getNodesInternal().filter(predicate).filter(Boolean));
 }
 
+function getRootNodeInternal(wrapper) {
+  if (wrapper[ROOT].length !== 1) {
+    throw new Error('getRootNodeInternal(wrapper) can only be called when wrapper wraps one node');
+  }
+  if (wrapper[ROOT] !== wrapper) {
+    return wrapper[ROOT_NODES][0];
+  }
+  return wrapper[ROOT][NODE];
+}
+
 function nodeParents(wrapper, node) {
-  return parentsOfNode(node, wrapper[ROOT].getNodeInternal());
+  return parentsOfNode(node, getRootNodeInternal(wrapper));
 }
 
 function privateSetNodes(wrapper, nodes) {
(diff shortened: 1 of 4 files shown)


jgzuke:
@ljharb how to rerun tests if they seem flaky?

ljharb:
@jgzuke you should be able to click through to the job and rerun the specific builds that errored; i've just done it for this one.

packages/enzyme/src/ReactWrapper.js
```diff
@@ -57,8 +58,18 @@ function filterWhereUnwrapped(wrapper, predicate) {
...
+  }
+  if (wrapper[ROOT] !== wrapper) {
+    return wrapper[ROOT_NODES][0];
+  }
+  return wrapper[ROOT][NODE];
+}
+
 function nodeParents(wrapper, node) {
-  return parentsOfNode(node, wrapper[ROOT].getNodeInternal());
+  return parentsOfNode(node, getRootNodeInternal(wrapper));
```
ljharb: should this also be replaced in `simulateError`?

packages/enzyme/src/ReactWrapper.js
```diff
@@ -737,7 +749,7 @@ class ReactWrapper {
    */
   parents(selector) {
     return this.single('parents', (n) => {
-      const allParents = this.wrap(this.single('parents', n => nodeParents(this, n)));
```
ljharb: why is the `single` removed here?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -973,7 +982,7 @@ class ShallowWrapper {
    */
   parents(selector) {
     return this.single('parents', (n) => {
-      const allParents = this.wrap(this.single('parents', n => nodeParents(this, n)));
```
ljharb: and here?

packages/enzyme/src/ReactWrapper.js
jgzuke: This is already being wrapped on the line before

jgzuke: `Single` wrapped I mean

packages/enzyme/src/ShallowWrapper.js
jgzuke: same as above, `this.single` is already called in the line above

packages/enzyme/src/ReactWrapper.js
ljharb: oh, duh, good catch

Issue #1809, opened 2018-09-07
Component is not updated after simulating error


aliaksandr-yermalayeu:
**Describe the bug**
Component is not rendered correctly after [error simulation](https://airbnb.io/enzyme/docs/api/ReactWrapper/simulateError.html) 

Possibly related issues: [#1757](https://github.com/airbnb/enzyme/issues/1757), [#1283](https://github.com/airbnb/enzyme/issues/1283)

**To Reproduce**
I created minimal [repo ](https://github.com/aliaksandr-yermalayeu/enzyme_mount_simulate/) with contained issue. Clone it, run _npm i_, then run tests with _npm test_ command. One snapshot will fail. 

![2018-09-07 19_33_09-dimscreen screen](https://user-images.githubusercontent.com/14282348/45231631-44644400-b2d5-11e8-8c91-66c96e227521.png)

**Expected behavior**
Error message should be rendered instead of normal content. It works fine with **shallow** and fails with **mount**

**Desktop:**
 - OS: win7
 - Version 3.6.0

ljharb:
enzyme has no snapshot support; the closest would be `.debug()`.

Can you update your test cases to assert that either `Component` or `Error` have a length of 1, as you'd expect?

asapach:
@ljharb, could you please explain why it works with `shallow()` and doesn't work with `mount()`?

aliaksandr-yermalayeu:
> enzyme has no snapshot support; the closest would be .debug().

There was no difference after I replaced snapshots with .debug(), except formatting.

> Can you update your test cases to assert that either Component or Error have a length of 1, as you'd expect?

No problem. I updated repo, removed all unnecessary tests, added one similar test with pure error for comparison.

![2018-09-08 01_11_21-dimscreen screen](https://user-images.githubusercontent.com/14282348/45245234-3548ba80-b304-11e8-95a3-92f0e2963b84.png)

ljharb:
@asapach `mount` uses react itself to render; `shallow` uses react-test-renderer. It's possible/likely that there's bugs in react-test-renderer, or in enzyme's use of it.

aliaksandr-yermalayeu:
@jgzuke I suspected something like that. Thank you for the clarification.

ljharb:
Reopening until #1812 lands.

PR #1812, opened 2018-09-08
Add update after simulateError

jgzuke:
Fixes https://github.com/airbnb/enzyme/issues/1809

Note that
```js
it('rerenders on an error during render', () => {
  const wrapper = mount(<ErrorBoundary spy={sinon.stub()} />);

  expect(wrapper.find({ children: 'HasThrown' })).to.have.lengthOf(0);
  expect(wrapper.find({ children: 'HasNotThrown' })).to.have.lengthOf(1);

  return new Promise((resolve) => {
    wrapper.setState({ throws: true }, () => {
      expect(wrapper.find({ children: 'HasThrown' })).to.have.lengthOf(1);
      expect(wrapper.find({ children: 'HasNotThrown' })).to.have.lengthOf(0);
      resolve();
    });
  });
});
```
Will still fail, I will fix that separately.

@ljharb

### DIFF
--- a/packages/enzyme/src/ReactWrapper.js
+++ b/packages/enzyme/src/ReactWrapper.js
@@ -643,6 +643,7 @@ class ReactWrapper {
       const nodeHierarchy = [thisNode].concat(nodeParents(this, thisNode));
       renderer.simulateError(nodeHierarchy, rootNode, error);
 
+      this[ROOT].update();
       return this;
     });
   }
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -5033,13 +5033,19 @@ describeWithDOM('mount', () => {
         class ErrorBoundary extends React.Component {
           constructor(...args) {
             super(...args);
-            this.state = { throws: false };
+            this.state = {
+              throws: false,
+              didThrow: false,
+            };
           }
 
           componentDidCatch(error, info) {
             const { spy } = this.props;
             spy(error, info);
-            this.setState({ throws: false });
+            this.setState({
+              throws: false,
+              didThrow: true,
+            });
           }
 
           render() {
@@ -5049,6 +5055,9 @@ describeWithDOM('mount', () => {
                 <MaybeFragment>
                   <span>
                     <Thrower throws={throws} />
+                    <div>
+                      {this.state.didThrow ? 'HasThrown' : 'HasNotThrown'}
+                    </div>
                   </span>
                 </MaybeFragment>
               </div>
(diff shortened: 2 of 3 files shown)


packages/enzyme/src/ReactWrapper.js
```diff
@@ -643,6 +643,7 @@ class ReactWrapper {
       const nodeHierarchy = [thisNode].concat(nodeParents(this, thisNode));
       renderer.simulateError(nodeHierarchy, rootNode, error);
 
+      this[ROOT].update();
```
ljharb: Even when I comment out this line, the tests still pass :-/

jgzuke: Which react version are you using? I tested with 16.0 but will test with the others when I get back

ljharb: ah, duh, with 15 :-p let me try with 16

jgzuke: @ljharb this change does fix the test in the example repo, but will try to rewrite the test here so that it fails properly on later versions

jgzuke: Actually, running `npm run react:16.3 && npm run test:watch` fails for me when this line is commented

ljharb: If you rebase on master, you can run `npm run react 16.2` etc, and tests will also run against more react versions.

jgzuke: Rebased, on react 16.{2-5} the added tests fail before the change and pass after.

ljharb:
@jgzuke the tests pass before your fix; see https://travis-ci.org/airbnb/enzyme/builds/427439455 vs https://travis-ci.org/airbnb/enzyme/builds/427439589 :-/

jgzuke:
@ljharb Hmm... When i check out https://github.com/airbnb/enzyme/pull/1812/commits/3113da372166f389dbc5daa66fdf89dca313a239 it fails for me, im not sure how to start debugging this

jgzuke:
@ljharb I reverted the change and just left the tests, resulting in https://travis-ci.org/airbnb/enzyme/builds/427510199 (which is the expected failures). Im not really sure what the problem here might be

jgzuke:
Are there any additional changes to be made here?

ljharb:
@jgzuke this build: https://travis-ci.org/airbnb/enzyme/builds/429796257 still didn't fail (https://travis-ci.org/airbnb/enzyme/builds/429796341 is the one with the fix) :-/

basically, i want to see a failing build, that the fix fixes, to have confidence to merge this.

jgzuke:
Im a bit confused why https://travis-ci.org/airbnb/enzyme/builds/427510199 fails but https://travis-ci.org/airbnb/enzyme/builds/429796257 passes given they contain the same changes, are you sure Jenkins is actually running on the displayed commit and not on the end of that branch

ljharb:
hmm, i'm pretty sure. i'll try pushing just the first commit, and i won't push the second til the test is done.

ljharb:
guess that was it! i'll merge this shortly.

PR #1817, opened 2018-09-11
[enzyme-adapter-react-{16,16.3}] [fix] `forwardRef`: respect `.displayName`

### LINKED ISSUE #1810: displayName set on forwardRef is not honored
**Describe the bug**
In styled-components v4 beta we started decorating the forwardRef wrapper with displayName and various other statics to reduce the amount of interim component layers during rendering and preserve various library behaviors. https://github.com/styled-components/styled-components/blob/30c372cf635e0db77f74c0aaacf60deda729581c/src/models/StyledComponent.js#L249-L311

However, it doesn't seem like enzyme surfaces a `displayName` set on a forwardRef component.

**To Reproduce**
See this codesandbox: https://codesandbox.io/s/yv7l2p76q9

**Expected behavior**
The test expectations should work since name and displayName are set on the forwardRef node.

It would be really great if this could work for a few reasons:

1. Shallow rendering does not currently render the immediate child of ForwardRef, so if we modified our library so the wrapped child of ForwardRef had all the statics etc it still wouldn't work without diving

2. The components don't show up at all when using "find" off of "mount" since ForwardRef isn't honoring displayName

Sidebar: I really wish ref forwarding was implemented as a decorator or class mixin... _sigh_.

ljharb:
Fixes #1810.

### DIFF
--- a/packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
+++ b/packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
@@ -461,6 +461,9 @@ class ReactSixteenThreeAdapter extends EnzymeAdapter {
       case ContextConsumer || NaN: return 'ContextConsumer';
       case ContextProvider || NaN: return 'ContextProvider';
       case ForwardRef || NaN: {
+        if (type.displayName) {
+          return type.displayName;
+        }
         const name = type.render.displayName || functionName(type.render);
         return name ? `ForwardRef(${name})` : 'ForwardRef';
       }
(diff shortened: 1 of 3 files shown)


ljharb:
@jgzuke because the 16 adapter is for 16.0 - 16.5, but the 16.3 adapter is only for 16.3, and those special cases are only in 16.3+.

ljharb:
I don't think it's really worth it; the original issue's repro requires styled-components, but in general, if it shows up in `.debug()`, it's findable.

probablyup:
Thanks Jordan!

darkowic:
Hello there! Will it be released anytime soon?

PR #1819, opened 2018-09-14
Add getDOMNode() into the Table of Contents

mikeyhu:


ljharb:
Only mount has getDOMNode.

PR #1832, opened 2018-09-21
[fix] `shallow`: A nested ForwardRef is not recognized as a valid component by `dive()`

### LINKED ISSUE #1830: A nested ForwardRef is not recognized as a valid component by shallow dive()
**Describe the bug**
A nested ForwardRef is not recognized as a valid component when I do shallow dive() to it.

**To Reproduce**
1. Sample code:
```javascript
  import * as React from "react";

  const Component = () => <div/>
  const HOCOne = (WrappedComponent) => React.forwardRef(() => <WrappedComponent/>)
  const HOCTwo = (WrappedComponent) => React.forwardRef(() => <WrappedComponent/>)
  const WithNestedForwarRefComponent = HOCTwo(HOCOne(Component)))
  
  const vDomWrapper = shallow(<WithNestedForwarRefComponent/>);
  
```
2. Then debug() and dive() vDomWrapper:
```javascript
  vDomWrapper.debug()
  vDomWrapper.dive()
```
3. Debug output:
`<ForwardRef(HOCOne(Component)) />`
4. See dive() error:
`TypeError: ShallowWrapper::dive() can only be called on components`

**Expected behavior**
A ForwardRef component must be recognized as such and it must be possible to do dive () on it.

**Desktop:**
 - OS: Ubuntu 18.04
 - Node: 8.11.3
 - Enzyme: 3.6.0
 - React: 16.5.0

SinHouse:
Fixes #1830

Other: 
- Added Utils.isCustomComponentElement tests
- Pass test:all
- Pass lint

### DIFF
--- a/packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
+++ b/packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
@@ -11,6 +11,7 @@ import TestUtils from 'react-dom/test-utils';
 import {
   isElement,
   isPortal,
+  isForwardRef,
   isValidElementType,
   AsyncMode,
   Fragment,
@@ -481,6 +482,13 @@ class ReactSixteenThreeAdapter extends EnzymeAdapter {
 
   isFragment(fragment) {
     return typeOfNode(fragment) === Fragment;
+  }
+
+  isCustomComponentElement(inst) {
+    if (!inst || !this.isValidElement(inst)) {
+      return false;
+    }
+    return typeof inst.type === 'function' || isForwardRef(inst);
   }
 
   createElement(...args) {
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -11,6 +11,7 @@ import TestUtils from 'react-dom/test-utils';
 import {
   isElement,
   isPortal,
+  isForwardRef,
   isValidElementType,
   AsyncMode,
   Fragment,
@@ -523,6 +524,13 @@ class ReactSixteenAdapter extends EnzymeAdapter {
 
   isFragment(fragment) {
     return typeOfNode(fragment) === Fragment;
+  }
+
+  isCustomComponentElement(inst) {
+    if (!inst || !this.isValidElement(inst)) {
+      return false;
+    }
+    return typeof inst.type === 'function' || isForwardRef(inst);
   }
 
   createElement(...args) {
--- a/packages/enzyme/src/Utils.js
+++ b/packages/enzyme/src/Utils.js
@@ -42,6 +42,9 @@ export function makeOptions(options) {
 }
 
 export function isCustomComponentElement(inst, adapter) {
+  if (adapter.isCustomComponentElement) {
+    return !!adapter.isCustomComponentElement(inst);
+  }
   return !!inst && adapter.isValidElement(inst) && typeof inst.type === 'function';
 }
 
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -6923,6 +6923,33 @@ describe('shallow', () => {
       expect(underwater.is(RendersDOM)).to.equal(true);
     });
 
+    describeIf(is('>=16.3.0'), 'forwardRef Elements', () => {
+      const ForwardRefWrapsRendersDOM = forwardRef && forwardRef(() => <WrapsRendersDOM />);
+      const NestedForwarRefsWrapsRendersDom = forwardRef
+        && forwardRef(() => <ForwardRefWrapsRendersDOM />);
+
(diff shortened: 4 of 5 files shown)


packages/enzyme/src/Utils.js
```diff
@@ -42,7 +42,7 @@ export function makeOptions(options) {
 }
 
 export function isCustomComponentElement(inst, adapter) {
-  return !!inst && adapter.isValidElement(inst) && typeof inst.type === 'function';
```
ljharb: Please add tests that ensure `isCustomComponentElement` returns `false` for HTML elements.

SinHouse: Hi @ljharb,

I thing the changes are ready but fails the Travis job that build react 16.5 with node 8.

My develop machine use node 8.11.3 with Ubuntu 18.04 and I pass test:all and the lint tests.

Is it possible that it' s a specific problem with the Travis environment or is the problem in my code?

Thanks,

ljharb: Since it errored, instead of failed, it's an issue with travis.

packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
```diff
@@ -483,6 +484,10 @@ class ReactSixteenThreeAdapter extends EnzymeAdapter {
     return typeOfNode(fragment) === Fragment;
   }
 
+  isForwardRef(forwardRef) {
```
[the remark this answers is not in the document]
SinHouse: I agree that isCustomComponentElement can be an adapter competence because its implementation changes between React versions.

The problem is that I have to identify a forwardRef in a robust way and therefore I need to access the isForwardRef dependency of the react-is package available in the current dependencies of the adapter.

A forwardRef component is an object and not a function like other custom components, this is why isCustomComponentElement fails with them.

The current solution just add changes to 16 and 16.3 adapters, it's a similar solution of the current isFragment adapter method present only in the React 16 adapters.

The proposed solution to expand the current interface adapter to include the isCustomComponentElement method add changes to all the adapters but I agree can be a better solution.

Right now I can' t think of a cleaner or more robust solution than the previous ones, I'm sorry.

I wait for you to know your opinion about it.

SinHouse:
Hi @ljharb,

Utils.isCustomComponentElement seems used only by ShallowWrapper.dive()

Is it a good idea remove isCustomComponentElement from Utils and call adapter.isCustomComponentElement directly?

SinHouse:
Hi @ljharb,

Please, can you update the pull request state? Thanks a lot,

packages/enzyme-adapter-react-13/src/ReactThirteenAdapter.js
```diff
@@ -271,6 +271,10 @@ class ReactThirteenAdapter extends EnzymeAdapter {
     return typeof object === 'string' || typeof object === 'function';
   }
 
+  isCustomComponentElement(inst) {
+    return !!inst && this.isValidElement(inst) && typeof inst.type === 'function';
```
ljharb: for now, let's have every adapter where it matches, directly use the implementation from utils

packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
```diff
@@ -483,6 +484,11 @@ class ReactSixteenThreeAdapter extends EnzymeAdapter {
     return typeOfNode(fragment) === Fragment;
   }
 
+  isCustomComponentElement(inst) {
+    const isFunctionOrForwardRef = () => typeof inst.type === 'function' || isForwardRef(inst);
```
ljharb: i think these can be inlined; no need to create a function for it

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -485,6 +486,11 @@ class ReactSixteenAdapter extends EnzymeAdapter {
     return typeOfNode(fragment) === Fragment;
   }
 
+  isCustomComponentElement(inst) {
+    const isFunctionOrForwardRef = () => typeof inst.type === 'function' || isForwardRef(inst);
```
ljharb: same here

packages/enzyme/src/Utils.js
```diff
@@ -42,7 +42,7 @@ export function makeOptions(options) {
...
+  return adapter.isCustomComponentElement(inst);
```
ljharb: this needs a fallback for the case where an adapter lacks `isCustomComponentElement`.

packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
SinHouse: done, the result is a little more confusing but it's faster :+1:

packages/enzyme/src/Utils.js
SinHouse: done,

SinHouse:
Thanks to you @ljharb

PR #1856, opened 2018-10-08
Invoke() api

koba04:
This PR is to pick up from #945
If @milesj is still working on #945, feel free to close this!

PR #1863, opened 2018-10-11
Add renderProp to ShallowWrapper

### LINKED ISSUE #1444: Feature request: wrap JSX without any rendering
Sometimes I find that I need to run assertions on raw `ReactElement`s, without needing to shallow render or mount. This is especially common with the hip new "render prop" pattern. Here's an example based on React Router:

```js
// Using Route's render prop to inject the `match`, as seen in this example:
// https://reacttraining.com/react-router/web/example/custom-link
const MyCustomLink = (children, to) => (
  <Route
    render={({ match }) => (
      <ObnoxiousHighlight on={Boolean(match)}>
        <Link to={to}>{children}</Link>
      </ObnoxiousHighlight>
    )}
  />
);

describe('MyCustomLink', () => {
  it('renders', () => {
    const outerWrapper = shallow(<MyCustomLink to="/home">Home</MyCustomLink>);
    const route = outerWrapper.find(Route);
    const innerJsx = route.prop('render')({ match: { __testMatch: true } });
    const innerWrapper = __someNewEnzymeFunction(innerJsx);
    expect(innerWrapper.find(ObnoxiousHighlight)).toHaveProp('on', true);
    expect(innerWrapper.find(Link)).toHaveProp('to', '/home');
  });
});
```

Note that if I called `shallow(innerJsx)`, it would be treated as a shallow render of `ObnoxiousHighlight`, which isn't what I'm interested in testing in this suite.

As a workaround, I've been using this dirty function:
```js
const wrap = element => shallow(<div>{element}</div>).childAt(0);
```

But it would be nice to have it as part of Enzyme.

dferber90:
Adds a `renderProp` method to `ShallowWrapper`s to ease testing render props.

This PR basically extracts `renderProp`  out of [@commercetools/enzyme-extensions](https://github.com/commercetools/enzyme-extensions). We would deprecate and drop it there if `renderProp` makes it into `enzyme` itself.

I wrote [this article](https://medium.com/@dferber90/test-a-render-prop-6a44e02f4c39) a while ago detailing how to use `renderProp` and why.

I'm curious for feedback of whether you'd want to accept it or not after some interest was shown in https://github.com/commercetools/enzyme-extensions/issues/8.

---

Thanks to @tdeekens who helped set up and bounce ideas back&forth with for `renderProp` in `enzyme-extensions`!

Closes #1444. Related to #1856.

### DIFF
--- a/docs/api/ReactWrapper/renderProp.md
+++ b/docs/api/ReactWrapper/renderProp.md
@@ -1,0 +1,85 @@
+# `.renderProp(propName, ...args) => ReactWrapper`
+
+Calls the current wrapper's property with name `propName` and the `args` provided.
+Returns the result in a new wrapper.
+
+NOTE: can only be called on wrapper of a single non-DOM component element node.
+
+#### Arguments
+
+1.  `propName` (`String`):
+1.  `...args` (`Array<Any>`):
+
+This essentially calls `wrapper.prop(propName)(...args)`.
+
+#### Returns
+
+`ReactWrapper`: A new wrapper that wraps the node returned from the render prop.
+
+#### Examples
+
+##### Test Setup
+
+```jsx
+class Mouse extends React.Component {
+  constructor() {
+    super();
+    this.state = { x: 0, y: 0 };
+  }
+
+  render() {
+    const { render } = this.props;
+    return (
+      <div
+        style={{ height: '100%' }}
+        onMouseMove={(event) => {
+          this.setState({
+            x: event.clientX,
+            y: event.clientY,
+          });
+        }}
+      >
+        {render(this.state)}
+      </div>
+    );
+  }
+}
+
+Mouse.propTypes = {
+  render: PropTypes.func.isRequired,
+};
+```
+
+```jsx
+const App = () => (
+  <div style={{ height: '100%' }}>
+    <Mouse
+      render={(x = 0, y = 0) => (
+        <h1>
+          The mouse position is ({x}, {y})
+        </h1>
+      )}
+    />
+  </div>
+);
+```
+
+##### Testing with no arguments
+
+```jsx
+const wrapper = mount(<App />)
+  .find(Mouse)
+  .renderProp('render');
+
+expect(wrapper.equals(<h1>The mouse position is 0, 0</h1>)).to.equal(true);
+```
+
+##### Testing with multiple arguments
+
+```jsx
+const wrapper = mount(<App />)
+  .find(Mouse)
+  .renderProp('render', [10, 20]);
+
+expect(wrapper.equals(<h1>The mouse position is 10, 20</h1>)).to.equal(true);
+```
--- a/docs/api/ShallowWrapper/renderProp.md
+++ b/docs/api/ShallowWrapper/renderProp.md
@@ -1,0 +1,85 @@
+# `.renderProp(propName, ...args) => ShallowWrapper`
+
+Calls the current wrapper's property with name `propName` and the `args` provided.
+Returns the result in a new wrapper.
+
+NOTE: can only be called on wrapper of a single non-DOM component element node.
+
+#### Arguments
+
+1.  `propName` (`String`):
+1.  `...args` (`Array<Any>`):
+
+This essentially calls `wrapper.prop(propName)(...args)`.
+
+#### Returns
+
+`ShallowWrapper`: A new wrapper that wraps the node returned from the render prop.
+
+#### Examples
+
+##### Test Setup
+
+```jsx
+class Mouse extends React.Component {
+  constructor() {
+    super();
+    this.state = { x: 0, y: 0 };
+  }
+
+  render() {
+    const { render } = this.props;
+    return (
+      <div
+        style={{ height: '100%' }}
+        onMouseMove={(event) => {
+          this.setState({
+            x: event.clientX,
+            y: event.clientY,
+          });
+        }}
+      >
+        {render(this.state)}
+      </div>
+    );
+  }
+}
+
+Mouse.propTypes = {
+  render: PropTypes.func.isRequired,
+};
+```
+
+```jsx
+const App = () => (
+  <div style={{ height: '100%' }}>
+    <Mouse
+      render={(x = 0, y = 0) => (
+        <h1>
+          The mouse position is ({x}, {y})
+        </h1>
+      )}
+    />
+  </div>
+);
+```
+
+##### Testing with no arguments
+
+```jsx
+const wrapper = shallow(<App />)
+  .find(Mouse)
+  .renderProp('render');
+
+expect(wrapper.equals(<h1>The mouse position is 0, 0</h1>)).to.equal(true);
+```
+
+##### Testing with multiple arguments
+
+```jsx
+const wrapper = shallow(<App />)
+  .find(Mouse)
+  .renderProp('render', [10, 20]);
+
+expect(wrapper.equals(<h1>The mouse position is 10, 20</h1>)).to.equal(true);
+```
--- a/packages/enzyme-adapter-react-13/src/ReactThirteenAdapter.js
+++ b/packages/enzyme-adapter-react-13/src/ReactThirteenAdapter.js
@@ -14,6 +14,7 @@ import {
   createMountWrapper,
   propsWithKeysAndRef,
   ensureKeyOrUndefined,
+  wrap,
 } from 'enzyme-adapter-utils';
 import mapNativeEventNames from './ReactThirteenMapNativeEventNames';
 import elementToTree from './ReactThirteenElementToTree';
@@ -240,6 +241,10 @@ class ReactThirteenAdapter extends EnzymeAdapter {
       default:
         throw new Error(`Enzyme Internal Error: Unrecognized mode: ${options.mode}`);
     }
+  }
+
+  wrap(element) {
+    return wrap(element);
   }
 
   // converts an RSTNode to the corresponding JSX Pragma Element. This will be needed
--- a/packages/enzyme-adapter-utils/src/wrapWithSimpleWrapper.jsx
+++ b/packages/enzyme-adapter-utils/src/wrapWithSimpleWrapper.jsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import { intersects } from 'semver';
+import PropTypes from 'prop-types';
+
+const propTypes = {
+  children: PropTypes.element.isRequired,
+};
+
+const Wrapper = (intersects('>= 0.14', React.version)
+  // eslint-disable-next-line prefer-arrow-callback
+  ? () => Object.assign(function SimpleSFCWrapper({ children }) {
+    return children;
+  }, { propTypes })
+  : () => {
+    class SimpleClassWrapper extends React.Component {
+      render() {
+        const { children } = this.props;
+        return children;
+      }
+    }
+    SimpleClassWrapper.propTypes = propTypes;
+    return SimpleClassWrapper;
+  }
+)();
+
+export default function wrap(element) {
+  console.log(React.version, Wrapper);
+  return <Wrapper>{element}</Wrapper>;
+}
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -4743,6 +4743,68 @@ describe('shallow', () => {
     });
   });
 
+  describe('.renderProp()', () => {
+    it('returns a wrapper around the node returned from the render prop', () => {
+      class Foo extends React.Component {
+        render() {
+          return <div className="in-foo" />;
+        }
+      }
+      class Bar extends React.Component {
+        render() {
+          const { render: r } = this.props;
+          return <div className="in-bar">{r()}</div>;
+        }
+      }
+
+      const wrapperA = shallow(<div><Bar render={() => <div><Foo /></div>} /></div>);
+      const renderPropWrapperA = wrapperA.find(Bar).renderProp('render');
+      expect(renderPropWrapperA.find(Foo)).to.have.lengthOf(1);
+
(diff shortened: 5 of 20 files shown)


dferber90:
The tests are fine for the most part. However, I keep getting errors in React 0.13 and 0.14, and I'm not sure how to resolve them. Maybe somebody has seen them before and knows what to do?


React 0.14

```
  1) shallow .renderProp() returns a ShallowWrapper wrapper around the node returned from the render prop:
     TypeError: inst.render is not a function
      at ShallowComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (node_modules/react/lib/ReactCompositeComponent.js:587:34)
      at ShallowComponentWrapper.mountComponent (node_modules/react/lib/ReactCompositeComponent.js:220:30)
      at ShallowComponentWrapper.wrapper [as mountComponent] (node_modules/react/lib/ReactPerf.js:66:21)
      at ReactShallowRenderer._render (node_modules/react/lib/ReactTestUtils.js:366:14)
      at _batchedRender (node_modules/react/lib/ReactTestUtils.js:348:12)
      at ReactDefaultBatchingStrategyTransaction.perform (node_modules/react/lib/Transaction.js:136:20)
      at Object.batchedUpdates (node_modules/react/lib/ReactDefaultBatchingStrategy.js:62:19)
      at Object.batchedUpdates (node_modules/react/lib/ReactUpdates.js:94:20)
      at ReactShallowRenderer.render (node_modules/react/lib/ReactTestUtils.js:343:16)
      at packages/enzyme-adapter-react-14/build/ReactFourteenAdapter.js:1:1
      at withSetStateAllowed (packages/enzyme-adapter-utils/build/Utils.js:1:1)
      at Object.render (packages/enzyme-adapter-react-14/build/ReactFourteenAdapter.js:1:1)
      at new ShallowWrapper (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.wrap (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.<anonymous> (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.single (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.shallow (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.<anonymous> (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.single (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.renderProp (packages/enzyme/build/ShallowWrapper.js:1:1)
      at Context.<anonymous> (packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx:4759:26)
      at process.topLevelDomainCallback (domain.js:121:23)
```

React 0.13

```
1) shallow .renderProp() returns a ShallowWrapper wrapper around the node returned from the render prop:
     TypeError: Cannot add property context, object is not extensible
      at ShallowComponentWrapper.mountComponent (node_modules/react/lib/ReactCompositeComponent.js:153:18)
      at ShallowComponentWrapper.wrapper [as mountComponent] (node_modules/react/lib/ReactPerf.js:70:21)
      at ReactShallowRenderer._render (node_modules/react/lib/ReactTestUtils.js:391:14)
      at ReactShallowRenderer.render (node_modules/react/lib/ReactTestUtils.js:375:8)
      at ReactShallowRenderer.contextCompatibleRender [as render] (packages/enzyme-adapter-react-13/build/ReactThirteenAdapter.js:1:1)
      at Object.render (packages/enzyme-adapter-react-13/build/ReactThirteenAdapter.js:1:1)
      at new ShallowWrapper (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.wrap (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.<anonymous> (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.single (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.shallow (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.<anonymous> (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.single (packages/enzyme/build/ShallowWrapper.js:1:1)
      at ShallowWrapper.renderProp (packages/enzyme/build/ShallowWrapper.js:1:1)
      at Context.<anonymous> (packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx:4752:33)
      at process.topLevelDomainCallback (domain.js:121:23)
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -4743,6 +4743,24 @@ describe('shallow', () => {
     });
   });
 
+  describe('.renderProp()', () => {
+    it('returns a ShallowWrapper wrapper around the node returned from the render prop', () => {
+      const Foo = () => (<div className="in-foo" />);
+      const Bar = props => (<div className="in-bar">{props.render()}</div>);
```
ljharb: let's use class-based components here, so that it'll pass in react 13.

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -220,7 +220,7 @@ class ShallowWrapper {
           });
         }
       }
-    // creating a child component through enzyme's ShallowWrapper APIs.
+      // creating a child component through enzyme's ShallowWrapper APIs.
```
ljharb: let's revert these unrelated indentation changes

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -1033,6 +1033,34 @@ class ShallowWrapper {
...
   }
 
+  /**
+   * Returns a wrapper of the node rendered by the provided render prop.
+   *
+   * @param {String} propName
+   * @returns {ShallowWrapper}
+   */
+  renderProp(propName, ...args) {
+    const Wrapper = props => props.children;
```
ljharb: This won't work in React 13.

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -1033,6 +1033,34 @@ class ShallowWrapper {
...
+    return this.single('renderProp', () => {
+      if (typeof propName !== 'string') {
+        throw new TypeError('`propName` must be a string');
+      }
+      const prop = this.prop(propName);
+      if (!prop) {
```
ljharb: `prop` could be falsy with `0`, `false`, and a number of other options. Perhaps instead this should do `const props = this.props()`, and then do `has(props, propName)`?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -1033,6 +1033,34 @@ class ShallowWrapper {
...
+        throw new Error(`no prop called “${propName}“ found`);
+      }
+      if (typeof prop !== 'function') {
+        throw new Error(
```
ljharb: this should be a TypeError

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -1033,6 +1033,34 @@ class ShallowWrapper {
...
+          `expected prop “${propName}“ to contain a function, but it holds “${typeof prop}“`,
+        );
+      }
+
+      const element = prop(...args);
+      const wrapped = getAdapter(this[OPTIONS]).createElement(Wrapper, null, element);
```
ljharb: we may need to add a method to all the adapters "wrap", so that the adapter itself can handle the wrapper implementation.

packages/enzyme/src/ShallowWrapper.js
ljharb: that's a reasonable alternative; altho it's nice to be able to give a better error message "this prop doesn't exist" versus "this prop isn't a function".

packages/enzyme/src/ShallowWrapper.js
ljharb: I think it'd need to be in this PR (in a separate commit, ofc).

dferber90:
> I tried to add `wrap` to all adapters but I'm not familiar enough with the internals of `enzyme` to get the tests to pass. I still keep bumping into `TypeError: inst.render is not a function` on React 14 😞  Other versions of React seem to be passing the tests now though!

The issue was `const stub = sinon.stub().returns(null);` effectively rendering `null` from a stateless functional component, which is not supported in React 14. Changing it to `const stub = sinon.stub().returns(<div />);` solved the issue. [This comment](https://github.com/styleguidist/react-styleguidist/issues/498#issue-236608981) tipped me off.

packages/enzyme-adapter-react-13/src/ReactThirteenAdapter.js
```diff
@@ -242,6 +242,17 @@ class ReactThirteenAdapter extends EnzymeAdapter {
...
+  wrap(element) {
+    class Wrapper extends React.Component {
+      render() {
+        // eslint-disable-next-line react/prop-types
+        const { children } = this.props;
+        return children;
+      }
+    }
+    return React.createElement(Wrapper, null, element);
+  }
```
dferber90: @ljharb Could you confirm this is what you had in mind for the `wrap` function on the adapters?

ljharb:
I've updated and rebased the branch; I'll give it another review and then we can add the method for `mount` as well.

docs/api/ShallowWrapper/renderProp.md
```diff
@@ -0,0 +1,85 @@
+# `.renderProp(propName, ...args) => ShallowWrapper`
+
+Calls the current wrapper's property with name `propName` and the `args` provided.
+Returns the result in a new shallow wrapper.
+
+NOTE: can only be called on wrapper of a single non-DOM component element node.
```
ljharb: where is "non-DOM" being enforced? we might need to check the node type isn't "host" in `renderProp`.

dferber90: Nice catch. It wasn't actually enforced so far. I now added an explicit check and amended the last commit. Copied that one from `simulateError`.

dferber90:
I'm still a bit puzzled by how `renderProp` would work with `mount` - or what it would be useful for (I haven't actually used `mount` myself). Depending on the system under test, the `renderProp` will likely get called anyways during rendering, so users can test the final output.

The only useful case for having `renderProp` in `mount` I could come up with would be to influence the arguments the actual render-prop gets called with. But there should be other means to do so by setting the parent component up properly. I'm questioning how useful having `renderProp` in `mount` would be after all.

I attempted to add `mount` anyways, but I failed at coming up with a good example for docs/tests and at actually making the implementation work. I'm not sure how to proceed as I feel kinda lost without understanding the end-goal of having `renderProp` in `mount`.

It would be nice if you could give me some direction or take over from here.

ljharb:
It's so you can unit-test the render prop without needing to know all the permutations of props that lead to certain input combinations.

dferber90:
@ljharb Thanks for driving this home!

I visited a friend this weekend so I didn't get to continue working on it. I really appreciate your help!

We'll deprecate the `renderProp` function in [@commercetools/enzyme-extensions](https://github.com/commercetools/enzyme-extensions) once this is released.

dferber90:
The current API works like this `wrapper.renderProp(propName, arg1, arg2)`.

This will call the prop `propName` on `wrapper` with `arg1` and `arg2`, like `wrapper[propName](arg1, arg2)`.

This API is nice to work with but prevents us from ever adding something like an options object or any other param to `renderProp` itself, as all arguments after the first one get forwarded to the render prop.

It might be a better idea to use an API like `wrapper.renderProp(propName, [arg1, arg2])` or `wrapper.renderProp(propName)(arg1, arg2)`.

So far this has not been a problem for us but I wanted to give a heads-up now that it's becoming official. I'm undecided of which API is suited best.

ljharb:
@dferber90 those are really good points. The explicit array is cleaner, until you want no arguments and you want to pass options. The “return a function” approach is cleaner when you have arguments, but perhaps less so when you don’t.

I think I’m slightly leaning towards the “returning a function” API. I’ll be releasing the adapters sooner than enzyme itself, so we have a bit of time to make the change, but not much :-)

docs/api/ReactWrapper/renderProp.md
```diff
@@ -0,0 +1,85 @@
...
+
+expect(wrapper.equals(<h1>The mouse position is 0, 0</h1>)).to.equal(true);
+```
+
+##### Testing with multiple arguments
+
+```jsx
+const wrapper = mount(<App />)
+  .find(Mouse)
+  .renderProp('render', [10, 20]);
```
dferber90: This is under "multiple arguments", so the code should be `.renderProp('render', 10, 20)`

docs/api/ShallowWrapper/renderProp.md
```diff
@@ -0,0 +1,85 @@
...
+
+expect(wrapper.equals(<h1>The mouse position is 0, 0</h1>)).to.equal(true);
+```
+
+##### Testing with multiple arguments
+
+```jsx
+const wrapper = shallow(<App />)
+  .find(Mouse)
+  .renderProp('render', [10, 20]);
```
dferber90: Same here, this should also be `.renderProp('render', 10, 20)`.

packages/enzyme-adapter-utils/src/wrapWithSimpleWrapper.jsx
```diff
@@ -0,0 +1,29 @@
...
+        return children;
+      }
+    }
+    SimpleClassWrapper.propTypes = propTypes;
+    return SimpleClassWrapper;
+  }
+)();
+
+export default function wrap(element) {
+  console.log(React.version, Wrapper);
```
dferber90: Is this a debugging-leftover? This should go away as far as I can tell.

dferber90:
@ljharb I now also opened https://github.com/airbnb/enzyme/pull/1891 to suggest the "return a function" approach instead.

Issue #1436, opened 2017-12-15
Create Adapter for React Native & React 16


mcdankyl:
# Problem Statement
Currently we have no support for using a react native adapter with Enzyme. This is because the current adapters are built for react native web. 

Attempting to use mount with the current adapter will cause various issues from polluting the user's console with a ton of logs to inability to import necessary libraries.

While this is the case you can get some testing done by filtering the logs and using jsdom.

Most of the console error issues seem to be coming from the ReactDOM call for the React 16 adapter [here](https://github.com/airbnb/enzyme/blob/master/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js#L173)

# Purpose
The purpose of this ticket is to track work related to implementation and decisions regarding implementation of a react native enzyme adapter. 

# TODO
1. Find an alternative to rendering with ReactDOM
2. Decide best way to use the solution for (1) with existing adapters
3. Decide whether we would like to do all necessary react native stubbing in the adapter a-la [react-native-mock-render](https://github.com/Root-App/react-native-mock-render), do a RYO implementation, or not take a firm stance to the native mocking
3. Add unit tests to verify functionality
4. Update Necessary Docs

# Related Issues
https://github.com/airbnb/enzyme/issues/1390
https://github.com/airbnb/enzyme/issues/831

ljharb:
We have one for React 16; but we definitely do need one for react-native.

Implementing it here in this repo would be great; anyone interested in opening a PR?

ljharb:
Since we’d want a different adapter as needed for different React-native versions, we might also want a utils package. We’d also want to add a new binary to the adapter helper for react-native.

mcdankyl:
Yup I'm already aware of the regular react 16 adapter. I was more saying one that works with react native apps built with react 16 as the react dependency.  I've been ripping apart and experimenting with the 16 adapter as a way to boost familiarity with the adapter system. 

I'm planning to start working on implementation after holiday travels. 

Do you have some ideas as to what the adapter would use instead of ReactDOM? I saw that react test utils also has a render method in addition to a shallow render method. I was thinking of investigating that first.

ljharb:
I'm really not sure, unfortunately. I would assume/hope that the react-native team would provide something for this use case.

mcdankyl:
[...]

Issue #1375, opened 2017-11-21
Fix documentation for React Native


Right2Drive:
# Description
The [documentation](http://airbnb.io/enzyme/docs/guides/react-native.html) instructs users to install [react-native-mock](https://github.com/RealOrangeOne/react-native-mock), which does not work with more recent versions of `react` due to it's [usage of a deprecated API](https://github.com/RealOrangeOne/react-native-mock/issues/139)

# Proposal
I'm not really sure what the appropriate solution for this is, whether to add a footnote to the documentation regarding recent versions of `react-native`/`react`, or finding a more reliable library/solution for the problem.

Thanks for the awesome utility! You've made a lot of lives easier! :1st_place_medal:

iwilsonq:
+1

I 'm using Expo to setup a React Native app but I want to be able to use Enzyme to help me with testing rather than `react-test-renderer`. I'm having difficulty setting this up and it looks like `react-native-mock` is not compatible with recent versions of RN?

shawnaxsom:
I can help with the docs here. I have React Native working with Enzyme, and actually with deep rendering with `mount()` too using jsdom. I didn't have to use `react-native-mock` or `react-native-mock-renderer` to get it to work (and those solutions made the DOM and snapshots harder to read so I wouldn't recommend them).

I'm assuming it should work with Expo also @iwilsonq, let me know if you still need help getting Enzyme to work.

shawnaxsom:
We opened sourced our boilerplate project where I have Enzyme set up with React Native, if anyone wants an example to look at:

React Native mobile boilerplate with Enzyme

- https://github.com/ProminentEdge/mobile-boilerplate


Relevant files in the repo for Jest / Enzyme:

- https://github.com/ProminentEdge/mobile-boilerplate/blob/master/jest.config.js
- https://github.com/ProminentEdge/mobile-boilerplate/blob/master/setup-tests.js
- https://github.com/ProminentEdge/mobile-boilerplate/blob/master/src/screens/todo-list/todo-list.test.tsx

shawnaxsom:
Documentation PR #1873 is up.

PR #1873, opened 2018-10-21
Documentation update for testing React Native using jsdom

shawnaxsom:
As discussed in #1375, I was also unable to get React Native testing working with Jest/Enzyme with the current documentation instructions.

I see that a React Native adapter is still being discussed in #1436, but I found that some users have had success with jsdom. I was able to get both react-native-mock-renderer and jsdom working, but react-native-mock-renderer didn't result in very clear snapshots when using Jest snapshot testing.

I have therefore provided some instructions here for how I was able to get React Native testing working with jsdom, and I added some examples from the [React Native boilerplate project](https://github.com/ProminentEdge/mobile-boilerplate) where it is currently working. I decided to leave in a few helpful tips of other issues I encountered along the way as well.

### DIFF
--- a/docs/guides/react-native.md
+++ b/docs/guides/react-native.md
@@ -9,32 +9,180 @@ a host device.
 This can be difficult when you want your test suite to run with typical Continuous Integration servers
 such as Travis.
 
-A pure JS mock of React Native exists and can solve this problem in the majority of use cases.
+To use enzyme to test React Native, you currently need to configure an adapter, and load an emulated DOM.
 
-To install it, run:
+## Configuring an Adapter
 
-```bash
-npm i --save-dev react-native-mock
+While a React Native adapter is [in discussion](https://github.com/airbnb/enzyme/issues/1436),
+a standard adapter may be used, such as 'enzyme-adapter-react-16':
+
+```jsx
+import Adapter from 'enzyme-adapter-react-16';
+
+Enzyme.configure({ adapter: new Adapter() });
 ```
 
-Requiring or importing the `/mock` entry file of this project will input the mock `react-native`
-export into the require cache, so that your application uses the mock instead.
+## Loading an emulated DOM with JSDOM
 
-If you are using a test runner such as mocha, this means that you can use the `--require` flag
-before you run your test suite, and enzyme should "just work":
+To use enzyme's `mount` until a React Native adapter exists, an emulated DOM must be loaded.
 
+While some have had success with [react-native-mock-renderer](https://github.com/Root-App/react-native-mock-render),
+the recommended approach is to use [https://github.com/tmpvar/jsdom](JSDOM),
+as documented for enzyme at the [JSDOM](https://airbnb.io/enzyme/docs/guides/jsdom.html) documentation page.
 
-### Mocha CLI
+JSDOM will allow all of the `enzyme` behavior you would expect. While Jest snapshot testing can be used with
+this approach as well, it isn't encouraged and is only supported through `wrapper.debug()`.
 
-```bash
-mocha --require react-native-mock/mock --recursive path/to/test/dir
+## Using enzyme's find when lacking className props
+
+It is worth noting that React Native allows for a [testID](https://facebook.github.io/react-native/docs/view#testid)
+prop, that can be used a selector similar to `className` in standard React:
+
+<!-- eslint no-unused-expressions: 0, semi: 0 -->
+```jsx
+    <View key={key} style={styles.todo} testID="todo-item">
+      <Text testID="todo-title" style={styles.title}>{todo.title}</Text>
+    </View>
 ```
 
-### In Code
+```jsx
+expect(wrapper.findWhere(node => node.prop('testID') === 'todo-item')).toExist();
+```
 
-```js
-/* file-that-runs-before-all-of-my-tests.js */
+## Example configuration for Jest
 
-// This will mutate `react-native`'s require cache with `react-native-mock`'s.
-require('react-native-mock/mock'); // <-- side-effects!!!
+To perform the necessary configuration in your testing framework, it is recommended to use a setup script,
+such as with Jest's `setupTestFrameworkScriptFile` setting.
+
+Create or update a `jest.config.js` file at the root of your project to include the `setupTestFrameworkScriptFile` setting:
+
+```jsx
+// jest.config.js
+
+module.exports = {
+  // Load setup-tests.js before test execution
+  setupTestFrameworkScriptFile: '<rootDir>setup-tests.js',
+
+  // ...
+};
+```
+
+Then create or update the file specified in `setupTestFrameworkScriptFile`, in this case `setup-tests.js` in the project root:
+
+```jsx
+// setup-tests.js
+
+import 'react-native';
+import 'jest-enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import Enzyme from 'enzyme';
+
+/**
+ * Set up DOM in node.js environment for Enzyme to mount to
+ */
+const { JSDOM } = require('jsdom');
+
+const jsdom = new JSDOM('<!doctype html><html><body></body></html>');
+const { window } = jsdom;
+
+function copyProps(src, target) {
+  Object.defineProperties(target, {
+    ...Object.getOwnPropertyDescriptors(src),
+    ...Object.getOwnPropertyDescriptors(target),
+  });
+}
+
+global.window = window;
+global.document = window.document;
+global.navigator = {
+  userAgent: 'node.js',
+};
+copyProps(window, global);
+
+/**
+ * Set up Enzyme to mount to DOM, simulate events,
+ * and inspect the DOM in tests.
+ */
+Enzyme.configure({ adapter: new Adapter() });
+
+/**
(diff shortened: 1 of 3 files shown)


docs/guides/react-native.md
```diff
@@ -9,32 +9,188 @@ a host device.
...
+
+```
+import Adapter from "enzyme-adapter-react-16";
+
+Enzyme.configure({ adapter: new Adapter() });
+```
+
+## Loading an emulated DOM with JSDOM
+
+To use enzyme's `deep` until a React Native adapter exists, an emulated DOM must be loaded.
```
ljharb: what is "deep"? Do you mean `mount`?

docs/guides/react-native.md
```diff
@@ -9,32 +9,188 @@ a host device.
...
+
+While some have had success with [react-native-mock-renderer](https://github.com/Root-App/react-native-mock-render),
+the recommended approach is to use [https://github.com/tmpvar/jsdom](JSDOM),
+as documented for enzyme at the [JSDOM](https://airbnb.io/enzyme/docs/guides/jsdom.html) documentation page.
+
+JSDOM will allow all of the `enzyme` behavior you would expect, and it will result in reasonable snapshots if
```
ljharb: let's note here that enzyme has no snapshot support besides `.debug()`

docs/guides/react-native.md
```diff
@@ -9,32 +9,188 @@ a host device.
...
+ * Set up DOM in node.js environment for Enzyme to mount to
+ */
+const { JSDOM } = require("jsdom");
+
+const jsdom = new JSDOM("<!doctype html><html><body></body></html>");
+const { window } = jsdom;
+
+function copyProps(src, target) {
+  const props = Object.getOwnPropertyNames(src)
+    .filter(prop => typeof target[prop] === "undefined")
```
ljharb: let's use single quotes here, so the JS would pass eslint

docs/guides/react-native.md
```diff
@@ -9,32 +9,188 @@ a host device.
...
+/**
+ * Set up DOM in node.js environment for Enzyme to mount to
+ */
+const { JSDOM } = require("jsdom");
+
+const jsdom = new JSDOM("<!doctype html><html><body></body></html>");
+const { window } = jsdom;
+
+function copyProps(src, target) {
+  const props = Object.getOwnPropertyNames(src)
```
ljharb: is `Object.getOwnPropertyDescriptors` not an option?

docs/guides/react-native.md
shawn-pe: Do you mean if jsdom isn't used? In our React Native boilerplate where we are using jsdom, we get better output from just expecting on the wrapper object itself:

https://github.com/ProminentEdge/mobile-boilerplate/blob/4994f554e64c63c7b10252baeb62c31a7cc6c5c1/src/screens/todo-list/todo-list.test.tsx#L141

If expecting with `expect(wrapper.debug())` we get e.g.

```
      <Container virtual={[undefined]} style={{...}}>
```

Where as if you expect with `expect(wrapper)` it will show us more information, like what styles were applied

```
      <Container
        style={
          Object {
            "backgroundColor": "#fff",
            "flex": 1,
            "height": 1334,
          }
        }
      >
```

docs/guides/react-native.md
shawnaxsom: Yes mount, good catch, I'll update that.

docs/guides/react-native.md
shawnaxsom: Do you mean that it isn't supported without `jsdom`? Because we get better output with `jsdom` if we just expect on the wrapper.

In our React Native boilerplate where we are using jsdom, we get better output from just expecting on the wrapper object itself:

https://github.com/ProminentEdge/mobile-boilerplate/blob/4994f554e64c63c7b10252baeb62c31a7cc6c5c1/src/screens/todo-list/todo-list.test.tsx#L141

If expecting with `expect(wrapper.debug())` we get e.g.

```
      <Container virtual={[undefined]} style={{...}}>
```

Where as if you expect with `expect(wrapper)` it will show us more information, like what styles were applied

```
      <Container
        style={
          Object {
            "backgroundColor": "#fff",
            "flex": 1,
            "height": 1334,
          }
        }
      >
```

docs/guides/react-native.md
shawnaxsom: Will do!

docs/guides/react-native.md
shawnaxsom: That makes sense, I updated the wording, let me know how that sounds: 

```
JSDOM will allow all of the `enzyme` behavior you would expect. While Jest snapshot testing can be used with
this approach as well, it isn't encouraged and is only supported through `wrapper.debug()`.
```

docs/guides/react-native.md
ljharb: Let's update to use it; it's cleaner anyways.

docs/guides/react-native.md
shawnaxsom: I've updated all the double quotes in the page to single quotes

docs/guides/react-native.md
ljharb: Sounds great, thanks.

docs/guides/react-native.md
shawnaxsom: This has been updated to `mount`

docs/guides/react-native.md
shawnaxsom: Do we want to support es5 / IE?

I updated for now to:

```
function copyProps(src, target) {
  Object.defineProperties(target, {
    ...Object.getOwnPropertyDescriptors(src),
    ...Object.getOwnPropertyDescriptors(target)
  });
}
```

I think that does what we want cleanly, but VSCode is complaining:

```
[ts] Property 'getOwnPropertyDescriptors' does not exist on type 'ObjectConstructor'. Did you mean 'getOwnPropertyDescriptor'?
lib.es5.d.ts(172, 5): 'getOwnPropertyDescriptor' is declared here.
```

Because Object.getOwnPropertyDescriptors isn't in the es5 spec and unsupported also for IE:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptors

shawnaxsom: Let me know whether we should keep as is to support es5 / IE, or whether the updated code I have above with `getOwnPropertyDescriptors` is preferred.

ljharb: gOPDs is polyfillable, so it should work just fine. https://www.npmjs.com/package/object.getownpropertydescriptors

The spread can be transpiled to Object.assign which is also polyfillable https://www.npmjs.com/package/object.assign

so i think that update works great

shawnaxsom: Sounds good, I updated copyProps here and in two other places in the codebase:

docs/guides/jsdom.md
packages/enzyme-example-mocha/test/.setup.js

jpaas:
Hey @shawnaxsom I'm trying out what you have in the docs, but I'm getting `Cannot find module 'StyleSheet' from 'react-native-implementation.js'`

shawnaxsom:
@jpaas Is the code public so I can see? Are you importing StyleSheet anywhere? StyleSheet should come from "react-native" instead of "react-native-implementation" generally.

Take a look at the boilerplate code we have here to compare: 

https://github.com/ProminentEdge/mobile-boilerplate

jpaas:
@shawnaxsom no sorry its a private repo. I did have a look at your repo. My `setup-tests.js` is identical. Even though I'm using typescript I wasn't able to get your `jest.config.js` to work. This is my `jest.config.js` that works for my non-enzyme tests:
```
module.exports = {
  preset: "ts-jest",
  setupTestFrameworkScriptFile: "<rootDir>/setupTests.js",
  transform: {
    // Required to work around this problem: https://github.com/facebook/react-native/issues/19859
    "^.+\\.(js|tx|tsx)$": "<rootDir>/node_modules/react-native/jest/preprocessor.js",
  },
}
```

shawnaxsom:
@jpaas Have you compared versions in package.json? What version of "react-native" are you on?

jpaas:
@shawnaxsom I'm using react-native 0.57.4 & react 16.6.0-alpha.8af6728 vs your 0.57.2 and react 16.5.0. I tried rolling back to your version, but no difference.

I tried that jest preset in the `package.json` but it didn't help, I suspect because my `jest.config.js` overrides it with the `ts-jest` preset.

When I try to use your `jest.config.js`, I get the error `Couldn't find preset "module:metro-react-native-babel-preset"` relative to my root dir`

shawn-pe:
You might need to install metro-react-native-babel-preset and use our
babel.config.js if you want to try our jest.config.js.

jpaas:
@shawnaxsom I am using `metro-react-native-babel-preset` as my preset in my `.babelrc`, so I imagine there's something about your `jest.config.js` that is preventing it from being resolved.

shawnaxsom:
If you have time to set up a public project that replicates the issue I
would be happy to take a look, or I might be able to screen share at some
point. It's hard to diagnose unfortunately without being able to look at
the project.

PR #1890, opened 2018-11-06
renderProp followup (doc corrections & console.log removal)

dferber90:
I found some issues while reviewing #1863 after it was merged. This follow-up resolves them.

*If we accept #1891 we don't need to merge this PR as those fixes are contained in there as well.*


dferber90:
One test run is failing, but it doesn't seem related to my changes. Maybe restarting the test run in CI is enough.

PR #1891, opened 2018-11-06
Change renderProp API

dferber90:
This PR changes the `renderProp` API as described in https://github.com/airbnb/enzyme/pull/1863#issuecomment-436175166.

This is merely a suggestion at the moment and I'm happy to close it again if we settle on another approach or even keep the current one.

The motivation for "returning a function" is that the arguments to enzyme's `renderProp` won't get mixed with the arguments to the render prop function itself. This allows us to eventually add more options (like context) which we can apply to the newly created wrapper.

An example usage (formatted with Prettier) would look like this:

```jsx
const wrapper = shallow(<App />)
  .find(Mouse)
  .renderProp("render")(10, 20)
  .find(Mouse)
  .renderProp("foo")(3, {
    long: {
      yeah: true
    }
  })
  .find(Bar)
  .renderProp("x")();
```

### Sidenote

At this point, one might wonder what the advantage of `wrapper.find(Foo).renderProp('render')()` over `wrapper.find(Foo).prop('render')()` is, and that's a fair question.

When using `renderProp`, the render props' return value gets nested into a new wrapper which enables this convenient chaining shown above - whereas `wrapper.find(Foo).prop('render')()` simply returns the React element without an enzyme wrapper. More about this is hinted at in [this article](https://medium.com/@dferber90/test-a-render-prop-6a44e02f4c39).

### DIFF
--- a/docs/api/ReactWrapper/renderProp.md
+++ b/docs/api/ReactWrapper/renderProp.md
@@ -1,7 +1,6 @@
-# `.renderProp(propName, ...args) => ReactWrapper`
+# `.renderProp(propName)(...args) => ReactWrapper`
 
-Calls the current wrapper's property with name `propName` and the `args` provided.
-Returns the result in a new wrapper.
+Returns a function that, when called with arguments `args`, will return a new wrapper based on the render prop in the original wrapper's prop `propName`.
 
 NOTE: can only be called on wrapper of a single non-DOM component element node.
 
@@ -69,7 +68,7 @@ const App = () => (
 ```jsx
 const wrapper = mount(<App />)
   .find(Mouse)
-  .renderProp('render');
+  .renderProp('render')();
 
 expect(wrapper.equals(<h1>The mouse position is 0, 0</h1>)).to.equal(true);
 ```
@@ -79,7 +78,7 @@ expect(wrapper.equals(<h1>The mouse position is 0, 0</h1>)).to.equal(true);
 ```jsx
 const wrapper = mount(<App />)
   .find(Mouse)
-  .renderProp('render', 10, 20);
+  .renderProp('render')(10, 20);
 
 expect(wrapper.equals(<h1>The mouse position is 10, 20</h1>)).to.equal(true);
 ```
--- a/docs/api/ShallowWrapper/renderProp.md
+++ b/docs/api/ShallowWrapper/renderProp.md
@@ -1,7 +1,6 @@
-# `.renderProp(propName, ...args) => ShallowWrapper`
+# `.renderProp(propName)(...args) => ShallowWrapper`
 
-Calls the current wrapper's property with name `propName` and the `args` provided.
-Returns the result in a new wrapper.
+Returns a function that, when called with arguments `args`, will return a new wrapper based on the render prop in the original wrapper's prop `propName`.
 
 NOTE: can only be called on wrapper of a single non-DOM component element node.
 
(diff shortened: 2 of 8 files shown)


docs/api/ReactWrapper/renderProp.md
```diff
@@ -1,4 +1,4 @@
-# `.renderProp(propName, ...args) => ReactWrapper`
+# `.renderProp(propName)(...args) => ReactWrapper`
 
 Calls the current wrapper's property with name `propName` and the `args` provided.
```
ljharb: Let's reword this; something like "Returns a function that, when called with arguments `args`, will return a new wrapper based on the render prop in the original wrapper's prop `propName`."

docs/api/ShallowWrapper/renderProp.md
```diff
@@ -1,4 +1,4 @@
-# `.renderProp(propName, ...args) => ShallowWrapper`
+# `.renderProp(propName)(...args) => ShallowWrapper`
 
 Calls the current wrapper's property with name `propName` and the `args` provided.
```
ljharb: this too.

packages/enzyme-test-suite/package.json
```diff
@@ -39,7 +39,8 @@
     "object-inspect": "^1.6.0",
     "prop-types": "^15.6.2",
     "semver": "^5.5.0",
-    "sinon": "^4.1.6"
+    "sinon": "^4.1.6",
+    "enzyme-adapter-react-13": "1.1.2"
```
ljharb: this should be reverted.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -4998,7 +4998,7 @@ describeWithDOM('mount', () => {
 
       const wrapper = mount(<Div />).childAt(0);
       expect(wrapper.is('div')).to.equal(true);
-      expect(() => wrapper.renderProp('foo')).to.throw();
+      expect(() => wrapper.renderProp('foo')()).to.throw();
```
ljharb: I would expect this to throw as-is - ie, before invoking the returned function.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -5017,7 +5017,7 @@ describeWithDOM('mount', () => {
         }
 
         const wrapper = mount(<div><Bar render={() => <div><Foo /></div>} /></div>);
-        expect(() => wrapper.find(Bar).renderProp('render')).to.throw(RangeError);
+        expect(() => wrapper.find(Bar).renderProp('render')()).to.throw(RangeError);
```
ljharb: similarly, i'd expect this to throw immediately, and not inside the returned function.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -4782,7 +4782,7 @@ describe('shallow', () => {
 
       const wrapper = shallow(<Div />);
       expect(wrapper.is('div')).to.equal(true);
-      expect(() => wrapper.renderProp('foo')).to.throw();
+      expect(() => wrapper.renderProp('foo')()).to.throw();
```
ljharb: and here

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -4801,7 +4801,7 @@ describe('shallow', () => {
         }
 
         const wrapper = shallow(<div><Bar render={() => <div><Foo /></div>} /></div>);
-        expect(() => wrapper.find(Bar).renderProp('render')).to.throw(RangeError);
+        expect(() => wrapper.find(Bar).renderProp('render')()).to.throw(RangeError);
```
ljharb: and here

dferber90:
👍 I amended the last commit with those fixes

PR #1897, opened 2018-11-13
Enable Gitbook search plugin

### LINKED ISSUE #1892: Can't search https://airbnb.io/enzyme/
### Is your feature request related to a problem? Please describe.

I find it frustrating to navigate the documentation at https://airbnb.io/enzyme/.

It's not just difficult to find the page you're looking for – I find the basic navigation ergonomics really poor as well. For instance, if I'm looking at [ReactWrapper.prototype.hasClass](https://airbnb.io/enzyme/docs/api/ReactWrapper/hasClass.html), it's not so easy to tell what other methods are available to me on a `ReactWrapper` object (e.g., if I were looking for a `className` method).

### Describe the solution you'd like

I would like to be able to search for text in the documentation.

### Describe alternatives you've considered

I can just go through Google search to the documentation, but I believe you folks are using GitBooks and there is a search plugin for it, so please consider using that.

filoxo:
Resolves #1892 by enabling the Gitbook search plugin. 

There is a risk that upgrading to the latest version of gitbook-cli may break building the docs (see https://github.com/GitbookIO/gitbook-cli/issues/74) but as I understand the issue, since this won't be changing the version of gitbook (only the gitbook-cli) I'm hopeful that the previously mentioned issue won't arise here.


filoxo:
Is there something I can do to inspire greater confidence on deploying this change?

ljharb:
@filoxo sorry for the delay here, i haven't had time to get to this. I'll test it and try to merge it asap.

filoxo:
Huh, the menu thing is interesting. The documentation for collapsible-menu actually says that having the menu items closed is the expected behavior.

https://github.com/rtCamp/gitbook-plugin-collapsible-menu#how-this-work

> First it hides all pages, and then using `.active` css class show's its parent and grandparent nodes (chapters).

I also noticed it behaves the same way (all menu items collapsed) when navigating away from the "Introduction" page and back on the live site.

I'll keep digging around and see if there's an option or config that will allow all menu items to be open by default, but would removing `collapsed-menu` plugin be a better solution if you prefer not having the menu items closed?

ljharb:
I don't mind the initial state so much, but i can't even click on the menus to expand them. Whatever it takes to make that happen would be great :-)

filoxo:
I found that even if search isn't enabled, gitbook includes the search-lunr plugin by default (in the `.gitbooks` folder) and attempts to set `gitbooks.search` engine, but search isn't defined so it throws this error.

```
jQuery.Deferred exception: Cannot read property 'getEngine' of undefined TypeError: Cannot read property 'getEngine' of undefined
    at Object.<anonymous> (https://airbnb.io/enzyme/gitbook/gitbook-plugin-lunr/search-lunr.js:54:37)
    at Object.dispatch (https://airbnb.io/enzyme/gitbook/gitbook.js:2:19320)
    at Object.m.handle (https://airbnb.io/enzyme/gitbook/gitbook.js:2:17339)
    at Object.trigger (https://airbnb.io/enzyme/gitbook/gitbook.js:3:8538)
    at Object.<anonymous> (https://airbnb.io/enzyme/gitbook/gitbook.js:3:9059)
    at Function.each (https://airbnb.io/enzyme/gitbook/gitbook.js:1:15309)
    at de.fn.init.each (https://airbnb.io/enzyme/gitbook/gitbook.js:1:13477)
    at de.fn.init.trigger (https://airbnb.io/enzyme/gitbook/gitbook.js:3:9034)
    at Object.r [as hasChanged] (https://airbnb.io/enzyme/gitbook/gitbook.js:4:8170)
    at https://airbnb.io/enzyme/:2240:26 undefined
``` 

After that error is thrown, the collapsible-menu plugin doesn't seem to be executed until the next page change event. I manually fixed this and ran the old version of the docs locally, and the menu collapses just like the new version does.
That's just a long winded way of saying that the menu was broken without search having been enabled and now that it is, its working as intended. 

But definitely, the menu should be clickable! I've tested this locally on latest Chrome, Firefox, and Safari on Mac, and on Windows 10 using Edge and IE11; and all of them work for me. Could you share which browser/version/platform you're using? As well as any console errors. The collapse-menu plugin uses jQuery so I wonder if some other error is being thrown and causing other plugins to fail.

ljharb:
Safari on Mac - but I'm using `file://`, and it seems gitbook makes cross-origin requests for some reason?

When I use a local http server, things work, but it'd still be nice if the menus were collapsible/expandable without navigation.

ljharb:
For the record, you're right - on the current site, `TypeError: undefined is not an object (evaluating 'gitbook.search.getEngine')` seems to be the reason they're initially expanded, but other pages show them collapsed.

ljharb:
Also, on the new site, i see twitter/facebook/share icons - let's remove those :-)

ljharb:
The CSS also seems a bit different; the old site is a bit more compact (in a good way).

filoxo:
I'm going to go ahead and decline this PR then, since it's introducing more changes than expected. I might try getting the old version working and see if I'm just dumb and can workaround the issues I'd noted in #1897 to avoid upgrading gitbook-cli.

Enabling search will cause all menu items to be closed on initial navigation because the collapsible-menu plugin will work correctly. Unfortunately there is no config/setting that will achieve what you like without forking it, which I'd rather not do because I won't maintain it. I just wanted search-ability on the docs site.

ljharb:
I’ll reopen this then and see what i can figure out.

filoxo:
Closing my own stale PRs.

ljharb:
I'm still hoping to get back to this; hopefully it's ok that it remains open.

ljharb:
@filoxo unfortunately by deleting the fork this PR is unrecoverable :-( i wish you hadn't done that.

PR #1903, opened 2018-11-14
Added support for shallow rendering children-as-a-function component

### LINKED ISSUE #1902: Shallow rendering component with children-as-a-function [render props]
**Is your feature request related to a problem? Please describe.**

I was frustrated when I had to test a component which contained another one with so-called children as a function (or render prop). 

Let's recap for a minute how children-as-a-function works:

```javascript
class Mouse extends React.Component {
  constructor(props) {
    super(props);
    this.handleMouseMove = this.handleMouseMove.bind(this);
    this.state = { x: 0, y: 0 };
  }

  handleMouseMove(event) {
    this.setState({
      x: event.clientX,
      y: event.clientY
    });
  }

  render() {
    return this.props.children(this.state);
  }
}
```
Let's check out this usage: 
```javascript
// App.jsx
const App = ({ someFlag }) => (
    <div>
    {someFlag && <SomeAlert />}
    <Mouse>{ ({ x, y }) => 
        <div>
            {someFlag && x> 10 && y>10 && <SomeComponent />}
            Current mouse position is {x}:{y}
        </div>
    }</Mouse>
    </div>
)
```

The test for App.jsx can look like (best solution available on the web):
```javascript

describe('App', () => {
    it('renders SomeComponent and SomeAlert when someFlag is true and mouse is at least 10px from the left and top screen borders', () => {
        const topComp = shallow(<App someFlag/>); 
        // topComp has [function] as a child, enzyme don't go further in shallow rendering
        // so we need to manually invoke that function
        const insideTheMouseComp = shallow(topComp.find(Mouse).prop('children')({x: 11, y: 11})) 

        //Lets try to lookup for the SomeComponent
        expect(insideTheMouseComp.find(SomeComponent).length).toEqual(1);

        // Okay, not bad
        // Let's try to lookup for the SomeAlert

        expect(insideTheMouseComp.find(SomeAlert).length).toEqual(1);
        // Oh no! Nothing there!
        // That's because SomeAlert is outside of Mouse 
        // we still can expect(topComp.find(SomeAlert).length).toEqual(1);
        // but it looks a bit ugly since we need to have knowledge
        // about both Mouse component implementation
        // and where it is located in App
        //
        
    });
});
```
Also what would happen if there are nested render-children-prop components,e.g.
```javascript
<Mouse>{
    ({x,y}) => <KeyPressed>{
        (isKeyPressed) => isKeyPressed ? <SomeComponent/> : <SomeAnotherComponent/>
    }</KeyPressed>
}</Mouse>
```
If you are to refactor this file (e.g. swap Mouse and KeyPressed components positions), you will need to rewrite all tests regarding that .find(...).prop('children')(...)
to match the new structure.
Bad.

How about declaratively create a mocks for the children-as-a-function component?
```javascript
describe('App', () => {
    it('renders SomeComponent and SomeAlert when someFlag is true and mouse is at least 10px from the left and top screen borders', () => {
        const MockedMouse = ({ children }) => children({ x: 11, y: 11 });
        jest.mock('./Mouse', MockedMouse);
        const justComp = shallow(<App someFlag />); 

        expect(justComp.find(SomeComponent).length).toEqual(1);



        expect(justComp.find(SomeAlert).length).toEqual(1);
        // Looks much cooler, we have single tree having all the info
        // But no. We still have that [function] as a child and enzyme not going deeper
        // These 2 expects would fail
        // It seems we need a solution for this case
    })
});
```

With this approach we still have to know how is Mouse working, but we are free now from the position it has in the tree.

**Describe the solution you'd like**

`packages/enzyme-adapter-utils/src/Utils.js`(Line 206):
```
export function elementToTree(el, recurse = elementToTree) {
    ...
    else if (typeof children !== 'undefined') {
    rendered = recurse(children);
  }
}
```
Let's change that to 
```
export function elementToTree(el, recurse = elementToTree) {
    ...
    else if (typeof children !== 'undefined') {
        if (typeof children === 'function' && nodeTypeFromType(type) === 'function') {
        // Handling chilren-as-a-function component
        rendered = recurse(type({ children }));
        } else {
        rendered = recurse(children);
        }
    }
```
Basically, we are saying that if children prop is function and currentElement is just function too (not ClassComponent), we are gonna pass those children to this component.

It's based upon two assumptions:
1. Children-as-a-function mock component should be a pure function of children.
2. Children-as-a-function real component should NOT be a function, but rather a class component (makes sense since children-as-a-function approach is used for stateful components)

**Describe alternatives you've considered**

• Writing a lot of unmaintainable tests
• Not using children-as-a-function

**Additional context**

This probably should be revisited with upcoming React Hooks.

Also I can provide a PR with these changes, but I thought we should discuss it first.

PS. My first issue, be nice to me please :)

shmidt-i:
Fixes #1902

### DIFF
--- a/packages/enzyme-adapter-utils/src/Utils.js
+++ b/packages/enzyme-adapter-utils/src/Utils.js
@@ -204,7 +204,12 @@ export function elementToTree(el, recurse = elementToTree) {
   if (isArrayLike(children)) {
     rendered = flatten(children).map(x => recurse(x));
   } else if (typeof children !== 'undefined') {
-    rendered = recurse(children);
+    if (typeof children === 'function' && nodeTypeFromType(type) === 'function') {
+      // Handling children-as-a-prop component
+      rendered = recurse(type({ children }));
+    } else {
+      rendered = recurse(children);
+    }
   }
 
   const nodeType = nodeTypeFromType(type);


shmidt-i:
A `.renderProp` method basically would be just a syntax sugar around first approach, it will still have the same problem that you can only get either the outsides of children-as-a-function component or the insides but not both. 

When you do `shallow` render you expect that all insides of the component would be rendered, so one can [according to docs]

> ensure that your tests aren't indirectly asserting on behavior of child components.

You either can't do it without invoking children-as-a-function or it adds a lot of cognitive overload to support these tests.

PR #1904, opened 2018-11-14
Zero-node specs and warning message

taylorkearns:
- Adds specs for wrapping zero nodes with `getDOMNode`, `dive`, and `name!`.
- Modifies the warning language to better accommodate zero nodes:
```
Method x is only meant to be run on a single node. 2 found instead.
Method x is only meant to be run on a single node. 0 found instead.
```
becomes
```
Method x is meant to be run on 1 node. 2 found instead.
Method x is meant to be run on 1 node. 0 found instead.
```

### DIFF
--- a/packages/enzyme/src/ReactWrapper.js
+++ b/packages/enzyme/src/ReactWrapper.js
@@ -1095,7 +1095,7 @@ class ReactWrapper {
     const fnName = typeof name === 'string' ? name : 'unknown';
     const callback = typeof fn === 'function' ? fn : name;
     if (this.length !== 1) {
-      throw new Error(`Method “${fnName}” is only meant to be run on a single node. ${this.length} found instead.`);
+      throw new Error(`Method “${fnName}” is meant to be run on 1 node. ${this.length} found instead.`);
     }
     return callback.call(this, this.getNodeInternal());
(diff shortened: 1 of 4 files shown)


packages/enzyme/src/ReactWrapper.js
```diff
@@ -1095,7 +1095,7 @@ class ReactWrapper {
     const fnName = typeof name === 'string' ? name : 'unknown';
     const callback = typeof fn === 'function' ? fn : name;
     if (this.length !== 1) {
-      throw new Error(`Method “${fnName}” is only meant to be run on a single node. ${this.length} found instead.`);
+      throw new Error(`Method “${fnName}” is meant to be run on 1 node. ${this.length} found instead.`);
```
ljharb: Can you elaborate on why this change is clearer to you? "single" means "1".

taylorkearns: When I scan the message, I pick up faster on the number 1 than the words "a single".  Because `this.length` is an integer, I find it easier to compare the two integers.

taylorkearns:
Updated this PR to just add the zero node specs, removes the suggested change to warning message.

Issue #1798, opened 2018-08-28
Documentation does not make it clear when to use `.dive()` vs `.shallow()`


GreenGremlin:
**Describe the bug**

It's not clear to me what the differences between `.dive()` and `.shallow()` are. From the documentation, and the code, it looks like they do the same thing except that `.dive()` errors if the element is anything but a component. It also looks like `.dive()` inherits the options of the container it is called on, where `.shallow()` does not. 

Why would one choose to use `.dive()` over `.shallow()`, or vice versa?  I imagine, it `.dive()` is preferred over `.shallow()`, since it's more restrictive, but it would be nice if the documentation called that out explicitly.

ljharb:
`.dive()` is sugar for "throw if there's more than one child, throw if that child isn't a custom component, call `.shallow` on it". It was a common enough pattern that it warranted first-class inclusion.

Improvements to the docs to make this more clear are quite welcome.

ljharb:
In particular, the mantra i often use is "when shallow rendering, one .dive() per HOC"

dschinkel:
I use dive() but even after reading above I still don't get why it matters vs a double shallow.  It's still not clear to me what dive() is giving me, the use case for it and why it's warranted in x situations.  If you need to get past an HOC, that works with double shallow OR dive() I've found.

AmirhosseinBayat:
this example show everything clearly:
`it('calls actions as expected when toggling switches', () => {
    const wrapper = shallow(
      <CameraSettings />,
      { context: { store: mockStore(initialState) } },
    );
    const render = wrapper.dive();
    render.find('Switch').forEach(child => {
      child.simulate('valueChange');
    });
});`

GreenGremlin:
> this example show everything clearly:

That’s a decent example of how to use dive, but shallow would work there just the same.

dschinkel:
No what I'm asking are use cases for using **double shallow** vs **shallow + dive**.

dschinkel:
Removed my questions doesn't look like I'm going to get a response....I'll figure it out and share it in my course.

ljharb:
i don't always have time to get to questions right away; but deleting them is a pretty good way to ensure i can't answer them :-p

dschinkel:
It's alright I need to move ahead anyway with stuff..I need to come to a conclusion one way or another and felt my post was pretty lengthy anyway.  I could put it back up, if you feel like you want to answer it later when you have time

ljharb:
[...]

PR #1905, opened 2018-11-15
[Docs] Add a NOTE line in docs of `.dive()` about it would throw error

markselby9:
Add a NOTE line in docs of `.dive()` about it would throw error under some cases.

See #1798

### DIFF
--- a/docs/api/ShallowWrapper/dive.md
+++ b/docs/api/ShallowWrapper/dive.md
@@ -2,7 +2,7 @@
 
 Shallow render the one non-DOM child of the current wrapper, and return a wrapper around the result.
 
-NOTE: can only be called on wrapper of a single non-DOM component element node.
+NOTE: can only be called on a wrapper of a single non-DOM component element node, otherwise it will throw an error. If you have to shallow-wrap a wrapper with multiple child nodes, use [.shallow()](ShallowWrapper/shallow.md).
 
 
 #### Arguments


docs/api/ShallowWrapper/dive.md
```diff
@@ -2,6 +2,8 @@
 
 Shallow render the one non-DOM child of the current wrapper, and return a wrapper around the result.
 
+NOTE: `dive` would throw an error if the wrapper contains more than one child, or if the (only one) child isn't a custom React component.
```
ljharb: This note is a duplicate of the one that appears right below it.

markselby9: I just made a change by amending the commit, please check.

docs/api/ShallowWrapper/dive.md
```diff
@@ -2,7 +2,8 @@
 
 Shallow render the one non-DOM child of the current wrapper, and return a wrapper around the result.
 
-NOTE: can only be called on wrapper of a single non-DOM component element node.
+NOTE: `dive` would throw an error if the wrapper contains more than one child, or if the (only one) child isn't a custom React component element node.
```
ljharb: Does "can only be called" not imply to you that it will throw an error?

"single" means "only one", and "non-DOM" means "custom React component".

markselby9: I agree, no need to change this sentence. I made a modification again.

Issue #388, opened 2016-05-12
When shallow rendering, .find, .findWhere, etc only go one node deep


dclowd9901:
I'm finding that even though I can output all content with wrapper.html(), when I try to dig through a rendered shallow component with .find and .findWhere, they stop at the top level. 

In this example:
```
wrapper.findWhere(function(el){ console.log(el); });
```
for instance, only runs once, even for a component with deeply nested inline elements.

Aweary:
That is what shallow rendering does:

https://facebook.github.io/react/docs/test-utils.html#shallow-rendering

>Shallow rendering is an experimental feature that lets you render a component "one level deep" and assert facts about what its render method returns, without worrying about the behavior of child components, which are not instantiated or rendered. This does not require a DOM

If you want to assert on the fully rendered DOM, you can use `mount`.

dclowd9901:
Sorry, I think I didn't make myself clear:

```
class Component extends React.Component {
    inline() {
        return (<div id="i-am-inline"></div>);
    }

    render() {
        return (<div id="i-am-the-parent">{this.inline()}</div>);
    }
}

var wrapper = Enzyme.shallow(<Component />);

console.log(wrapper.html());
// outputs
// <div id="i-am-the-parent"><div id="i-am-inline"></div></div>

console.log(wrapper.find('#i-am-inline').length);
// outputs 0
```

If I perform `wrapper.find('#i-am-inline')` it won't be found, despite the fact that it is clearly being rendered.

Aweary:
@dclowd9901 I'll reopen this for now until I can reproduce and clarify what the expected behavior here should be.

dclowd9901:
Thanks; it may just be my misunderstanding of shallow render. My understanding was always that it will render an entire *component* (depth notwithstanding), just not other components that might inhabit it.

brad-decker:
the find method seems to be working only against the nodes that are compiled via shallow, or mount, and in the case of shallow only the root node will be found. The html() method will render the full html but the find method is not finding against the html output. Ran into this issue earlier today and just using mount fixed the problem.

ljharb:
That's intentional - `html()` just does a full `render`, but `.find` operates on the shallow tree.

brad-decker:
@ljharb yeah i agree with you. I think this issue can be closed.

Aweary:
Then I will close this out again 😄

jwbay:
[...]

PR #1912, opened 2018-11-21
[Docs] `html`: shallow renders full tree

zbycz:
I was very confused, why `shallow().html()` returns the whole HTML tree. This is kind of unxepected, so this PR can help others undestand.

I found that its expected behavior here: https://github.com/airbnb/enzyme/issues/388#issuecomment-250234719

btw, how does it really work? Does it call the same render method for both `mount()` and `shallow()`? Possibly even for `render()`? Thanks!


ljharb:
`.html()` calls into `render`.

What is probably wanted most of the time is `.debug()`.

docs/api/ShallowWrapper/html.md
```diff
@@ -1,10 +1,11 @@
 # `.html() => String`
 
-Returns a string of the rendered HTML markup of the current render tree.
+Returns a string of the rendered HTML markup of the current render tree. The full tree is rendered, not only the Shallow part.
```
ljharb:
```suggestion
Returns a string of the rendered HTML markup of the entire current render tree (not just the shallow-rendered part).
```

zbycz:
@ljharb Thanks for the clarification! I accepted your suggestion and added a sentence about internal workings and hint for debug.

I hope several commits are fine - github can squash em, right?

PR #1914, opened 2018-11-22
Added support for React.memo

### LINKED ISSUE #1875: New `memo` feature from React 16.6 is not supported
### Current behavior
Shallow render will fail with error:
```
Invariant Violation: ReactShallowRenderer render(): Shallow rendering works only with custom components, but the provided element type was 'object'
```

### Expected behavior
Render without error.

### Your environment
No need, happenes everywhere. I new adaper version will be needed I guess.

#### API

- [x] shallow
- [x] mount
- [ ] render

#### Version

| enzyme              |  3.7.0
| react               | 16.6.0

#### Adapter

- [X] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] others ( )

jintoppy:
This is based on and fixes #1875.

### DIFF
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -22,6 +22,8 @@ import {
   ForwardRef,
   Profiler,
   Portal,
+  isMemo,
+  Memo,
 } from 'react-is';
 import { EnzymeAdapter } from 'enzyme';
 import { typeOfNode } from 'enzyme/build/Utils';
@@ -155,7 +157,26 @@ function toTree(vnode) {
         instance: null,
         rendered: childrenToTree(node.child),
       };
-
+    case FiberTags.MemoClass:
+      return {
+        nodeType: 'class',
+        type: node.elementType.type,
+        props: { ...node.memoizedProps },
+        key: ensureKeyOrUndefined(node.key),
+        ref: node.ref,
+        instance: node.stateNode,
+        rendered: childrenToTree(node.child.child),
+      };
+    case FiberTags.MemoSFC:
+      return {
+        nodeType: 'function',
+        type: node.elementType.type,
+        props: { ...node.memoizedProps },
+        key: ensureKeyOrUndefined(node.key),
+        ref: node.ref,
+        instance: null,
+        rendered: childrenToTree(node.child.child),
+      };
     case FiberTags.HostComponent: {
       let renderedNodes = flatten(nodeAndSiblingsArray(node.child).map(toTree));
       if (renderedNodes.length === 0) {
@@ -371,6 +392,16 @@ class ReactSixteenAdapter extends EnzymeAdapter {
           );
 
           const context = getMaskedContext(Component.contextTypes, unmaskedContext);
+
+          if (!isStateful && isMemo(el.type)) {
+            const InnerComp = el.type.type;
+            const wrappedEl = Object.assign(
+              (...args) => InnerComp(...args), // eslint-disable-line new-cap
+              InnerComp,
+            );
+            return withSetStateAllowed(() => renderer.render({ ...el, type: wrappedEl }, context));
+          }
+
           if (!isStateful && typeof Component === 'function') {
             const wrappedEl = Object.assign(
               (...args) => Component(...args), // eslint-disable-line new-cap
@@ -540,6 +571,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
     switch ($$typeofType) {
       case ContextConsumer || NaN: return 'ContextConsumer';
       case ContextProvider || NaN: return 'ContextProvider';
+      case Memo || NaN: return displayNameOfNode(type.type);
       case ForwardRef || NaN: {
         if (type.displayName) {
           return type.displayName;
--- a/packages/enzyme-adapter-react-16/src/detectFiberTags.js
+++ b/packages/enzyme-adapter-react-16/src/detectFiberTags.js
@@ -18,6 +18,7 @@ module.exports = function detectFiberTags() {
   const supportsMode = typeof React.StrictMode !== 'undefined';
   const supportsContext = typeof React.createContext !== 'undefined';
   const supportsForwardRef = typeof React.forwardRef !== 'undefined';
+  const supportsMemo = typeof React.memo !== 'undefined';
 
   function Fn() {
     return null;
@@ -44,6 +45,12 @@ module.exports = function detectFiberTags() {
     ClassComponent: getFiber(React.createElement(Cls)).tag,
     Fragment: getFiber([['nested']]).tag,
     FunctionalComponent: getFiber(React.createElement(Fn)).tag,
+    MemoSFC: supportsMemo
+      ? getFiber(React.createElement(React.memo(Fn))).tag
+      : -1,
+    MemoClass: supportsMemo
+      ? getFiber(React.createElement(React.memo(Cls))).tag
+      : -1,
     HostPortal: getFiber(ReactDOM.createPortal(null, global.document.createElement('div'))).tag,
     HostComponent: getFiber(React.createElement('span')).tag,
     HostText: getFiber('text').tag,
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -29,6 +29,7 @@ import {
   createRef,
   Fragment,
   forwardRef,
+  memo,
   PureComponent,
 } from './_helpers/react-compat';
 import {
@@ -60,6 +61,7 @@ describeWithDOM('mount', () => {
           return <div className="box">{children}</div>;
         }
       }
+
       class Foo extends React.Component {
         render() {
           return (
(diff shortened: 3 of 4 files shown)


alexgurr:
When I used the code directly from https://github.com/airbnb/enzyme/issues/1875 (that you have ported to enzyme) it works, but only for top level memo'd components. If there are nested components   that are wrapped in memo(), they still display in the DOM as [object Object]. Does this code address that?

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -249,6 +249,10 @@ function getEmptyStateValue() {
   return testRenderer._instance.state;
 }
 
+function isMemo(node) {
+  return typeof node.type === 'object' && node.type.$$typeof === Symbol.for('react.memo');
```
ljharb: This must not use `Symbol`; instead, can you import `REACT_MEMO_TYPE` from `react-is`, and compare to that? Specifically:
```suggestion
  return node.type && node.type.$$typeof === REACT_MEMO_TYPE;
```

Separately, can you file an issue on `react-is` to add an `isMemo` function?

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -513,6 +527,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
     switch ($$typeofType) {
       case ContextConsumer || NaN: return 'ContextConsumer';
       case ContextProvider || NaN: return 'ContextProvider';
+      case Symbol.for('react.memo') || NaN: return 'Memo';
```
ljharb:
```suggestion
      case REACT_MEMO_TYPE || NaN: return 'Memo';
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1273,6 +1274,47 @@ describe('shallow', () => {
...
+        console.log(wrapper.debug());
+        expect(wrapper.find('div')).to.have.lengthOf(1);
+      });
+
+      it('renders memoised component and pass the props', () => {
+        const Component = ({ title }) => <div><span>{title}</span></div>;
+        const ComponentWithMemo = memo(Component);
+
+        const wrapper = shallow(<ComponentWithMemo title="memo test" />);
+        console.log(wrapper.debug());
```
```suggestion
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1273,6 +1274,47 @@ describe('shallow', () => {
...
+      it('renders memoised component', () => {
+        const Component = () => <div />;
+        const ComponentWithMemo = memo(Component);
+
+        const wrapper = shallow(<ComponentWithMemo />);
+        console.log(wrapper.debug());
+        expect(wrapper.find('div')).to.have.lengthOf(1);
+      });
+
+      it('renders memoised component and pass the props', () => {
```
```suggestion
      it('renders memoized component and passes the props', () => {
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1273,6 +1274,47 @@ describe('shallow', () => {
...
       });
     });
+
+    describeIf(is('>= 16.6'), 'memo', () => {
+      it('renders memoised component', () => {
+        const Component = () => <div />;
+        const ComponentWithMemo = memo(Component);
+
+        const wrapper = shallow(<ComponentWithMemo />);
+        console.log(wrapper.debug());
```
```suggestion
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1273,6 +1274,47 @@ describe('shallow', () => {
...
+              <div>
+                <Component />
+                <Component />
+              </div>
+            );
+          }
+        }
+
+        const wrapper = shallow(<Foo />);
+        console.log(wrapper.debug());
```
```suggestion
```

jintoppy:
@alexgurr Yes. It handles nested memoized components as well.

jintoppy:
@ljharb Thanks so much for the quick response and taking time to review. I have made the changes. 
REACT_MEMO_TYPE was coming as undefined. So, I added a null check also. Not sure if I am doing that in the right way. Please check and let me know

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -49,6 +50,7 @@ const is166 = is165 && !React.unstable_AsyncMode; // 16.6+
 
 // Lazily populated if DOM is available.
 let FiberTags = null;
+const REACT_MEMO_TYPE_SYMBOL = REACT_MEMO_TYPE || Symbol.for('react.memo');
```
ljharb: There's no need at all for the `||` fallback. `react-is` always provides it.
```suggestion
```

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -249,6 +251,10 @@ function getEmptyStateValue() {
   return testRenderer._instance.state;
 }
 
+function isMemo(node) {
+  return typeof node.type === 'object' && node.type.$$typeof === REACT_MEMO_TYPE_SYMBOL;
```
ljharb:
```suggestion
  return typeof node.type === 'object' && REACT_MEMO_TYPE && node.type.$$typeof === REACT_MEMO_TYPE;
```

packages/enzyme-adapter-react-16/src/detectFiberTags.js
```diff
@@ -44,6 +45,9 @@ module.exports = function detectFiberTags() {
     ClassComponent: getFiber(React.createElement(Cls)).tag,
     Fragment: getFiber([['nested']]).tag,
     FunctionalComponent: getFiber(React.createElement(Fn)).tag,
+    Memo: supportsMemo
+      ? getFiber(React.createElement(React.memo(Fn))).tag
+      : -1,
```
ljharb: this seems like it could all go on one line?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -461,6 +462,40 @@ describeWithDOM('mount', () => {
...
+          <InnerComp />
+          <div className="bar">bar</div>
+          <div className="qoo">{foo}</div>
+        </div>
+      ));
+
+      const wrapper = mount(<Foo foo="qux" />);
+      expect(wrapper.debug()).to.equal(`<Component foo="qux">
+  <div>
+    <InnerComp>
```
ljharb: hmm - i'd expect some sort of indication here that it's memoized.

ljharb:
ahhhhh i see that `REACT_MEMO_TYPE` isn't even exported :-/ https://unpkg.com/react-is@16.6.3/cjs/react-is.development.js

That means we'll have to temporarily hardcode it with `var hasSymbol = typeof Symbol === 'function' && Symbol.for; var REACT_MEMO_TYPE = hasSymbol ? Symbol.for('react.memo') : 0xead3;`, which is a shame.

jintoppy:
@ljharb  I will then wait for the PR in react to be merged, and will update this PR.

koragan:
@jintoppy [looks like you're clear to update this PR](https://github.com/facebook/react/pull/14313)

ljharb:
@koragan the PR is merged, but react-is still hasn't been updated.

phenax:
When can we expect this patch to be published? A few of the production project that we're working on depend on it.

ljharb:
@phenax first we have to wait until react-is v16 gets updated.

amite:
react-is seems to have support for React Memo: https://github.com/facebook/react/blob/master/packages/react-is/src/ReactIs.js#L120

What exactly is pending? 🤔

ljharb:
@amite yes, it's been merged, but not yet released to npm by the react team.

rpearce:
Update: latest `react-is` version `16.6.3` was published to npm 2 days ago

rpearce:
@jintoppy My mistake! Thank you for the correction

MarceloAlves:
`react-is` [16.7.0](https://www.npmjs.com/package/react-is/v/16.7.0) was finally released 🎉

ljharb:
I've updated `react-is` in master and rebased this branch. I'm not sure what's up with the test failures; it might be an issue in enzyme's overall test setup. I'll look into it.

ljharb:
The underlying problem has been fixed and this branch rebased.

ljharb:
No update yet. I've rebased this; I'll give it another pass tomorrow.

RoystonS:
Btw, there appear to be _two_ React memo tags, not just one: MemoComponent (currently tag 14) and SimpleMemoComponent (currently tag 15). I just got bitten by this on our in-house custom-patched enzyme where we were only coping with tag 15, and I suddenly hit a 14 tag.

I'm not at all sure of the exact details as I only got hit by this a few minutes ago, but it _looks_ like React.memo(fn) gets you a simple component (tag 15) and React.memo(class) gets you a non-simple component (tag 14).

I think the PR at present would only cope with tag 15.

RoystonS:
I've done a bit more digging, and it **is** the case that there's one tag type for `React.memo(FunctionalComponent)` and a different tag type for `React.memo(ClassComponent)`

Details: https://codesandbox.io/s/kwvx4qwxyo  (tags are output to the console)

ljharb:
I've rebased and updated this to have tests for both class-based and SFC components.

https://github.com/facebook/react/issues/14807 may be a blocker.

RoystonS:
Btw, React are trying to move away from the term 'SFC':
https://twitter.com/dan_abramov/status/1057625147216220162

ljharb:
Per https://github.com/facebook/react/issues/14807#issuecomment-462147431, the shallow renderer will need to be updated before this can go in as is.

VincentLanglet:
Any update ?

If it's not easy / possible to make it works for class component, can not we go incrementally ?

Actually it does not work for both functional component and class component. A first implementation working only for functional component is still better.

ljharb:
@VincentLanglet no, it'd be better to have zero support than partial support, and either way this is blocked on react's shallow renderer change.

mattdarveniza:
Support for the case that 99% of people actually want to use memo for (functional components) seems to be a pretty good reason to merge partial support to me. Usage with classes, while supported by React, isn't actually documented anywhere, and I would say it's an edge case or at least very niche.

The failing test is only related to class components so far as I can tell, so if support for that was moved to a separate PR this would be mergeable, we're not waiting on any dependencies for functional components, right?

ljharb:
OK, that's a reasonable argument. I suppose it'll just "start working" once react-test-renderer fixes https://github.com/facebook/react/issues/14807#issuecomment-462147431, at which point we can update the tests here.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -540,6 +571,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
     switch ($$typeofType) {
       case ContextConsumer || NaN: return 'ContextConsumer';
       case ContextProvider || NaN: return 'ContextProvider';
+      case Memo || NaN: return displayNameOfNode(type.type);
```
TymchenkoOleksandr: displayName is broken here!

ljharb: Can you elaborate? a new issue would be ideal.

TymchenkoOleksandr: always getting `displayName` "Component" with `React.memo()`. need to set correct `displayName` each time after applying `React.memo()`.

ljharb: @TymchenkoOleksandr would you mind filing a new issue, and fill out the template? that way I'll be able to fix it more quickly.

PR #1924, opened 2018-12-04
Update: Adds recursive check for isEmptyRender

### LINKED ISSUE #1866: isEmptyRender Not Working for Children
**Describe the bug**

It seems like `isEmptyRender` fails when a component renders children, even if the children themselves are empty.

**To Reproduce**

This is a simplified test case that illustrates the problem:

```
class RenderChildren extends React.Component {
  render() {
    return this.props.children;
  }
}

class RenderNull extends React.Component {
  render() {
    return null;
  }
}

it("is an empty render", () => {
  let component = mount(
    <RenderChildren>
      <RenderNull />
    </RenderChildren>
  );

  expect(component.html()).toEqual(null);
  expect(component.isEmptyRender()).toEqual(true);
});
```

In this example, the first matcher passes but the second does not.

**Expected behavior**

If the component's HTML is `null`, then the component should be considered an empty render.

**Screenshots**

screen shot 2018-10-12 at 11 53 54 am

 - OS: macOS 10.13.6
 - Browser N/A
 - Version 3.7.0

sstern6:
Fixes #1866

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -3223,6 +3223,109 @@ describeWithDOM('mount', () => {
       expect(wrapper.isEmptyRender()).to.equal(data.expectResponse);
     });
 
+    describe('nested nodes', () => {
+      class RenderChildren extends React.Component {
+        render() {
+          return this.props.children;
+        }
+      }
+
+      class RenderNull extends React.Component {
+        render() {
+          return null;
+        }
+      }
+
+      it('returns true for nested elements that return null', () => {
+        const wrapper = mount((
+          <RenderChildren>
+            <RenderNull />
+          </RenderChildren>
+        ));
+
+        expect(wrapper.isEmptyRender()).to.equal(true);
+      });
+
+      it('returns false for multiple nested elements that all return null', () => {
+        const wrapper = mount((
+          <RenderChildren>
+            <div />
+          </RenderChildren>
+        ));
+
+        expect(wrapper.isEmptyRender()).to.equal(false);
+      });
+
+      it('returns false for multiple nested elements where one fringe returns a non null value', () => {
+        const wrapper = mount((
+          <RenderChildren>
+            <div>Hello</div>
+          </RenderChildren>
+        ));
+
+        expect(wrapper.isEmptyRender()).to.equal(false);
+      });
+
+      itIf(is('>= 16'), 'returns false for multiple nested elements that all return null', () => {
+        const wrapper = mount((
+          <RenderChildren>
+            <RenderNull />
+            <RenderChildren>
+              <RenderNull />
+              <div />
+            </RenderChildren>
+          </RenderChildren>
+        ));
+
+        expect(wrapper.isEmptyRender()).to.equal(false);
+      });
+
+      itIf(is('>= 16'), 'returns false for multiple nested elements where one fringe returns a non null value', () => {
+        const wrapper = mount((
+          <RenderChildren>
+            <RenderNull />
+            <RenderChildren>
+              <RenderNull />
+              <RenderNull />
+            </RenderChildren>
+            <RenderChildren>
+              <RenderNull />
+              <RenderChildren>
+                <RenderNull />
+                <RenderNull />
+                <RenderNull />
+                <div>Hello</div>
+              </RenderChildren>
+            </RenderChildren>
+          </RenderChildren>
+        ));
+
+        expect(wrapper.isEmptyRender()).to.equal(false);
+      });
+
+      itIf(is('>= 16'), 'returns true for multiple nested elements where all values are null', () => {
+        const wrapper = mount((
+          <RenderChildren>
+            <RenderNull />
+            <RenderChildren>
+              <RenderNull />
+              <RenderNull />
+            </RenderChildren>
+            <RenderChildren>
+              <RenderNull />
+              <RenderChildren>
+                <RenderNull />
+                <RenderNull />
+                <RenderNull />
+              </RenderChildren>
+            </RenderChildren>
+          </RenderChildren>
+        ));
+
+        expect(wrapper.isEmptyRender()).to.equal(true);
+      });
+    });
+
     it('does not return true for HTML elements', () => {
       const wrapper = mount(<div className="bar baz" />);
       expect(wrapper.isEmptyRender()).to.equal(false);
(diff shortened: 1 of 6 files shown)


packages/enzyme-test-suite/test/Utils-spec.jsx
```diff
@@ -693,6 +700,45 @@ describe('Utils', () => {
...
+        });
+      });
+    });
+  });
+
+  describe('renderedDive', () => {
+    const emptyNodetestData = emptyNodes();
+    const nonEmptyNodeData = nonEmptyNodes();
+
+    itWithData(emptyNodetestData, 'when renderedDive receives nodes that render validEmpty values it returns: ', (data) => {
```
ljharb: i'm not sure i'm a fan of generative test cases - i'd rather have them all laboriously written out, or, a hardcoded test that iterates through a hardcoded array of data.

packages/enzyme/src/Utils.js
```diff
@@ -320,3 +320,49 @@ export function spyMethod(instance, methodName) {
...
+export function validEmptyValues(renderedValue) {
+  return renderedValue === null || renderedValue === false;
+}
+
+export function renderedDive(nodes) {
+  let isEmptyRender = true;
+  const renderedKey = 'rendered';
+
+  function dive(renderedNodes) {
+    const isList = renderedNodes && renderedNodes.length > 0 && Array.isArray(renderedNodes);
```
ljharb: `= Array.isArray(renderedNodes) && renderedNodes.length > 0`?

packages/enzyme/src/Utils.js
```diff
@@ -320,3 +320,49 @@ export function spyMethod(instance, methodName) {
...
+    if (isList) {
+      return renderedNodes.some((n) => {
+        if (n) {
+          if ((has(n, renderedKey))) {
+            dive(n.rendered);
```
ljharb: should this be `n[renderedKey]`?

packages/enzyme/src/Utils.js
```diff
@@ -320,3 +320,49 @@ export function spyMethod(instance, methodName) {
...
+          } else if (!validEmptyValues(n.rendered)) {
```
ljharb: also here?

packages/enzyme/src/Utils.js
ljharb: or maybe we don't need `renderedKey` at all, and we can just hardcode `.rendered`

packages/enzyme/src/Utils.js
```diff
@@ -320,3 +320,49 @@ export function spyMethod(instance, methodName) {
...
+        isEmptyRender = false;
+      } else {
+        dive(renderedNodes.rendered);
+      }
+    }
+
+    return isEmptyRender;
+  }
+
+  dive(nodes);
```
ljharb: could this not be `return dive(nodes)`?

packages/enzyme-test-suite/test/Utils-spec.jsx
sstern6: Sounds good, I had originally done that but saw the generative test pattern so I went with that.  Ill fix.

packages/enzyme/src/Utils.js
sstern6: So initially I did have that.  And then in my utils test cases, I was getting inconsistent return values.  I will change back, bc this is something I prefer and research why.

packages/enzyme/src/Utils.js
sstern6: if we remove renderedKey could we keep this?

ljharb: yeah def; we should either use renderedKey everywhere, or nowhere

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -3223,6 +3223,133 @@ describeWithDOM('mount', () => {
...
+        }
+      }
+
+      class RenderNull extends React.Component {
+        render() {
+          return null;
+        }
+      }
+
+      describeIf(is('< 16'), 'rendered nodes', () => {
```
ljharb: these tests should still pass on 16+, no? we can run them unconditionally?
```suggestion
      describe('rendered nodes', () => {
```

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -3223,6 +3223,133 @@ describeWithDOM('mount', () => {
...
+              <RenderNull />
+            </RenderChildren>
+          ));
+
+          expect(wrapper.isEmptyRender()).to.equal(true);
+        });
+      });
+
+      describeIf(is('>= 16'), 'rendered nodes', () => {
+        it('returns true for nested elements that return null', () => {
```
ljharb: if the above describe runs unconditionally, this test becomes redundant and can be removed

packages/enzyme-test-suite/test/Utils-spec.jsx
```diff
@@ -693,6 +697,54 @@ describe('Utils', () => {
     });
   });
 
+  describe('validEmptyValues', () => {
+    describe('when recieves "false" or null', () => {
```
ljharb:
```suggestion
    describe('when receives `false` or `null`', () => {
```

packages/enzyme/src/Utils.js
```diff
@@ -351,3 +351,48 @@ export function shallowEqual(objA, objB) {
...
+
+export function renderedDive(nodes) {
+  let isEmptyRender = true;
+
+  function dive(renderedNodes) {
+    const isList = Array.isArray(renderedNodes) && renderedNodes.length > 0;
+    if (isList) {
+      renderedNodes.some((n) => {
+        if (n) {
+          if ((has(n, 'rendered'))) {
```
ljharb:
```suggestion
          if (has(n, 'rendered')) {
```

packages/enzyme/src/Utils.js
```diff
@@ -351,3 +351,48 @@ export function shallowEqual(objA, objB) {
...
+            isEmptyRender = false;
+          }
+        } else if (!validEmptyValues(n)) {
+          isEmptyRender = false;
+        }
+
+        return !isEmptyRender;
+      });
+    } else {
+      /* eslint-disable no-lonely-if */
```
ljharb: why can't this be `else if`?

packages/enzyme/src/Utils.js
```diff
@@ -351,3 +351,48 @@ export function shallowEqual(objA, objB) {
 
   return true;
 }
+
+export function validEmptyValues(renderedValue) {
```
ljharb: since this is a predicate, can we name it something like `isEmptyValue`?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
sstern6: good catch.

Issue #1872, opened 2018-10-18
Child component props aren't updated after call to setState in async handler


Bargs:
I'm having trouble testing an event handler that does an async operation followed by a setState. Even when I wait for the setState to complete by awaiting a promise resolved from inside its callback, the component doesn't seem to fully re-render. What's weird is that I can confirm that the parent state is updated but props on a sub component that rely on that state are not. Even stranger, text that comes from those props inside the sub component is updated like normal.  I created a small script to reproduce the behavior. Note that after the onClick handler and setState finish this child component props are still `{ id: 'childDiv', counter: 1 }`. Only after a forced update do they change.

```jsx
import { mount } from 'enzyme';
import React, { Component } from 'react';
import Promise from 'bluebird';

class TestComponent extends Component {

  state = {
    counter: 1,
  };

  handleClick = async () => {
    await Promise.delay(100);
    return await new Promise((resolve) => {
      this.setState({ counter: 2 }, () => {
        resolve();
      });
    });
  };

  render() {
    return <div id={'parentDiv'} onClick={this.handleClick}>
      <TestSubComponent id={'childDiv'} counter={this.state.counter}/>
    </div>;
  }
}

class TestSubComponent extends Component {

  render() {
    return <div>{this.props.counter}</div>;
  }
}

describe('foo', () => {

  it('should foo', async () => {
    const component = mount(<TestComponent/>);

    // prints {"counter":1}
    console.log('=== parent state before click');
    console.log(JSON.stringify(component.state()));

    // prints { id: 'childDiv', counter: 1 }
    console.log('=== sub component props before click');
    console.log(component.find(TestSubComponent).props());

    // prints 1
    console.log('=== sub component props before click');
    console.log(component.find(TestSubComponent).text());

    await component.find('#parentDiv').props().onClick();

    // prints {"counter":2}
    console.log('=== parent state after click');
    console.log(JSON.stringify(component.state()));

    // prints { id: 'childDiv', counter: 1 }
    console.log('=== sub component props after click');
    console.log(component.find(TestSubComponent).props());

    // prints 2
    console.log('=== sub component text after click');
    console.log(component.find(TestSubComponent).text());

    component.update();

    // prints { id: 'childDiv', counter: 2 }
    console.log('=== sub component props after update');
    console.log(component.find(TestSubComponent).props());

  });

});
```

#### API

- [x] shallow
- [x] mount
- [ ] render

#### Version

| library             | version
| ------------------- | -------
[...]

PR #1930, opened 2018-12-07
Test Case for props in async handler

k3ithl1m:
Written the test case to test our props when there is an async handler #1872 @ljharb


packages/enzyme-test-suite/package.json
```diff
@@ -39,7 +39,8 @@
     "object-inspect": "^1.6.0",
     "prop-types": "^15.6.2",
     "semver": "^5.5.0",
-    "sinon": "^4.1.6"
+    "sinon": "^4.1.6",
+    "bluebird": "^3.5.3"
```
ljharb: Let’s not add this dep.

packages/enzyme-test-suite/package.json
```diff
@@ -55,4 +56,4 @@
     "eslint-plugin-react": "^7.11.1",
     "react-is": "^16.6.0"
   }
-}
+}
```
ljharb: All files should have trailing newlines.

PR #1931, opened 2018-12-08
wrote the test case for props in async handler(new PR)

### LINKED ISSUE #1872: Child component props aren't updated after call to setState in async handler

k3ithl1m:
Written the test case to test our props when there is an async handler #1872 @ljharb

Replaces #1930.

Closes #1872.

### DIFF
--- a/packages/enzyme-test-suite/test/_helpers/index.jsx
+++ b/packages/enzyme-test-suite/test/_helpers/index.jsx
@@ -150,3 +150,9 @@ export function generateEmptyRenderData() {
     { message: 'React element', value: <noscript />, expectResponse: false },
   ];
 }
+
+export function delay(ms) {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+}
--- a/packages/enzyme-test-suite/test/shared/methods/props.jsx
+++ b/packages/enzyme-test-suite/test/shared/methods/props.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { expect } from 'chai';
 
 import {
+  delay,
   describeIf,
   itIf,
 } from '../../_helpers';
@@ -119,6 +120,51 @@ export default function describeProps({
         expect(wrapper.props()).to.be.an('object').that.has.all.keys({
           'data-is-global': false,
           'data-is-undefined': true,
+        });
+      });
+    });
+
+    describe('props in async handler', () => {
+      class TestComponent extends React.Component {
+        constructor(props) {
+          super(props);
+          this.state = { counter: 1 };
+          this.handleClick = this.handleClick.bind(this);
+        }
+
+        handleClick() {
+          return delay(100).then(() => new Promise((resolve) => {
+            this.setState({ counter: 2 }, () => {
+              resolve();
+            });
+          }));
+        }
+
+        render() {
+          const { counter } = this.state;
+          return (
+            <div id="parentDiv" onClick={this.handleClick}>
+              <TestSubComponent id="childDiv" counter={counter} />
+            </div>
+          );
+        }
+      }
+
+      class TestSubComponent extends React.Component {
+        render() {
+          const { counter } = this.props;
+          return <div>{counter}</div>;
+        }
+      }
+
+      it('child component props should update after call to setState in async handler', () => {
+        const wrapper = Wrap(<TestComponent />);
+        expect(wrapper.find(TestSubComponent).props()).to.eql({ id: 'childDiv', counter: 1 });
+        const promise = wrapper.find('#parentDiv').props().onClick();
+        expect(wrapper.find(TestSubComponent).props()).to.eql({ id: 'childDiv', counter: 1 });
+        return promise.then(() => {
+          wrapper.update();
+          expect(wrapper.find(TestSubComponent).props()).to.eql({ id: 'childDiv', counter: 2 });
         });
       });
     });


packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -3646,6 +3646,51 @@ describeWithDOM('mount', () => {
...
+    describe('props in async handler', () => {
+      class TestComponent extends React.Component {
+        constructor(props) {
+          super(props);
+          this.state = { counter: 1 };
+          this.handleClick = this.handleClick.bind(this);
+          this.delay = this.delay.bind(this);
+        }
+
+        async handleClick() {
```
ljharb: we can't actually use async functions in enzyme tests, but in this case it shouldn't matter since this function returns a promise.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -3646,6 +3646,51 @@ describeWithDOM('mount', () => {
...
+          this.delay(100);
```
ljharb: without `await this.delay(100)`, this line has no effect. Try this instead:
```suggestion
          return this.delay(100).then(() => {
            return new Promise(…)
          });
```

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -3646,6 +3646,51 @@ describeWithDOM('mount', () => {
...
+          return new Promise((resolve) => {
+            this.setState({ counter: 2 }, () => {
+              resolve();
+            });
+          });
+        }
+
+        delay(ms) {
```
ljharb: this can just be a standalone function; it doesn't need to be an instance method.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -3646,6 +3646,51 @@ describeWithDOM('mount', () => {
...
+          return new Promise((resolve) => {
+            setTimeout(resolve, ms);
+          });
+        }
+
+        render() {
+          return (
+            <div id={'parentDiv'} onClick={this.handleClick}>
```
```suggestion
            <div id="parentDiv" onClick={this.handleClick}>
```

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -3646,6 +3646,51 @@ describeWithDOM('mount', () => {
...
+              <TestSubComponent id={'childDiv'} counter={this.state.counter} />
```
```suggestion
              <TestSubComponent id="childDiv" counter={this.state.counter} />
```

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -3646,6 +3652,45 @@ describeWithDOM('mount', () => {
...
+          );
+        }
+      }
+
+      class TestSubComponent extends React.Component {
+        render() {
+          return <div>{this.props.counter}</div>;
+        }
+      }
+      it('child component props should update after call to setState in async handler', async () => {
```
ljharb: this also needs to not be an async function

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -3646,6 +3652,45 @@ describeWithDOM('mount', () => {
...
+
+      class TestSubComponent extends React.Component {
+        render() {
+          return <div>{this.props.counter}</div>;
+        }
+      }
+      it('child component props should update after call to setState in async handler', () => {
+        const component = mount(<TestComponent />);
+        component.find('#parentDiv').props().onClick();
+        expect(component.find(TestSubComponent).props()).to.eql({ id: 'childDiv', counter: 2 });
```
ljharb: so, i would expect this test to fail - because the `setState` has `delay(100)` on it.

It's impossible for the test to know when that async action is completed, unless it hardcodes a `delay(100)`, or unless the `onClick` returns a promise.

k3ithl1m: Yes. It will fail. I can work on trying to fix the bug too. Does the test look right though?

ljharb: No, to clarify - the test itself can't possibly pass. It's not possible for a test to magically wait for all async actions to be completed.

k3ithl1m: Ah okay. That makes sense. I was working on it last night and am trying to figure out.

I thought of setting a
‘Const promise = component.find(‘parentDiv’).props().onClick()’
And then promise.resolve with the expect in it. But it doesnt seem to work. I was thinking that it might be a reasonable approach because the handleClick returns a promise. 

Other than that, I assume that having a delay(100) is not the best way to approach this problem. 

What’s your thought?

ljharb: I'm starting to think there's not actually a bug here, and that the original issue is invalid.

ljharb: ok actually, the original issue in #1872 is still a bug - i think the test case needs a bit of tweaking to more closely match that (without using await or async functions)

k3ithl1m: I'm sorry but how should I go around writing this test? Or more, how am I able to know that I'm writing the test correctly?

PR #1934, opened 2018-12-10
Fix .exists not accepting any EnzymeSelector

btimo:
`.exists`, since [#1695](https://github.com/airbnb/enzyme/pull/1695), accepts an optional selector. 

Unfortunately, this optional parameter was limited to the `string` type whereas the documentation inform us it can be any valid [EnzymeSelector](https://airbnb.io/enzyme/docs/api/selector.html).

This commit fixes this limitation by accepting any type for the parameter and using `.find()` internal functions to throw any necessary `TypeError`.


packages/enzyme/src/ReactWrapper.js
```diff
@@ -1073,14 +1073,11 @@ class ReactWrapper {
    * Returns true if the current wrapper has nodes. False otherwise.
    * If called with a selector it returns `.find(selector).exists()` instead.
    *
-   * @param {String|Function} selector (optional)
+   * @param {String|Function|Object} selector (optional)
```
ljharb: The signature for `find` is `* @param {String|Function} selector` - should that be changed, or should this one stay the same?

btimo: Not sure about what you're asking:
- I changed `find` signature to `* @param {String|Function|Object} selector`
- Internally, `find` will use the `selector` with functions that accepts `{String|Function|Object}`
- `EnzymeSelector` and `find` documentation show that you can call it with an `Object`:
   - https://airbnb.io/enzyme/docs/api/selector.html#5-object-property-selector
   - https://airbnb.io/enzyme/docs/api/ShallowWrapper/find.html

ljharb: I mean that the jsdoc comment for `find` in both ShallowWrapper and ReactWrapper say "string or function" - so it seems like it should be "string or function or object" as well

PR #1945, opened 2018-12-16
[Fix]: `selectors`: support universal selector

### LINKED ISSUE #1158: Support universal selector
Right now we parse selectors with the universal selector `*` but we don't have any logic to handle it.

chenesan:
Fix #1158 .

Note that rst-selector-parser has universal selector support now so we only need to change selectors code in enzyme.


chenesan:
@ljharb added 4 more tests like `.top-div ~ *` `#wrap-div > *` and force pushed the commit

docs/api/selector.md
```diff
@@ -13,7 +13,9 @@ follows:
 - class syntax (`.foo`, `.foo-bar`, etc.)
 - element syntax (`input`, `div`, `span`, etc.)
 - id syntax (`#foo`, `#foo-bar`, etc.)
-- attribute syntax (`[href="foo"]`, `[type="text"]`, and the other attribute selectors listed [here](https://developer.mozilla.org/en-US/docs/Learn/CSS/Introduction_to_CSS/Attribute_selectors).)
+- attribute syntax (`[href="foo"]`, `[type="text"]`, and the other attribute
+- universal syntax (`*`)
+selectors listed [here](https://developer.mozilla.org/en-US/docs/Learn/CSS/Introduction_to_CSS/Attribute_selectors).)
```
eps1lon: This line needs to be moved back to "attribute syntax". It currently renders
>- attribute syntax ([href="foo"], [type="text"], and the other attribute
>- universal syntax (*) selectors listed here.)

eps1lon:
Does this only select DOM nodes or also components (class or function etc).

chenesan:
Currently it's only for DOM nodes like other string css selector.

docs/api/selector.md
chenesan: Oops, thanks!

eps1lon:
Could we add a test to ensure that this is not returning other components? I'm asking because I currently use a helper getter to find the outermost DOM node in a subtree with findWhere and this could potentially replace that.

ljharb:
@chenesan all the selectors in enzyme select on all kinds of nodes, not just DOM nodes.

ljharb:
@eps1lon you may want `.hostNodes()` to filter out non-DOM nodes

chenesan:
@ljharb Sorry for misleading! I just added one test to show that universal selector also select component node.

chenesan:
@ljharb  Is there anything I can do to progress this one? Thanks :-)

docs/api/selector.md
ljharb: This hasn't been updated yet.

ljharb:
@chenesan 
https://github.com/airbnb/enzyme/pull/1945#discussion_r251726993

docs/api/selector.md
chenesan: fixed in eee9542

Issue #1958, opened 2018-12-31
[Feature Request] createContext() not supported in enzyme.


hyochan:
I'd like to know how to test `provider` for `context api`. I've found some tutorials on testing `consumer` but no tutorial exists for testing `provider` itself.

```
import React, { createContext } from 'react';
import { Text, TouchableOpacity, View } from 'react-native';
import ProfileModal from '../components/shared/ProfileModal';

const ProfileModalContext = createContext();

export const ProfileModalConsumer = ProfileModalContext.Consumer;

export class ProfileModalProvider extends React.Component {
  static modal;

  state = {
    user: null,
  };

  actions = {
    setModal: (v) => {
      console.log('setModal', v);
      this.modal = v;
    },
    showModal: (user) => {
      console.log('showModal');
      this.setState({ user }, () => {
        if (this.modal) {
          console.log('openModal');
          this.modal.showAddBtn(true);
          this.modal.open();
        }
      });
    },
    onChatPressed: (navigation) => {
      if (this.modal) {
        this.modal.close();
      }
      navigation.navigate('Chat');
    }
  };

  render() {
    const { state, actions, modal } = this;
    const value = { state, actions, modal };
    return (
      <ProfileModalContext.Provider value={value}>
        {this.props.children}
        <ProfileModal
          id='profile_modal'
          ref={(v) => {
            console.log('v', v);
            this.modal = v;
          }}
          onChatPressed={ () => this.actions.onChatPressed(this.props.navigation) }
        />
      </ProfileModalContext.Provider>
    );
  }
}
```
I've tried making test code for above `provder`.
```
import 'react-native';
import * as React from 'react';
import { ProfileModalProvider } from '../ProfileModalProvider';

// Note: test renderer must be required after react-native.
import renderer from 'react-test-renderer';
import { shallow, render } from 'enzyme';

describe('rendering test', () => {
  const wrapper = shallow(
    <ProfileModalProvider />,
  ).dive();

  it('renders as expected', () => {
    expect(wrapper).toMatchSnapshot();
  });
});

describe('interaction', () => {
  const context = {
    state: {
      user: null,
    },
    actions: {
      setModal: jest.fn(),
      showModal: jest.fn(),
      onChatPressed: jest.fn(),
    }
  };
  let props;
  let wrapper;
  beforeEach(() => {
    props = {
      navigation: {
        navigate: jest.fn(),
      },
      ...context,
    };
[...]

PR #1960, opened 2019-01-02
Add ability to mount()/shallow() a node inside a wrapping component

minznerjosh:
### Overview

This PR adds an API for `mount()`ing a `node` inside of a custom `wrappingComponent`.

```jsx
import { Provider } from 'react-redux';
import { Router } from 'react-router';
import store from './my/app/store';
import mockStore from './my/app/mockStore';

function MyProvider(props) {
  const { children, customStore } = props;

  return (
    <Provider store={customStore || store}>
      <Router>
        {children}
      </Router>
    </Provider>
  );
}

const wrapper = mount(<MyComponent />, {
  wrappingComponent: MyProvider,
});
```

It is also possible to get a `ReactWrapper` around the `wrappingComponent`:

```jsx
const provider = wrapper.getWrappingComponent();
provider.setProps({ customStore: mockStore });
wrapper.update(); // update root wrapper after wrappingComponent props changed
```

Related to #1958.


docs/api/ReactWrapper/setProviders.md
```diff
@@ -0,0 +1,66 @@
...
+  <ContextC.Provider value="React" />,
+]);
+wrapper.text(); // We Love React
+```
+
+#### Common Gotchas
+
+- `.setProviders()` can only be used on a wrapper that was initially created with a call to `mount()`
+- `.setProviders()` is only supported in React >= 16.3
+- `.setProviders()` requires `enzyme-adapter-react-16@>=1.8` or `enzyme-adapter-react-16.3@>=1.5`
```
minznerjosh: These version will need to be updated depending on when/if this makes it in.

ljharb:
Adding createContext support in any form is a major effort; I want to set expectations early that this PR may not make it in at all, and that support is required to be identical for both mount and shallow.

docs/api/ReactWrapper/setProviders.md
ljharb: Let's remove version numbers entirely; there's no need to specify it here.

packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
```diff
@@ -312,6 +316,12 @@ class ReactSixteenThreeAdapter extends EnzymeAdapter {
         return fn();
         // return ReactDOM.unstable_batchedUpdates(fn);
       },
+      setProviders(newProviders) {
+        if (newProviders.some(provider => !isContextProvider(provider))) {
+          throw new Error('ReactWrapper::setProviders() may only be called with Provider elements');
+        }
+        instance.setProviders(newProviders);
+      },
```
minznerjosh: It seems like the more idiomatic way to do this would be to handle the update via the `render()` method. However, I'm pretty sure (correct me if I'm wrong) changing the signature of that method would need to force this feature into a new major version.

Adding a `setProviders()` method here at least allows enzyme to throw a helpful error if react/the adapter is at too old of a version.

docs/api/ReactWrapper/setProviders.md
```diff
@@ -0,0 +1,66 @@
...
+  <ContextA.Provider value="We" />,
+  <ContextC.Provider value="React" />,
+]);
+wrapper.text(); // We Love React
+```
+
+#### Common Gotchas
+
+- `.setProviders()` can only be used on a wrapper that was initially created with a call to `mount()`
+- `.setProviders()` is only supported in React >= 16.3
```
ljharb: mount itself has no direct tie to React; it should be able to work as long as the adapter supports it (which might not be react at all)

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
```diff
@@ -32,6 +32,29 @@ const makeValidElementType = (adapter) => {
...
+ *      </Bar>
+ *    </Foo>
+ *  )
+ */
+const reduceElements = (elements) => {
+  const reversed = [...elements];
+  reversed.reverse();
+
+  return reversed.reduce((childElement, parentElement) => (
+    React.cloneElement(parentElement, null, childElement)
```
ljharb: cloning elements is very slow and should be avoided. why is it needed here?

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
```diff
@@ -79,17 +103,27 @@ export default function createMountWrapper(node, options = {}) {
       return inst;
     }
 
+    setProviders(providers) {
+      const { providers: currentProviders } = this.state;
+      const providerTypes = providers.map(provider => provider.type);
+
+      this.setState({
+        providers: currentProviders.filter(provider => (
+          !providerTypes.includes(provider.type)
```
ljharb: this API means there's never any way to remove a provider, only to replace existing ones or add new ones.

Separately, `providers` being an array implies that the order matters - why and how? If it matters because it's nested into a tree, then it seems much simpler to only ever allow a *single* provider element, and come up with an alternative solution that lets the user determine their own nesting yet also indicate where children should be rendered.

minznerjosh:
@ljharb I figured that'd be the case! Just wanted to get something up here to get a conversation started.

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
minznerjosh: The goal is to convert the provider elements the user passes into a tree:

So this:
```jsx
mount(<MyComponent />, {
  providers: [
    <ContextA.Provider value="A" />,
    <ContextB.Provider value="B" />,
    <ContextC.Provider value="C" />,
  ],
});
```

Is rendered as:
```jsx
<ContextA.Provider value="A">
  <ContextB.Provider value="B">
    <ContextC.Provider value="C">
      <MyComponent />
    </ContextC.Provider>
  </ContextB.Provider>
</ContextA.Provider>
```

And I figured `cloneElement()` was the most idiomatic way to do this. I'm also working under the assumption that _mutating_ the providers the user passes in would be a bad idea. Would a more manual
```jsx
// not sure if this will actually work
{
  ...parentElement,
  props: {
    ...parentElement.props,
    children: childElement,
  },
}
```

be preferable, or is it the copying itself that is too slow to be acceptable?

docs/api/ReactWrapper/setProviders.md
minznerjosh: Sounds good. Is there any prior art for what to do in a case where an adapter does _not_ support a feature? For example, I'm assuming we'll never want to add support for this API to adapters that handle react@<16.3.

packages/enzyme/src/ReactWrapper.js
```diff
@@ -357,6 +363,64 @@ class ReactWrapper {
...
+   * ```
+   * @param {Array<ReactElement>} providers
+   * @returns {ReactWrapper}
+   */
+  setProviders(providers) {
+    if (this[ROOT] !== this) {
+      throw new Error('ReactWrapper::setProviders() can only be called on the root');
+    }
+    if (!this[RENDERER].setProviders) {
+      throw new Error('ReactWrapper::setProviders() can only be called with react >= 16.3 and a matching enzyme adapter.');
```
minznerjosh: @ljharb thanks for the clarification on the separation between enzyme itself and its adapters! This is the only file I've changed in the `enzyme` package, so I'm gathering that the problem here is just that I'm mentioning react versions explicitly in this error message (and the one at the top of the file.) Is the solution just to change these errors to only talk about _adapter_ compatibility and not _react_ compatibility?

minznerjosh:
@ljharb So, I'm sure you've noticed that this PR does not add support for `shallow()`. I'll give that a whirl, but I think it'd be best to hammer out an API before I do. Thanks again for such a speedy first review! Cheers!

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
minznerjosh: Ah got it! I was a bit confused by what you meant by "a single provider". I was thinking about the new context API in terms of individual values (as is the case with the old API), but perhaps that was me trying to make the new API into something that it is not.

The way you compose multiple contexts in the new React API is by nesting components—perhaps the enzyme API should reflect that, as you've pointed out! So, what do you think about this:

```jsx
mount(<MyComponent />, {
  providers: (
    <ContextA.Provider value="A">
      <ContextB.Provider value="B">
        <ContextC.Provider value="C" />
      </ContextB.Provider>
    </ContextA.Provier>
  ),
});
```

And then changing via context by

```jsx
wrapper.setProviders(
  <ContextA.Provider value="A">
    <ContextB.Provider value="foo">
      <ContextC.Provider value="C" />
    </ContextB.Provider>
  </ContextA.Provier>
);
```

My only concern is that this API makes it a bit challenging to change just a _single_ provider's value without having to re-specify the entire tree. I think it will be common to be dealing with a bunch of different contexts from different libraries, all of which are required for a component to render. But each individual _spec_ will deal with the how the component responds to just _one_ of those contexts changing. So, what do you think about (in addition to the `.setProviders()` API) having an API like this:

```jsx
wrapper.updateProvider(<ContextB.Provider value="foo" />);
```

Which would update _just_ that provider, wherever it is in the tree, and throw an error if the provider is not already in the tree.

Also, I don't feel strongly about any of these names if you'd like them to be different. I am leaning towards keeping the plural "providers" because I think it makes it a bit more clear that the element can have many nested providers, though I do realize you are only passing a single react element.

ljharb: Re-specifying the entire tree is already the case if you want to change a child of the element you wrapped, for example - i'm also not clear on the use case of changing providers after initially wrapping.

Maybe instead of this feature being specific to `createContext` providers, it could be more general - something like, "wrapWith". A single render prop function that gets passed a child, and wraps it as needed.

Example:
```jsx
const wrapper = mount(<Foo />, {
  wrapWith(root) {
    return (
      <ContextA.Provider value="A">
        <ContextB.Provider value="foo">
          <ContextC.Provider value="C">
            {root}
          </ContextC.Provider>
        </ContextB.Provider>
      </ContextA.Provider>
    );
  },
});
```
The API could throw if your wrapWith function failed to render the child (solo, ie, with no siblings), and then `wrapper` in this case would be `<Foo />`, not any of the providers, but they'd still be exercised in the rendering path.

packages/enzyme-test-suite/package.json
```diff
@@ -55,4 +55,4 @@
     "eslint-plugin-react": "^7.11.1",
     "react-is": "^16.7.0"
   }
-}
+}
```
ljharb: please revert this so that all files have trailing newlines

minznerjosh: Whoops. Didn't mean to `git add` this.

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
minznerjosh: I think the use-case is something like this:

Imagine you have a [`react-flag`](https://github.com/garbles/flag)-like library and some future version of `react-redux` that uses the new context API. You'd render your component like this (using the wrapWith API):

```jsx
// my app's feature flags
const flags = {
  foo: false,
  bar: false,
};

const wrapper = mount(<Foo />, {
  wrapWith: root => (
    <FlagProvider value={flags}>
      <ReduxProvider value={store}>{root}</ReduxProvider>
    </FlagProvider>
  ),
});
```

And now we want to test that some side-effect in `<Foo />`'s `componentDidUpdate` gets called when the feature flag changes. We need some way of changing `<FlagProvider />`'s `value`, and it would be nice if we didn't have to re-setup the `<ReduxProvider />` when we do that.

Also, if our component starts depending on a new provider in the future, it'd be really nice to only have to add it to the options we pass enzyme, and not every place where we're just changing a single provider's value.

I'm guessing the API for changing the `wrapWith` function could be something like

```jsx
wrapper.setWrapWith(root => (
  <FlagProvider value={{ ...flags, bar: true }}>
    <ReduxProvider value={store}>{root}</ReduxProvider>
  </FlagProvider>
));
```

But that doesn't address my concern about having to respecify the entire tree. One solution that comes to mind (just spitballing here) is making `wrapWith` a _component_:

```jsx
const wrapper = mount(<Foo />, {
  wrapper: {
    component: (props) => {
      const { flags, store, children } = props;

      return (
        <FlagProvider value={{ ...flags, bar: true }}>
          <ReduxProvider value={store}>{children}</ReduxProvider>
        </FlagProvider>
      );
    },
    props: { store: createStore, flags },
  },
});
```

And then having an API that lets you set the wrapper component's props:

```jsx
wrapper.setWrapperProps({
  flags: {
    ...flags,
    foo: true,
  },
});
```

(`wrapWith` seemed like a better name for a function than a component so I changed it to just `wrapper`; but it also seems a bit confusing because `wrapper` is the conventional name for a `ReactWrapper` instance.)

ljharb: (We can bikeshed the names later)

OK, so you're suggesting the user be able to provide a "wrapping component", and provide explicit `setProps`/`setState` access on it - what about instead, `wrapper.getWrappingComponent()`, that returned a normal enzyme wrapper around that component - providing you with the same APIs?

minznerjosh: I like that idea a lot! I'll implement and then drop you a comment here so we can discuss. I really appreciate the time you've taken to work through this with me. Thank you!

packages/enzyme-adapter-utils/src/RootFinder.jsx
```diff
@@ -0,0 +1,17 @@
...
+export default class RootFinder extends React.Component {
+  render() {
+    const { children } = this.props;
+
+    return children;
+  }
+}
+RootFinder.propTypes = {
+  children: PropTypes.node.isRequired,
+};
```
minznerjosh: This component is only ever used in `createMountWrapper`, so I think the code would actually be a bit easier to understand if it just lived in that file. But eslint forbids that so I stuck it here. Do you think this could warrant a rule override?

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
```diff
@@ -79,26 +87,53 @@ export default function createMountWrapper(node, options = {}) {
...
+      const { mount, props, wrappingComponentProps } = this.state;
       if (!mount) return null;
-      return (
-        <Component {...props} />
-      );
+      const component = <Component {...props} />;
+      if (WrappingComponent) {
+        return (
+          <WrappingComponent {...wrappingComponentProps}>
+            <RootFinder ref={this.setRootFinderInstance}>{component}</RootFinder>
```
minznerjosh: We need a reliable way to get the `node` that was passed as the first argument to `mount()` and this seemed like the simplest way. Is it acceptable? Are there going to be performance implications of using a `ref`?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -164,6 +164,156 @@ describeWithDOM('mount', () => {
       expect(() => wrapper.state('key')).to.throw('ReactWrapper::state("key") requires that `state` not be `null` or `undefined`');
     });
 
+    describeIf(is('>= 0.14') ,'wrappingComponent', () => {
```
minznerjosh: There's no reason this feature couldn't work in react 13, but the tests don't because react 13 uses owner-based context instead of parent-based. Is it worth coming up with a different test methodology for react 0.13, or are we okay with not supporting this feature in the 0.13 adapter? (I'm not sure how useful this feature is without parent-based context.)

packages/enzyme/src/ReactWrapper.js
```diff
@@ -104,19 +105,24 @@ class ReactWrapper {
...
       privateSetNodes(this, this[RENDERER].getNode());
     } else {
-      privateSet(this, UNRENDERED, null);
       privateSet(this, RENDERER, root[RENDERER]);
+      privateSet(this, WRAPPING_COMPONENT_RENDERER, root[WRAPPING_COMPONENT_RENDERER]);
       privateSet(this, ROOT, root);
       privateSetNodes(this, nodes);
       privateSet(this, ROOT_NODES, root[NODES]);
     }
+    privateSet(this, UNRENDERED, nodes);
```
minznerjosh: `UNRENDERED` needs to be set to call `.setProps()` on the `wrappingComponent`. It didn't seem like there was any harm in setting it to `nodes` regardless of `!root`. (All the tests still passed.)

packages/enzyme/src/ReactWrapper.js
```diff
@@ -221,6 +227,27 @@ class ReactWrapper {
...
+    if (this[ROOT] !== this || this[RENDERER] === this[WRAPPING_COMPONENT_RENDERER]) {
+      throw new Error('ReactWrapper::getWrappingComponent() can only be called on the root');
+    }
+    if (!this[OPTIONS].wrappingComponent) {
+      throw new Error('ReactWrapper::getWrappingComponent() can only be called on a wrapper that was originally passed a `wrappingComponent` option');
+    }
+    const node = this[WRAPPING_COMPONENT_RENDERER].getNode();
+    const wrapper = this.wrap(node);
+    privateSet(wrapper, ROOT, wrapper);
+    privateSet(wrapper, RENDERER, this[WRAPPING_COMPONENT_RENDERER]);
```
minznerjosh: This is a bit hacky. Is it acceptable? We _don't_ want this new `ReactWrapper` to render itself, but we _do_ want it to be a root (so we can `.setProps()` on it.)

We could also decide to not support `.setProps()` and make it a non-root. In that case, we could just get by using `.setState()` to change the context.

minznerjosh:
@ljharb

**TL;DR;**
- I implemented the API we discussed for `mount()`
- Some of the things I did might be a bit too hacky
- Things could be simplified and made less-hacky if `.getWrappingComponent()` did not return a root
- I'm not sure how much sense this API makes for `shallow()`; any possibility of moving forward without `shallow()`?

I took a stab at implementing the API we discussed for `mount()`. Because I wanted `.setProps()` to work, I made the `.getWrappingComponent()` wrapper a root (of sorts.) This led me to do some things that _might_ be a bit questionable. Because we can now `.setState()` non-roots, perhaps we can live with `.getWrappingComponent()` _not_ returning a root. That, however, would create complications around `.getWrappingComponent().update()`. (It would update the root, not the `wrappingComponent`.)

I don't use `shallow()` much in my day-to-day, but as I started digging into how it works, I started to doubt how useful this feature would be for `shallow()`. The only reason I can think of for the `wrappingComponent` feature to exist is for users can set up context (new or legacy) without changing the root wrapper. I think this is a need that arises from _integration_-style testing, whereas `shallow()` is strictly for unit-style tests. If we _were_ to just shallow-render the `wrappingComponent` and then effectively `.dive()` to the root, the effect would be as if we didn't have a `wrappingComponent` at all.

Perhaps this is because, as others have pointed out, `.dive()` doesn't currently call `.getChildContext()` or look at `.contextTypes`. I imagine we could do something where we dive through every component in the `wrappingComponent`, collect its context, and then pass that context when we shallow render the root node, but that seems a bit complicated and would only work for _legacy_ context, not the new context API. Perhaps we could do something similar with the new context `Provider`/`Consumer`, though.

But I still doubt how much sense `wrappingComponent` makes as a `shallow()` option. If `shallow()` exists to isolate components from each other, it doesn't seem intuitive to me to support wrapping your component in a `wrappingComponent` because shallow rendering, by definition, will shield the root component from the effects of `wrappingComponent`.

So, my question is, is there any possibility you'd be open to this being a `mount()`-only feature? If `shallow()` support _really_ is required, what should the behavior for it be? If we decide that the `wrappingComponent` feature _only_ exists for providing context, then there would be a lot of work required in `shallow()` to support that. If `wrappingComponent` exists literally just to wrap the enzyme root in another component, and follow the _existing behavior_ for doing that with `mount()`/`shallow()`, then I don't think it's a useful feature for `shallow()`.

packages/enzyme-adapter-utils/src/createMountWrapper.jsx
ljharb: we should avoid using `ref`, yes.

minznerjosh:
So, is this the strategy you're looking for in `shallow()`?

1. Do a _full_ render of the `wrappingComponent`, but without a dependency on the DOM, stopping when we reach the `node` passed to `shallow(node)`
2. Collect all the context values that were set in that full render of `wrappingComponent` (both new and legacy)
3. Shallow-render the `node`, passing in all the context we collected in step 2 (using some new functionality that doesn't exist yet for new context)

ljharb:
Yep! sounds right to me (where "full render" is actually multiple shallow renders and dives)

minznerjosh:
Yup yup. Okay, in that case, I think I'm going to hit the pause button on this PR and try to add support for `dive()`ing `<Consumer />`s and `<Provider />`s in a separate PR. That seems like a separate issue that should be fixed independently, and fixing that issue should lay the foundation for implementing this feature.

ljharb:
That sounds like an awesome plan!

kumar303:
This PR looks like it will close https://github.com/airbnb/enzyme/issues/1958 💃

ljharb:
@minznerjosh seems like this can be picked up again :-)

minznerjosh:
@ljharb I think we need #1966 first! (If we’re trying to solve the new context problem, which is how this all started haha.)

ljharb:
hmm, do we? it seems like they can go in parallel, and this one is more of a foundational API versus just being focused on Context.

minznerjosh:
@ljharb You're totally right. Though it won't solve the original use case, there's no reason they can't be developed in parallel. I'll pick this back up ASAP!

But first I'd like to get your thoughts on something. The question I'm trying to answer is: should the wrapper returned by `getWrappingComponent()` be a "root" wrapper, or a child of the main wrapper?

* Making it a root
  * Pros
    * You can call `setProps()` on it (one of the original motivations behind this API)
  * Cons
    * Calling things like `setProps()` would only update itself, not the main wrapper
    * You can also call things like `unmount()`/`detach()` which we probably don't want you to call on this wrapper
* Making it a child
  * Pros
    * Calling `setState()`/etc. would automatically update the main wrapper
  * Cons
    * Calling `update()` would _only_ update the main wrapper, not itself.
    * You can't call `setProps()`

We could also do things like, make it a root, but link it to the main wrapper somehow so that the main wrapper is updated whenever it is updated.

Or, make it a child, but make an exception that allows you to call `setProps()`. We could also just decide to make it a child, and you're not allowed to call `setProps()`. (I imagine you could always get by with using state instead of props.)

Do you have any thoughts on any of this? I'd love some direction. Thanks!

ljharb:
I think it’d be kind of something different - a root, but it’s only child would be a live link to the *actual* root.

minznerjosh:
@ljharb cool that’s the direction I was leaning towards too. So, when you `update()` it, the _actual_ root should update too?

Do you think it’s worth the trouble to make sure that methods like `detach()` only work on the _actual_ root, and not this special one?

ljharb:
Yes, it should update the actual root.

I think we’ll need to go through each wrapper method, and decide if it makes sense to work on the “wrapper root” or not - if it doesn’t, we may need to make it throw.

Another possibility is having the wrapper root be an entirely different class, that delegates to the real root.

ljharb:
ping :-) i'm planning to make a release soon, and it'd be great to get this into it.

minznerjosh:
Okay! I’ll carve some time out for it this week.

minznerjosh:
@ljharb I updated the implementation for `mount()` to not use refs, and I added an implementation for `shallow()`!

packages/enzyme-adapter-react-14/src/ReactFourteenAdapter.js
```diff
@@ -122,7 +130,21 @@ class ReactFourteenAdapter extends EnzymeAdapter {
...
+
+        if (!rootFinder) {
+          if (typeof options.wrappingComponent === 'function') {
+            throw new Error('`wrappingComponent` must render its children!');
+          } else {
+            throw new Error('Enzyme Internal Error: could not find <RootFinder />.');
+          }
+        }
+
+        return rootFinder.rendered;
```
ljharb: should we assert that `rootFinder.rendered` is a single react element matching the type of that passed to the wrapper?

packages/enzyme-adapter-react-15.4/src/ReactFifteenFourAdapter.js
```diff
@@ -106,6 +107,11 @@ function instanceToTree(inst) {
 
 const eventOptions = { animation: true };
 
+function RootFinder(props) {
```
ljharb:
```suggestion
function RootFinder({ children }) {
```

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -229,6 +229,26 @@ export function elementToTree(el, recurse = elementToTree) {
   };
 }
 
+export function findElement(el, predicate) {
+  if (el === null || typeof el !== 'object' || !('type' in el)) {
+    return undefined;
+  }
+  const { rendered } = el;
```
ljharb: this line can be moved down below the if

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -165,6 +165,178 @@ describeWithDOM('mount', () => {
       expect(() => wrapper.state('key')).to.throw('ReactWrapper::state("key") requires that `state` not be `null` or `undefined`');
     });
 
+    describeIf(is('>= 0.14'), 'wrappingComponent', () => {
```
ljharb: react 13 can't support this?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -5701,6 +5873,7 @@ describeWithDOM('mount', () => {
     in main (created by ErrorBoundary)`}
     in div (created by ErrorBoundary)
     in ErrorBoundary (created by WrapperComponent)
+    in RootFinder (created by WrapperComponent)
```
ljharb: this line shouldn't appear in debug output.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -5774,6 +5947,7 @@ describeWithDOM('mount', () => {
     in div (created by ErrorBoundary)
     in ErrorBoundary (created by ErrorSFC)
     in ErrorSFC (created by WrapperComponent)
+    in RootFinder (created by WrapperComponent)
```
ljharb: also this one.

packages/enzyme/src/ReactWrapper.js
```diff
@@ -104,20 +107,33 @@ class ReactWrapper {
...
       privateSet(this, ROOT_NODES, root[NODES]);
     }
+    privateSet(this, UNRENDERED, nodes);
     privateSet(this, OPTIONS, root ? root[OPTIONS] : options);
+    privateSet(this, LINKED_ROOTS, []);
+    privateSet(this, UPDATED_BY, null);
+
+    if (typeof options.wrappingComponent === 'function') {
+      if (typeof this[RENDERER].getWrappingComponentRenderer !== 'function') {
+        throw new Error('your adapter does not support `wrappingComponent`. Try upgrading it!');
```
ljharb:
```suggestion
        throw new TypeError('your adapter does not support `wrappingComponent`. Try upgrading it!');
```

packages/enzyme/src/ReactWrapper.js
```diff
@@ -1189,6 +1236,28 @@ class ReactWrapper {
   }
 }
 
+/**
+ * A *special* "root" wrapper that represents the component passed as `wrappingComponent`.
+ * It is linked to the primary root such that updates to it will update the primary,
```
ljharb: the comment here seems to contradict the sentence in the docs

packages/enzyme/src/ReactWrapper.js
```diff
@@ -1189,6 +1236,28 @@ class ReactWrapper {
...
+  constructor(root, renderer) {
+    super(renderer.getNode(), root);
+
+    privateSet(this, ROOT, this);
+    privateSet(this, RENDERER, renderer);
+    this[LINKED_ROOTS].push(root);
+  }
+
+  getWrappingComponent() {
+    throw new Error('ReactWrapper::getWrappingComponent() can only be called on the root');
```
ljharb:
```suggestion
    throw new TypeError('ReactWrapper::getWrappingComponent() can only be called on the root');
```

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -243,6 +245,76 @@ function privateSetChildContext(adapter, wrapper, instance, renderedNode, getChi
   }
 }
 
+/**
+ * Recursively dive()s every custom component in a wrapper until
```
ljharb: i'm not sure this is viable - a wrapping component could, for example, wrap in a div. if we don't want to allow that, we need to explicitly check for it in both mount and shallow, and throw.

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -261,6 +333,23 @@ class ShallowWrapper {
         throw new TypeError('ShallowWrapper can only wrap valid elements');
       }
 
+      if (typeof options.wrappingComponent === 'function') {
+        if (typeof adapter.RootFinder !== 'function') {
+          throw new Error('your adapter does not support `wrappingComponent`. Try upgrading it!');
```
ljharb:
```suggestion
          throw new TypeError('your adapter does not support `wrappingComponent`. Try upgrading it!');
```

docs/api/ReactWrapper/getWrappingComponent.md
```diff
@@ -0,0 +1,52 @@
...
+const provider = wrapper.getWrappingComponent();
+provider.setProps({ customStore: mockStore });
+wrapper.update(); // update root wrapper after wrappingComponent props changed
+```
+
+
+
+#### Common Gotchas
+
+- Your root wrapper _will not_ automatically `update()` when the state/props of your `wrappingComponent` are changed. If you expect your component to render differently after calling `.setState()`/`.setProps()` on your `wrappingComponent`, you'll need to manually call `update()` on your root wrapper.
```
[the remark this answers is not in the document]
minznerjosh: Whoops! Sorry this was leftover from my first implementation. This latest implementation handles the updates automatically!

packages/enzyme-adapter-react-14/src/ReactFourteenAdapter.js
minznerjosh: I don't think we need to. Because _we_ are the ones passing the root node as children to `<RootFinder />`, we can be confident that its children are correct. What `wrappingComponent` gets passed as `children` is actually `<RootFinder />`.

Now it _would_ be possible for the user to use `cloneElement()` or something to mess with the root by modifying `children.props.children` in their `wrappingComponent`, but I think, given that the user is highly unlikely to do that by accident, we don't need to go about making assertions for that case.

Does that seem reasonable?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
minznerjosh: We talked about this a bit way back in the early days of this PR (feels like ages ago haha) but the issue here is more just that this test suite is not capable of testing react 13 because of react 13's owner-based context API. This test suite requires parent-based context. (You'll notice that none of the `options.context` tests run on react 13 either.)

packages/enzyme/src/ReactWrapper.js
minznerjosh: Yup the docs are wrong. Sorry!

packages/enzyme/src/ShallowWrapper.js
minznerjosh: This implementation _should_ support non-custom components too. It iterates the entire component tree (including host components) and just `dive()`s on custom ones. But to be absolutely sure I'll throw some `<div />`s into the tests.

minznerjosh:
@ljharb I'll investigate where logic can/should be moved into adapters and will implement (along with the other changes you've requested) in my next iteration.

If I decide to keep something in enzyme, rather than moving it into an adapter, I'll leave you my reasoning in a comment so we can discuss further.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
ljharb: ok, fair enough :-) we'll postpone that until after landing all the current context PRs.

packages/enzyme/src/ShallowWrapper.js
ljharb: `.dive()` explicitly throws on divs, is what i mean :-)

ljharb:
@minznerjosh ping, any update?

minznerjosh:
Yup! I’ve been out of the country since last Thursday and am only getting back home tonight. I’ll get the changes made tomorrow. 😊

packages/enzyme-adapter-react-14/src/ReactFourteenAdapter.js
```diff
@@ -136,6 +142,20 @@ class ReactFourteenAdapter extends EnzymeAdapter {
...
+            return instance ? instanceToTree(instance._reactInternalInstance).rendered : null;
+          },
+          render(el, context, callback) {
+            if (!instance) {
+              throw new Error('The wrapping component may not be updated if the root is unmounted.');
+            }
+            instance.setWrappingComponentProps(el.props, callback);
+          },
+        };
+      },
```
minznerjosh: @ljharb This seemed a bit trivial to dedupe into `enzyme-adapter-utils`. Do you agree?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
minznerjosh: I'm gonna push back here. This isn't `.debug()` output, it's the stack react 16 prints when an error is thrown. `<RootFinder />` is an enzyme "internal" component, but `<WrapperComponent />`, which is also an internal component _does_ appear in this output. And I think that means `<RootFinder />` technically belongs here too.

packages/enzyme/src/ShallowWrapper.js
minznerjosh: I added some extra `<div />`/`<span />`s to the tests and can verify this still works. It works because this method uses the adapter to determine if a component can be `dive()`ed before diving it. If it can't be `dive()`ed it continues recursing to the next level of children. So, it _does_ iterate the whole tree (until it finds its target) but it only `dive()`s components that can be `dive()`ed.

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -243,6 +246,52 @@ function privateSetChildContext(adapter, wrapper, instance, renderedNode, getChi
...
+  }
+  const children = wrapper.children();
+  for (let i = 0; i < children.length; i += 1) {
+    const found = deepRender(children.at(i), target, adapter);
+    if (typeof found !== 'undefined') {
+      return found;
+    }
+  }
+  return undefined;
+}
```
minznerjosh: I believe this function _does_ belong in here. This function is all about taking in a `ShallowWrapper` and traversing it (while calling `dive()`.) So all its logic is around _enzyme_ `ShallowWrapper`s, not react itself. That tells me it belongs in enzyme, not an adapter.

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -243,6 +246,52 @@ function privateSetChildContext(adapter, wrapper, instance, renderedNode, getChi
...
+ * @param {Adapter} adapter An Enzyme adapter
+ * @returns {object} The context collected
+ */
+function getContextFromWrappingComponent(wrapper, adapter) {
+  const rootFinder = deepRender(wrapper, wrapper[ROOT_FINDER], adapter);
+  if (!rootFinder) {
+    throw new Error('`wrappingComponent` must render its children!');
+  }
+  return rootFinder[OPTIONS].context;
+}
```
minznerjosh: Same with this function. It operates entirely with `ShallowWrapper` APIs—nothing react-specific.

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -261,6 +310,25 @@ class ShallowWrapper {
         throw new TypeError('ShallowWrapper can only wrap valid elements');
       }
 
+      if (typeof options.wrappingComponent === 'function') {
+        if (typeof adapter.wrapWithWrappingComponent !== 'function') {
+          throw new TypeError('your adapter does not support `wrappingComponent`. Try upgrading it!');
+        }
+        const { node: wrappedNode, RootFinder } = adapter.wrapWithWrappingComponent(nodes, options);
```
minznerjosh: Previously the logic in `adapter.wrapWithWrappingComponent` was in enzyme. This method is all about element creation, so I think it did make more sense for it to live in the adapter.

minznerjosh:
@ljharb this is ready for another review!

packages/enzyme-adapter-react-14/src/ReactFourteenAdapter.js
ljharb: if the `getNode` logic is the same, and is unlikely to differ (or can be passed in via callback) then it seems worth deduping as much as possible.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
ljharb: I understand why it appears there; but *users* should not see the implementation details of enzyme. That there's one wrapper component is unavoidable (or else I'd happily avoid it), but there shouldn't be two.

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -261,6 +310,25 @@ class ShallowWrapper {
...
+        // eslint-disable-next-line no-use-before-define
+        const wrappingComponent = new WrappingComponentWrapper(wrappedNode, this, RootFinder);
+        const wrappingComponentContext = getContextFromWrappingComponent(
+          wrappingComponent, adapter,
+        );
+        options = {
```
ljharb: can any of this logic be moved to `makeOptions`, even if that means "is the root" needs to be passed?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -385,6 +453,23 @@ class ShallowWrapper {
     return this[RENDERER].getNode().instance;
   }
 
+  /**
+   * If a `wrappingComponent` was passed in `options`, this methods returns a `ShallowWrapper`
+   * around the rendered `wrappingComponent`. This `ShallowWrapper` can be used to update the
+   * `wrappingComponent`'s props, state, etc.
+   *
+   * @returns ReactWrapper
```
ljharb:
```suggestion
   * @returns ShallowWrapper
```

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -1531,6 +1619,60 @@ class ShallowWrapper {
...
+}
+
+/**
+ * A *special* "root" wrapper that represents the component passed as `wrappingComponent`.
+ * It is linked to the primary root such that updates to it will update the primary.
+ *
+ * @class WrappingComponentWrapper
+ */
+class WrappingComponentWrapper extends ShallowWrapper {
+  /* eslint-disable class-methods-use-this */
```
ljharb: only the `getWrappingComponent` method needs this override, it looks like?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -1531,6 +1619,60 @@ class ShallowWrapper {
...
+  });
+}
+
+/**
+ * A *special* "root" wrapper that represents the component passed as `wrappingComponent`.
+ * It is linked to the primary root such that updates to it will update the primary.
+ *
+ * @class WrappingComponentWrapper
+ */
+class WrappingComponentWrapper extends ShallowWrapper {
```
ljharb: since this is still accessible via `ShallowWrapper.prototype.getWrappingComponent`, let's move it to its own file?

minznerjosh: I don't think that's feasible because it would create an import cycle. `./ShallowWrapper` would try to import `./WrappingComponentWrapper` so that it can `new WrappingComponentWrapper()` in the `ShallowWrapper`'s `constructor`. And `./WrappingComponentWrapper` would try to import `./ShallowWrapper` so it can `extends ShallowWrapper`.

And we know that `ShallowWrapper` needs to be defined before we can do `extends ShallowWrapper` (but we can safely `new WrappingComponentWrapper()` in `ShallowWrapper`'s `constructor` even if it's defined down here.) At least if they're in the same file we can ensure that's the case.

packages/enzyme-adapter-react-14/src/ReactFourteenAdapter.js
minznerjosh: Extracted!

packages/enzyme/src/ShallowWrapper.js
minznerjosh: Because this is shallow-only logic but `makeOptions` is used in `ShallowWrapper` and `ReactWrapper` I instead opted to pull this logic out into a separate function in `ShallowWrapper.js`.

minznerjosh:
@ljharb updated!

packages/enzyme/src/ShallowWrapper.js
ljharb: ah, that's a good point.

packages/enzyme/src/ShallowWrapper.js
ljharb: wait - why is this shallow-only? `mount` should be able to have a wrapping component as well.

minznerjosh: `mount` can have a wrapping component! I just meant all the code that I moved out of the constructor of `ShallowWrapper` into a separate function is only for shallow. That code:

* Does the recursive `dive()`ing
* Pulls the accumulated context off of the `<RootFinder />`
* Transforms `options.context` to extend it with the accumulated context

None of that should happen when mounting because we just _actually_ render the wrapping component and let react just do its thing w/r/t the context.

packages/enzyme-test-suite/test/adapter-utils-spec.jsx
```diff
@@ -137,6 +144,7 @@ describe('enzyme-adapter-utils', () => {
       <span />,
       <B />,
       <C />,
+      <RootFinder />,
```
ljharb: as mentioned previously; this shouldn't show up.

packages/enzyme/src/ReactWrapper.js
```diff
@@ -235,6 +268,20 @@ class ReactWrapper {
       return root.update();
     }
     privateSetNodes(this, this[RENDERER].getNode());
+    this[LINKED_ROOTS].forEach((linkedRoot) => {
+      if (linkedRoot !== this[UPDATED_BY]) {
+        /* eslint-disable no-param-reassign */
+        // Only update a linked it root if it is not the originator of our update().
+        // This is needed to prevent infinite recursion when there is a bi-directional
```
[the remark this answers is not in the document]
minznerjosh: Most definitely! All the tests that trigger an `update()` on either the wrapping component or the root (if there is a wrapping component) would start failing.

packages/enzyme-test-suite/test/adapter-utils-spec.jsx
minznerjosh: Yup! This is part of the test data, not the expectation. I threw it in there to assert that it's getting filtered out like you've asked. 😊

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -287,3 +310,45 @@ export function getMaskedContext(contextTypes, unmaskedContext) {
   }
   return fromEntries(Object.keys(contextTypes).map(key => [key, unmaskedContext[key]]));
 }
+
+export function getNodeFromRootFinder(tree, options) {
+  const rootFinder = findElement(tree, node => node.type === RootFinder);
```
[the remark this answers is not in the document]
minznerjosh: Right right. Sorry the implementation has waffled back and forth between "always wrap in `<RootFinder />`" and "only wrap in `<RootFinder />` if there is a `wrappingComponent`." I think we're now settled on the latter. This was poorly-adapted code from when the former was true. Sorry for the sloppiness! I'll clean this up.

packages/enzyme-adapter-react-13/src/ReactThirteenAdapter.js
```diff
@@ -280,6 +285,13 @@ class ReactThirteenAdapter extends EnzymeAdapter {
   createElement(...args) {
     return React.createElement(...args);
   }
+
+  wrapWithWrappingComponent(node, options) {
+    return {
+      RootFinder,
```
ljharb: how come this one adapter doesn't need to `...this` in its new renderer?

Also, if it's not going to be supported in react 13 yet (since the tests don't cover it), let's have this method instead throw an explicit error?

packages/enzyme/src/ReactWrapper.js
```diff
@@ -104,20 +107,33 @@ class ReactWrapper {
...
       privateSet(this, ROOT, root);
       privateSetNodes(this, nodes);
       privateSet(this, ROOT_NODES, root[NODES]);
     }
+    privateSet(this, UNRENDERED, nodes);
     privateSet(this, OPTIONS, root ? root[OPTIONS] : options);
+    privateSet(this, LINKED_ROOTS, []);
+    privateSet(this, UPDATED_BY, null);
+
+    if (typeof options.wrappingComponent === 'function') {
```
ljharb: should maybe this be a function also defined on adapters, so that most of them can use `react-is`, and the react 16 ones will be able to support non-function component types, like forwardRef?

minznerjosh: Good call!

packages/enzyme-adapter-react-13/src/ReactThirteenAdapter.js
minznerjosh: Added a test for throwing an error in react 13!

minznerjosh:
@ljharb updated!

minznerjosh:
Whoops, I can see I broke some stuff. Will fix ASAP!

ljharb:
ping :-) any update? i'm anxious to get this one in.

minznerjosh:
@ljharb updated! The build succeeded!

minznerjosh:
@ljharb think this could be merged soon?

nelsonpecora:
Hi folks, I just ran into this issue and I'm so happy someone's already solved it! From the last comment, does this mean we'll be able to use this feature when the next version of enzyme is released?

ljharb:

PR #1966, opened 2019-01-05
Add ability to dive() and shallow-render-as-root Consumers and Providers

### LINKED ISSUE #1647: dive() breaks when using the new context API.
**Describe the bug**

The new React context elements (https://reactjs.org/docs/context.html) report their type as objects instead of functions (custom react components) or strings (native react components), which breaks `dive()`. I haven't tested whether other methods of `ShallowWrapper` break.

**To Reproduce**

https://codesandbox.io/s/zxj4898zxp?module=%2Fsrc%2FApp.test.js

**Expected behavior**

`shallow` and `dive` should implement actual render semantics when handling the new context API:
1. Automatically update current context whenever a context provider is encountered and remove the provider node from the tree.
2. Automatically pass the current context to childs of context consumer nodes and remove the consumer node from the tree.

_I'm sure the above are not the exact semantics, but at least the most intuitive._

### LINKED ISSUE #1908: Dive shallow in a Context Consumer
Hi,

### Current behavior

I'm trying to dive into a Component wrapped with a HOC that adds a Context Consumer but I get an error `TypeError: ShallowWrapper::dive() can only be called on components`. Here is my test case to reproduce :
```js
import React from 'react';
import { configure, shallow } from 'enzyme';

const ThemeContext = React.createContext('light');

const withTheme = (Component) => {
  class ComponentWithTheme extends React.Component {
    render() {
      return (
        <ThemeContext.Consumer>
          {theme => <Component {...this.props} theme={theme} />}
        </ThemeContext.Consumer>
      );
    }
  }
  ComponentWithTheme.displayName = `withTheme(${Component.displayName || Component.name})`;
  return ComponentWithTheme;
}

class MyComponent extends React.Component {
  render() {
    const { theme } = this.props;
    return <div>{theme}</div>;
  }
}
const MyComponentWithTheme = withTheme(MyComponent);

const wrapper = shallow(<MyComponentWithTheme />, { context: 'dark' }).dive(); // TypeError
```
I'm not familiar with React Contexts, but it seems related to the "function as a child", which is however the regular way to use contexts (cf [Context.Consumer](https://reactjs.org/docs/context.html#contextconsumer) in React doc).

### Expected behavior

In my test case, the `dive` should succeed and return a wrapper around `<div>dark</div>` I guess.

### Your environment
My `package.json` :
```json
{
  "dependencies": {
    "react": "^16.5.1",
    "react-dom": "^16.6.3"
  },
  "devDependencies": {
    "@babel/core": "^7.1.6",
    "@babel/node": "^7.0.0",
    "@babel/preset-env": "^7.1.6",
    "@babel/preset-react": "^7.0.0",
    "enzyme": "^3.7.0",
    "enzyme-adapter-react-16": "^1.7.0"
  }
}
```
#### API

- [x] shallow

#### Version

| enzyme              | 3.7.0
| react               | 16.5.1
| react-dom           | 16.6.3
| react-test-renderer | 16.3.2
| adapter (below)     | 1.7.0

#### Adapter

- [x] enzyme-adapter-react-16

Am I getting something wrong ? Is there anything particular to do when "shallow diving" in a Context Consumer ? I can't see where's my error on this case so any help would be appreciated !

### LINKED ISSUE #1958: [Feature Request] createContext() not supported in enzyme.

minznerjosh:
Fixes #1958
Fixes #1908
Fixes #1647

I think the most controversial choice I've made here is only collecting context when a `<Provider />` is the root (either because it is `.dive()ed` or passed directly to `shallow(<Provider />)`).

If we wanted to reliably make a `<Consumer />`'s `value` be that of the closest `<Provider />`, we'd need to recursively `.dive()` from the top until we find that `<Provider />`. I don't like that idea because we could end up rendering components that the end user did not want rendered.

I think, if the user _does_ want all their `<Provider />`s' values to be automatically collected, that's where `wrappingComponent` (from #1960) will come into play.

### DIFF
--- a/packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
+++ b/packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
@@ -15,6 +15,8 @@ import {
   Element,
   ForwardRef,
   Fragment,
+  isContextConsumer,
+  isContextProvider,
   isElement,
   isForwardRef,
   isPortal,
@@ -241,6 +243,18 @@ function nodeToHostNode(_node) {
   return mapper(node);
 }
 
+function getProviderDefaultValue(Provider) {
+  // React stores references to the Provider's defaultValue differently across versions.
+  if ('_defaultValue' in Provider._context) {
+    return Provider._context._defaultValue;
+  }
+  throw new Error('Enzyme Internal Error: can’t figure out how to get Provider’s default value');
+}
+
+function makeFakeElement(type) {
+  return { $$typeof: Element, type };
+}
+
 const eventOptions = { animation: true };
 
 class ReactSixteenThreeAdapter extends EnzymeAdapter {
@@ -357,11 +371,30 @@ class ReactSixteenThreeAdapter extends EnzymeAdapter {
     let isDOM = false;
     let cachedNode = null;
     return {
-      render(el, context) {
+      render(el, context, {
+        providerValues = new Map(),
+      } = {}) {
         cachedNode = el;
         /* eslint consistent-return: 0 */
         if (typeof el.type === 'string') {
           isDOM = true;
+        } else if (isContextProvider(el)) {
+          providerValues.set(el.type, el.props.value);
+          const MockProvider = Object.assign(
+            props => props.children,
+            el.type,
+          );
+          return withSetStateAllowed(() => renderer.render({ ...el, type: MockProvider }));
+        } else if (isContextConsumer(el)) {
+          const Provider = adapter.getProviderFromConsumer(el.type);
+          const value = providerValues.has(Provider)
+            ? providerValues.get(Provider)
+            : getProviderDefaultValue(Provider);
+          const MockConsumer = Object.assign(
+            props => props.children(value),
+            el.type,
+          );
+          return withSetStateAllowed(() => renderer.render({ ...el, type: MockConsumer }));
         } else {
           isDOM = false;
           const { type: Component } = el;
@@ -539,11 +572,17 @@ class ReactSixteenThreeAdapter extends EnzymeAdapter {
   }
 
   isCustomComponent(type) {
-    const fakeElement = { $$typeof: Element, type };
+    const fakeElement = makeFakeElement(type);
     return !!type && (
       typeof type === 'function'
       || isForwardRef(fakeElement)
+      || isContextProvider(fakeElement)
+      || isContextConsumer(fakeElement)
     );
+  }
+
+  isContextConsumer(type) {
+    return !!type && isContextConsumer(makeFakeElement(type));
   }
 
   isCustomComponentElement(inst) {
@@ -551,6 +590,14 @@ class ReactSixteenThreeAdapter extends EnzymeAdapter {
       return false;
     }
     return this.isCustomComponent(inst.type);
+  }
+
+  getProviderFromConsumer(Consumer) {
+    const { Provider } = Consumer || {};
+    if (Provider) {
+      return Provider;
+    }
+    throw new Error('Enzyme Internal Error: can’t figure out how to get Provider from Consumer');
   }
 
   createElement(...args) {
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -18,6 +18,8 @@ import {
   Element,
   ForwardRef,
   Fragment,
+  isContextConsumer,
+  isContextProvider,
   isElement,
   isForwardRef,
   isMemo,
@@ -303,6 +305,21 @@ function wrapAct(fn) {
   return returnVal;
 }
 
+function getProviderDefaultValue(Provider) {
+  // React stores references to the Provider's defaultValue differently across versions.
+  if ('_defaultValue' in Provider._context) {
+    return Provider._context._defaultValue;
+  }
+  if ('_currentValue' in Provider._context) {
+    return Provider._context._currentValue;
+  }
+  throw new Error('Enzyme Internal Error: can’t figure out how to get Provider’s default value');
+}
+
+function makeFakeElement(type) {
+  return { $$typeof: Element, type };
+}
+
 class ReactSixteenAdapter extends EnzymeAdapter {
   constructor() {
     super();
@@ -454,11 +471,30 @@ class ReactSixteenAdapter extends EnzymeAdapter {
     };
 
     return {
-      render(el, unmaskedContext) {
+      render(el, unmaskedContext, {
+        providerValues = new Map(),
+      } = {}) {
         cachedNode = el;
         /* eslint consistent-return: 0 */
         if (typeof el.type === 'string') {
           isDOM = true;
+        } else if (isContextProvider(el)) {
+          providerValues.set(el.type, el.props.value);
+          const MockProvider = Object.assign(
+            props => props.children,
+            el.type,
+          );
+          return withSetStateAllowed(() => renderer.render({ ...el, type: MockProvider }));
+        } else if (isContextConsumer(el)) {
+          const Provider = adapter.getProviderFromConsumer(el.type);
+          const value = providerValues.has(Provider)
+            ? providerValues.get(Provider)
+            : getProviderDefaultValue(Provider);
+          const MockConsumer = Object.assign(
+            props => props.children(value),
+            el.type,
+          );
+          return withSetStateAllowed(() => renderer.render({ ...el, type: MockConsumer }));
         } else {
           isDOM = false;
           const { type: Component } = el;
@@ -673,11 +709,17 @@ class ReactSixteenAdapter extends EnzymeAdapter {
   }
 
   isCustomComponent(type) {
-    const fakeElement = { $$typeof: Element, type };
+    const fakeElement = makeFakeElement(type);
     return !!type && (
       typeof type === 'function'
       || isForwardRef(fakeElement)
+      || isContextProvider(fakeElement)
+      || isContextConsumer(fakeElement)
     );
+  }
+
+  isContextConsumer(type) {
+    return !!type && isContextConsumer(makeFakeElement(type));
   }
 
   isCustomComponentElement(inst) {
@@ -685,6 +727,22 @@ class ReactSixteenAdapter extends EnzymeAdapter {
       return false;
     }
     return this.isCustomComponent(inst.type);
+  }
+
+  getProviderFromConsumer(Consumer) {
+    // React stores references to the Provider on a Consumer differently across versions.
+    if (Consumer) {
+      let Provider;
+      if (Consumer.Provider) {
+        ({ Provider } = Consumer);
+      } else if (Consumer._context) {
+        ({ Provider } = Consumer._context);
+      }
+      if (Provider) {
+        return Provider;
+      }
+    }
+    throw new Error('Enzyme Internal Error: can’t figure out how to get Provider from Consumer');
   }
 
   createElement(...args) {
--- a/packages/enzyme-test-suite/test/Adapter-spec.jsx
+++ b/packages/enzyme-test-suite/test/Adapter-spec.jsx
@@ -1169,7 +1169,40 @@ Warning: Failed Adapter-spec type: Invalid Adapter-spec \`foo\` of type \`string
     });
 
     itIf(is('>=16.3'), 'returns true for forward refs', () => {
-      expect(adapter.isCustomComponent(React.forwardRef(() => null))).to.equal(true);
+      expect(adapter.isCustomComponent(forwardRef(() => null))).to.equal(true);
+    });
+  });
+
+  describeIf(is('>= 16.3'), 'isContextConsumer(type)', () => {
+    it('returns true for createContext() Consumers', () => {
+      expect(adapter.isContextConsumer(createContext().Consumer)).to.equal(true);
+    });
+
+    it('returns false for everything else', () => {
+      expect(adapter.isContextConsumer(null)).to.equal(false);
+      expect(adapter.isContextConsumer(true)).to.equal(false);
+      expect(adapter.isContextConsumer(undefined)).to.equal(false);
+      expect(adapter.isContextConsumer(false)).to.equal(false);
+      expect(adapter.isContextConsumer(() => <div />)).to.equal(false);
+      expect(adapter.isContextConsumer(forwardRef(() => null))).to.equal(false);
+      expect(adapter.isContextConsumer(createContext().Provider)).to.equal(false);
+    });
+  });
+
+  describeIf(is('>= 16.3'), 'getProviderFromConsumer(Consumer)', () => {
+    it('gets a createContext() Provider from a Consumer', () => {
+      const Context = createContext();
+
+      expect(adapter.getProviderFromConsumer(Context.Consumer)).to.equal(Context.Provider);
+    });
+
+    it('throws an internal error if something that is not a Consumer is passed', () => {
+      expect(() => adapter.getProviderFromConsumer(null)).to.throw(
+        'Enzyme Internal Error: can’t figure out how to get Provider from Consumer',
+      );
+      expect(() => adapter.getProviderFromConsumer({})).to.throw(
+        'Enzyme Internal Error: can’t figure out how to get Provider from Consumer',
+      );
     });
   });
 });
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -261,6 +261,53 @@ describeWithDOM('mount', () => {
         expect(wrapper.text()).to.equal('Context says: I can be set!');
       });
 
+      describeIf(is('>= 16.3'), 'with createContext()', () => {
+        let Context1;
+        let Context2;
+
+        function WrappingComponent(props) {
+          const { value1, value2, children } = props;
+          return (
+            <Context1.Provider value={value1}>
+              <Context2.Provider value={value2}>
+                {children}
+              </Context2.Provider>
+            </Context1.Provider>
+          );
+        }
+
+        function Component() {
+          return (
+            <Context1.Consumer>
+              {value1 => (
+                <Context2.Consumer>
+                  {value2 => (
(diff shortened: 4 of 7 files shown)


packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js
```diff
@@ -321,11 +323,21 @@ class ReactSixteenThreeAdapter extends EnzymeAdapter {
     let isDOM = false;
     let cachedNode = null;
     return {
-      render(el, context) {
+      render(el, context, {
+        providerValues = new Map(),
+      } = {}) {
```
minznerjosh: I'd like to make the third argument be an object. This way we can easily pass additional params as enzyme/react evolves without introducing breaking changes.

ljharb:
enzyme doesn't yet have *any* support for createContext, so there shouldn't be any mention of "Consumer" or "Provider" in documentation until that's done.

Providers, prior to createContext being a feature, are components that provide child context.

minznerjosh:
I'm a bit confused. There are a couple of tests ([this one](https://github.com/airbnb/enzyme/blob/master/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx#L293-L307) and [this one](https://github.com/airbnb/enzyme/blob/master/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx#L204-L227)) that specifically deal with the `createContext()` API. When `mount()`ing, the proper value gets passed down and everything—it functions fully! W/r/t `shallow()`, a component with a `<Provider />` or `<Consumer />` will render, but you currently can't `.dive()` it to make it render for real. So it seems like enzyme does have _some_ support for `createContext()`. Once you are able to `.dive()` `<Consumer />`s and `<Provider />`s, as this PR implements, what more must enzyme do to support `.createContext()`?

I do think the gotcha about needing to `.dive()` a `<Provider />` to have it affect a downstream `<Consumer />` does need to be called out in documentation. Perhaps I could add some clarification that this regards the new `createContext()` API?

minznerjosh:
@ljharb I've created #1973 to try to come to consensus on what full support for `createContext()` in enzyme should look like.

dschinkel:
One of the biggest problems I have is that React and Redux are forcing everyone to use the Context API.  What if there are those who don't want to use that for persistence and actualy are fine with passing state as props and find that route "simplistic" regardless of the "inconvenience".  

The problem I see is that the way React and Redux are moving, they're continually changing the API in a way that undermines the ability to shallow and IMO they don't really care because they would have already noticed that enzyme shallow is breaking due to their major API changes.  They're not backward compatible anymore, new react features are literally breaking things and they don't care about isolated tests IMO generally which is my general take on it because if they did they'd already have seen this issue before releasing such major changes to their API. 

It's fundamental to be able to shallow components. That's what made react so create in the beginning. VueJS still has that ability in fact it's built right into their testing lib.

minznerjosh:
Btw I don’t think this implementation is _quite_ right yet. It effectively auto-dives through as many Consumers/Providers as are necessary until it reaches a non Consumer/Provider. I think the correct behavior is for a Consumer/Provider to behave like any other component. Does that make sense?

I’ll take a look at implementing what I believe is the correct functionality in a couple of days.

AnaRobynn:
Why limiting only diving Providers/Consumers. Same problems with shallow occur when using a HOC. 

Have you thought about a functionality like: diveUntil?

dschinkel:
@AnaRobynn agree.

**We should still be able to still:**
shallow + store prop / (still be able to dive when necessary) - for isolated tests
mount() + provider + store prop - for integration tests

and in either situation specify a store as prop.  This ability should never change on enzyme, React, or Redux side because people rely on writing tests this way.

minznerjosh:
@AnaRobynn a feature like that is definitely possible to implement. But that’s out of scope for this PR. Presumably `diveUntil()` (that supports consumers and providers) would be built on top of the work done in this PR.

You might also be interested in this library, which adds this functionality to enzyme today! https://www.npmjs.com/package/enzyme-shallow-until

dschinkel:
Yea though one issue with diveUntil() is that you are basically losing some design feedback though.  Because sure, it might make your tests less fragile if you add say another HOC or renderProps or something else, but you actually want dive() to be painful because that helps give you design pressure that says "Me the component is painful to test, because you need to make my simpler and break me apart".  What happens is people resort to that or mount() and then they never see other issues with their design and maybe why it "hurts" and so they start only creating integration style tests which does not give you the bang for your buck.

ljharb:
I definitely don't want a "diveUntil" API as part of shallow; that's something I'm planning to tackle as an entirely separate top-level API after we're caught up with React's featureset.

I'm not even confident that enzyme should auto-dive Providers and Consumers yet, to be honest.

AnaRobynn:
@ljharb @dschinkel Sounds reasonable.

Let’s say I have a component, I want to inject styles via a HOC (similar to material-ui). 
I’m wondering if my tests should break, just because I’m adding this styling HOC or not.

I not clear to me what the benefits/drawbacks are for for either failing/not failing a unit test, due to adding a HOC.

But that’s irrelevant in this PR.

minznerjosh:
@ljharb I don't think we should auto-dive Providers/Consumers. I think instead you should be able to manually `dive()` them like any other component. I'm going to fix-up this PR today so that calling `dive()` once does not end up rendering more than one level of Consumer/Provider.

dschinkel:
> I’m wondering if my tests should break, just because I’m adding this styling HOC or not
Good question I'll try to tackle that question in my upcoming TDD course...which will now be FREE!!!!! YAY!

ljharb:
@minznerjosh it makes sense that diving a Consumer/Provider should work as if it was a regular custom component with a function child - altho it's a bit surprising that it doesn't already do that without any enzyme changes.

AnaRobynn:
@AnaRobynn I know it’s a breaking change. I’m wondering if it should break a shallow test or not.

ljharb:
@AnaRobynn no, adding a well-behaved HOC should *not* be a breaking change for any consumers *except* `shallow`, for which it *should* require one additional `.dive()` call.

dschinkel:
We need to be careful saying that "everything is a problem".  Shallow, think about it.  It's very specific.  You have to expect that if your design is not clean and simple, you're going to have breaking changes if you introduce HOCs.  Again that's design feedback, and that's a good thing and one should listen to it.  Most devs don't understand that.  I know you do @AnaRobynn but just making a general statement here. This is a big reason why the context of TDD helps and why those who do TDD already see this and those who write tests after don't.  I'm not putting anyone down, it's just fact.  So I'm trying to make people aware of this, here and also in my upcoming vids.

AnaRobynn:
@dschinkel @ljharb
True, if you really think about shallow component is that props and what gets rendered are the api. So in theory wrapping some component in a HOC is breaking that API.

Anyways, I do not further want to derail this PR with unrelated questions. Thank you for replying/explaining my concerns/questions. Really helpful.

dschinkel:
@ljharb just wondering what you meant by:

>Solving that problem is something I'm interesting in doing after we're caught up to React's featureset.

minznerjosh:
@ljharb this PR is now ready for a review!

* `createContext()` consumers and providers are _not_ automatically `dive()`ed. You have to call `dive()` on them like any other custom component.
* You can shallow render a `createContext()` consumer or provider as the root. You can even call `.setProps({ value: ... })` on a provider to change the context value of downstream consumers.
* If you `dive()`/render a consumer _without_ rendering its provider further up in the tree, the default context value will be used.
* If you `dive()`/render a consumer _after_ having rendered its provider further up in the tree, the context value of that provider will be used.

Earlier, you said this:

> it's a bit surprising that it doesn't already do that without any enzyme changes

I have a couple of thoughts here. First, the react shallow renderer just doesn't support rendering providers/consumers right now. You could argue that it _should_ support that, but I think when you dig into it a bit, it's understandable why they didn't build support for that.

The new context API is all about the interaction between two components: the provider and consumer. A consumer is pretty much nothing without its provider, and a provider is useless without a consumer. (This is starting to sound like the "if a tree falls in the forrest" thought experiment.)

So the new context API is pretty much useless to the react shallow renderer because it is only ever concerned about the component you render with it. (I think the shallow renderer can't handle host nodes for similar philosophical reasons.)

_Enzyme_ `shallow()`, on the other hand, _does_ support rendering more than one component through `dive()`! And it supports collecting and passing context as the user `dive()`s through their component.

So, even if the react shallow renderer supported rendering providers/consumers, enzyme would still need to do work to handle collecting and passing that context as the user `dive()`s. And, seeing how useless consumers/providers are without the ability to `dive()`, I think it's fairly reasonable to add the rendering functionality _just_ in enzyme.

ljharb:
True, but I'd expect the Provider and Consumer to work the same as if I'd implemented them myself in JS - iow:
```js
function C() {
  return (
    <Provider value={v}>
        <Consumer>
          {value => {
            return <div>{value}</div>;
          }}
        </Consumer>
    </Provider>
  );
}

const wrapper = shallow(<C />);
wrapper.debug(); /*
   <Provider value={v}>
        <Consumer>
          {value => {
            return <div>{value}</div>;
          }}
        </Consumer>
    </Provider>
*/

const shallowProvider = wrapper.dive();
shallowProvider.debug(); /*
        <Consumer>
          {value => {
            return <div>{value}</div>;
          }}
        </Consumer>
*/
shallowProvider.context(); // something that contains the Provider value

const shallowConsumer = shallowProvider.dive();
shallowConsumer.debug(); /*
<div>{value}</div>
*/
```

If the shallow renderer doesn't actually handle the native Provider and Consumer, then that's fine, but I'd assume then that enzyme could simply supply a "mock" Provider and Consumer as needed, so as not to hardcode the logic.

minznerjosh:
> I'd assume then that enzyme could simply supply a "mock" Provider and Consumer as needed

Unless I'm misunderstanding you, and I think there's a good possibility I am, that's what this PR does. When it detects it is rendering a Provider/Consumer, [it creates a mock Provider/Consumer and renders that instead](https://github.com/airbnb/enzyme/blob/7b08b1d32ef6fdeb3c2c40b3ba0442c766dc3e85/packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js#L339-L355).

> so as not to hardcode the logic

Can you elaborate on what you mean by "the logic"?

Sorry if I'm being really slow here! I feel like there's something in your feedback that's not yet clicked for me.

ljharb:
No worries, if this was simple to do I’d have done it a long time ago :-D

by “the logic” i mean “whatever the render and lifecycle methods would do if the native Provider/Consumer were refuse componentd”

minznerjosh:
> refuse componentd

This is where you lost me.

tilgovi:
> refuse componentd

I assume that's a typo for "react components". In other words, do not treat consumers and providers as a different kind of object, but just as components that have manage and propagate state.

tilgovi:
I have to admit I can't really understand the ask either. Providers and consumer are special in that they pass state across any number of intervening components, not just parent to child. A provider can mostly be a component that manages state, but it also needs to pass that state to any descendant consumer.

Not understanding the internals of enzyme, I was thinking that maybe the ask was to replace any consumers and providers with components that use consumers and providers and hope that somehow React itself could do the actual context management with Enzyme only being exposed to these proxy components.

minznerjosh:
Ah, thanks for the assist @tilgovi! I was having trouble de-typo-ing refuse=>react.

@ljharb, are you asking that the "mock" Consumer/Provider implement "the logic" as if it were just a regular old react component? For example, you could implement the new context API using the legacy context API:

*Warning, pseudo code ahead*
```jsx
function createMockContext(defaultValue) {
  const providerId = uuid(); // get some kind of unique id for this context

  class Provider extends React.Component {
    getChildContext() {
      return {
        [providerId]: typeof this.props.value === 'undefined' ? defaultValue : this.props.value,
      };
    }
  }
  Provider.childContextTypes = {
    [providerId]: PropTypes.any,
  };

  class Consumer extends React.Component {
    render() {
      return this.props.children(this.context[providerId]);
    },
  }
  Consumer.contextTypes = Provider.childContextTypes;

  return { Consumer, Provider };
}
```

However, we should definitely _not_ rely on the legacy context API to mock the `createContext()` API because it is a deprecated API.

My understanding of Consumer/Provider behavior is the following:

* Rendering a Provider with a `value` prop sets the context's value to that prop's value.
* If somewhere further down the tree the context's consumer is rendered, its `children` prop is called with the context's value.
* If a consumer is rendered and the context value is not set (aka the consumer is not a descendent of its provider,) its `children` prop is called with the default context value.
* If a provider's `value` prop is changed, all the consumers' `children` props are called with the updated value.

The current [provider mock](https://github.com/airbnb/enzyme/blob/7b08b1d32ef6fdeb3c2c40b3ba0442c766dc3e85/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js#L389-L392) and [consumer mock](https://github.com/airbnb/enzyme/blob/7b08b1d32ef6fdeb3c2c40b3ba0442c766dc3e85/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js#L398-L401) in this PR satisfy all the requirements except the last one regarding the changing of a provider's value. However, I don't think the mocks needs to satisfy that requirement because we would expect the users to re-`dive()` their consumer after the provider's value changes. For example:

```jsx
class A extends React.Component {
  render() {
    return <div>{this.props.foo}</div>;
  }
}
class B extends React.Component {
  constructor() {
    this.state = {
      myProp: 'cool',
    };
  }

  render() {
    return <A foo={this.state.myProp} />;
  }
}

const b = shallow(<B />);
let a = b.dive();
expect(a.text()).to.equal('cool');

b.setProps({ myProp: 'beans' });

// <A />'s output is still the same
a.update();
expect(a.text()).to.equal('cool');
// We have to re-dive() A to see the update
a = b.dive();
expect(a.text()).to.equal('beans');
```

And that's exactly how things will work with the mock Provider and Consumer in this PR. When you re-`dive()` the consumer after changing the provider's `value`, the consumer will render with the updated value.

Perhaps I'm still misunderstanding what you said earlier:

> enzyme could simply supply a "mock" Provider and Consumer as needed

Please let me know if that's the case!

ljharb:
Sorry for the typo.

Yes, I'm saying that a Provider and Consumer should be *no different* from a user-created React component - and treated as if they were a user-created one.

I'm not talking about this PR specifically yet - just talking in generalities about how I expect things to work. It's highly likely that this PR already does exactly what I hope it to :-D i just haven't yet had the time to properly review it.

minznerjosh:
No worries! 😊 I was interpreting your earlier comments as “I’m not going to review this yet because there are major architectural/thematic flaws that need to be fixed before we get into the nitty-gritty of a code review.”

I’m hopeful this implementation does what you expect too! So now I’m going to chill out and wait for you to review. 😎

eugenet8k:
Guys, any progress on submitting this PR? It is a big blocker for migration to react-redux v6 😭

hally9k:
We too are blocked on this one, any update?

ljharb:
@minznerjosh i've freshly rebased this. Can we make sure to add test cases for both shallow and mount to cover this functionality, including both consumers and providers as the root, or with `.dive`, `.shallow`, and `.mount`?

minznerjosh:
@ljharb yup I'll give this PR some love tomorrow!

minznerjosh:
@ljharb Updated!

* Added `mount()` coverage for rendering a `<Provider />`/`<Consumer />` as root (there was already a spec for rendering them as _not_ root)
* The `shallow()` tests from the first iteration already had coverage for rendering both `<Provider />` and `<Consumer />` as root, and `dive()`ing them both
* Added the required logic for `wrappingComponent` in `shallow()` to support `createContext()` (just like legacy context has to be pulled off the `<RootFinder />`, so does `createContext()` context)

Looking forward to your review!

minznerjosh:
@ljharb P.S. the cause of the build failure appears to be flakiness—not failing tests.

PR #1968, opened 2019-01-07
Filter legacy context on childContextTypes when necessary

### LINKED ISSUE #1590: Context is read by components regardless of contextTypes definition
Consider the following, simplified example:

```js
class NotGonnaRead extends Component {
  render() {
    return (
      <p>My favorite color is {this.context.favoriteColor}</p>
    );
  }
}

class WillRead extends Component {
  render() {
    return (
      <p>My favorite color is {this.context.favoriteColor}</p>
    );
  }
}

WillRead.contextTypes = {
  favoriteColor: PropTypes.string,
}

class FavoriteColorProvider extends Component {
  getChildContext() {
    return {
      favoriteColor: 'purple',
    };
  }
}
```

If we render it like so:

```js
<FavoriteColorProvider>
    <NotGonnaRead />
</FavoriteColorProvider>
```

then the output is going to be `My favorite color is `.

Rendering this, however:

If we render it like so:

```js
<FavoriteColorProvider>
    <WillRead />
</FavoriteColorProvider>
```

will correctly output `My favorite color is purple`.

### Current behavior

If we would like to test if NotGonnaRead component is behaving properly (which it isn't), we would need to do something along the lines:

```js
shallow(
  <NotGonnaRead />,
  { context: { favoriteColor: 'purple' } }
);
```

This will render `<p>My favorite color is purple</p>`.

### Expected behavior

This example should render `<p>My favorite color is </p>` as contextTypes are not correctly defined.

#### API

- [x] shallow
- [ ] mount
- [ ] render

#### Version

| library         | version
| ---------------- | -------
| Enzyme           | 3.3.0
| React            |  16.2.0

#### Adapter

- [x] enzyme-adapter-react-16 1.1.1
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] others ( )

minznerjosh:
Fixes #1590 

In `react-test-renderer@16.0 - 16.2` there's an issue where `context` is not filtered to a component's `childContextTypes`. This issue is not present in `react-test-renderer@<16` or `react-test-renderer@>=16.3`. This PR implements the same fix as `react-test-renderer@16.3` for enzyme adapters using `react-test-renderer` that have the issue.

### DIFF
--- a/packages/enzyme-adapter-utils/src/Utils.js
+++ b/packages/enzyme-adapter-utils/src/Utils.js
@@ -1,4 +1,5 @@
 import functionName from 'function.prototype.name';
+import fromEntries from 'object.fromentries';
 import createMountWrapper from './createMountWrapper';
 import createRenderWrapper from './createRenderWrapper';
 import wrap from './wrapWithSimpleWrapper';
@@ -278,4 +279,11 @@ export function simulateError(
   );
 
   componentDidCatch.call(catchingInstance, error, { componentStack });
+}
+
+export function getMaskedContext(contextTypes, unmaskedContext) {
+  if (!contextTypes || !unmaskedContext) {
+    return {};
+  }
+  return fromEntries(Object.keys(contextTypes).map(key => [key, unmaskedContext[key]]));
 }
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -4480,6 +4480,30 @@ describe('shallow', () => {
         expect(wrapper.context().name).to.equal(context.name);
         expect(wrapper.context('name')).to.equal(context.name);
       });
+
+      it('filters context to childContextTypes', () => {
+        class Bar extends React.Component {
+          render() {
+            return <div />;
+          }
+        }
+        Bar.contextTypes = {
+          name: PropTypes.string,
+        };
+        class Foo extends React.Component {
+          render() {
+            return (
+              <div>
+                <Bar />
+              </div>
+            );
+          }
+        }
+
+        const context = { name: 'foo', hello: 'world' };
+        const wrapper = shallow(<Foo />, { context });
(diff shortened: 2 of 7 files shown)


packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -4480,6 +4480,30 @@ describe('shallow', () => {
...
+              <div>
+                <Bar />
+              </div>
+            );
+          }
+        }
+
+        const context = { name: 'foo', hello: 'world' };
+        const wrapper = shallow(<Foo />, { context });
+        expect(JSON.parse(wrapper.find(Bar).dive().text())).to.eql({ name: 'foo' });
```
ljharb: instead of using json and text here, could the test assert that `wrapper.find(Bar).context()` was the expected object?

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -279,3 +279,14 @@ export function simulateError(
 
   componentDidCatch.call(catchingInstance, error, { componentStack });
 }
+
+export function getMaskedContext(contextTypes, unmaskedContext) {
+  if (!contextTypes || !unmaskedContext) {
+    return {};
+  }
+  const context = {};
+  Object.keys(contextTypes).forEach((key) => {
```
ljharb: Let's do this:
```js
return fromEntries(Object.keys(contextTypes).map(key => [key, unmaskedContext[key]]));
```
(with `import fromEntries from 'object.fromentries'`)

minznerjosh:
@ljharb feedback addressed and rebased!

ljharb:
I've rebased this and added some adapter-utils unit tests.

minznerjosh:
Whoops! Sorry for leaving out tests for the adapter utils. I would’ve been happy to add them, but I understand it was probably easier for you to just do it yourself. 😊

I’ll double-check my other PRs to make sure I haven’t left any other modules untested.

ljharb:
@minznerjosh nbd, i appreciate it :-)

PR #1971, opened 2019-01-07
Support getChildContext() in shallow render

### LINKED ISSUE #664: shallow-rendering component with `getChildContext` should work
```jsx
  getChildContext() {
    return { a: 1 };
  }
  render() {
    return <Bar b={this.context.b} />;
  }
}
Foo.contextTypes = {
  b: React.PropTypes.number,
};

class Bar extends React.Component {
  render() {
    return <div />;
  }
}
Bar.contextTypes = {
  a: React.PropTypes.number,
  b: React.PropTypes.number,
};
```

```jsx
const wrapper = shallow(<Foo />, { context: { b: 2 });

expect(wrapper).props()).to.eql({ b: 2 }); // works
// .props() provides me with the first argument that will be passed to `Bar`'s constructor

expect(wrapper).context()).to.eql({ a: 1, b: 2 }); // does not work
// I expect .context() to provide me with the second argument that will be passed to `Bar`'s constructor
```

minznerjosh:
Fixes #664

This PR adds support for `getChildContext()` when shallow-rendering. I've decided to only implement parent-based context, not owner-based context (as I'm not exactly sure what that behavior is, tbh) which means I've excluded `react@0.13` from this implementation. (Hope that's okay!)

In `react@<=15`, the react shallow renderer actually calls `getChildContext()` for us! This means, as is the case with `shouldComponentUpdate()`, we just spy on that method's return value.

In `react@>=16`, `getChildContext()` is _not_ called (understandable—it's a deprecated API), so we have to call it ourselves.

P.S. Sorry the diff is a little ugly. I ended up messing with the nesting of some conditionals to avoid duplicated checks.

### DIFF
--- a/packages/enzyme/src/ShallowWrapper.js
+++ b/packages/enzyme/src/ShallowWrapper.js
@@ -40,6 +40,7 @@ const ROOT = sym('__root__');
 const OPTIONS = sym('__options__');
 const SET_STATE = sym('__setState__');
 const ROOT_NODES = sym('__rootNodes__');
+const CHILD_CONTEXT = sym('__childContext__');
 
 /**
  * Finds all nodes in the current wrapper nodes' render trees that match the provided predicate
@@ -136,6 +137,10 @@ function getAdapterLifecycles({ options }) {
     setState: {
       ...lifecycles.setState,
     },
+    getChildContext: {
+      calledByRenderer: true,
+      ...lifecycles.getChildContext,
+    },
     ...(componentDidUpdate && { componentDidUpdate }),
   };
 }
@@ -180,6 +185,65 @@ function isPureComponent(instance) {
   return instance && instance.isPureReactComponent;
 }
 
+function getChildContext(node, hierarchy, renderer) {
+  const { instance, type: Component } = node;
+  const componentName = displayNameOfNode(node);
+  // Warn like react if childContextTypes is not defined:
+  // https://github.com/facebook/react/blob/1454a8be03794f5e0b23a7e7696cbbbdcf8b0f5d/packages/react-dom/src/server/ReactPartialRenderer.js#L639-L646
+  if (typeof Component.childContextTypes !== 'object') {
+    // eslint-disable-next-line no-console
+    console.warn(
+      `${componentName}.getChildContext(): childContextTypes must be defined in order to use getChildContext().`,
+    );
+    return {};
+  }
+  // Check childContextTypes like react:
+  // https://github.com/facebook/react/blob/1454a8be03794f5e0b23a7e7696cbbbdcf8b0f5d/packages/react-dom/src/server/ReactPartialRenderer.js#L630-L637
+  const childContext = instance.getChildContext();
+  Object.keys(childContext).forEach((key) => {
+    if (!(key in Component.childContextTypes)) {
+      throw new Error(
+        `${componentName}.getChildContext(): key "${key}" is not defined in childContextTypes.`,
+      );
+    }
+  });
+  if (typeof renderer.checkPropTypes === 'function') {
+    renderer.checkPropTypes(Component.childContextTypes, childContext, 'child context', hierarchy);
+  }
+  return childContext;
+}
+
+function spyOnGetChildContextInitialRender(nodes, adapter) {
+  if (
+    !isCustomComponentElement(nodes, adapter)
+    || !nodes.type.prototype
+    || typeof nodes.type.prototype.getChildContext !== 'function'
+  ) {
+    return null;
+  }
+
+  return spyMethod(nodes.type.prototype, 'getChildContext');
+}
+
+function privateSetChildContext(adapter, wrapper, instance, renderedNode, getChildContextSpy) {
+  const renderer = wrapper[RENDERER];
+  // We only support parent-based context.
+  if (adapter.options.legacyContextMode !== 'parent') { return; }
+  if (getChildContextSpy) {
+    privateSet(wrapper, CHILD_CONTEXT, getChildContextSpy.getLastReturnValue());
+    getChildContextSpy.restore();
+  } else if (typeof instance.getChildContext === 'function') {
+    // If there's no spy but getChildContext is a function, that means our renderer
+    // is not going to call it for us, so we need to call it ourselves.
+    const nodeHierarchy = [wrapper[NODE]].concat(nodeParents(wrapper, wrapper[NODE]));
+    const childContext = getChildContext(renderedNode, nodeHierarchy, renderer);
+    privateSet(wrapper, CHILD_CONTEXT, childContext);
+  } else {
+    privateSet(wrapper, CHILD_CONTEXT, null);
+  }
+}
+
+
 /**
  * @class ShallowWrapper
  */
@@ -197,6 +261,9 @@ class ShallowWrapper {
         throw new TypeError('ShallowWrapper can only wrap valid elements');
       }
 
+      const getChildContextSpy = lifecycles.getChildContext.calledByRenderer
+        ? spyOnGetChildContextInitialRender(nodes, adapter)
+        : null;
       privateSet(this, ROOT, this);
       privateSet(this, UNRENDERED, nodes);
       const renderer = adapter.createRenderer({ mode: 'shallow', ...options });
@@ -221,6 +288,7 @@ class ShallowWrapper {
             instance.componentDidMount();
           });
         }
+        privateSetChildContext(adapter, this, instance, renderedNode, getChildContextSpy);
       }
     // creating a child component through enzyme's ShallowWrapper APIs.
     } else {
@@ -363,7 +431,8 @@ class ShallowWrapper {
         // NOTE(lmr): In react 16, instances will be null for SFCs, but
         // rerendering with props/context is still a valid thing to do. In
         // this case, state will be undefined, but props/context will exist.
-        const instance = this.instance() || {};
+        const node = this[RENDERER].getNode();
+        const instance = node.instance || {};
         const { state } = instance;
         const prevProps = instance.props || this[UNRENDERED].props;
         const prevContext = instance.context || this[OPTIONS].context;
@@ -374,15 +443,25 @@ class ShallowWrapper {
         this[RENDERER].batchedUpdates(() => {
           // When shouldComponentUpdate returns false we shouldn't call componentDidUpdate.
           // so we spy shouldComponentUpdate to get the result.
+          const lifecycles = getAdapterLifecycles(adapter);
           let shouldRender = true;
-          let spy;
+          let shouldComponentUpdateSpy;
+          let getChildContextSpy;
           if (
             !this[OPTIONS].disableLifecycleMethods
             && instance
-            && typeof instance.shouldComponentUpdate === 'function'
           ) {
-            spy = spyMethod(instance, 'shouldComponentUpdate');
-          } else if (isPureComponent(instance)) {
+            if (typeof instance.shouldComponentUpdate === 'function') {
+              shouldComponentUpdateSpy = spyMethod(instance, 'shouldComponentUpdate');
+            }
+            if (
+              lifecycles.getChildContext.calledByRenderer
+              && typeof instance.getChildContext === 'function'
+            ) {
+              getChildContextSpy = spyMethod(instance, 'getChildContext');
+            }
+          }
+          if (!shouldComponentUpdateSpy && isPureComponent(instance)) {
             shouldRender = pureComponentShouldComponentUpdate(
               prevProps,
               props,
@@ -392,17 +471,16 @@ class ShallowWrapper {
           }
           if (props) this[UNRENDERED] = cloneElement(adapter, this[UNRENDERED], props);
           this[RENDERER].render(this[UNRENDERED], nextContext);
-          if (spy) {
-            shouldRender = spy.getLastReturnValue();
-            spy.restore();
+          if (shouldComponentUpdateSpy) {
+            shouldRender = shouldComponentUpdateSpy.getLastReturnValue();
+            shouldComponentUpdateSpy.restore();
           }
           if (
             shouldRender
             && !this[OPTIONS].disableLifecycleMethods
             && instance
           ) {
-            const lifecycles = getAdapterLifecycles(adapter);
-
+            privateSetChildContext(adapter, this, instance, node, getChildContextSpy);
             if (lifecycles.getSnapshotBeforeUpdate) {
               let snapshot;
               if (typeof instance.getSnapshotBeforeUpdate === 'function') {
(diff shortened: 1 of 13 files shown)


minznerjosh:
P.S. I'm pretty sure the build will succeed once is #1968 is in.

ljharb:
@minznerjosh I've rebased this and split up the commits a bit more atomically; it'd be great if you could add some tests to each commit/change :-)

minznerjosh:
@ljharb you’ve got it! I’ll take care of it when I get back from vacation on Wednesday. Thanks for taking a look at this!

minznerjosh:
@ljharb I added tests to every commit except ```[enzyme-adapter-react-*] [new] add `legacyContextMode`, and `getChildContext` lifecycle method``` (which is just a configuration change.) This is ready for another round of review!

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -180,6 +185,55 @@ function isPureComponent(instance) {
...
+  }
+  const childContext = instance.getChildContext();
+  Object.keys(childContext).forEach((key) => {
+    if (!(key in Component.childContextTypes)) {
+      const componentName = displayNameOfNode(node);
+      throw new Error(`${componentName}.getChildContext(): key "${key}" is not defined in childContextTypes.`);
+    }
+  });
+  if (typeof renderer.checkPropTypes === 'function') {
+    renderer.checkPropTypes(Component.childContextTypes, childContext, 'child context', hierarchy);
```
ljharb: are childContextTypes new in React 16?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -180,6 +185,55 @@ function isPureComponent(instance) {
...
 }
 
+function getChildContext(node, hierarchy, renderer) {
+  const { instance, type: Component } = node;
+  if (!Component.childContextTypes) {
+    return {};
+  }
+  const childContext = instance.getChildContext();
+  Object.keys(childContext).forEach((key) => {
+    if (!(key in Component.childContextTypes)) {
```
ljharb: are we using `in` here because React does? it seems like this should be an own property check.

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -180,6 +185,55 @@ function isPureComponent(instance) {
...
+  ) {
+    return null;
+  }
+
+  return spyMethod(nodes.type.prototype, 'getChildContext');
+}
+
+function privateSetChildContext(adapter, wrapper, instance, renderedNode, getChildContextSpy) {
+  const renderer = wrapper[RENDERER];
+  // We only support parent-based context.
```
ljharb: what does this mean for react 0.13? is everything the same as before the PR?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -180,6 +185,55 @@ function isPureComponent(instance) {
...
+  if (adapter.options.legacyContextMode !== 'parent') { return; }
```
ljharb: what about using older versions of the react 0.13 adapter, that won't have this property?

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -197,6 +251,8 @@ class ShallowWrapper {
         throw new TypeError('ShallowWrapper can only wrap valid elements');
       }
 
+      const getChildContextSpy = lifecycles.getChildContext.calledByRenderer
+        ? spyOnGetChildContextInitialRender(nodes, adapter) : null;
```
ljharb: if the entire ternary can't fit on one line, then the `?` and `:` need to both be on their own lines

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -509,17 +577,28 @@ class ShallowWrapper {
...
+            shouldComponentUpdateSpy = spyMethod(instance, 'shouldComponentUpdate');
+          }
+          if (
+            lifecycles.getChildContext.calledByRenderer
+            && typeof instance.getChildContext === 'function'
+          ) {
+            getChildContextSpy = spyMethod(instance, 'getChildContext');
+          }
+        }
+        if (!shouldComponentUpdateSpy && isPureComponent(instance)) {
```
ljharb: I believe PureComponents completely ignore any shouldComponent methods, so if it's a PureComponent, it should always fall into this branch

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -1411,7 +1493,15 @@ class ShallowWrapper {
       if (!isCustomComponentElement(el, adapter)) {
         throw new TypeError(`ShallowWrapper::${name}() can only be called on components`);
       }
-      return this.wrap(el, null, { ...this[OPTIONS], ...options });
+      return this.wrap(el, null, {
+        ...this[OPTIONS],
+        ...options,
+        context: {
+          ...this[OPTIONS].context,
+          ...options.context,
```
ljharb: i wonder if we should instead consider the `context` option as *overriding* the proper child context - ie, only do the correct thing when a context option isn't supplied? (or alternatively, not pass forward the context option?

packages/enzyme/src/ShallowWrapper.js
minznerjosh: Not at all! I did a bit of hunting and the [first instance of `childContextTypes` I can find in the source](https://github.com/facebook/react/blob/v0.9.0/src/core/ReactCompositeComponent.js#L968-L996) is in `v0.9.0`. I believe this means it's been part of the context API since the beginning.

packages/enzyme/src/ShallowWrapper.js
minznerjosh: Yeah I tried to match react as much as possible [and they are using `in`](https://github.com/facebook/react/blob/1454a8be03794f5e0b23a7e7696cbbbdcf8b0f5d/packages/react-dom/src/server/ReactPartialRenderer.js#L633).

packages/enzyme/src/ShallowWrapper.js
minznerjosh: Correct!

packages/enzyme/src/ShallowWrapper.js
minznerjosh: It should behave exactly as before `undefined !== 'parent'` is `true` so this function will just early-return. But I will add some tests to increase our confidence.

packages/enzyme/src/ShallowWrapper.js
minznerjosh: Interesting! Based on my reading of the [current implementation in master](https://github.com/airbnb/enzyme/blob/c7282ebc52aafd3442f7b52dde75354bae973674/packages/enzyme/src/ShallowWrapper.js#L379-L392), we don't do the `isPureComponent` check if `instance.shouldComponentUpdate` is a function (as per the first `if ()` condition.

So this implementation more closely matches the implementation in master. However, perhaps what's in master is wrong. In that case, would you like me to fix it here, or should that fix happen separately?

packages/enzyme/src/ShallowWrapper.js
minznerjosh: I like the idea of the `context` option being an override!

minznerjosh:
@ljharb 

* Fixed the formatting of that ternary
* Added tests that simulate an old adapter
* Made it so that the `context` option overrides a parent's context (and added a test)

This is ready for another round of review!

packages/enzyme/src/ShallowWrapper.js
ljharb: Then why wouldn't every adapter, not just the 16+ ones, implement `checkPropTypes`? or, why wouldn't we implement it in enzyme-adapter-utils instead of in each adapter?

packages/enzyme/src/ShallowWrapper.js
ljharb: Let's add a comment then, above line 193, linking to https://github.com/facebook/react/blob/1454a8be03794f5e0b23a7e7696cbbbdcf8b0f5d/packages/react-dom/src/server/ReactPartialRenderer.js#L630-L637

packages/enzyme/src/ShallowWrapper.js
ljharb: Let's file a new issue after this is merged, then, to track adding support for that.

packages/enzyme/src/ShallowWrapper.js
ljharb: Well, first, let's double check me by confirming what React does. If React doesn't match master, then we probably should get in a PR to fix that prior to this one.

packages/enzyme/src/ShallowWrapper.js
minznerjosh: The <16 shallow renderers call `getChildContext()` for us and implement this logic themselves. So there's no need for the <16 renderers to implement `checkPropTypes` because nothing uses them.

However, if you think the method should be there just for the sake of consistency, I can add it (and refactor it into enzyme-adapter-utils.)

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -180,6 +185,63 @@ function isPureComponent(instance) {
...
 }
 
+function getChildContext(node, hierarchy, renderer) {
+  const { instance, type: Component } = node;
+  const componentName = displayNameOfNode(node);
+  if (!Component.childContextTypes) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      `${componentName}.getChildContext(): childContextTypes `
+      + 'must be defined in order to use getChildContext().',
```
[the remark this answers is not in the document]
minznerjosh: [React does warn](https://github.com/facebook/react/blob/1454a8be03794f5e0b23a7e7696cbbbdcf8b0f5d/packages/react-dom/src/server/ReactPartialRenderer.js#L639-L646)! I added a comment with a link to the react impl.

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -1411,7 +1502,15 @@ class ShallowWrapper {
...
         throw new TypeError(`ShallowWrapper::${name}() can only be called on components`);
       }
-      return this.wrap(el, null, { ...this[OPTIONS], ...options });
+      return this.wrap(el, null, {
+        ...this[OPTIONS],
+        ...options,
+        context: {
+          ...this[OPTIONS].context,
+          ...this[ROOT][CHILD_CONTEXT],
+          ...options.context,
```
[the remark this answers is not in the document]
minznerjosh: That wouldn't quite work, since we're relying on passing `options.context` in order to combine contexts as we `dive()` through multiple levels of providers.

I've updated the code to this:
```javascript
context: options.context || {
  ...this[OPTIONS].context,
  ...this[ROOT][CHILD_CONTEXT],
}
```

So now `options.context` (passed to `dive()`) always takes precedence, but context is still combined when `dive()`ing multiple providers.

minznerjosh:
@ljharb 

* Should we implement `checkPropTypes` in all the adapters, even though it's only used in the >=16 adapters?
* Removed multi-line strings
* Added comment to explain warning if `childContextTypes` is not defined, with link to react src
* Added comment to explain checking of propTypes implementation, with link to react src
* Adjusted implementation of `dive()` so that `options.context` (passed to `dive()`) overrides anything else.

This is ready for another round!

packages/enzyme/src/ShallowWrapper.js
minznerjosh: It looks like master (and this PR) has the correct behavior. [React first checks if `shouldComponentUpdate` is a function, and _then_ checks if the component is a `PureComponent`](https://github.com/facebook/react/blob/4846809370737f13c0465ac7b83764b408c6e732/packages/react-reconciler/src/ReactFiberClassComponent.js#L259-L284).

packages/enzyme/src/ShallowWrapper.js
ljharb: I'm confused; this problem (enzyme doesn't provide proper child context) definitely predates react 16 - since the shallow renderer only renders one level, i'm not sure how it would take child context into account?

packages/enzyme/src/ShallowWrapper.js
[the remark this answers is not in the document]
ljharb: hmm. I'm not sure that's what we want. if it's not in the child context, it shouldn't be passed to the new child.

packages/enzyme/src/ShallowWrapper.js
ljharb: Gotcha. So we still have to do work for every react version, only *more* work in 16+.

minznerjosh: Exactly!

ljharb:
@minznerjosh is there any chance the functionality is actually accessible inside one of the react packages, without copy/paste?

minznerjosh:
@ljharb Unfortunately, I don't think so. 🙁Both pieces of logic live in the `react-reconciler` package [here](https://github.com/facebook/react/blob/10a7a5b5ced683125d68584a78084faac3197846/packages/react-reconciler/src/ReactFiberClassComponent.js#L492-L499) and [here](https://github.com/facebook/react/blob/10a7a5b5ced683125d68584a78084faac3197846/packages/react-reconciler/src/ReactFiberContext.js#L149-L213), but neither are public exports, and the public exports that use those functions (like [`updateContainer`](https://github.com/facebook/react/blob/10a7a5b5ced683125d68584a78084faac3197846/packages/react-reconciler/src/ReactFiberReconciler.js#L283-L299)) do too much to be useful for us.

minznerjosh:
P.S. Just wanna ping [the conversation happening over here](https://github.com/airbnb/enzyme/pull/1971#discussion_r249040520) as that thread still needs to be closed. 😊

packages/enzyme/src/ShallowWrapper.js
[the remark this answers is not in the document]
ljharb: ahhhh ok - then yes, the larger object should indeed be passed down, as long as the component only sees the filtered context. Thanks for explaining.

ljharb:
@minznerjosh what about https://unpkg.com/prop-types@15.6.2/checkPropTypes.js ?

packages/enzyme/src/ShallowWrapper.js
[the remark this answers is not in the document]
minznerjosh: My pleasure!

minznerjosh:
@ljharb Can we live with the duplication of react logic here? If so, anything else blocking?

ljharb:
It's not ideal, but we can live with it :-) I'm reviewing and making some tweaks locally now.

ljharb:
Thanks for fixing this long-standing bug!

PR #1972, opened 2019-01-08
Upgrade eslint-plugin-react in packages

minznerjosh:
I started getting tired of seeing the red x next to my builds so I decided to figure out why the linter was failing. Lo and behold, @ljharb fixed the issue in `eslint-plugin-react` a few days ago!

It looks like you upgraded the dependency in the top-level `package.json`, but not in the child packages, so I figured I'd help you out with that chore.

P.S. I believe it is also a common monorepo pattern to only declare `devDependencies` in the top-level `package.json`. We could also move to that pattern to avoid chores like this in the future.

### DIFF
--- a/packages/enzyme-adapter-react-13/package.json
+++ b/packages/enzyme-adapter-react-13/package.json
@@ -52,7 +52,7 @@
     "eslint-config-airbnb": "^17.1.0",
     "eslint-plugin-import": "^2.14.0",
     "eslint-plugin-jsx-a11y": "^6.1.2",
-    "eslint-plugin-react": "^7.11.1",
+    "eslint-plugin-react": "^7.12.3",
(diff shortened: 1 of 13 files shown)


ljharb:
Personally I think that's a terrible pattern; each package should be able to operate as if it's in a standalone, separate repo, as would be preferred.

minznerjosh:
I think it depends on what your goals are. If the goal is to make sure that the dependencies are the same across the entire _project_, no matter what, then it seems like a reasonable choice. You know what’s best for enzyme. 😊

And whoops. I’ve been in yarn world for the past 2 years (where versions are locked in yarn.lock) so I forgot the bump wasn’t necessary.

ljharb:
Only apps should have lockfiles :-)

PR #1975, opened 2019-01-10
Support React.lazy and React.Suspense

### LINKED ISSUE #1917: Support Suspense
### Current behavior

I'm currently getting the following error using a Suspense component in my tests.

    Enzyme Internal Error: unknown node with tag 13

Looking at the current react work tags, this is for the suspense component: https://github.com/facebook/react/blob/v16.6.0/packages/shared/ReactWorkTags.js#L44

Looking at the current implementation of `detectFiberTag` it doesn't look for Suspense components: https://github.com/airbnb/enzyme/blob/enzyme%403.7.0/packages/enzyme-adapter-react-16/src/detectFiberTags.js#L42

I believe the exception is coming here: https://github.com/airbnb/enzyme/blob/enzyme%403.7.0/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js#L103 after it hits the default for the switch.

### Expected behavior

Can handle suspense components

#### API

- [ ] shallow
- [x] mount
- [ ] render

#### Version

| enzyme              |  3.7.0
| react               |  16.6.3
| react-dom           | 16.6.3
| react-test-renderer | 16.6.3
| adapter (below)     | 1.7.0

#### Adapter

- [x] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] others ( )

chenesan:
Fixes #1917 .

Given a component wrapped by `React.lazy` in `<Suspense />`
It'll plainly render a `<Lazy />` in `shallow` and render fallback component in mount.

There are something I'm not sure / still working on:

1. What should `displayName` of the component returned by `React.lazy` be? Currently I directly named it as `Lazy`. Not sure if it's something we could define by ourselves.

2. I'm trying to add an `waitUntilLazyLoaded()` on `ReactWrapper`, which will return a promise resolving when the dynamic import loaded and React trigger the re-render, so we can write some tests like:

```jsx

const LazyComponent = lazy(() => import('/path/to/dynamic/component'));
const Fallback = () => <div />;
const SuspenseComponent = () => (
    <Suspense fallback={<Fallback />}>
      <LazyComponent />
    </Suspense>
);

const wrapper = mount(<SuspenseComponent />)
await wrapper.waitUntilLazyLoaded()

expect(wrapper.find('DynamicComponent').to.have.lengthOf(1)

```

But I don't know how to detect if all the lazy loaded component inside `<Suspense />` has completeted loading. It looks like that we have to hack around react fiber. @ljharb Would you know any way to detect this?

Also note that this PR add `babel-plugin-dynamic-import-node` and `babel-plugin-syntax-dynamic-import` for using `import()`, `babel-eslint` in `enzyme-adapter-react-16` and `enzyme-test-suite` for dynamic import support of eslint.

### DIFF
--- a/docs/api/shallow.md
+++ b/docs/api/shallow.md
@@ -50,8 +50,9 @@ describe('<MyComponent />', () => {
   - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, `componentDidMount`
 is not called on the component, and `componentDidUpdate` is not called after
 [`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md). Default to `false`.
-- `options.wrappingComponent`: (`ComponentType` [optional]): A component that will render as a parent of the `node`. It can be used to provide context to the `node`, among other things. See the [`getWrappingComponent()` docs](ShallowWrapper/getWrappingComponent.md) for an example. **Note**: `wrappingComponent` _must_ render its children.
-- `options.wrappingComponentProps`: (`Object` [optional]): Initial props to pass to the `wrappingComponent` if it is specified.
+  - `options.wrappingComponent`: (`ComponentType` [optional]): A component that will render as a parent of the `node`. It can be used to provide context to the `node`, among other things. See the [`getWrappingComponent()` docs](ShallowWrapper/getWrappingComponent.md) for an example. **Note**: `wrappingComponent` _must_ render its children.
+  - `options.wrappingComponentProps`: (`Object` [optional]): Initial props to pass to the `wrappingComponent` if it is specified.
+  - `options.suspenseFallback`: (`Boolean` [optional]): If set to true, when rendering `Suspense` enzyme will replace all the lazy components in children with `fallback` element prop. Otherwise it won't handle fallback of lazy component. Default to `true`. Note: not supported in React < 16.6.
 
 #### Returns
 
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -10,6 +10,7 @@ import { version as testRendererVersion } from 'react-test-renderer/package.json
 import TestUtils from 'react-dom/test-utils';
 import semver from 'semver';
 import checkPropTypes from 'prop-types/checkPropTypes';
+import has from 'has';
 import {
   AsyncMode,
   ConcurrentMode,
@@ -22,13 +23,17 @@ import {
   isContextProvider,
   isElement,
   isForwardRef,
+  isLazy,
   isMemo,
   isPortal,
+  isSuspense,
   isValidElementType,
+  Lazy,
   Memo,
   Portal,
   Profiler,
   StrictMode,
+  Suspense,
 } from 'react-is';
 import { EnzymeAdapter } from 'enzyme';
 import { typeOfNode } from 'enzyme/build/Utils';
@@ -228,6 +233,19 @@ function toTree(vnode) {
         rendered: childrenToTree(node.child),
       };
     }
+    case FiberTags.Suspense: {
+      return {
+        nodeType: 'function',
+        type: Suspense,
+        props: { ...node.memoizedProps },
+        key: ensureKeyOrUndefined(node.key),
+        ref: node.ref,
+        instance: null,
+        rendered: childrenToTree(node.child),
+      };
+    }
+    case FiberTags.Lazy:
+      return childrenToTree(node.child);
     default:
       throw new Error(`Enzyme Internal Error: unknown node with tag ${node.tag}`);
   }
@@ -273,6 +291,25 @@ function nodeToHostNode(_node) {
     return node.rendered.map(mapper);
   }
   return mapper(node);
+}
+
+function replaceLazyWithFallback(node, fallback) {
+  if (!node) {
+    return null;
+  }
+  if (Array.isArray(node)) {
+    return node.map(el => replaceLazyWithFallback(el, fallback));
+  }
+  if (isLazy(node.type)) {
+    return fallback;
+  }
+  return {
+    ...node,
+    props: {
+      ...node.props,
+      children: replaceLazyWithFallback(node.props.children, fallback),
+    },
+  };
 }
 
 const eventOptions = {
@@ -351,6 +388,9 @@ class ReactSixteenAdapter extends EnzymeAdapter {
 
   createMountRenderer(options) {
     assertDomAvailable('mount');
+    if (has(options, 'suspenseFallback')) {
+      throw new TypeError('`suspenseFallback` is not supported by the `mount` renderer');
+    }
     if (FiberTags === null) {
       // Requires DOM.
       FiberTags = detectFiberTags();
@@ -445,9 +485,13 @@ class ReactSixteenAdapter extends EnzymeAdapter {
     };
   }
 
-  createShallowRenderer(/* options */) {
+  createShallowRenderer(options = {}) {
     const adapter = this;
     const renderer = new ShallowRenderer();
+    const { suspenseFallback } = options;
+    if (typeof suspenseFallback !== 'undefined' && typeof suspenseFallback !== 'boolean') {
+      throw TypeError('`options.suspenseFallback` should be boolean or undefined');
+    }
     let isDOM = false;
     let cachedNode = null;
 
@@ -498,8 +542,20 @@ class ReactSixteenAdapter extends EnzymeAdapter {
           return withSetStateAllowed(() => renderer.render({ ...el, type: MockConsumer }));
         } else {
           isDOM = false;
-          const { type: Component } = el;
-
+          let renderedEl = el;
+          if (isLazy(renderedEl)) {
+            throw TypeError('`React.lazy` is not supported by shallow rendering.');
+          }
+          if (isSuspense(renderedEl)) {
+            let { children } = renderedEl.props;
+            if (suspenseFallback) {
+              const { fallback } = renderedEl.props;
+              children = replaceLazyWithFallback(children, fallback);
+            }
+            const FakeSuspenseWrapper = () => children;
+            renderedEl = React.createElement(FakeSuspenseWrapper, null, children);
+          }
+          const { type: Component } = renderedEl;
           const isStateful = Component.prototype && (
             Component.prototype.isReactComponent
             || Array.isArray(Component.__reactAutoBindPairs) // fallback for createClass components
@@ -517,7 +573,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
 
           if (!isStateful && typeof Component === 'function') {
             return withSetStateAllowed(() => renderer.render(
-              { ...el, type: wrapFunctionalComponent(Component) },
+              { ...renderedEl, type: wrapFunctionalComponent(Component) },
               context,
             ));
           }
@@ -546,7 +602,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
               });
             }
           }
-          return withSetStateAllowed(() => renderer.render(el, context));
+          return withSetStateAllowed(() => renderer.render(renderedEl, context));
         }
       },
       unmount() {
@@ -609,6 +665,9 @@ class ReactSixteenAdapter extends EnzymeAdapter {
   }
 
   createStringRenderer(options) {
+    if (has(options, 'suspenseFallback')) {
+      throw new TypeError('`suspenseFallback` should not be specified in options of string renderer');
+    }
     return {
       render(el, context) {
         if (options.context && (el.type.contextTypes || options.childContextTypes)) {
@@ -676,6 +735,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
         case StrictMode || NaN: return 'StrictMode';
         case Profiler || NaN: return 'Profiler';
         case Portal || NaN: return 'Portal';
+        case Suspense || NaN: return 'Suspense';
         default:
       }
     }
@@ -692,6 +752,13 @@ class ReactSixteenAdapter extends EnzymeAdapter {
         }
         const name = displayNameOfNode({ type: type.render });
         return name ? `ForwardRef(${name})` : 'ForwardRef';
+      }
+      case Lazy || NaN: {
+        if (type.displayName) {
+          return type.displayName;
+        }
+        const name = displayNameOfNode({ type: type._result });
+        return name ? `lazy(${name})` : 'lazy';
       }
       default: return displayNameOfNode(node);
     }
@@ -716,6 +783,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
       || isForwardRef(fakeElement)
       || isContextProvider(fakeElement)
       || isContextConsumer(fakeElement)
+      || isSuspense(fakeElement)
     );
   }
 
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -754,11 +754,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
         return name ? `ForwardRef(${name})` : 'ForwardRef';
       }
       case Lazy || NaN: {
-        if (type.displayName) {
-          return type.displayName;
-        }
-        const name = displayNameOfNode({ type: type._result });
-        return name ? `lazy(${name})` : 'lazy';
+        return 'lazy';
       }
       default: return displayNameOfNode(node);
     }
--- a/packages/enzyme-test-suite/test/Adapter-spec.jsx
+++ b/packages/enzyme-test-suite/test/Adapter-spec.jsx
@@ -9,20 +9,22 @@ import {
 } from 'react-is';
 import PropTypes from 'prop-types';
 import wrap from 'mocha-wrap';
-import { wrapWithWrappingComponent, RootFinder } from 'enzyme-adapter-utils';
+import { fakeDynamicImport, wrapWithWrappingComponent, RootFinder } from 'enzyme-adapter-utils';
 
 import './_helpers/setupAdapters';
 import Adapter from './_helpers/adapter';
 import {
-  renderToString,
+  AsyncMode,
+  ConcurrentMode,
   createContext,
   createPortal,
   forwardRef,
   Fragment,
-  StrictMode,
-  AsyncMode,
-  ConcurrentMode,
+  lazy,
   Profiler,
+  renderToString,
+  StrictMode,
+  Suspense,
 } from './_helpers/react-compat';
 import { is } from './_helpers/version';
 import { itIf, describeWithDOM, describeIf } from './_helpers';
@@ -1062,6 +1064,56 @@ describe('Adapter', () => {
 
     itIf(is('>= 16.6'), 'supports ConcurrentMode', () => {
       expect(getDisplayName(<ConcurrentMode />)).to.equal('ConcurrentMode');
+    });
+
+    itIf(is('>= 16.6'), 'supports Suspense', () => {
+      expect(getDisplayName(<Suspense />)).to.equal('Suspense');
+    });
+
+    itIf(is('>= 16.6'), 'supports lazy', () => {
+      class DynamicComponent extends React.Component {
+        render() {
+          return <div>DynamicComponent</div>;
+        }
+      }
+      const LazyComponent = lazy(() => fakeDynamicImport(DynamicComponent));
+      expect(getDisplayName(<LazyComponent />)).to.equal('lazy');
+    });
+
+    itIf(is('>= 16.6'), 'show explicitly defined display name of lazy component', () => {
+      class DynamicComponent extends React.Component {
+        render() {
+          return <div>DynamicComponent</div>;
+        }
+      }
+      const theDisplayName = 'SOMETHING';
+      const LazyComponent = Object.assign(lazy(() => fakeDynamicImport(DynamicComponent)), { displayName: theDisplayName });
+      expect(getDisplayName(<LazyComponent />)).to.equal(theDisplayName);
+    });
+
+    itIf(is('>= 16.6'), 'show display name of wrapped component of lazy', () => {
+      class ComponentWithDisplayName extends React.Component {
+        render() {
+          return <div>DynamicComponent</div>;
+        }
+      }
+      ComponentWithDisplayName.displayName = 'Something';
+      const LazyComponent = lazy(() => fakeDynamicImport(ComponentWithDisplayName));
+      /* eslint-disable no-underscore-dangle */
+      LazyComponent._result = ComponentWithDisplayName;
+      expect(getDisplayName(<LazyComponent />)).to.equal(`lazy(${ComponentWithDisplayName.displayName})`);
+    });
+
+    itIf(is('>= 16.6'), 'show name of wrapped component of lazy if its displayName is empty', () => {
+      class ComponentWithoutDisplayName extends React.Component {
+        render() {
+          return <div>DynamicComponent</div>;
+        }
+      }
+      const LazyComponent = lazy(() => fakeDynamicImport(ComponentWithoutDisplayName));
+      /* eslint-disable no-underscore-dangle */
+      LazyComponent._result = ComponentWithoutDisplayName;
+      expect(getDisplayName(<LazyComponent />)).to.equal('lazy(ComponentWithoutDisplayName)');
     });
   });
 
--- a/packages/enzyme-test-suite/test/Adapter-spec.jsx
+++ b/packages/enzyme-test-suite/test/Adapter-spec.jsx
@@ -1079,42 +1079,6 @@ describe('Adapter', () => {
       const LazyComponent = lazy(() => fakeDynamicImport(DynamicComponent));
       expect(getDisplayName(<LazyComponent />)).to.equal('lazy');
     });
-
-    itIf(is('>= 16.6'), 'show explicitly defined display name of lazy component', () => {
-      class DynamicComponent extends React.Component {
-        render() {
-          return <div>DynamicComponent</div>;
-        }
-      }
-      const theDisplayName = 'SOMETHING';
-      const LazyComponent = Object.assign(lazy(() => fakeDynamicImport(DynamicComponent)), { displayName: theDisplayName });
-      expect(getDisplayName(<LazyComponent />)).to.equal(theDisplayName);
-    });
-
-    itIf(is('>= 16.6'), 'show display name of wrapped component of lazy', () => {
-      class ComponentWithDisplayName extends React.Component {
-        render() {
-          return <div>DynamicComponent</div>;
-        }
-      }
-      ComponentWithDisplayName.displayName = 'Something';
-      const LazyComponent = lazy(() => fakeDynamicImport(ComponentWithDisplayName));
-      /* eslint-disable no-underscore-dangle */
-      LazyComponent._result = ComponentWithDisplayName;
-      expect(getDisplayName(<LazyComponent />)).to.equal(`lazy(${ComponentWithDisplayName.displayName})`);
-    });
-
-    itIf(is('>= 16.6'), 'show name of wrapped component of lazy if its displayName is empty', () => {
-      class ComponentWithoutDisplayName extends React.Component {
-        render() {
-          return <div>DynamicComponent</div>;
-        }
-      }
-      const LazyComponent = lazy(() => fakeDynamicImport(ComponentWithoutDisplayName));
-      /* eslint-disable no-underscore-dangle */
-      LazyComponent._result = ComponentWithoutDisplayName;
-      expect(getDisplayName(<LazyComponent />)).to.equal('lazy(ComponentWithoutDisplayName)');
-    });
   });
 
   describeIf(is('>= 16.2'), 'determines if node isFragment', () => {
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -15,17 +15,20 @@ import {
   withSetStateAllowed,
 } from 'enzyme/build/Utils';
 import getAdapter from 'enzyme/build/getAdapter';
+import { fakeDynamicImport } from 'enzyme-adapter-utils';
 
 import './_helpers/setupAdapters';
 import {
   createClass,
   createContext,
   createPortal,
-  Fragment,
   forwardRef,
+  Fragment,
+  lazy,
   memo,
   Profiler,
   PureComponent,
+  Suspense,
   useEffect,
   useState,
 } from './_helpers/react-compat';
@@ -34,6 +37,7 @@ import {
   describeIf,
   itIf,
 } from './_helpers';
+import getLoadedLazyComponent from './_helpers/getLoadedLazyComponent';
 import describeMethods from './_helpers/describeMethods';
 import {
   is,
@@ -1077,6 +1081,132 @@ describeWithDOM('mount', () => {
     'unmount',
     'wrap',
   );
+
+  describeIf(is('>= 16.6'), 'Suspense & lazy', () => {
+    class DynamicComponent extends React.Component {
+      render() {
+        return (
+          <div>Dynamic Component</div>
+        );
+      }
+    }
+    class Fallback extends React.Component {
+      render() {
+        return (
+          <div>Fallback</div>
+        );
+      }
+    }
+    it('finds Suspense and its children when no lazy component', () => {
+      class Component extends React.Component {
+        render() {
+          return (
+            <div>test</div>
+          );
+        }
+      }
+
+      const SuspenseComponent = () => (
+        <Suspense fallback={Fallback}>
+          <Component />
+        </Suspense>
+      );
+
+      const wrapper = mount(<SuspenseComponent />);
+
+      expect(wrapper.is(SuspenseComponent)).to.equal(true);
+      expect(wrapper.find(Component)).to.have.lengthOf(1);
+      expect(wrapper.find(Fallback)).to.have.lengthOf(0);
+    });
+
+    it('can mount Suspense directly', () => {
+      const wrapper = mount(<Suspense fallback={Fallback} />);
+      expect(wrapper.is(Suspense)).to.equal(true);
+    });
+
+    it('finds fallback when given lazy component in initial mount', () => {
+      const LazyComponent = lazy(() => fakeDynamicImport(DynamicComponent));
+      const SuspenseComponent = () => (
+        <Suspense fallback={<Fallback />}>
+          <LazyComponent />
+        </Suspense>
+      );
+
+      const wrapper = mount(<SuspenseComponent />);
+
+      expect(wrapper.is(SuspenseComponent)).to.equal(true);
+      expect(wrapper.find(LazyComponent)).to.have.lengthOf(0);
+      expect(wrapper.find(Fallback)).to.have.lengthOf(1);
+    });
+
+    it('return fallback string when given lazy component in initial mount and call .debug()', () => {
+      const LazyComponent = lazy(() => fakeDynamicImport(DynamicComponent));
+      const SuspenseComponent = () => (
+        <Suspense fallback={<Fallback />}>
+          <LazyComponent />
+        </Suspense>
+      );
+
+      const wrapper = mount(<SuspenseComponent />);
+
+      expect(wrapper.debug()).to.equal(`<SuspenseComponent>
+  <Suspense fallback={{...}}>
+    <Fallback>
+      <div>
+        Fallback
+      </div>
+    </Fallback>
+  </Suspense>
+</SuspenseComponent>`);
+    });
+
+    it('return wrapped component when given loaded lazy component in initial mount', () => {
+      const LazyComponent = getLoadedLazyComponent(DynamicComponent);
+      const SuspenseComponent = () => (
+        <Suspense fallback={<Fallback />}>
+          <LazyComponent />
+        </Suspense>
+      );
+
+      const wrapper = mount(<SuspenseComponent />);
+
+      expect(wrapper.is(SuspenseComponent)).to.equal(true);
+      expect(wrapper.find(LazyComponent)).to.have.lengthOf(0);
+      expect(wrapper.find(DynamicComponent)).to.have.lengthOf(1);
+      expect(wrapper.find(Fallback)).to.have.lengthOf(0);
+    });
+
+    it('return wrapped component string when given loaded lazy component in initial mount and call .debug()', () => {
+      const LazyComponent = getLoadedLazyComponent(DynamicComponent);
+      const SuspenseComponent = () => (
+        <Suspense fallback={<Fallback />}>
+          <LazyComponent />
+        </Suspense>
+      );
+
+      const wrapper = mount(<SuspenseComponent />);
+
+      expect(wrapper.debug()).to.equal(`<SuspenseComponent>
+  <Suspense fallback={{...}}>
+    <DynamicComponent>
+      <div>
+        Dynamic Component
+      </div>
+    </DynamicComponent>
+  </Suspense>
+</SuspenseComponent>`);
+    });
+
+    it('throws if options.suspenseFallback is specified', () => {
+      const LazyComponent = lazy(fakeDynamicImport(DynamicComponent));
+      const SuspenseComponent = () => (
+        <Suspense fallback={<Fallback />}>
+          <LazyComponent />
+        </Suspense>
+      );
+      expect(() => mount(<SuspenseComponent />, { suspenseFallback: false })).to.throw();
+    });
+  });
 
   describe('.mount()', () => {
     it('calls componentWillUnmount()', () => {
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -1688,20 +1688,6 @@ describe('shallow', () => {
       expect(wrapper.debug()).to.equal('<lazy />');
     });
 
-    it('returns lazy component string including wrapped component when debug() is called and lazy component is loaded', () => {
-      const LazyComponent = lazy(fakeDynamicImport(DynamicComponent));
-      /* eslint-disable no-underscore-dangle */
-      LazyComponent._result = DynamicComponent;
-
-      const wrapper = shallow((
-        <Suspense fallback={<Fallback />}>
-          <LazyComponent />
-        </Suspense>
-      ));
-
-      expect(wrapper.debug()).to.equal('<lazy(DynamicComponent) />');
-    });
-
     it('replaces LazyComponent with Fallback when render Suspense if options.suspenseFallback=true', () => {
       const LazyComponent = lazy(fakeDynamicImport(DynamicComponent));
 
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -14,15 +14,18 @@ import {
   withSetStateAllowed,
 } from 'enzyme/build/Utils';
 import getAdapter from 'enzyme/build/getAdapter';
+import { fakeDynamicImport } from 'enzyme-adapter-utils';
 
 import './_helpers/setupAdapters';
 import {
   createClass,
   createContext,
   createPortal,
-  Fragment,
   forwardRef,
+  Fragment,
+  lazy,
   PureComponent,
+  Suspense,
   useEffect,
   useState,
   Profiler,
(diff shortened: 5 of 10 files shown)


rodoabad:
@chenesan are you not able to determine the display name of the component that is being loaded by `React.lazy`? Once the promise has resolved you should have full access to the lazy loaded component correct?

chenesan:
@rodoabad I'm not sure what the `displayName` of the wrapper component returned by `React.lazy` is. See the test in https://github.com/airbnb/enzyme/blob/master/packages/enzyme-test-suite/test/Adapter-spec.jsx#L987-L1011 and https://github.com/airbnb/enzyme/blob/master/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js#L493-L525

chenesan:
@ljharb I just force pushed the branch and I think it's ok to review it now.

Basically this PR will supports shallow/mount with `Suspense` and `React.lazy`.

.babelrc
```diff
@@ -2,6 +2,8 @@
   "presets": ["airbnb"],
   "plugins": [
     ["transform-replace-object-assign", { "moduleSpecifier": "object.assign" }],
+    "syntax-dynamic-import",
+    "dynamic-import-node"
```
ljharb: let's only add these in the "test" env; i don't think this is ever a safe transform except at the app level.

packages/enzyme-adapter-react-16/.eslintrc
```diff
@@ -1,5 +1,6 @@
 {
   "extends": "airbnb",
+  "parser": "babel-eslint",
```
ljharb: this is also not acceptable; babel-eslint enables all syntax currently, so airbnb projects only use the default parser.
```suggestion
```

packages/enzyme-adapter-react-16/src/detectFiberTags.jsx
```diff
@@ -38,6 +67,9 @@ module.exports = function detectFiberTags() {
     // eslint-disable-next-line no-unused-vars
     FwdRef = React.forwardRef((props, ref) => null);
   }
+  if (supportsLazy) {
+    LazyComponent = React.lazy(() => import('./_helpers/dynamicImportedComponent'));
```
ljharb: There's zero point in using `import()` here - it has to be transpiled, and it's never safe to transpile it below the app level, so this would have to go untranspiled, and that'd be a breaking change.

It'd be much simpler to use `() => Promise.resolve().then(() => ({ default: require(…) }))`.

Separately, I'm not sure why this can't be even simpler - `React.lazy(() => Promise.resolve().then(() => ({ default() { return null; } }))`

packages/enzyme-test-suite/.eslintrc
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "airbnb",
   "root": true,
+  "parser": "babel-eslint",
```
ljharb:
```suggestion
```

packages/enzyme-test-suite/package.json
```diff
@@ -35,8 +35,8 @@
     "jsdom": "^6.5.1",
     "lodash.isequal": "^4.5.0",
     "mocha-wrap": "^2.1.2",
-    "object.assign": "^4.1.0",
     "object-inspect": "^1.6.0",
+    "object.assign": "^4.1.0",
```
ljharb: not sure why this got re-sorted, what version of npm did you use to install it?

packages/enzyme-test-suite/test/Adapter-spec.jsx
```diff
@@ -1008,6 +1010,15 @@ describe('Adapter', () => {
     itIf(is('>= 16.6'), 'supports ConcurrentMode', () => {
       expect(getDisplayName(<ConcurrentMode />)).to.equal('ConcurrentMode');
     });
+
+    itIf(is('>= 16.6'), 'supports Suspense', () => {
+      expect(getDisplayName(<Suspense />)).to.equal('Suspense');
+    });
+
+    itIf(is('>= 16.6'), 'supports lazy', () => {
+      const LazyComponent = lazy(() => import('./_helpers/dynamicImportedComponent'));
```
ljharb: similarly, this can/should use `new Promise`, not `import()`.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -560,6 +562,44 @@ describeWithDOM('mount', () => {
...
+    it('finds Suspense and its children when no lazy component', () => {
+      class Component extends React.Component {
+        render() {
+          return (
+            <div>test</div>
+          );
+        }
+      }
+      const SuspenseComponent = () => (
+        <Suspense fallback={false}>
```
ljharb: can we pass `Fallback` here, and assert that none are found?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -560,6 +562,44 @@ describeWithDOM('mount', () => {
...
+        <Suspense fallback={<Fallback />}>
+          <LazyComponent />
+        </Suspense>
+      );
+
+      const wrapper = mount(<SuspenseComponent />);
+
+      expect(wrapper.find('SuspenseComponent')).to.have.lengthOf(1);
+      expect(wrapper.find(LazyComponent)).to.have.lengthOf(0);
+      expect(wrapper.find('Fallback')).to.have.lengthOf(1);
```
```suggestion
      expect(wrapper.find(Fallback)).to.have.lengthOf(1);
```

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -560,6 +562,44 @@ describeWithDOM('mount', () => {
...
+      const Fallback = () => <div />;
+      const SuspenseComponent = () => (
+        <Suspense fallback={<Fallback />}>
+          <LazyComponent />
+        </Suspense>
+      );
+
+      const wrapper = mount(<SuspenseComponent />);
+
+      expect(wrapper.find('SuspenseComponent')).to.have.lengthOf(1);
```
ljharb: `wrapper` here *is* the SuspenseComponent, so `find` isn't the proper way to assert on it.
```suggestion
      expect(wrapper.is(SuspenseComponent)).to.equal(true);
```

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -560,6 +562,44 @@ describeWithDOM('mount', () => {
...
+      }
+      const SuspenseComponent = () => (
+        <Suspense fallback={false}>
+          <Component />
+        </Suspense>
+      );
+
+      const wrapper = mount(<SuspenseComponent />);
+
+      expect(wrapper.find('SuspenseComponent')).to.have.lengthOf(1);
```
ljharb: `wrapper` here *is* the SuspenseComponent, so `find` isn't the proper way to assert on it.
```suggestion
      expect(wrapper.is(SuspenseComponent)).to.equal(true);
```

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -560,6 +562,44 @@ describeWithDOM('mount', () => {
...
+      expect(wrapper.find(Component)).to.have.lengthOf(1);
+    });
+
+    it('finds fallback when given lazy component in initial mount', () => {
+      const LazyComponent = lazy(() => import('./_helpers/dynamicImportedComponent.jsx'));
```
ljharb: let's avoid `import` here.

packages/enzyme-test-suite/test/_helpers/dynamicImportedComponent.jsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const DynamicComponent = () => <div>Dynamic Component</div>;
```
ljharb: this entire file seems unnecessary if it exists in `enzyme`

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1273,6 +1275,45 @@ describe('shallow', () => {
...
+        const Fallback = () => <div />;
+        const SuspenseComponent = () => (
+          <Suspense fallback={<Fallback />}>
+            <LazyComponent />
+          </Suspense>
+        );
+
+        const wrapper = shallow(<SuspenseComponent />);
+
+        expect(wrapper.find('Suspense')).to.have.lengthOf(1);
```
ljharb: Since a shallow wrapper is "what the wrapped thing *renders*", i would expect this to be:
```suggestion
        expect(wrapper.is(Suspense)).to.equal(true);
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1273,6 +1275,45 @@ describe('shallow', () => {
...
+        expect(wrapper.find(LazyComponent)).to.have.lengthOf(1);
+        // won't render fallback in shallow renderer
```
ljharb: can we add another test that `.dive()`s the `Suspense`, and then becomes the fallback?

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1273,6 +1275,45 @@ describe('shallow', () => {
...
+        }
+        const SuspenseComponent = () => (
+          <Suspense fallback={false}>
+            <Component />
+          </Suspense>
+        );
+
+        const wrapper = shallow(<SuspenseComponent />);
+
+        expect(wrapper.find('Suspense')).to.have.lengthOf(1);
```
ljharb: Since a shallow wrapper is "what the wrapped thing *renders*", i would expect this to be:
```suggestion
        expect(wrapper.is(Suspense)).to.equal(true);
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1273,6 +1275,45 @@ describe('shallow', () => {
...
+      it('finds Suspense and its children when no lazy component', () => {
+        class Component extends React.Component {
+          render() {
+            return (
+              <div>test</div>
+            );
+          }
+        }
+        const SuspenseComponent = () => (
+          <Suspense fallback={false}>
```
ljharb: also here, let's pass a `Fallback`

packages/enzyme-adapter-react-16/src/detectFiberTags.jsx
chenesan: Ah... yeah, the `Promise.resolve().then(() => ({ default() { return null; } }))` indeed works. If so I think all the added eslint plugins for `import` syntax and `babel-eslint` can be removed. Thanks for this solution!

ljharb: For what it's worth, this way is *much* cleaner in the sense that it preserves a single source of truth of "how `import()` should work in node" - but due to the underspecified nature of ESM, until node ships unflagged ESM, it's not safe for anything but the top-level app to define "how `import()` works", unfortunately.

chenesan: I think maybe it's worth to add a utility to make it clearer like

```js
const fakeDynamicImport = (module) => Promise.resolve().then(() => ({ default: module}))

// so we can write
const LazyComponent = React.lazy(fakeDynamicImport(SomeComponent))
```
for all the places using `React.lazy` in enzyme packages. But I'm not sure where to put this utility.

ljharb: Sounds great! If it'd be used by only adapters, it should be in adapter-utils; if by enzyme as well, then probably in enzyme itself; since it's likely to only be used in the 16 adapter for now but the 17 adapter later, probably in adapter-utils?

chenesan: `React.lazy` only appears in 16 adapter and tests. Looks like adapter-utils is a good place.

packages/enzyme-test-suite/package.json
chenesan: It looks like I accidentally add new package with yarn. will fix this.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
chenesan: Oops, I forgot to test shallow render `Suspense` directly, and I found out that for now `react-test-renderer/shallow` doesn't allow render `Suspese`:

```js
import ShallowRenderer from "react-test-renderer/shallow"

const renderer = new ShallowRenderer()
renderer.render(<Suspense fallback={false} />)

```

will throw:

```
Invariant Violation: ReactShallowRenderer render(): Shallow rendering works only with custom components, but the provided element type was `symbol`.
    at invariant (/Users/yishan/Documents/Projects/enzyme/packages/enzyme-adapter-react-16/node_modules/react-test-renderer/cjs/react-test-renderer-shallow.development.js:54:15)
    at ReactShallowRenderer.render (/Users/yishan/Documents/Projects/enzyme/packages/enzyme-adapter-react-16/node_modules/react-test-renderer/cjs/react-test-renderer-shallow.development.js:382:78)
```

It looks like that react shallow renderer cannot recoginze the `Suspense` type symbol. I should check this at first... So unless react-test-renderer supports this I think we can only support `Suspense` and `lazy` in mount. 😢

chenesan:
@ljharb I'm wondering if I should work on shallow renderer in react side to make progress for this PR.
If so, I'm not sure how shallow renderer should handle `fallback` of Suspense. Should we just return the children, or traverse the children of `Suspense` to find all `lazy` component and transform this to fallback element? I know this seems a bit off topic here, though.

chenesan:
Hi @ljharb I tried to start a PR in facebook/react#14638 to support shallow rendering Suspense in react shallow renderer. I'm not sure if I did it right for enzyme. I guess you may be interested in looking into this ;)

chenesan:
ok, I'll add them back. If there's still no reply in react this week I'll start to work on support Suspense in adapter 16.

And if we have to support Suspense in enzyme side, how should we handle the fallback inside Suspense? I think we should keep behavior of shallow as same as mount, and we could just turn unresolved lazy component node in children to fallback node. But if we keep `.dive()` down into the children and get a lazy component node, should we handle fallback for it? Or we should just throw error? I'd prefer to the latter since React doesn't support render lazy component independently, but I'm not sure if this behavior would confuse user.

ljharb:
@chenesan it kind of depends on what react's planning to do.

In other words, in general, we want shallow and mount to be as close as possible. However, when you're shallow rendering something that *renders* a Suspense or a lazy, those things should just show up in the tree - when you dive through one of them, or when you shallow render one of them directly, then their implementation should be "invoked", whatever that ends up meaning.

chenesan:
So when we shallow render Suspense
```js
const LazyComponent = React.lazy(() => import('InternalComponent'))
shallow(<Suspense fallback={<Fallback />}>
  <LazyComponent />
</Suspense>)
```
then should it render a fallback, or the `<LazyComponent />` -- but not invoke the lazy load function (which should be invoked when we shallow render `<LazyComponent />`?

And another problem is when we `shallow(<LazyComponent />)`, there seems to be some possible behaviors:

1. throw error, since React not allow singly render lazy component
2. wait it load until it resolved or error, if error we have to give it a fallback. But it's tricky to wait for loading in tests.
3. add option in `shallow`, or we could provide some utility `resetLoader` to replace `LazyComponent` loader function in run time, to replace `LazyComponent` with implementation when shallow rendering.

I think 1+3 sounds reasonable.

> Similarly, it should be possible (in both shallow and mount) to force a Suspense to render the lazy component or the fallback, so both paths can be tested.

it sounds like we can pass an option when shallow rendering to determine how to render `Suspense` - maybe a "normal" mode to render them as same as what React does and a "fallback" mode to render them all as fallback.

ljharb:
> then should it render a fallback, or the <LazyComponent /> -- but not invoke the lazy load function (which should be invoked when we shallow render <LazyComponent />?

Yes. Whether it should render the fallback or the LaxyComponent, I have no idea - (note that this only comes up when the Provider is the component being rendered).

As for `LazyComponent`, that's up to the author of the test - it seems like perhaps we only need a mechanism for the developer to instruct enzyme whether to use the fallback or to use the child.

chenesan:
ok, so I'll add an option (maybe `suspenseFallback: boolean`) into `shallow` and pass it down to adapter. When `shallow(<Suspense fallback={<Fallback />}>/* children */</Suspense>)`, if it's `true`, it will traverse through the children and replace all of `<LazyComponent />` with  `<Fallback />`, else we will keep them there.
And for `shallow(<LazyComponent />)` just let shallow renderer throw error(maybe rethrow with an clearer error message).
I'll work on them after this week if there's still no feedback from react side. Thanks for discussion :)

ljharb:
Sounds great. The only concern I think that's left is "what happens when you shallow render a lazy component", and it'd be great to know what React plans in the shallow renderer before we implement it here (we don't have to wait on them releasing that, we just need to know what they'll do)

chenesan:
Thanks for reply :-) I agree that it's easier to directly test the wrapped component itself and just want to make sure how React will handle this. Will keep waiting for reply from gaearon.

Also note some people may shallow render`<Suspense><LazyComponent /></Suspense>`, and call [.dive()](https://airbnb.io/enzyme/docs/api/ShallowWrapper/dive.html) to see what would be rendered inside lazy component. Not sure if we should support such use case. If React will not support shallow rendering lazy component then I think we could also not support this in enzyme.

bvaughn:
I don't feel comfortable saying what React _will_ or _will not_ do with regard to shallow rendering and lazy, but it does seem like a flawed premise. Really all that such a test case (`<Suspense><LazyComponent /></Suspense>`) would be testing is the renderer itself. What you'd really want to test is the component _returned by_ `LazyComponent`. (I don't personally see much value in shallow testing that either, but at least you're "testing" your own renderer code at that point.)

ljharb:
@bvaughn the question is about testing a lazy component in the absence of suspense; the answer might very well be “pretend the lazy wrapper isn’t there”.

bvaughn:
Yeah, I understand the question. I'm questioning the premise.

ljharb:
A module might export a lazy component, and the tests ideally shouldn’t have to know or care that it’s lazy - and if they do, they still need a way to test the unexported wrapped component, without a dom and without rendering the entire tree.

bvaughn:
`lazy` was designed to wrap an import, rather than to be the export. I think this pattern is more practical (since it defers the decision of code splitting to the caller) and easier to test.

chenesan:
hmm, (from my view) it looks like currently react is not interested in handling lazy in shallow renderer(or thinks it's a bad idea to shallow render lazy component for testing and users should directly shallow render the wrapped component). @gaearon (sorry for pinging you here again 😓 ) do you have any comments on @ljharb last comment? We might need a reply (even it's really no plan/interests on this) to go ahead (to determine what enzyme should do when people `shallow(<LazyComponent />`) for this PR, thanks :-)

bvaughn:
> While that makes perfect sense as advice for the author of the code, as a testing framework, we can’t limit ourselves to what might be a better pattern - we have to have a reasonable answer for every edge case.

We will have to disagree here then. I think that saying, "this is not a good pattern to use because of these reasons ___", is a reasonable answer.

chenesan:
> "this is not a good pattern to use because of these reasons ___"
Does it mean "So to discourage this pattern, currently we have no plan to support this and will throw error when shallow rendering lazy component" ?

ljharb:
@bvaughn that's fine to have as a stance from React - but that has to translate to explicit, defined behavior for the shallow renderer and react dev tools, which enzyme will follow. So, what's the explicit behavior react is committing to here, even if it's "throwing"?

bvaughn:
> Does it mean "So to discourage this pattern, currently we have no plan to support this and will throw error when shallow rendering lazy component" ? Or it would be an undefined behavior?

I've already mentioned that I don't think I'm the best one to say what the React team will do with regard to this behavior in the future, because shallow rendering is not a feature I've paid much attention to. 😄 That being said, the shallow renderer currently only supports custom components (class or function) and `fowrardRef`. Other types (including `React.Suspense`) will tigger an error. I'm not aware of any plans in the near/mid future to change this.

> @bvaughn that's fine to have as a stance from React - but that has to translate to explicit, defined behavior for the shallow renderer and react dev tools

Not sure I understand how React dev tools is related to this discussion. (They support `lazy`, as do ancillary packages like `react-is`.) The above statements are only scoped to shallow rendering and `React.lazy`.

chenesan:
Thanks for confirmation and response @bvaughn :-)
@ljharb if react has no plan for this, should we still try to handle lazy component in shallow? Personally I think we can throw for now because I'm not sure how many pepole would shallow rendering lazy component in their tests. If there are more feature requests and use cases then we could consider what to do next.

ljharb:
It's fine if, for now, we throw on shallow rendering a lazy component, as long as we can ensure the same description of it in the debug tree as `mount` would have.

chenesan:
>  as long as we can ensure the same description of it in the debug tree as `mount` would have.

Sorry that I couldn't get what you mean 😓 . Did you mean `.debug()` output of mounting (including lazy component) should be as same as shallowing?

```js
const element = <Suspense fallback={<Fallback />}>
  <LazyComponent />
</Suspense>
const mountWrapper = mount(element)
const shallowWrapper = shallow(element)

expect(mountWrapper.debug()).to.equal(shallowWrapper.debug())
```

in `mount` the `<LazyComponent />` will be rendered as `<Fallback />` in first render, and in shallow that will depends on the option(as above discussion) we passed into `shallow()` .

ljharb:
Let's move forward with this plan and with your understanding and we can tweak it as needed :-)

chenesan:
@ljharb I added an option `suspenseFallback` into second arg of `shallow`. When it's `true` (default) it will replace all lazy component in children with fallback when rendering `Suspense`, or it will keep all lazy component there. Also throw error inside enzyme when shallow rendering lazy component.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -559,7 +593,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
     if (!inst || !this.isValidElement(inst)) {
       return false;
     }
-    return typeof inst.type === 'function' || isForwardRef(inst);
+    return typeof inst.type === 'function' || isForwardRef(inst) || isSuspense(inst);
```
chenesan: I think `isCustomComponentElement` should be renamed after adding `Suspense`, but cannot figure out a better name :(

chenesan:
@ljharb Is there anything I could do to progress this?

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -344,9 +380,11 @@ class ReactSixteenAdapter extends EnzymeAdapter {
     };
   }
 
-  createShallowRenderer(/* options */) {
+  createShallowRenderer(options) {
     const adapter = this;
     const renderer = new ShallowRenderer();
+    const { suspenseFallback } = options;
+    const shouldReplaceLazyWithFallback = typeof suspenseFallback === 'undefined' ? true : !!suspenseFallback;
```
ljharb: let's throw here if `typeof suspenseFallback` isn't `undefined` or `boolean`

ljharb: additionally, let's throw if this option is specified at *all* in the other two renderers.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -357,20 +395,34 @@ class ReactSixteenAdapter extends EnzymeAdapter {
           isDOM = true;
         } else {
           isDOM = false;
-          const { type: Component } = el;
-
+          let renderedEl = el;
+          if (isLazy(renderedEl)) {
+            throw Error('Enzyme doesn\'t support lazy component because React doesn\'t support it.');
```
ljharb:
```suggestion
            throw TypeError('`React.lazy` is not supported by shallow rendering.');
```

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -344,9 +380,11 @@ class ReactSixteenAdapter extends EnzymeAdapter {
     };
   }
 
-  createShallowRenderer(/* options */) {
+  createShallowRenderer(options) {
```
ljharb:
```suggestion
  createShallowRenderer(options = {}) {
```
this keeps the `.length` the same.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -537,6 +590,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
         const name = type.render.displayName || functionName(type.render);
         return name ? `ForwardRef(${name})` : 'ForwardRef';
       }
+      case Lazy || NaN: return 'LazyComponent';
```
ljharb: `React.lazy` doesn't seem to provide a display name - let's use:
```suggestion
      case Lazy || NaN: {
        if (type.displayName) {
          return type.displayName;
        }
        const name = displayNameOfNode({ type: type._ctor });
        return name ? `React.lazy(${name})` : `lazy`;
      }
```
let's also add tests for an explicit `Object.assign(React.lazy(Component), { displayName: 'something' })`, as well as a lazy component wrapping something *with* a displayName, one without, and one with no name at all.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -537,6 +590,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
         const name = type.render.displayName || functionName(type.render);
```
ljharb: might as well:
```suggestion
        const name = displayNameOfNode({ type: type.render });
```

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
ljharb: this is basically `isNonHostElement`, but let's not worry about the name now

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
chenesan: It's tricky to test if displayName of wrapped component is included in displayName of lazy component. The `_ctor` is not the wrapped component itself but the function dynamically importing wrapped component. Before it's loaded we can't get the displayName of wrapped component.

ljharb: interesting, ok - that actually does make sense, considering that one of the use cases for `lazy` is a component that hasn't been fetched yet.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -279,6 +316,10 @@ class ReactSixteenAdapter extends EnzymeAdapter {
 
   createMountRenderer(options) {
     assertDomAvailable('mount');
+    assertSuspenseFallbackOptionNotExists(options, 'mount');
+    if (Object.prototype.hasOwnProperty.call(options, 'suspenseFallback')) {
```
ljharb:
```suggestion
    if (has(options, 'suspenseFallback')) {
```

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -344,9 +385,14 @@ class ReactSixteenAdapter extends EnzymeAdapter {
...
 
-  createShallowRenderer(/* options */) {
+  createShallowRenderer(options = {}) {
     const adapter = this;
     const renderer = new ShallowRenderer();
+    const { suspenseFallback } = options;
+    if (typeof suspenseFallback !== 'undefined' && typeof suspenseFallback !== 'boolean') {
+      throw TypeError('`options.suspenseFallback` should be boolean or undefined');
+    }
+    const shouldReplaceLazyWithFallback = typeof suspenseFallback === 'undefined' ? true : !!suspenseFallback;
```
ljharb:
```suggestion
    const shouldReplaceLazyWithFallback = typeof suspenseFallback === 'boolean' ? suspenseFallback : true;
```

altho i wonder if maybe we could invert this, so that "absent" and `false` are the same?

chenesan: That makes sense since now we throw error when it's not boolean or undefined.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -641,6 +641,51 @@ describeWithDOM('mount', () => {
...
+      LazyComponent._status = 1;
+      const SuspenseComponent = () => (
+        <Suspense fallback={<Fallback />}>
+          <LazyComponent />
+        </Suspense>
+      );
+
+      const wrapper = mount(<SuspenseComponent />);
+
+      expect(wrapper.debug()).to.equal(`<SuspenseComponent>
```
chenesan: Note that the lazy component node is disappeared in mount when its loaded (`_result` get loaded and `_status` = 1). And that will make the `.debug()` different between shallow and mount when rendering `Suspense`. I'm not sure if we should also ignore the loaded lazy component node when shallow rendering `Suspense` to make behavior consistent.

chenesan:
Has inverted the `suspenseFallback` option, added debug tests and also tests for loaded lazy component for mount. @ljharb can review this again now 😄

docs/api/shallow.md
```diff
@@ -50,6 +50,7 @@ describe('<MyComponent />', () => {
   - `options.disableLifecycleMethods`: (`Boolean` [optional]): If set to true, `componentDidMount`
 is not called on the component, and `componentDidUpdate` is not called after
 [`setProps`](ShallowWrapper/setProps.md) and [`setContext`](ShallowWrapper/setContext.md). Default to `false`.
+  - `options.suspenseFallback`: (`Boolean` [optional]): If set to true, when rendering `Suspense` enzyme will replace all the lazy components in children with `fallback` element prop. Otherwise it won't handle fallback of lazy component. Default to `true`.
```
ljharb: is this still correct?

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
ljharb: ping on this one

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -534,9 +595,16 @@ class ReactSixteenAdapter extends EnzymeAdapter {
...
-        const name = type.render.displayName || functionName(type.render);
+        const name = displayNameOfNode({ type: type.render });
         return name ? `ForwardRef(${name})` : 'ForwardRef';
       }
+      case Lazy || NaN: {
+        if (type.displayName) {
+          return type.displayName;
+        }
+        const name = displayNameOfNode({ type: type._result });
+        return name ? `React.lazy(${name})` : 'lazy';
```
ljharb: how do the dev tools show a lazy component?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
ljharb: can you elaborate on this?

I would expect that a Suspense is in one of two states: render the children, or render the fallback, and that shallow and mount would be consistent for each of these states.

docs/api/shallow.md
chenesan: Ah, yeah, the default is `false` now.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
chenesan: After look into what dev tools show I found it either shows the fallback component or the loaded wrapped component. It won't show the lazy component. But in shallow rendering because we keep the lazy component node, this is still used.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
chenesan: For example, the following jsx
```js
const LazyComponent = React.lazy(() => import('WrappedComponent'))

<Suspense fallback={<Fallback />}>
  <LazyComponent />
</Suspense>
```
in mount, react dev tool shows either fallback when it's not loaded:

```js
<Suspense fallback={...}>
  <Fallback />
</Suspense>
```

or the wrapped component

```js
<Suspense fallback={...}>
  <WrappedComponent />
</Suspense>
```

In both cases the `LazyComponent` node is not shown.

But in shallow with `options.suspenseFallback = false` we will keep the `LazyComponent` node there no matter it's loaded or not because (I think) we might want to check if the lazy component node exists. And this makes it inconsistent with mount.

chenesan:
@ljharb sorry that I put this PR aside for a long time.
Where should I start or clarify to progress this? Thanks :-)

IgorLeshchenko:
any updates on this?

ljharb:
@chenesan there's a few unresolved questions; i'll give it a fresh pass.

packages/enzyme-adapter-react-16/src/detectFiberTags.jsx
```diff
@@ -14,10 +15,38 @@ function getFiber(element) {
   return inst._reactInternalFiber.child;
 }
 
+function getLazyFiber(LazyComponent) {
+  const container = global.document.createElement('div');
```
ljharb: what happens when this runs on the server?

packages/enzyme-test-suite/test/Adapter-spec.jsx
```diff
@@ -1010,6 +1013,56 @@ describe('Adapter', () => {
...
+    itIf(is('>= 16.6'), 'show display name of wrapped component of lazy', () => {
+      class ComponentWithDisplayName extends React.Component {
+        render() {
+          return <div>DynamicComponent</div>;
+        }
+      }
+      ComponentWithDisplayName.displayName = 'Something';
+      const LazyComponent = lazy(() => fakeDynamicImport(ComponentWithDisplayName));
+      /* eslint-disable no-underscore-dangle */
+      LazyComponent._result = ComponentWithDisplayName;
```
ljharb: why is this needed?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -561,6 +564,134 @@ describeWithDOM('mount', () => {
...
+      </div>
+    </Fallback>
+  </Suspense>
+</SuspenseComponent>`);
+    });
+
+    it('return wrapped component when given loaded lazy component in initial mount', () => {
+      const LazyComponent = lazy(() => fakeDynamicImport(DynamicComponent));
+      /* eslint-disable no-underscore-dangle */
+      LazyComponent._result = DynamicComponent;
```
ljharb: also here?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -561,6 +564,134 @@ describeWithDOM('mount', () => {
...
+      /* eslint-disable no-underscore-dangle */
+      LazyComponent._status = 1;
```
ljharb: also here?

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
ljharb: We shouldn't need to mutate underscore-prefixed things to be able to test this.

packages/enzyme-test-suite/test/Adapter-spec.jsx
chenesan: I'd like to test what displayName will be if the wrapped component is loaded. And in lazy component we could get the wrapped component in `_result`.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
chenesan: I'd like to test the case where the wrapped component is loaded here.

packages/enzyme-adapter-react-16/src/detectFiberTags.jsx
chenesan: `getFiber` (https://github.com/airbnb/enzyme/pull/1975/files#diff-bbd8c2174348cfcf3910cf8876ae1d7cR5) also use `global.document` so I think it's fine.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
chenesan: I agree that it's not good to use `_result` to test this. I'd try to [pass synchronous thenable](https://github.com/facebook/react/pull/14626) to replace `fakeDynamicImport` to see if that works.

packages/enzyme-test-suite/test/Adapter-spec.jsx
ljharb: I'd prefer not to include these implementation details in the test. Can it be tested by providing an actually loaded component, or by awaiting the promise, rather than this?

chenesan:
Hi @ljharb I just changed two things:

(1) I decided not to have wrapped component display name in lazy component node, just show `lazy`.
Because:

* In shallow, lazy components are never loaded since the shallow renderer won't handle `lazy`. I think it's enough to just show `<lazy>` for the lazy component node.
* In mount, either the fallback or the loaded component will show in react developer tool (See [previous comment](https://github.com/airbnb/enzyme/pull/1975#discussion_r255287463)). The lazy component never shows.

(2) I add `fakeSyncThenable` in `enzyme-adapter-utils` to pass synchronous thenable to `lazy()` so we can directly test loaded lazy component in mount tests without `_result` trick.

chenesan:
I just found out that the ci failed in react 16.7 & 16.6 because they don't support passing sync thenable to `lazy()`. I rely on this to test the loaded lazy component. we also could just `await` a while to wait for the wrapped component loading in the tests but I don't think it's a good idea.  I guess we could just run the tests upon react 16.8?

ljharb:
I'm not sure why we couldn't wait on the promise (since we have direct access to it), rather than needing a sync thenable (either way, let's keep a sync thenable test for 16.8+)

chenesan:
The promise return the component module is loaded by React so we couldn't await it. In `lazy()` we just pass a function returning a promise and React will call it. My previous comment may be misleading since I say we "pass sync thenable into lazy()", but what we pass actually is a function returning promise.

ljharb:
We provide the promise, though - so we can import that into the tests and await it, no? In other words, not `() => makeNewPromise()`, but provide the same `===` promise to both lazy and the tests.

chenesan:
Ah I see 😓 , I would try that today or tomorrow.

chenesan:
Hmmm, await the promise before call `lazy()` doesn't work, it still doesn't load it in first render -- I think it's the same as `fakeDynamicImport()` since the implementation of `fakeDynamicImport` is also just return a resolved promise:

```javascript
export function fakeDynamicImport(moduleToImport) {
  return Promise.resolve({ default: moduleToImport });
}
```

And here is the code how react handle lazy component loading (https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberLazyComponent.js#L30-L87) It looks like even though the promise is resolved, it still have to wait for the first synchronous rendering to change the `_status` and `_result`.

Solutions I could think of are:
1. synchronous thenable (but only works in react v16.8)
2. `await delay(a short period of time)` after the mount.
3. hard coded `_result` and `_status` before mount.

The (1) is the simpler one I think, but if we want to run the test in v16.6, I think the (3) is the best we could do.

chenesan:
@ljharb I just made a `getLoadedLazyComponent` in `enzyme-adapter-utils` so:

* if React is > 16.8 it will utilize `fakeSyncThenable`.
* if React is 16.6 / 16.7 it will set `_result` and `_status` -- not ideal, but seems the best way under these versions.

I noticed the coverage check failed; not sure if I should also add a test for `getLoadedLazyComponent`.

danielsogl:
I don't want to be rough but this is a dealbreaker for me to use Enzyme as library in my enterprise application. Is there a special reason why the enzyme team cannot fix the todos in this PR so it can be merged?

DonikaV:
@ danielsogl +1, i am waiting this update. Because i cannot use Suspense, Lazy without tests... :|

chenesan:
Hi @ljharb , sorry for pinging, any news after my change in https://github.com/airbnb/enzyme/pull/1975#issuecomment-475927325 ?

ljharb:
Rebased; will do a more complete review shortly.

ljharb:
@chenesan looks good overall; please re-check-out the branch to my latest rebase, and we'll need to increase test coverage.

chenesan:
Just fixed some typos and rebased again. As for the coverage I found the coverage of `enzyme-adapter-utils` dropped a lot, however I'm sure we only added a `fakeDynamicImport` and not removed any old tests. I also saw the coverage is different between [travis ci log](https://travis-ci.org/airbnb/enzyme/jobs/516768192#L3492) and [coverall](https://coveralls.io/jobs/47069188). Not sure what's happening here. @ljharb Do you have any idea? thanks :-)

ljharb:
@chenesan coveralls merges coverage across many travis jobs; it should be higher than any individual travis job.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -228,6 +233,19 @@ function toTree(vnode) {
         rendered: childrenToTree(node.child),
       };
     }
+    case FiberTags.Suspense: {
+      return {
+        nodeType: 'suspense',
```
ljharb: hmm, is this nodeType needed? it should probably be `function`, no?

packages/enzyme-test-suite/test/_helpers/getLoadedLazyComponent.js
```diff
@@ -0,0 +1,31 @@
...
+  return {
+    then(resolve) {
+      return resolve({ default: result });
+    },
+  };
+}
+
+export default function getLoadedLazyComponent(wrappedComponent) {
+  if (is('>= 16.8')) {
+    return React.lazy(() => fakeSyncThenable(wrappedComponent));
```
ljharb: i missed this in the refactoring; `lazy` should be imported from `react-compat` here.

packages/enzyme-test-suite/test/_helpers/getLoadedLazyComponent.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+  if (is('>= 16.6')) {
+    const LazyComponent = React.lazy(() => fakeDynamicImport(wrappedComponent));
```
ljharb: also here

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
chenesan: Frankly speaking I'm not quite certain what `nodeType` it should be. I just think (probably I'm wrong) it's not a functional component so it couldn't be `function` and should be other type. It seems that `nodeType` is only used in enzyme internally so maybe it's OK to just have it `function`, though?

chenesan:
Just fixed the lazy import in `getLoadedLazyComponent`; Still not figured out why coverage of unchanged files (for example `createMountWrapper.js` `wrapWithSimpleWrapper.js`)  in `enzyme-adapter-utils/src` dropped. The missed branch is weird, for example https://coveralls.io/builds/22647796/source?filename=packages/enzyme-adapter-utils/src/createMountWrapper.jsx :

螢幕快照 2019-04-08 上午9 12 27

I think `createMountWrapper` is definitely hit in the React adapter, but I don't know why coverall marked the L45 as uncover and member function of `WrappedComponent` as missed branch.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
ljharb: unless there's a reason for it to be different, I think it's fine to make it appear to be a functional component.

chenesan: Fixed in 173001f .

chenesan:
@ljharb  I ran coverage on my machine and found that the coverage dropped in 7fdd617 . Will try to look into it this week.

chenesan:
@ljharb Still not figured this out :-( Sometimes the coverage recovered after I removed some lines of code, but when I tried to reproduce again it still dropped. Wondering if it's an issue in istanbul side -- Do you have any idea around this?

chenesan:
@ljharb Get this done in e60788c. It seems in some places we import `fakeDynamicImport` from `enzyme-adapter-utils/src/Utils` rather than `enzyme-adapter-utils`, which cause `nyc` count coverage incorrectly.

chenesan:
@ljharb Sorry for pinging again :-| Just fixed the coverage issue in e60788c so I think this could be reviewed / merged now?

DonikaV:
Sorry, i have a question. When i can use enzyme, shallow with lazy and Suspense? Is there any updates of library in the future? Merge was 8 days ago. But i see that the version of enzyme is the same...
Thanks in advance.

ljharb:
This will be in the next release; there’s no timeline for it.

GitHub-Mar:
I noticed this was released 3 days ago in v 1.13 of enzyme-adapter-react-16. Does Enzyme now fully support dynamically imported components and <Suspense/> wrapping? We are still struggling to get this working for our local tests. 

Is there any documentation or working examples someone could point me to? I've been trying to test a dynamically imported component that is not wrapped in a Suspense component as we are using a Suspense higher up the component tree to wrap everything, however, I'm unsure how to tell enzyme to resolve the promise and load through the component? If I debug I can see the node as a child of my mounted wrapper but it's just listed as a react.lazy. 

I believe this is the correct strategy when using Suspense components, though I'm also quite new to this and may well be wrong!

ljharb:
@GitHub-Mar no, this pr has two components, enzyme itself (not released) and the adapter (released).

GitHub-Mar:
Ah my mistake! So this feature is not currently supported in Enzyme, then? I'm guessing your above comment still stands and there is no timeline as to when this will be released?

ljharb:
Correct.

ljharb:

DonikaV:
Still failing. Do you have some docs for this? I mean maybe i should update my tests or something like this?
Thanks

ljharb:
@DonikaV all the docs are published. If you've fully updated enzyme and your adapter, please file a new issue and fill out the template.

shridharkalagi:
Can I use Enzyme `mount` API with Lazy and Suspense yet?

ron23:
I'm still a bit confused on what's the best way to test that a component wrapped in lazy renders inside Suspense.
Say I have:
```
function myComponent = (props) => {
   const MyLazy = React.lazy(() => import('./MyLazyComponent');
   return <Suspense fallback={...}><MyLazy/></Suspense>
}
```

If I use the method mentioned above to mock react the debug:
```
jest.mock('react', () => {
  const React = jest.requireActual('react');
  React.Suspense = ({ children }) => children
  return React;
});
```
 I get something like that:
```
<Component fallback={{...}}>
      <lazy />
    </Component>
```

I want to test that MyComponent actually renders MyLazyComponent.
I tried mount and shallow and neither works.

```
enzyme@3.11.0
enzyme-adapter-react-16@1.15.2
```

linwei0201:
is this feature available in the latest version of enzyme and react-adapter?
We are using `enzyme v3.11.0` and  `enzyme-adapter-react-16 v1.15.5`, but when I tried to use `wrapper.waitUntilLazyLoaded`, it saids "Property 'waitUntilLazyLoaded' does not exist on type 'ReactWrapper"

ljharb:
@linwei0201 yes, it was released in v3.10.0 - but this PR doesn't include anything called `waitUntilLazyLoaded` - that's not a thing that exists.

orlandovallejos:
So, what is the new name @ljharb ?
@ron23 were you able to test your Lazy component?

ljharb:
@orlandovallejos you can see it in the PR diff itself.

FirstWhack:
Making this comment for historical purposes.

@orlandovallejos no such property (by name or functionality) was implemented in this PR.


@ljharb, any chance you can delete your comment above? It's still causing confusion after 11 months :)

ljharb:
@FirstWhack i said `waitUntilLazyLoaded` was renamed, meaning, `waitUntilLazyLoaded` doesn't exist (but used to, in the original draft of this PR, but was rebased out).

However, I do see that it never in fact existed - the OP actually says "so i _can_ write tests like", meaning, that code snippet is imaginary.

"You can see it in the PR diff itself" means you can look at the PR diff to see what *did* land - but you're right that there's no similar awaitable thing.

PR #1981, opened 2019-01-18
[Fix] `shallow`: ensure SCU is called after GDSFP returns null

### LINKED ISSUE #1970: State updated before call to shouldComponentUpdate in shallow render
### Current behavior

Given a simple component like this:

```typescript
import * as React from "react";

interface Data {
    value: string;
}

export class SimpleComponent extends React.Component<Data, Data> {

    protected static getDerivedStateFromProps(props: Data, state: Data) {
        if (props.value === state.value) {
            return null;
        }

        console.log("updating state: ", props.value);
        return {
            value: props.value
        };
    }

    public state: Data = {
        value: this.props.value
    };

    public shouldComponentUpdate(nextProps: Data, nextState: Data) {
        console.log("state: ", this.state);
        console.log("nextState: ", nextState);
        console.log("props: ", this.props);
        console.log("nextProps: ", nextProps);
        const shouldUpdate = nextState.value !== this.state.value;
        console.log("shouldUpdate: ", shouldUpdate);
        return shouldUpdate;
    }

    public render() {
        console.log("render: ", this.state.value);
        return (
            <input value={this.props.value} />
        );
    }
}
```

I would expect the following test to pass:

```typescript
import * as React from "react";
import { Expect, Test, TestFixture, FocusTest } from "alsatian";
import { shallow } from "enzyme";

import { SimpleComponent } from "./SimpleComponent";

@TestFixture("SimpleComponent")
export class SimpleComponentTests {

    @FocusTest
    @Test("should update value from props")
    public shouldUpdateValueFromProps() {

        const wrapper = shallow(<SimpleComponent value="initial" />);
        Expect(wrapper.find("input").prop("value")).toEqual("initial");
        wrapper.setProps({ value: "updated" });
        Expect(wrapper.find("input").prop("value")).toEqual("updated");
    }
}
```

However it fails as `shouldComponentUpdate` returns `false`.  Looking at the logs we see the following:

```
# FIXTURE SimpleComponent
render:  initial
updating state:  updated
state:  { value: 'updated' }
nextState:  { value: 'updated' }
props:  { value: 'initial' }
nextProps:  { value: 'updated' }
shouldUpdate:  false
not ok 1 should update value from props
 ---
   message: "Expected \"initial\" to be equal to \"updated\"."
   severity: fail
   data:
     got: "initial"
     expect: "updated"
 ...
```

As you can see from the above `this.state` appears to have already been updated by the time `shouldComponentUpdate` gets called resulting in it returning `false` and the test failing as it never re-renders.

If I run the same component in the browser I get the following output:

```
render:  initial
updating state:  updated
state:  {value: "initial"}
nextState:  {value: "updated"}
props:  {value: "initial"}
nextProps:  {value: "updated"}
shouldUpdate:  true
render:  updated
```

Which is what I would expect.

### Expected behavior

The state of the component should not be updated before calling shouldComponentUpdate.

### Your environment

> not sure whats needed here

#### API

- [x] shallow
- [ ] mount
- [ ] render

#### Version

| enzyme              |  ^3.8.0
| react               |  ^16.3.0
| react-dom           |  ^16.5.0
| react-test-renderer |  ~16.5.2
| adapter (below)     |  ^1.7.1

#### Adapter

- [x] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] others ( )

chenesan:
Prevent forgetting PR for #1970 😓  . I've added test case in this PR(should fail now).
Wait for release of facebook/react#14613 and we can update react-test-renderer in dependency.

Fixes #1970.

### DIFF
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -5,8 +5,10 @@ import ReactDOM from 'react-dom';
 import ReactDOMServer from 'react-dom/server';
 // eslint-disable-next-line import/no-unresolved
 import ShallowRenderer from 'react-test-renderer/shallow';
+import { version as testRendererVersion } from 'react-test-renderer/package';
 // eslint-disable-next-line import/no-unresolved
 import TestUtils from 'react-dom/test-utils';
+import semver from 'semver';
 import checkPropTypes from 'prop-types/checkPropTypes';
 import {
   isElement,
@@ -51,6 +53,8 @@ const is164 = !!TestUtils.Simulate.touchStart; // 16.4+
 const is165 = !!TestUtils.Simulate.auxClick; // 16.5+
 const is166 = is165 && !React.unstable_AsyncMode; // 16.6+
 const is168 = is166 && typeof TestUtils.act === 'function';
+
+const hasShouldComponentUpdateBug = semver.satisfies(testRendererVersion, '< 16.8');
 
 // Lazily populated if DOM is available.
 let FiberTags = null;
@@ -303,7 +307,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
           onSetState: true,
         },
         getDerivedStateFromProps: {
-          hasShouldComponentUpdateBug: !is168,
+          hasShouldComponentUpdateBug,
         },
         getSnapshotBeforeUpdate: true,
         setState: {
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -50,6 +50,7 @@ import detectFiberTags from './detectFiberTags';
 const is164 = !!TestUtils.Simulate.touchStart; // 16.4+
 const is165 = !!TestUtils.Simulate.auxClick; // 16.5+
 const is166 = is165 && !React.unstable_AsyncMode; // 16.6+
+const is168 = is166 && typeof TestUtils.act === 'function';
 
 // Lazily populated if DOM is available.
 let FiberTags = null;
@@ -280,7 +281,7 @@ function getEmptyStateValue() {
 }
 
 function wrapAct(fn) {
-  if (typeof TestUtils.act !== 'function') {
+  if (!is168) {
     return fn();
   }
   let returnVal;
@@ -301,7 +302,9 @@ class ReactSixteenAdapter extends EnzymeAdapter {
         componentDidUpdate: {
           onSetState: true,
         },
-        getDerivedStateFromProps: true,
+        getDerivedStateFromProps: {
+          hasShouldComponentUpdateBug: !is168,
+        },
         getSnapshotBeforeUpdate: true,
         setState: {
           skipsComponentDidUpdateOnNullish: true,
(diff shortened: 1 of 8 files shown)


chenesan:
@ljharb react 16.8 is out. I have tried manually upgrade `react-test-renderer` to 16.8 in `enzyme-react-adapter-16` and the tests passed :-) but I don't know how to handle the `react-test-renderer` dependency in enzyme.

ljharb:
@chenesan i'm not sure what you mean; it should be auto installed

ljharb:
The error showing up in the tests is specific to React 16.3. It seems like this might be fixable in the 16.3 adapter?

chenesan:
@ljharb I think it's not fixable in 16.3 because it's fixed in react-test-renderer@16.8. I'll add an `itIf('>= 16.8')` for the test.

chenesan:
@ljharb I saw that the build failed in React=16.0 and node=4 after my push. I looked into the log and it failed in `before_install` step. I'm certain that there's no dependency change in this PR so I'm wondering why that happened.

ljharb:
@chenesan i'd prefer if we could add whatever changes are necessary to fix it in 16.3, with react-test-renderer 16.3. We've done similar fixes before.

As for the correct version of react-test-renderer, that's something I might have to fix - testing should be done with react-test-renderer at the same minor version as react itself.

chenesan:
@ljharb Ok, I'll try to fix this in adapter 16.3. However, since the lifecycle logic is running in the react shallow renderer, I think it would be hard or not possible to fix this (maybe I'm wrong).

> We've done similar fixes before.

Could you guide me where the related issues / fixes are? Maybe I can get some ideas from this. Thanks :-)

ljharb:
Mainly, we can spy on component methods, so that we should be able to determine when individual lifecycle methods occur.

chenesan:
@ljharb I think this could be reviewed now :). I tried to wrap sCU and set instance state as correct old state so `this.state` won't be the state updated after gDSFP in shallow renderer (which should be set after sCU).

ljharb:
I've updated/rebased this; but it seems like https://github.com/facebook/react/commit/4f332885a2541e68c573d19ca94c875fdf988a38 doesn't actually fix the bug in react-test-renderer 16.8+ - once I made it so that the fix only applied to 16.3 - 16.7, the 16.8 tests failed locally :-/ altho everything seems to pass in CI, so I'm a bit confused.

chenesan:
Hmm interesting :( I'd try to take a look in this week.

ljharb:
k, figured it out.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -5,8 +5,10 @@ import ReactDOM from 'react-dom';
 import ReactDOMServer from 'react-dom/server';
 // eslint-disable-next-line import/no-unresolved
 import ShallowRenderer from 'react-test-renderer/shallow';
+import { version as testRendererVersion } from 'react-test-renderer/package';
```
benjycui: This is no `react-test-renderer/package` in the latest [`react-test-renderer`](https://github.com/facebook/react/tree/master/packages/react-test-renderer)

PR #1987, opened 2019-01-24
Adds to hasClass the ability to search using regex expression

stackchain:
When using a framework that injects JSS (decorators) like material-ui, it becomes complicated to deal with class searching. 
There are some conditions, especially when dealing with transitions, when adding and removing a react component affects the UX. Ex. there is no fade-out animation when the component just vanishes away from the user screen.

`{ show && <MyFadeOutComponent /> }`

So one way to tackle this problem is to add and remove classes to the component, instead of rendering off the react component, when removing the class the transitions will work fine.

`{ <MyFadeInAndFadeOutComponent  className="hide" />`

Testing this situation using decorators is tricky; You can export a version of the component before the injection as most people do in the storybook to show the correct propTypes, but sometimes it will require more work to be accomplished.

`export const UndecoredMyComponent = MyComponent;
export default withStyles(styles, { theme: true })(MyComponent);`

Another approach: you can mock the decorator.

Therefore would be great to add to hasClass function the ability to search using a regex, it makes those conditions above easier to test.

`
// Should pass if is hiding the component (using class)
expect(hasClassName(node, /(MyComponent)-(hide)-(\d+)/)).to.equal(true);`


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Change Log
 
+## 3.8.1
```
ljharb: Thanks; i prefer to write the changelog entries myself based on the commit titles and PR titles :-)

packages/enzyme-test-suite/package.json
```diff
@@ -55,4 +55,4 @@
     "eslint-plugin-react": "^7.12.4",
     "react-is": "^16.7.0"
   }
-}
+}
```
ljharb: let's revert this file

packages/enzyme-test-suite/test/RSTTraversal-spec.jsx
```diff
@@ -43,6 +43,20 @@ describe('RSTTraversal', () => {
       const node = $(<div className={classes} />);
       expect(hasClassName(node, 'foo-bar')).to.equal(true);
     });
+
+    it('works if searching with a RegExp', () => {
+      function classes() {}
+      classes.toString = () => 'ComponentName-classname-123';
+      const node = $(<div className={classes} />);
```
ljharb: i'm really confused why we need a stringifiable function instead of just a string literal

packages/enzyme/src/RSTTraversal.js
```diff
@@ -34,6 +34,7 @@ export function childrenOfNode(node) {
 export function hasClassName(node, className) {
   let classes = propsOfNode(node).className || '';
   classes = String(classes).replace(/\s/g, ' ');
+  if (className instanceof RegExp) return className.test(classes);
```
ljharb: `instanceof` isn't a reliable means of testing things for any builtin; please use the https://www.npmjs.com/package/is-regex package instead.

packages/enzyme-test-suite/test/RSTTraversal-spec.jsx
stackchain: Copy from a more complex example.

packages/enzyme/src/RSTTraversal.js
stackchain: Done

docs/api/ReactWrapper/hasClass.md
```diff
@@ -5,22 +5,28 @@ Returns whether or not the wrapped node has a `className` prop including the pas
 
 #### Arguments
 
-1. `className` (`String`): A single class name.
+1. `className` (`String` | `RegExp`): A single class name or a regex expression.
```
ljharb: Let's change the docs for ShallowWrapper too.

stackchain: Done!

PR #1995, opened 2019-01-31
Make sure findWhere doesn't get stuck on empty string

### LINKED ISSUE #1994: .findWhere failing to parse all nodes in wrapper
### Current behavior
I have a component that does conditional rendering in the form of: 
```jsx
<header>
  <h1>{title}</h1>
  {subtitle && <div>{subtitle}</div>}
</header>
```
_Note: subtitle is defaulted to an empty string, so that block is returning an empty string_

When that component is inside of a `ReactWrapper` and I use `.findWhere`, the function doesn't get past that conditional render.

### Expected behavior
I would expect `.findWhere` to be able to see that empty string and still navigate to the next node in the wrapper

Now, I'd say this could be considered user error, since we can write that condition differently and get the test to pass. However, it also seems strange to me that `findWhere` would get stuck on this. If this isn't something you consider an issue just let me know 👍 Otherwise I can take a crack at fixing it if y'all are cool with that.

### Your environment
You can see this replicated in [this codesandbox](https://codesandbox.io/s/jp4vnqy2n3?expanddevtools=1&module=%2Fsrc%2FHello.js&previewwindow=tests)

#### API

- [ ] shallow
- [x] mount
- [ ] render

#### Version

| enzyme              | 3.8.0
| react               | 16.7.0
| react-dom           | 16.7.0
| adapter (below)     | 1.8.0

#### Adapter

- [x] enzyme-adapter-react-16

#### Related issues
I saw these two issues that seemed like they might be relevant, but after looking into it I don't think they are quite related:
#1825 
#1993

MikaelCarpenter:
# What
- Add a `findWhere` test case that replicates the use case I ran into
- Update `RSTTraversal.treeForEach` to not call `fn` on empty strings
- Update `treeForEach` test case that references empty strings

# Why
In the event that the DOM has an empty string, make sure `findWhere` can iterate past it and still find elements that are further down the tree.

## Links
- Fixes #1994
- [Codesandbox](https://codesandbox.io/s/jp4vnqy2n3?expanddevtools=1&module=%2Fsrc%2FHello.js&previewwindow=tests) that replicates the initial issue I had

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -1721,6 +1721,39 @@ describeWithDOM('mount', () => {
       expect(foundNotSpan).to.have.lengthOf(0);
     });
 
+    it('does not get trapped when conditionally rendering using an empty string variable as the condition', () => {
+      const emptyString = '';
+
+      class Foo extends React.Component {
+        render() {
+          return (
+            <div>
+              <header>
+                <span />
+                {emptyString && <i />}
+              </header>
+              <div>
+                <span data-foo={this.props.selector}>Test</span>
+              </div>
+            </div>
+          );
+        }
+      }
+
+      const selector = 'blah';
+      const wrapper = mount(<Foo selector={selector} />);
+      const foundSpan = wrapper.findWhere(n => (
+        n.type() === 'span'
+        && n.props()['data-foo'] === selector
+      ));
+
+      expect(foundSpan.debug()).to.equal((
+        `<span data-foo="${selector}">
+  Test
+</span>`
+      ));
+    });
+
     it('returns props object when props() is called', () => {
       class Foo extends React.Component {
         render() {
--- a/packages/enzyme/src/RSTTraversal.js
+++ b/packages/enzyme/src/RSTTraversal.js
@@ -40,7 +40,7 @@ export function hasClassName(node, className) {
 }
 
 export function treeForEach(tree, fn) {
-  if (tree !== null && tree !== false && typeof tree !== 'undefined') {
+  if (tree) {
     fn(tree);
   }
   childrenOfNode(tree).forEach(node => treeForEach(node, fn));
--- a/packages/enzyme-test-suite/test/RSTTraversal-spec.jsx
+++ b/packages/enzyme-test-suite/test/RSTTraversal-spec.jsx
@@ -350,10 +350,11 @@ describe('RSTTraversal', () => {
       const node = $((
         <div>
           <p>{''}</p>
+          <p>Test</p>
         </div>
       ));
       treeForEach(node, spy);
-      expect(spy).to.have.property('callCount', 3);
+      expect(spy).to.have.property('callCount', 4); // div, p, p, 'Test'
     });
 
     it('passes in the node', () => {
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -1584,7 +1584,7 @@ describe('shallow', () => {
       const stub = sinon.stub();
       wrapper.findWhere(stub);
       const passedNodeLengths = stub.getCalls().map(({ args: [firstArg] }) => firstArg.length);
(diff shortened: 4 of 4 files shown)


packages/enzyme/src/RSTTraversal.js
```diff
@@ -40,7 +40,7 @@ export function hasClassName(node, className) {
 }
 
 export function treeForEach(tree, fn) {
-  if (tree !== null && tree !== false && typeof tree !== 'undefined') {
+  if (tree !== null && tree !== false && typeof tree !== 'undefined' && tree !== '') {
```
MikaelCarpenter: would it make sense to just use `if (tree)` here, rather than blacklisting specific values?

sstern6: Thanks for reporting and contributing.  I think this is a good start.  Though as newer versions of react come out, more and more valid "empty" values become natively supported by react.  Ie emptyString, emptyArray, emptyFragments, undefined.  Adding a fix for just an empty string seems like a good start and solves your problem but i think we need to consider other use cases including the ones i mentioned above. 

@ljharb thoughts on this?

ljharb: i mean, realistically here, this is deciding when to bypass calling `childrenOfNode`.

A node can only have children if it's an object, right? Could this change be effectively `if (isPrimitive(tree)) {`? (with `function isPrimitive(val) { return !val || Object(val) !== val; }`)

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -1706,6 +1706,33 @@ describeWithDOM('mount', () => {
...
+          );
+        }
+      }
+
+      const selector = 'blah';
+      const wrapper = mount(<Foo selector={selector} />);
+      const foundSpan = wrapper.findWhere(n => (
+        n.type() === 'span' && n.props()['data-foo'] === selector
+      ));
+      expect(foundSpan.type()).to.equal('span');
```
ljharb: it might be more robust here to instead make an assertion on `foundSpan.debug()`?

packages/enzyme/src/RSTTraversal.js
MikaelCarpenter: So I tried this out and it looks like calling `fn(tree)` on things like strings and numbers is a desired functionality. Here are some of the tests that fail (trimmed out some of the output):
```bash
1) RSTTraversal treeForEach does not get trapped from empty strings:

2) (uses jsdom) mount .findWhere(predicate) does not pass in null or false nodes:

    AssertionError: expected [ Array(3) ] to deeply equal [ Array(4) ]
    + expected - actual

        [
          "<div>\n  foo bar\n</div>"
          true
        ]
    +  [
    +    "foo bar"
    +    [null]
    +  ]
      ]
    
    at Context.<anonymous> (packages/enzyme-test-suite/test/ReactWrapper-spec.jsx:1880:30)

3) (uses jsdom) mount .findWhere(predicate) allows `.text()` to be called on text nodes:

4) shallow .contains(node) works with strings:

5) shallow .contains(node) works with numbers:

6) shallow .contains(node) works with nested strings & numbers:

7) shallow .findWhere(predicate) allows `.text()` to be called on text nodes:

    AssertionError: expected [ Array(3) ] to deeply equal [ Array(4) ]
    + expected - actual

        [
          "<div>\n  foo bar\n</div>"
          "foo bar"
        ]
    +  [
    +    "foo bar"
    +    "foo bar"
    +  ]
      ]
    
    at Context.<anonymous> (packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx:1829:31)

8) selectors (uses jsdom) mount :empty pseudo selector:

9) selectors shallow :empty pseudo selector:

    AssertionError: expected {} to have a length of 0 but got 1
    + expected - actual

    -1
    +0
    
    at Context.<anonymous> (packages/enzyme-test-suite/test/selector-spec.jsx:371:52)
```

MikaelCarpenter: So it looks like we only want to prevent falsey values. I tried out `if (tree) { fn(tree) }` and the tests will still pass.

ljharb: ah, ok - let's do that then :-)

PR #2000, opened 2019-02-04
Fix a typo

IvanDimanov:


docs/api/ShallowWrapper/render.md
```diff
@@ -28,7 +28,7 @@ function Bar() {
 ```
 
 ```jsx
-const wrapper = shallow(<Bar />);
```
ljharb: This isn't a typo; this is a ShallowWrapper doc, using the `.render` method - not the top-level `render` API.

IvanDimanov: Excuse my mistake :)

PR #2007, opened 2019-02-06
1862/fix setprops on cdu

### LINKED ISSUE #1862: Calling `setProps` on wrapperresults in `componentDidUpdate` being called twice
**Describe the bug**
Calling `setProps` results in `componentDidUpdate` being called twice (instead of once).

**To Reproduce**
Run the test:

```
import React from 'react';


export default class Dummy extends React.Component {
    componentWillReceiveProps(nextProps) {
        this.setState({someState: nextProps.myProp});
    }

    componentDidUpdate() {
        this.foo();
    }

    foo() {}

    render() {
        return (
            <div>
                <div>{`myProp: ${this.props.myProp}`}</div>
            </div>
        );
    }
}
```

```
import React from 'react';
import {shallow} from 'enzyme';

import Dummy from 'questions/components/dummy.js';


describe('<Dummy/>', () => {
    it('works', () => {
        const props = {};
        const wrapper = shallow(<Dummy {...props} />);
        wrapper.instance().foo = jest.fn();
        wrapper.setProps({myProp: 'Prop Value'});
        expect(wrapper.instance().foo).toHaveBeenCalledTimes(1);
    });
});
```

**Expected behavior**
`componentDidUpdate` should only be called once.

 - OS: macOS Mojave
 - Browser n/a
 - Version 3.7.0

**Smartphone (please complete the following information):**
n/a

peanutenthusiast:
Fixes #1862


peanutenthusiast:
No problem!!

willdurand:
We noticed a regression in our test suite (between Enzyme 3.8 and 3.9), likely due to this patch. Downgrading to Enzyme 3.8 makes our test suite pass again.

We have some code that changes a component's `state` in `componentWillReceiveProps()` and, depending on this new `state`, a function is called in `componentDidUpdate()`. This is not working anymore and I believe this patch is the "culprit".

This PR shows the regression as well as the revert to Enzyme 3.8 to get a green CI again: https://github.com/mozilla/addons-frontend/pull/7603.

I am not sure what you or I could do though. Moving away from CWRP is on our roadmap and might "fix" this incompatibility, but upgrading from 3.8 to 3.9 should not break our test suite. I don't know what's the most "correct" behavior between no batch updates, the <=3.8 behavior and the new one introduced by this patch 🤷‍♂️

ljharb:
@willdurand thanks for the report, could you file a new issue and fill out the template?

PR #2010, opened 2019-02-07
Updating setState Documentation

EstebanMarin:


docs/api/ShallowWrapper/setState.md
```diff
@@ -7,8 +7,7 @@ your component's external API (which is often accessible via [`.instance()`](ins
 to get it into whatever state you want to test, in order to be as accurate of a test as possible.
 This is not always practical, however.
 
-NOTE: can only be called on a wrapper instance that is also the root instance.
-
+As for the version(3.^8) of Enzyme you can now setState() to non-root elements
```
ljharb: this is the case for ReactWrapper, but not for ShallowWrapper. Can you update that one instead?

Also, let's use the text "NOTE: prior to v3.8.0 of enzyme, can only be called on a wrapper instance that is also the root instance."

EstebanMarin: I made the change in the correct file and left `setState` in `ShallowWrapper` untouched.

Issue #2011, opened 2019-02-08
React hooks support checklist


chenesan:
Sorry for removing issue template. Since the React 16.8 is out we can start to work on hooks support in enzyme now. Previously I tried to open PR in #2008 to help with this but found out that there are [too many apis in React hooks](https://reactjs.org/docs/hooks-reference.html). After some thinking I feel it's better to track all of the supports / PR in one issues, and create one PR for each hooks api (`useState`, `useEffect`, etc).

- [ ] useState (#1938, #1996, PR #2008)
- [ ] useEffect
- [ ] useContext
- [ ] useReducer
- [ ] useCallback
- [ ] useMemo
- [ ] useRef (#1938)
- [ ] useImperativeHandle
- [ ] useLayoutEffect
- [ ] useDebugValue
- [ ] custom hooks (#1996)

Currently I'm working in `useState` (#2008) and I'm glad to see if anyone else would like to help to support other apis. For now we need to add many tests to make sure what is working in enzyme, and what is not, then add patch / feature into it. I can help to maintain the above checklist if there are any related issues / PR . 

I'm simply volunteer to do this and not a main maintainer here though so @ljharb if you have had any plan / schedule on this feel free to tell me and I can change or close this :-)

k3ithl1m:
Thanks for making this list @chenesan!
@ljharb, I'm interested in tackling some of this features too. It might take me some time to finish implementing it as I'm getting a little busy in the next two weeks, but I am very interested in helping out. 
I've currently worked on two test cases #2029

If no one has took up the useEffect feature. I would like to try that out!

AnaRobynn:
@chenesan Also note that React seemed to have an issue with shallow rendering hooks right now. Not sure if it was addressed by the React time already or not.

nixstrom:
@AnaRobynn Didn't they solve this with https://github.com/facebook/react/pull/14567 ?

AnaRobynn:
It’s supported, but there are issues: https://twitter.com/dan_abramov/status/1103304963445403648?s=21

chenesan:
I think the issue should be facebook/#14841, which #2008 would depends on.

pgangwani:
@chenesan  @k3ithl1m : Added test cases #2041 for useEffect, useState & custom hooks along with combination of those.

pgangwani:
@chenesan : Please add #2041 in custom hooks checklist as I added test case in that PR

pgangwani:
[...]

PR #2008, opened 2019-02-07
[Tests] add hooks tests

chenesan:
This PR adds support for [basic hooks API] (https://reactjs.org/docs/hooks-reference.html#basic-hooks) (`useState`, `useEffect`, `useContext`, related issue #1938 #1996). Currently I'm not sure what hooks API is already working in enzyme shallow / mount and what isn't, so I'll start from adding test cases to confirm the current supports. I'll try to finish the following checklist in the next few days. Also tell me if it's better to have single PR for each api :-)

- [ ] useState
  - [x] mount
    - [x] get value from hooks
    - [x] set value from hooks and rendered element should be updated
  - [ ] shallow
    - [x] get value from hooks
    - [x] set value from hooks and rendered element should be updated (that's a bug in react shallow renderer, fixed in react after react@16.8.5, see https://github.com/facebook/react/pull/15120)

I've added some test cases for `useState` and found it doesn't work in `shallow` to setState.

(Updated: Created a checklist in #2011 to try to track any issue / PR in hooks, and concentrate on support `useState` in this PR)

Issue #1996, opened 2019-02-01
Testing Hooks with mount: onChange not called


ekaradon:
### Current behavior

When simulating some changes on input with components using hooks, depending of the component, it looks like the `onChange` is not called.

### Expected behavior

Whatever the component, when simulating changes on input, the `onChange` method must be called.

#### API
 mount
#### Version

| library             | version
| ------------------- | -------
| enzyme              | 3.4.1
| react               | 16.8.0-alpha.1
| react-dom           | 16.8.0-alpha.1
| react-test-renderer | 15.6.2
| adapter (below)     | 1.3.0

#### Adapter

enzyme-adapter-react-16

```
import React, { useState, useEffect } from 'react';
import sinon from 'sinon';
import Adapter from 'enzyme-adapter-react-16';
import { mount, configure } from 'enzyme';

configure({ adapter: new Adapter() });

function useFormInput(initial_value = '') {
    const [value, setValue] = useState(initial_value);

    function handle_change(e) {
        setValue(e.target.value);
    }

    return {
        value
        , onChange: handle_change
    };
}

function Input(props) {
    return (
        <div>
            <input {...props} />
        </div>
    );
}

function ControlledInputWithEnhancedInput({ searchSomething }) {
    const search = useFormInput();

    useEffect(
        () => { searchSomething(search.value); }
        , [search.value]
    );

    return (
        <Input {...search} />
    );
}

function ControlledInputWithNativeInput({ searchSomething }) {
    const search = useFormInput();

    useEffect(
        () => { searchSomething(search.value); }
        , [search.value]
    );

    return (
        <input {...search} />
    );
}

it('does not succeed', () => {
    jest.useFakeTimers();

    const spy = sinon.spy();
    const component = mount(<ControlledInputWithEnhancedInput searchSomething={spy} />);

    component.simulate('change', { target: { value: 'foo' } });
    jest.runAllTimers();

    expect(spy.withArgs('foo').calledOnce).toBe(true);
});

it('does succeed', () => {
    jest.useFakeTimers();

    const spy = sinon.spy();
    const component = mount(<ControlledInputWithNativeInput searchSomething={spy} />);

    component.simulate('change', { target: { value: 'foo' } });
    jest.runAllTimers();

    expect(spy.withArgs('foo').calledOnce).toBe(true);
});

```

The single fact to wrap the input with a `div` is hurting the test. Really Weird. Is there anyone having an idea about this?

ljharb:
[...]

Issue #1683, opened 2018-06-19
Stateless functional components have "this" defined


halkeye:
**Describe the bug**
Looks like stateless functional components work differently in browser vs enzyme. 
I'm not sure if this is a enzyme, react, react-test-utils or babel issue.
 
**To Reproduce**

Run this jest test
```
const React = require('react');
const Enzyme = require('enzyme');
const Adapter = require('enzyme-adapter-react-16');
const { shallow } = require('enzyme');

// React 16 Enzyme adapter
Enzyme.configure({ adapter: new Adapter() });

const ArrowComponent = () => {
  console.log(this.props);
  return <div />;
};

// shouldn't work either
function FuncComponent () {
  console.log(this.props);
  return <div />;
}

test('FuncComponent', () => {
  shallow(<FuncComponent abc="123" />);
});

test('ArrowComponent', () => {
  shallow(<ArrowComponent abc="123" />);
});
```

**Expected behavior**
Both tests should fail (not sure how to properly write that, maybe expect throws?) but only the arrow function has this as undefined.

Module Versions: 
```
├── enzyme@3.3.0
└── enzyme-adapter-react-16@1.1.1
└── react@16.4.1
```
Node Verison: v8.11.3

LSB Version:    :core-4.1-amd64:core-4.1-noarch
Distributor ID: CentOS
Description:    CentOS Linux release 7.5.1804 (Core)
Release:        7.5.1804
Codename:       Core

halkeye:
It might just be "This is how javascript works" with functions and this, but it was a shock to me because we refactored a few classes into stateless functions and had one break because we missed the this.props but the tests were still passing. 

Functional components shouldn't have this.props defined, so i think its enzyme's shallow stuff

ljharb:
Hmm, that does seem like an issue - `this` should *not* work in SFCs.

ljharb:
This is a bug in `react-test-renderer`; I've filed it there. https://github.com/facebook/react/issues/13141

ljharb:
ok, confirmed it's only a bug in shallow, and only in react 16.

PR #2014, opened 2019-02-08
Don't wrap functional components in react >=16.5 to shallow render

rodrigopr:
This is necessary because ReactShallowRender checks for equality in function components to decide if it should keep hooks state between renders.

This along with https://github.com/facebook/react/pull/14802 / https://github.com/facebook/react/issues/14840 should help in test components with some of the core hooks.
link #2011 , #2008 and #1996

### DIFF
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -405,6 +405,27 @@ class ReactSixteenAdapter extends EnzymeAdapter {
     const renderer = new ShallowRenderer();
     let isDOM = false;
     let cachedNode = null;
+
+    let lastComponent = null;
+    let wrappedComponent = null;
+
+    // Wrap functional components on versions prior to 16.5,
+    // to avoid inadvertently pass a `this` instance to it.
+    const wrapFunctionalComponent = (Component) => {
+      if (is165) {
+        return Component;
+      }
+
+      if (lastComponent !== Component) {
+        wrappedComponent = Object.assign(
+          (...args) => Component(...args), // eslint-disable-line new-cap
+          Component,
+        );
+        lastComponent = Component;
+      }
+      return wrappedComponent;
+    };
+
     return {
       render(el, unmaskedContext) {
         cachedNode = el;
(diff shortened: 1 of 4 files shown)


rodrigopr:
@ljharb sure, I guess I'll wait for the react PR to be merged, to be able to do a more meaningful test for it.

gaearon:
I don't understand what this fix is supposed to do or why it makes sense. Could you please explain it?

gaearon:
Huh I see. I didn't realize Enzyme wraps the passed type... In that case this fix makes sense.
(Why does Enzyme do this?)

rodrigopr:
Seems like It was introduced to solve #1683, that is linked to issue https://github.com/facebook/react/issues/13141 fixed since react 16.5.0.
On version 16.8.5 the tests passes without any wrapper.
What do you think @ljharb, @gaearon?

For now, I've updated the PR with tests

gaearon:
I'd remove the wrapper if we're targeting 16.5+.

ljharb:
That’s definitely something we can and should do - and it can be done in the adapter. We have all sorts of version-conditional hacks and bugfixes patches in the adapters.

rodrigopr:
@ljharb I've updated the PR approach, to only wrap the component fn on version < 16.5

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -426,22 +437,17 @@ class ReactSixteenAdapter extends EnzymeAdapter {
...
+            ));
           }
 
           if (!isStateful && typeof Component === 'function') {
-            if (Component !== cachedComponent) {
-              wrappedEl = Object.assign(
-                (...args) => Component(...args), // eslint-disable-line new-cap
-                Component,
-              );
-              cachedComponent = Component;
```
ljharb: this line is important here; we don't want to create a new wrapper every time. Can we preserve it?

rodrigopr: PR updated with the requested change. I can also simplify the wrap-fuction using lodash.memoize, if you think is worth adding the dep.

Also, a test started failing, I've tracked to this change:
https://github.com/airbnb/enzyme/pull/2014/commits/abed5df0d69afa9eee66cd8ac719f4a213b9b58c (it started failing because it no longer tries to wrap the component)
I've reverted it since React.Memo also work on class component, but let me know if there is a better approach.

ljharb: Nah, no need to add the dep.

Your change seems good :-)

ljharb:
k, i've rebased this on top of a major tests refactor; i'm not sure if your tests are still passing.

rodrigopr:
Thanks, seems the test still pass after the rebase

PR #2015, opened 2019-02-10
docs: update Jest to setupFilesAfterEnv

AndersDJohnson:
Per https://jestjs.io/docs/en/configuration#setupfilesafterenv-array, `setupTestFrameworkScriptFile` has been renamed to `setupFilesAfterEnv`, so updating these docs.

PR #2024, opened 2019-02-25
Update deprecated jest config key in docs

VGoose:
Per[ jest docs](https://jestjs.io/docs/en/configuration#setupfilesafterenv-array), `setupTestFrameworkScriptFile` has been renamed to `setupFilesAfterEnv`


ljharb:
#2013 and #2015 also made the same change; can you please confirm there's nowhere else in the repo that this change is needed?

VGoose:
I've found 3 more instances after checking every page of the doc excluding the API Reference.  I assumed a jest configuration wouldn't be in those pages.

Issue #2020, opened 2019-02-22
Regression on 'componentDidUpdate' not being called when calling 'setProps' on shallow wrapped component


BertrandFritsch:
Calling `setProps` on a shallow wrapper may result in not calling `componentDidUpdate` of the wrapped component. It's the case when the state depends on a prop and when that specific `setProps` call does change the state. If that call doesn't change the state, `componentDidUpdate` is called. I suspect this is a regression of #2007. It worked fine in version 3.8.0

### To reproduce

```javascript
import { shallow } from 'enzyme';
import * as React from 'react';

class DummyComp extends React.PureComponent {
  constructor(props) {
    super(props);
    this.state = { stateCounter: 0 };
  }

  static getDerivedStateFromProps(nextProps) {
    // set stateCounter if asked for
    if (nextProps.propCounter > -1) {
      return {
        stateCounter: nextProps.propCounter
      };
    }

    return null;
  }

  componentDidUpdate() {
    console.log('Counter:', this.state.stateCounter);
  }

  render() {
    return <p>Counter: { this.state.stateCounter }</p>;
  }
}

describe('Counter', () => {
  it('should call cDU on setProps', () => {

    const wrapper = shallow(<DummyComp propCounter={ 0 } />);

    const cDUSpy = jest.spyOn(wrapper.instance(), 'componentDidUpdate');

    // does not change the state, so cDU will be called
    wrapper.setProps({ propCounter: -1 });

    expect(cDUSpy).toBeCalledTimes(1);

    // does change the state, cDU should be called anyway
    wrapper.setProps({ propCounter: 1 });

    expect(cDUSpy).toBeCalledTimes(2); // this test will fail
  });
});
```

#### API

- [x] shallow
- [ ] mount
- [ ] render

#### Version

| library             | version
| ------------------- | -------
| enzyme              | 3.9.0
| react               | 16.8.2
| react-dom           | 16.8.2
| react-test-renderer | 16.8.2
| adapter (below)     | 

#### Adapter

- [x] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-16.3
- [ ] enzyme-adapter-react-16.2
- [ ] enzyme-adapter-react-16.1
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-15.4
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] enzyme-adapter-react-helper
- [ ] others ( )

[...]

PR #2027, opened 2019-02-27
[TEST] Regression on CDU on setProps on shallow wrapped component

peanutenthusiast:
For #2020.

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -2502,8 +2502,8 @@ describeWithDOM('mount', () => {
       ]);
     });
 
-    describe('setProps should not call componentDidUpdate twice', () => {
-      it('first test case', () => {
+    describe('setProps does not call componentDidUpdate twice', () => {
+      it('when setState is called in cWRP', () => {
         class Dummy extends React.Component {
           constructor(...args) {
             super(...args);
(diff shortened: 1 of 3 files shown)


ljharb:
Thanks! I've updated the tests and split out the "no state change" and "state change" cases, and added some more assertions.

ljharb:
I suspect this is a bug in react's shallow renderer - I've filed https://github.com/facebook/react/issues/15075 to investigate.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -6854,6 +6854,123 @@ describeWithDOM('mount', () => {
...
+
+      describe('setProps calls `componentDidUpdate` when `getDerivedStateFromProps` is defined', () => {
+        class DummyComp extends PureComponent {
+          constructor(...args) {
+            super(...args);
+            this.state = { state: -1 };
+          }
+
+          static getDerivedStateFromProps({ changeState, counter }, prevState) {
+            return changeState ? { state: counter * 10 } : prevState;
```
miraage: Shouldn't we return `null` when state does not need to be updated? https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

miraage: > `getDerivedStateFromProps` is invoked right before calling the render method, both on the initial mount and on subsequent updates. It should return an object to update the state, or **null** to update nothing

ljharb: Yes, that's true. However, returning the previous state should result in state not being updated either. Updating to null still fails the same.

ljharb:
I rebased this on top of #1981, which seems related, but one of the shallow tests still fails.

peanutenthusiast:
Added `describeIfs` to ensure the tests only run for `16.3`and up

ljharb:
Good catch :-) we still need to make the shallow test pass, though. Got any ideas for a fix?

peanutenthusiast:
The new commit satisfies the failing test, but are there any other conditions/test cases we should cover?

ljharb:
Probably :-) but if all the tests are passing, and it looks reasonable, let's merge it!

peanutenthusiast:
Alrighty 👍

dallonf:
What would it take for this fix to get released to NPM? I think I just got bitten by this bug when updating dependencies

ljharb:
I'm waiting on a few more PRs to land before publishing the next semver-minor of enzyme itself.

bdwain:
@ljharb are you still waiting for more versions? It would be nice to get bugfixes out ASAP since the latest release is presumably broken. I'd be happy to do a branch of 3.9.0 with just this if it would get released as 3.9.1

ljharb:
I might be able to pull together a patch tomorrow; but there’s still a minor I’m hoping to land first.

jdk243:
@ljharb any estimate of when this patch is going to make it to NPM? getting some test failures related to this issue after updating to 3.9.0

joshmanders:
Is there any update on this? This is breaking our tests.

cedosw:
Same here ... tests are failing because of this :(

bdwain:
@ljharb seems like releases of things are going out, but not the main enzyme package itself. Any reason why there can't be a release? The latest version has been broken for 2 months while master has been fixed. Seems like it shouldn't be terrible to release a patch version of whatever is in master.

sulemanof:
Really need this fix to be released 😢

joshmanders:
Doesn't seem like a priority to airbnb, best to probably move on to [React Testing Library](https://testing-library.com/docs/react-testing-library/intro)

ljharb:
@joshmanders it is very much a priority, but thanks for your quite hostile comment. If you make another one, I'll need to block you from the airbnb organization on github.

ljharb:
@bdwain every main enzyme release potentially creates a flurry of new issues, so it is in fact quite terrible to release it before it's ready. I'm still planning to get a few fixes in before releasing v3.10.

paulkmiller:
> @joshmanders it is very much a priority, but thanks for your quite hostile comment. If you make another one, I'll need to block you from the airbnb organization on github.

Hostile? Wow. You need to chill.

zackify:
This is quite unfair to threaten banning  someone for stating an alternative, after waiting over  a month for a release...

ljharb:
@paulkmiller @zackify you may be unaware of this person's history making similar comments that warrants this response, but regardless, the hostility wasn't for posting an alternative, it was for the first part.

joshmanders:
@ljharb both of these people know me very well. I'd like for you to point out where I've made similar comments.

ljharb:

PR #2034, opened 2019-03-06
Wrap mount renderer render with reactTestUtils.act()

chenesan:
This PR wrap the render function of mount renderer with `act()` so that use cases with `useEffect` would work. The test is the test case in https://github.com/threepointone/react-act-examples/blob/master/README.md#effects .

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -31,6 +31,8 @@ import {
   forwardRef,
   memo,
   PureComponent,
+  useEffect,
+  useState,
 } from './_helpers/react-compat';
 import {
   describeWithDOM,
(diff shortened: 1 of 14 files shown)


packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -1660,6 +1662,24 @@ describeWithDOM('mount', () => {
         });
       });
     });
+    describeIf(is('>= 16.8'), 'hooks', () => {
+      it('handle useEffect', () => {
+        const ComponentUsingEffectHook = () => {
```
```suggestion
        function ComponentUsingEffectHook() {
```

pgangwani:
@chenesan : Glad to see this PR merged.

pgangwani:
@chenesan @ljharb Can we implement same changes for ShallowRenderer too considering facebook/react#15120 ? Please correct me if my understanding is wrong.

ljharb:
@pgangwani I had to revert the act-wrapping in shallow because it requires a DOM, which shallow must not.

pgangwani:
@ljharb Got it. Main intension was to know if there is impact after facebook/react#15120 is merged.

PR #2036, opened 2019-03-06
Support getDerivedStateFromError

barmac:
This PR adds support for `static getDerivedStateFromError`.

### DIFF
--- a/docs/api/ReactWrapper/simulateError.md
+++ b/docs/api/ReactWrapper/simulateError.md
@@ -2,7 +2,7 @@
 
 Simulate a component throwing an error as part of its rendering lifecycle.
 
-This is particularly useful in combination with React 16 error boundaries (ie, the `componentDidCatch` lifecycle method).
+This is particularly useful in combination with React 16 error boundaries (ie, the `componentDidCatch` and `static getDerivedStateFromError` lifecycle methods).
 
 
 #### Arguments
@@ -26,6 +26,17 @@ function Something() {
 }
 
 class ErrorBoundary extends React.Component {
+  static getDerivedStateFromError(error) {
+    return {
+      hasError: true,
+    };
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { hasError: false };
+  }
+
   componentDidCatch(error, info) {
     const { spy } = this.props;
     spy(error, info);
@@ -33,9 +44,10 @@ class ErrorBoundary extends React.Component {
 
   render() {
     const { children } = this.props;
+    const { hasError } = this.state;
     return (
       <React.Fragment>
-        {children}
+        {hasError ? 'Error' : children}
       </React.Fragment>
     );
   }
@@ -50,6 +62,7 @@ const wrapper = mount(<ErrorBoundary spy={spy}><Something /></ErrorBoundary>);
 const error = new Error('hi!');
 wrapper.find(Something).simulateError(error);
 
+expect(wrapper.state()).to.have.property('hasError', true);
 expect(spy).to.have.property('callCount', 1);
 expect(spy.args).to.deep.equal([
   error,
--- a/docs/api/ShallowWrapper/simulateError.md
+++ b/docs/api/ShallowWrapper/simulateError.md
@@ -2,7 +2,7 @@
 
 Simulate a component throwing an error as part of its rendering lifecycle.
 
-This is particularly useful in combination with React 16 error boundaries (ie, the `componentDidCatch` lifecycle method).
+This is particularly useful in combination with React 16 error boundaries (ie, the `componentDidCatch` and `static getDerivedStateFromError` lifecycle methods).
 
 
 #### Arguments
@@ -26,6 +26,17 @@ function Something() {
 }
 
 class ErrorBoundary extends React.Component {
+  static getDerivedStateFromError(error) {
+    return {
+      hasError: true,
+    };
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = { hasError: false };
+  }
+
   componentDidCatch(error, info) {
     const { spy } = this.props;
     spy(error, info);
@@ -33,9 +44,10 @@ class ErrorBoundary extends React.Component {
 
   render() {
     const { children } = this.props;
+    const { hasError } = this.state;
     return (
       <React.Fragment>
-        {children}
+        {hasError ? 'Error' : children}
       </React.Fragment>
     );
   }
@@ -50,6 +62,7 @@ const wrapper = shallow(<ErrorBoundary spy={spy}><Something /></ErrorBoundary>);
 const error = new Error('hi!');
 wrapper.find(Something).simulateError(error);
 
+expect(wrapper.state()).to.have.property('hasError', true);
 expect(spy).to.have.property('callCount', 1);
 expect(spy.args).to.deep.equal([
   error,
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -316,6 +316,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
         getChildContext: {
           calledByRenderer: false,
         },
+        getDerivedStateFromError: is166,
       },
     };
   }
@@ -362,8 +363,17 @@ class ReactSixteenAdapter extends EnzymeAdapter {
         return instance ? toTree(instance._reactInternalFiber).rendered : null;
       },
       simulateError(nodeHierarchy, rootNode, error) {
-        const { instance: catchingInstance } = nodeHierarchy
-          .find(x => x.instance && x.instance.componentDidCatch) || {};
+        const isErrorBoundary = ({ instance: elInstance, type }) => {
+          if (is166 && type && type.getDerivedStateFromError) {
+            return true;
+          }
+          return elInstance && elInstance.componentDidCatch;
+        };
+
+        const {
+          instance: catchingInstance,
+          type: catchingType,
+        } = nodeHierarchy.find(isErrorBoundary) || {};
 
         simulateError(
           error,
@@ -372,6 +382,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
           nodeHierarchy,
           nodeTypeFromType,
           adapter.displayNameOfNode,
+          is166 ? catchingType : undefined,
         );
       },
       simulateEvent(node, event, mock) {
@@ -482,6 +493,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
           nodeHierarchy.concat(cachedNode),
           nodeTypeFromType,
           adapter.displayNameOfNode,
+          is166 ? cachedNode.type : undefined,
         );
       },
       simulateEvent(node, event, ...args) {
--- a/packages/enzyme-adapter-utils/src/Utils.js
+++ b/packages/enzyme-adapter-utils/src/Utils.js
@@ -266,19 +266,27 @@ export function simulateError(
   hierarchy,
   getNodeType = nodeTypeFromType,
   getDisplayName = displayNameOfNode,
+  catchingType = {},
 ) {
-  const { componentDidCatch } = catchingInstance || {};
-  if (!componentDidCatch) {
+  const instance = catchingInstance || {};
+
+  const { componentDidCatch } = instance;
+
+  const { getDerivedStateFromError } = catchingType;
+
+  if (!componentDidCatch && !getDerivedStateFromError) {
     throw error;
   }
 
-  const componentStack = getComponentStack(
-    hierarchy,
-    getNodeType,
-    getDisplayName,
-  );
+  if (getDerivedStateFromError) {
+    const stateUpdate = getDerivedStateFromError.call(catchingType, error);
+    instance.setState(stateUpdate);
+  }
 
-  componentDidCatch.call(catchingInstance, error, { componentStack });
+  if (componentDidCatch) {
+    const componentStack = getComponentStack(hierarchy, getNodeType, getDisplayName);
+    componentDidCatch.call(instance, error, { componentStack });
+  }
 }
 
 export function getMaskedContext(contextTypes, unmaskedContext) {
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -5946,14 +5946,18 @@ describeWithDOM('mount', () => {
           }
 
           render() {
-            const { throws } = this.state;
+            const {
+              didThrow,
+              throws,
+            } = this.state;
+
             return (
               <div>
                 <MaybeFragment>
                   <span>
                     <Thrower throws={throws} />
                     <div>
-                      {this.state.didThrow ? 'HasThrown' : 'HasNotThrown'}
+                      {didThrow ? 'HasThrown' : 'HasNotThrown'}
                     </div>
                   </span>
                 </MaybeFragment>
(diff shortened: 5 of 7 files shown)


docs/api/ReactWrapper/simulateError.md
```diff
@@ -35,7 +46,7 @@ class ErrorBoundary extends React.Component {
     const { children } = this.props;
     return (
       <React.Fragment>
-        {children}
+        { this.state.hasError ? "Error" : children }
```
ljharb:
```suggestion
        {hasError ? 'Error' : children}
```

docs/api/ReactWrapper/simulateError.md
```diff
@@ -35,7 +46,7 @@ class ErrorBoundary extends React.Component {
     const { children } = this.props;
```
ljharb:
```suggestion
    const { children } = this.props;
    const { hasError } = this.state;
```

docs/api/ReactWrapper/simulateError.md
```diff
@@ -50,6 +61,7 @@ const wrapper = mount(<ErrorBoundary spy={spy}><Something /></ErrorBoundary>);
 const error = new Error('hi!');
 wrapper.find(Something).simulateError(error);
 
+expect(wrapper.state().hasError).to.equal(true);
```
ljharb:
```suggestion
expect(wrapper.state()).to.have.property('hasError', true);
```

docs/api/ShallowWrapper/simulateError.md
```diff
@@ -35,7 +46,7 @@ class ErrorBoundary extends React.Component {
     const { children } = this.props;
     return (
       <React.Fragment>
-        {children}
+        { this.state.hasError ? "Error" : children }
```
ljharb:
```suggestion
        {hasError ? 'Error' : children}
```

docs/api/ShallowWrapper/simulateError.md
```diff
@@ -35,7 +46,7 @@ class ErrorBoundary extends React.Component {
     const { children } = this.props;
```
ljharb:
```suggestion
    const { children } = this.props;
    const { hasError } = this.state;
```

docs/api/ShallowWrapper/simulateError.md
```diff
@@ -50,6 +61,7 @@ const wrapper = shallow(<ErrorBoundary spy={spy}><Something /></ErrorBoundary>);
 const error = new Error('hi!');
 wrapper.find(Something).simulateError(error);
 
+expect(wrapper.state().hasError).to.equal(true);
```
ljharb:
```suggestion
expect(wrapper.state()).to.have.property('hasError', true);
```

packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
```diff
@@ -295,8 +295,12 @@ class ReactSixteenOneAdapter extends EnzymeAdapter {
         return instance ? toTree(instance._reactInternalFiber).rendered : null;
       },
       simulateError(nodeHierarchy, rootNode, error) {
+        const isErrorBoundary = el => (
+          el.componentDidCatch || (el.constructor && el.constructor.getDerivedStateFromError)
```
ljharb: Since the argument isn't an element, it's an instance, let's call it `instance`.

packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
```diff
@@ -295,8 +295,12 @@ class ReactSixteenOneAdapter extends EnzymeAdapter {
...
+        );
+
         const { instance: catchingInstance } = nodeHierarchy
-          .find(x => x.instance && x.instance.componentDidCatch) || {};
+          .find(x => x.instance && isErrorBoundary(x.instance)) || {};
```
ljharb: Also, rather than relying on `constructor`, let's pass in `x.instance, x.type` (and update the hierarchy code to pass it)

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -267,18 +267,35 @@ export function simulateError(
   getNodeType = nodeTypeFromType,
   getDisplayName = displayNameOfNode,
 ) {
-  const { componentDidCatch } = catchingInstance || {};
-  if (!componentDidCatch) {
+  const instance = catchingInstance || {};
+
+  const { componentDidCatch } = instance;
+
+  const { getDerivedStateFromError } = instance.constructor;
```
ljharb: similarly, let's pass in the constructor rather than relying on the `.constructor` property

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -267,18 +267,35 @@ export function simulateError(
...
+    const componentStack = getComponentStack(hierarchy, getNodeType, getDisplayName);
+    componentDidCatch.call(catchingInstance, error, { componentStack });
+  }
+
+  if (getDerivedStateFromError) {
+    setDerivedStateFromError();
+  }
+
+  if (componentDidCatch) {
+    callComponentDidCatch();
```
ljharb: can you confirm that this is the order these get called by react itself?

barmac:
Thanks for the feedback! I'll apply the changes/resolve conversations today evening or on Friday (UTC+1).

packages/enzyme-adapter-utils/src/Utils.js
barmac: `ComponentDidCatch` is always called as a callback to state update from `getDerivedStateFromError`, so the order is the same. See [ReactFiberUndwindWork#createClassErrorUpdate](https://github.com/facebook/react/blob/1e3b6192b54df20ac117a2af56afbe00ac9487b7/packages/react-reconciler/src/ReactFiberUnwindWork.js#L103) for comparison. 
However, I am not sure if calling `instance.setState` with state update from `gDSFE` is an accurate simulation. Do you have an idea how to make sure that `componentDidCatch` has access to the freshly updated state?

ljharb: Hmm - let's start by setting up very explicit test cases. We already have some that show that cDC is called, with the right info; we now have some that gDSFE is called, with the right info - now we need to ensure that cDC is called (with the right info) both when gDFSE returns new state, and also when it does not return changed state.

Once we get those set up (verified against actual React behavior), then it'll be easier to decide how to implement the behavior :-)

barmac: So actually React first tries to rerender after gDSFE and it fails when state remains the same, so cDC is never called. I pushed tests for this and have also a small codesandbox with such behaviour (https://codesandbox.io/s/q7r2z35y4w?fontsize=14).

barmac:
OK, so the tests revealed that simulateError does not cause an error boundary failure which can be triggered via real error during render.

ljharb:
@barmac can you elaborate on that last comment? `simulateError` is intended to be able to cause the same behavior as an error during actual rendering. If it doesn't, it may need to be fixed.

barmac:
Thanks for the `shallow` fix.

I updated PR so that we actually pass the type to simulate error instead of relying on constructor property. 

The previous comment is irrelevant as I forgot that with `simulateError` the component's state itself is not the cause of error, so even without state update the component should be able to rerender and be stable again. This is not the case regarding the real errors, where usually it is the component's state which is broken and not updating it forces React to unmount the whole tree.

ljharb:
I'm not sure that's necessarily true; you could easily have a rendering error without corrupting the state.

docs/api/ShallowWrapper/simulateError.md
```diff
@@ -26,16 +26,28 @@ function Something() {
...
 class ErrorBoundary extends React.Component {
+  static getDerivedStateFromError(error) {
+    return {
+      hasError: true,
+    };
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = {};
```
ljharb:
```suggestion
    this.state = { hasError: false };
```
(it's a best practice to always initialize every state field in the constructor)

docs/api/ReactWrapper/simulateError.md
```diff
@@ -26,16 +26,28 @@ function Something() {
...
 class ErrorBoundary extends React.Component {
+  static getDerivedStateFromError(error) {
+    return {
+      hasError: true,
+    };
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = {};
```
ljharb:
```suggestion
    this.state = { hasError: false };
```

packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
```diff
@@ -295,12 +295,17 @@ class ReactSixteenOneAdapter extends EnzymeAdapter {
         return instance ? toTree(instance._reactInternalFiber).rendered : null;
       },
       simulateError(nodeHierarchy, rootNode, error) {
-        const { instance: catchingInstance } = nodeHierarchy
-          .find(x => x.instance && x.instance.componentDidCatch) || {};
+        const isErrorBoundary = ({ instance: elInstance, type }) => (
```
ljharb: why is the `elInstance` rename needed?

packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
```diff
@@ -295,12 +295,17 @@ class ReactSixteenOneAdapter extends EnzymeAdapter {
...
+          (elInstance && elInstance.componentDidCatch) || (type && type.getDerivedStateFromError)
+        );
+
+        const { instance: catchingInstance, type: catchingType } = nodeHierarchy
+          .find(isErrorBoundary) || {};
 
         simulateError(
           error,
           catchingInstance,
+          catchingType,
```
ljharb: so something here. we have to always be able to handle "old enzyme + new adapter", "new enzyme + old adapter", and "new enzyme + new adapter" - so adding an argument in the middle like this is a breaking change.

It'll have to be added at the end.

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -262,23 +262,40 @@ export function getComponentStack(
 export function simulateError(
   error,
   catchingInstance,
+  catchingType,
```
ljharb: similarly, this makes the commit a breaking change - it'll have to be added at the end.

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -262,23 +262,40 @@ export function getComponentStack(
...
   getDisplayName = displayNameOfNode,
 ) {
-  const { componentDidCatch } = catchingInstance || {};
-  if (!componentDidCatch) {
+  const instance = catchingInstance || {};
+  const type = catchingType || {};
+
+  const { componentDidCatch } = instance;
+
+  const { getDerivedStateFromError } = type;
```
ljharb: i think this will work?
```suggestion
  const { getDerivedStateFromError } = {} = catchingType;
```

packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
barmac: OK, I'll move it to the end.

packages/enzyme-adapter-utils/src/Utils.js
barmac: I'll try this out.

packages/enzyme-adapter-react-16.1/src/ReactSixteenOneAdapter.js
barmac: ESLint rule `no-shadow` wouldn't allow plain `instance`.

packages/enzyme-adapter-utils/src/Utils.js
barmac: Ended up with `catchingType` at the end of params and thus with default value assigned already there.

barmac:
Rolled back react-adapter-16.1-3 changes as `getDerivedStateFromError` was introduced in React@16.6.

barmac:
Updated branch. @ljharb please check whether it's ready for merging :).

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -362,8 +362,12 @@ class ReactSixteenAdapter extends EnzymeAdapter {
         return instance ? toTree(instance._reactInternalFiber).rendered : null;
       },
       simulateError(nodeHierarchy, rootNode, error) {
-        const { instance: catchingInstance } = nodeHierarchy
-          .find(x => x.instance && x.instance.componentDidCatch) || {};
+        const isErrorBoundary = ({ instance: elInstance, type }) => (
+          (elInstance && elInstance.componentDidCatch) || (type && type.getDerivedStateFromError)
```
ljharb: probably only in 16.6+ should gDSFE's presence matter?
```suggestion
          (elInstance && elInstance.componentDidCatch) || (is166 && type && type.getDerivedStateFromError)
```

packages/enzyme-adapter-utils/src/Utils.js
```diff
@@ -266,19 +266,35 @@ export function simulateError(
...
 ) {
-  const { componentDidCatch } = catchingInstance || {};
-  if (!componentDidCatch) {
+  const instance = catchingInstance || {};
+
+  const { componentDidCatch } = instance;
+
+  const { getDerivedStateFromError } = catchingType;
+
+  if (!componentDidCatch && !getDerivedStateFromError) {
```
ljharb: I think we should also pass in a boolean, `checkForGetDerivedStateFromError` or something - because in react < 16.6, that logic shouldn't apply.

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -5891,14 +5891,18 @@ describeWithDOM('mount', () => {
...
+            } = this.state;
+
             return (
               <div>
                 <MaybeFragment>
                   <span>
                     <Thrower throws={throws} />
                     <div>
-                      {this.state.didThrow ? 'HasThrown' : 'HasNotThrown'}
+                      { didThrow ? 'HasThrown' : 'HasNotThrown' }
```
ljharb: jsx curlies don't get bounding spaces
```suggestion
                      {didThrow ? 'HasThrown' : 'HasNotThrown'}
```

packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -6026,6 +6030,385 @@ describeWithDOM('mount', () => {
       });
     });
 
+    describeIf(is('>= 16.6'), 'getDerivedStateFromError', () => {
+      describe('errors inside an error boundary', () => {
+        const errorToThrow = new EvalError('threw an error!');
+
+        const hasFragments = is('>= 16.2');
```
ljharb: this will always be true in 16.6+, so i think we can inline a `true` for this, and optimize further as a result.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -6225,14 +6225,17 @@ describe('shallow', () => {
...
+              throws,
+            } = this.state;
             return (
               <div>
                 <MaybeFragment>
                   <span>
                     <Thrower throws={throws} />
                     <div>
-                      {this.state.didThrow ? 'HasThrown' : 'HasNotThrown'}
+                      { didThrow ? 'HasThrown' : 'HasNotThrown' }
```
```suggestion
                      {didThrow ? 'HasThrown' : 'HasNotThrown'}
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -6311,6 +6314,352 @@ describe('shallow', () => {
       });
     });
 
+    describeIf(is('>= 16.6'), 'getDerivedStateFromError', () => {
+      describe('errors inside an error boundary', () => {
+        const errorToThrow = new EvalError('threw an error!');
+
+        const hasFragments = is('>= 16.2');
```
ljharb: also can inline here

ljharb:
I've made a few of the changes, and rebased; some tests are failing locally.

barmac:
Thanks :)

I ran `test:all` on clean install and did not get any test failures. Are you sure this is not related to any files left after previous work?

ljharb:
Looks like react v16.8.3 fails, and v16.8.4 passes.

barmac:
Still same, `npm run react 16.8.3 && npm run test:only` is all green. I confirmed with `node_modules/react` that I'm testing with the right version.

ljharb:
hmm, interesting. maybe it's the test renderer? I'll check again.

ljharb:
k, now i can't reproduce :-) i'll rebase this, and once tests pass, merge. Thanks for your patience!

Issue #2054, opened 2019-03-17
support unstable_Profiler


eps1lon:
### Current behavior
mounting a tree using `React.unstable_Profiler` throws
`Enzyme Internal Error: unknown node with tag 12`

https://codesandbox.io/s/jxjoj2p0v

### Expected behavior
`mount(<React.unstable_Profiler id="test" onRender={id => console.log(id)} />);` does not throw

#### API

- [x] shallow
- [x] mount
- [ ] render

#### Version

| library             | version
| ------------------- | -------
| enzyme              | 3.9.0
| react               | 16.8.4
| react-dom           | 16.8.4
| react-test-renderer | 16.8.4
| adapter (below)     | 1.11.2

#### Adapter

- [x] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-16.3
- [ ] enzyme-adapter-react-16.2
- [ ] enzyme-adapter-react-16.1
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-15.4
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] enzyme-adapter-react-helper
- [ ] others ( ) 

As far as I can tell enzyme is simply not supporting this (yet). Happy to start working on a PR that adds this to the mount API. 

I'm personally not that interested in shallow rendering. I guess we would need to establish first how the Profiler should work with shallow rendering.

This should also be included in #1553.

ljharb:
You’re correct, it doesn’t yet. A PR would be great!

ljharb:
(and yes, to be added to mount we’d need to decide what the behavior is for shallow, and ideally it’d be identical)

ljharb:
I'd say this is closed by #2055.

PR #2055, opened 2019-03-17
feat: Add support for unstable_Profiler to mount API

eps1lon:
Part of #2054

Opening this with failing tests. Will need to establish if we accept those for now or fix before merge. I'm also missing some context what work in `mount` is done by React and what is re-implemented in this package. 

## TODO
1. [x] revert `pretest` removal
2. [x] add find by display name support
3. [x] add find by `id` support
4. [x] investigate missing timings
5. [x] prop warnings for Profiler

### about find by *

~I guess this is related to `Profiler` not appearing in `wrapper.debug()` output. Might be caused by `Profiler` being one of those "mode-components" (e.g. StrictMode or ConcurrentMode)~ Solved and explained in https://github.com/airbnb/enzyme/pull/2055#discussion_r272426306

### missing timings

These are implented in the fiber related code (ReactFiberCommitWork):
- https://github.com/facebook/react/blob/0c03a474362c27aa9222ae27088782a01402067c/packages/react-reconciler/src/ReactFiberCommitWork.js#L564

`onRender` isn't even called with `ReactTestRenderer.create`. Maybe this is the actual blocker here?
Update: Timings weren't missing. The test environment didn't have a `performance` API available at which point React uses `Date.now` which isn't precise enough to display any duration.

### prop warnings
There is currently a test case that included a bare `<Profiler />` (existed before this PR). Those should trigger warnings about missing `id` and `onRender` props. Probably never triggered since this is not actual prop types validation but happens when creating the fiber: https://github.com/facebook/react/blob/4186952a6f3558eb4fae9f6c5f669bd898dc1d97/packages/react-reconciler/src/ReactFiber.js#L597
Update: Warnings are logged perfectly fine with `mount`. I don't think we need to shim those in `getDisplayName`.

### DIFF
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -216,6 +216,7 @@ function toTree(vnode) {
     case FiberTags.ContextProvider:
     case FiberTags.ContextConsumer:
       return childrenToTree(node.child);
+    case FiberTags.Profiler:
     case FiberTags.ForwardRef: {
       return {
         nodeType: 'function',
--- a/packages/enzyme-adapter-react-16/src/detectFiberTags.js
+++ b/packages/enzyme-adapter-react-16/src/detectFiberTags.js
@@ -19,6 +19,7 @@ module.exports = function detectFiberTags() {
   const supportsContext = typeof React.createContext !== 'undefined';
   const supportsForwardRef = typeof React.forwardRef !== 'undefined';
   const supportsMemo = typeof React.memo !== 'undefined';
+  const supportsProfiler = typeof React.unstable_Profiler !== 'undefined';
 
   function Fn() {
     return null;
@@ -65,6 +66,9 @@ module.exports = function detectFiberTags() {
       : -1,
     ForwardRef: supportsForwardRef
       ? getFiber(React.createElement(FwdRef)).tag
+      : -1,
+    Profiler: supportsProfiler
+      ? getFiber(React.createElement(React.unstable_Profiler, { id: 'mock', onRender: () => {} })).tag
       : -1,
   };
 };
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -24,6 +24,7 @@ import {
   Fragment,
   forwardRef,
   memo,
+  Profiler,
   PureComponent,
   useEffect,
   useState,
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -21,7 +21,6 @@
     "build:watch": "lerna run --parallel watch",
     "pretest": "lerna run lint",
     "test": "npm run test:only",
-    "pretest:only": "npm run build",
```
ljharb: i usually run `npx mocha --recursive packages/enzyme-test-suite/test` directly, to avoid this (because i run the build watcher in another tab)

eps1lon: Ah `npx`. I'm used to `yarn` where I it automatically uses the binary if the script can't be found. Thanks for the tip. Might be good to include in CONTRIBUTING.md.

eps1lon:
>should the "mode" components exist in .debug()? ideally yes, but we'd want it to be the same answer for mount and shallow, and react may have made the decision for us.

It does appear in the `react-devtools` browser extension and Reacts component stack traces which is why I thought it should appear in `.debug()` as well. But if the `debug` output should correspond to `shallow` levels? then it should match those.

I also had `find` in mind (currently 2 skipped test cases). I don't know if there are use cases to find a certain Profiler.

> There very well might be a bug with the test renderer, or with server rendering - ideally the profiler has actual meaningful behavior (even if that's "it just noops and renders its children) on the server, and thus in the shallow renderer

Added some more test cases to the codesandbox in #2054. It looks like for `ReactTestRenderer.create` `onRender` is never called. I guess `mount` should behave similar? 

Going to look into this again and open an issue on the react repo. It's not obvious to me how the Profiler should behave in `react-test-renderer`.

ljharb:
`<Profiler>` is a "shallow level"; i think that means it must appear in debug. You definitely should be able to `.find(Profiler)` if you like, altho I can't conceive of why you would :-)

The first source of truth is React itself in a browser - this determines how `mount` works. Next, we look at React on the server (`ReactDOM.renderToString`). This determines how SSR works. `shallow` tries to mimic `mount` as much as possible, and only deviates where a DOM is required, and possibly based on how SSR works. So, if react-test-renderer doesn't do something that React does, and there's an argument to be made that it should, then it's a bug with the test renderer.

eps1lon:
Update:
Timings weren't missing. The test environment didn't have a `performance` API available at which point React uses Date.now which isn't precise enough to display any duration.

Warnings are logged perfectly fine with mount. I don't think we need to shim those in `getDisplayName`.

`react-test-renderer/shallow` currently throws if you pass a valid `Profiler` component. [I asked in the profiler RFC](https://github.com/reactjs/rfcs/pull/51#issuecomment-474228457) what the plans are for `react-test-renderer`. I guess until then it's ok for the Profiler to not appear in `.debug()` or be available in `.find()`.

eps1lon:
Anything blocking this? The purpose is only to support `mount` in this PR. `shallow` support requires the [feature flag to be enabled in React](https://github.com/reactjs/rfcs/pull/51#issuecomment-474402923). Once this is released we can revisit the `shallow` API. It would be unfortunate if this would block partial support of the profiler in enzyme.

ljharb:
Given that the [implication](https://github.com/reactjs/rfcs/pull/51#issuecomment-474402923) is that it would work in the shallow renderer once it becomes stable, it seems like we'd want to add (skipped) shallow tests now, as well.

ljharb:
@eps1lon ok, i've rebased this and tweaked some things. Some tests are now failing for me locally, only because `Profile` *is* appearing in the shallow renderer - for consistency with mount, I think either we need to fix `mount` so that Profiler *does* show up in `.debug` - or, we need to fix `shallow` so that it does not (and update both the shallow and mount tests to consistently reflect that reality).

eps1lon:
I would include it in the debug() output since it would be closer to the actual component stack (in react and devtools). If it appears in debug() it should probably be findable too?

ljharb:
To clarify, these axioms would ideally remain true:
 1. If it appears in `.debug`, it is findable
 1. if it is findable, it appears in `.debug`
 1. if it appears in `.debug` for `mount`, it should for `shallow`
 1. if it appears in `.debug` for `shallow`, it should for `mount`
 1. if the react dev tools show something, it should show up in `.debug` (if the react dev tools do not show something, we can still choose to include it, if it passes the other criteria)

Thoughts?

eps1lon:
Sounds good. Follows principle of least surprise.

ljharb:
To be even more explicit :-) once the debug/findability/shallow-mount-consistency issue is resolved, I'm stoked to merge and release this. The currently-skipped shallow tests, pending changes in the shallow renderer, don't need to block full support for mount (as long as we have a path to identical support in the future, which we do)

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -207,6 +207,7 @@ function toTree(vnode) {
     case FiberTags.ContextProvider:
     case FiberTags.ContextConsumer:
       return childrenToTree(node.child);
+    case FiberTags.Profiler:
```
eps1lon: This was the mistake. I previously included this in the same branch as the context tags, strict mode etc. That branch will not be considered a node. At least this is how I understood it.

There's probably an argument to be made that they should also be considered as valid nodes (since they also appear in reacts component stack and devtools) but that's for another time (personally I'm not interested in those components WRT to find/debug).

Not sure if Profiler should be handled *exactly* like `forwardRef` but tests are passing for now so :man_shrugging:

eps1lon:
And it's green.

shallow and mount now create the same component tree WRT to the Profiler.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
ljharb: hmm - if they appear in react devtools, and the stack, then they *definitely* need to appear in `debug`.

andrewholsted:
Is unstable profiler different than the latest Profiler? We're looking for support in our jest tests and are running into issues

ljharb:
@andrewholsted if it's newly renamed in 16.9, enzyme hasn't added support for it yet.

andrewholsted:
@ljharb Ok, thanks. I'm working on a PR now. It seems pretty straight forward (as it's just removing the .unstable_) and making sure it's backwards compatible, too. Running into some issues building enzyme-adapter-react-16 locally and linking to my project to verify. I'm assuming this is a use case enzyme wants to support?

ljharb:
@andrewholsted note that the master branch is failing tests since the 16.9 release, so the first task to unblock all enzyme development is to make a PR fixing that :-/

Yes, the renamed Profiler should definitely be supported.

andrewholsted:
@ljharb - I think this PR fixes all tests on master due to those two issues

https://github.com/airbnb/enzyme/pull/2233

PR #2060, opened 2019-03-19
Update documentation

annie-annie:
change method call from `to.equal` to `toEqual`

### DIFF
--- a/docs/api/ShallowWrapper/last.md
+++ b/docs/api/ShallowWrapper/last.md
@@ -12,7 +12,7 @@ Reduce the set of matched nodes to the last in the set, just like `.at(length - 
 
 ```jsx
 const wrapper = shallow(<MyComponent />);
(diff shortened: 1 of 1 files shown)


ljharb:
Indeed, all our docs use chai syntax, so this is correct as is.

scott-ln:
Okay, but that's not made clear anywhere, which is why you've had multiple PRs from people on this topic. It's hostile to learners. Could you please consider improving your documentation to highlight that fact on individual documentation pages?

ljharb:
It would be hostile to learners in the same way to use jest matcher syntax.

we could import expect from chai in all our examples, i suppose.

scott-ln:
Yes of course, I'm not suggesting that you switch to the other syntax, just asking that you make it clear which syntax you're using. Thank you.

PR #2064, opened 2019-03-24
[doc]fix error example

Tankunpeng:
```
const total = wrapper.find(Bar).reduce((amount, n) => amount + n.prop('amount'));
```
In this example, total is an Object, because of there is no initial reduce value. we must set the initial value with 0 to get the right result.
And, the right result is 14.


Tankunpeng:
Sorry, the mount docs also need fix. I start a new pull request. #2066

PR #2066, opened 2019-03-24
the mount docs also need fix

Tankunpeng:


ljharb:
Merged into / duplicate of #2064.

PR #2074, opened 2019-03-28
Sort orgs list and add an entry

fatso83:
This sorts the list of organizations using a neutral alphabetical ordering and adds a company to the list.


ljharb:
These are sorted in insertion order; I'll update your PR to add your own organization to the bottom.

PR #2083, opened 2019-04-05
fix invariant violation in provider

ooHmartY:
Hello!

I'm running into an issue when using the latest version of react. 

```
Invariant Violation: Provider(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.
```

Turns out the context provider in `deterctFiberTags` needs to return a value.


ooHmartY:
No problem - thank you for the quick merge :D

PR #2100, opened 2019-04-16
Fix simulate being a no-op for memo components

eps1lon:
Fixes `mount(<MemoizedButton />).simulate('click')` not calling the `onClick` handler.

I strongly suspect this fixes issues with `simulate` not necessarily related to `React.memo` as well.

Alternate implementation without arrow functions and outer `this`: 9ef0776bc3fee3d539558c1ba486feb99273b223


eps1lon:
@ljharb That's even better, thanks.

PR #2116, opened 2019-05-06
Ensure the right versions of React / adapters are installed

ahuth:
There's a weird issue where partial React versions can result in the wrong versions of React being installed. This happens because `semver.intersects` is returning something weird when both are true:

- A partial version of React is used - for example, 16.3
- And it's compared against a version with prerelease specifiers in it

See [this truth table](https://stackblitz.com/edit/semver-intersection-truth-table) which highlights the issue.

In this PR we:
- Separate out a function for determining the right adapter to install (which makes testing it easier)
- Convert to using `semver.satisfies` from `semver.intersects`

@ljharb , what do you think?

### DIFF
--- a/packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
+++ b/packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
@@ -1,0 +1,32 @@
+import semver from 'semver';
+
+export default function getAdapterForReactVersion(reactVersion) {
+  const normalizedVersion = semver.coerce(reactVersion);
+
+  if (semver.satisfies(normalizedVersion, '^16.4.0-0')) {
+    return 'enzyme-adapter-react-16';
+  }
+  if (semver.satisfies(normalizedVersion, '~16.3.0-0')) {
+    return 'enzyme-adapter-react-16.3';
+  }
+  if (semver.satisfies(normalizedVersion, '~16.2')) {
+    return 'enzyme-adapter-react-16.2';
+  }
+  if (semver.satisfies(normalizedVersion, '~16.0.0-0 || ~16.1')) {
+    return 'enzyme-adapter-react-16.1';
+  }
+  if (semver.satisfies(normalizedVersion, '^15.5.0')) {
+    return 'enzyme-adapter-react-15';
+  }
+  if (semver.satisfies(normalizedVersion, '15.0.0 - 15.4.x')) {
+    return 'enzyme-adapter-react-15.4';
+  }
+  if (semver.satisfies(normalizedVersion, '^0.14.0')) {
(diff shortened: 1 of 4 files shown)


packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
```diff
@@ -0,0 +1,32 @@
+import semver from 'semver';
+
+export default function getAdapterForReactVersion(reactVersion) {
+  const normalizedVersion = semver.coerce(reactVersion);
```
ahuth: Need this because something like "16.3" isn't a valid version, per semver.

Their docs state that `coerce` is guaranteed to generate a version as long as there is a number in the string.

packages/enzyme-test-suite/test/enzyme-adapter-react-install-spec.js
```diff
@@ -0,0 +1,86 @@
+import { expect } from 'chai';
+import getAdapterForReactVersion from '../../enzyme-adapter-react-helper/src/getAdapterForReactVersion';
```
ahuth: Feel like I should've been able to do 

```
import getAdapterForReactVersion from 'enzyme-adapter-react-helper/src/getAdapterForReactVersion';
```

packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
```diff
@@ -0,0 +1,32 @@
...
+    return 'enzyme-adapter-react-15.4';
+  }
+  if (semver.satisfies(normalizedVersion, '^0.14.0')) {
+    return 'enzyme-adapter-react-14';
+  }
+  if (semver.satisfies(normalizedVersion, '^0.13.0')) {
+    return 'enzyme-adapter-react-13';
+  }
+
+  return 'enzyme-adapter-react-16';
```
ljharb: we didn't used to have this fallback here - should it throw instead?

packages/enzyme-test-suite/test/enzyme-adapter-react-install-spec.js
ljharb: i think if you add `enzyme-adapter-react-helper` as a dev dep of `enzyme-test-suite`, that this will work?

ahuth: Oh, that would make sense

packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
ahuth: Yeah, that's probably better than randomly getting the -16 adapter

PR #2121, opened 2019-05-11
React hooks test cases part 1 -- useLayoutEffect

chenesan:
Related to #2011 .

Added `useLayoutEffect` tests with almost the same flow except that the component using `useLayoutEffect` instead `useEffect`.

The shallow test is also skipped due to https://github.com/facebook/react/issues/15275 . After that resolves we should reopen those skipped tests.

Also reset the timeout of `setTimeout` in test from `1e3` to `100` -- hope to make tests a little sooner.

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -30,6 +30,7 @@ import {
   PureComponent,
   Suspense,
   useEffect,
+  useLayoutEffect,
   useState,
 } from './_helpers/react-compat';
 import {
@@ -973,7 +974,7 @@ describeWithDOM('mount', () => {
           setCtr(1);
           setTimeout(() => {
             setCtr(2);
-          }, 1e3);
+          }, 100);
         }, []);
         return (
           <div>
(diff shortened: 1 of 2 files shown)


packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -997,7 +998,40 @@ describeWithDOM('mount', () => {
   </div>
 </ComponentUsingEffectHook>`);
         done();
-      }, 1e3);
+      }, 100);
+    });
+    it('works with `useLayoutEffect`', (done) => {
```
```suggestion

    it('works with `useLayoutEffect`', (done) => {
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1131,13 +1132,44 @@ describe('shallow', () => {
...
+      setTimeout(() => {
+        wrapper.update();
+        expect(wrapper.debug()).to.equal(`<div>
+  2
+</div>`);
+        done();
+      }, 100);
+    });
+
+    // TODO: enable when the shallow renderer fixes its bug
```
```suggestion
    // TODO: enable when https://github.com/facebook/react/issues/15275 is fixed
```

PR #2122, opened 2019-05-11
Removed react-16.8 from allowed_failures in travis

### LINKED ISSUE #2113: hooks: fix failing tests with mismatched renderer
There's a few tests failing, currently marked as allowed failures in https://github.com/airbnb/enzyme/commit/3db6013d9f5b9034594b78f5c182fc4146af4841, that shouldn't be failing.

Ideally, the functionality can be fixed; if not, then the tests should be skipped when the breakages are expected.

AlokTakshak:
Closes #2113.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,10 +93,6 @@ matrix:
       env: KARMA=true REACT=15
     - node_js: "6"
       env: KARMA=true REACT=16
-    - node_js: "lts/*"
-      env: REACT=16.8.5 RENDERER=16.8.3
-    - node_js: "lts/*"
-      env: REACT=16.8 RENDERER=16.7
   exclude:
     - node_js: "6"
       env: REACT=0.13


AlokTakshak:
@ljharb 
I have removed react 16.8 from allowed_failures

.travis.yml
```diff
@@ -33,7 +33,7 @@ matrix:
     - node_js: "lts/*"
       env: REACT=16.8.3
     - node_js: "lts/*"
-      env: REACT=16.8 RENDERER=16.7
+      env: REACT=16.8 RENDERER=16.8
```
ljharb: this is breaking the purpose of the test, which is to test react 16.8 with the 16.7 renderer.
```suggestion
      env: REACT=16.8 RENDERER=16.7
```

ljharb (changes requested):

.travis.yml
AlokTakshak: But  we need corresponding version of react and react-dom if there is version mismatch - hooks test will fail as writtern on react website one of reason for invalid hooks is
`Mismatching Versions of React and React DOM`

ljharb: react and react-dom will both be 16.8 with this. `RENDERER` is about `react-test-renderer`, which doesn't necessarily have to match.

AlokTakshak: ok , then i can use `render, fireEvent, getByTestId` for updating the failing test cases?

ljharb: The work that needs to be done to close the linked issue is to determine if the existing tests are valid - if so, the actual code needs to be fixed; or, if they're not valid, change them so they are valid.

AlokTakshak: thanks, I will look into it and update because in some test case we are simulating and clicking and expecting updated value but it won't `rerender` in the `older version` of `react-test-renderer` so we have to `manually call` the` rerender` function.
thanks again for replying

ljharb: cool, maybe that's the fix we need then :-)

AlokTakshak: @ljharb  I was going through a lot of articles the react-test-renderer before 16.8.5 have problem with shallow rendering for hooks , as they won't re-render after setting State , and we are using same the same in Adaptors so shallow won't work  for version below 16.8.5,
now what we should do, modify the adaptor?

ljharb: yes, ideally we'd modify the react 16 adapter to have fallback behavior for an older renderer version.

ljharb: @AlokTakshak any interest in pursuing this fix?

AlokTakshak: I want to but i don't have any clue of how to fix it

ljharb: My thinking is that the react 16 adapter would add some way (only for react-test-renderer prior to 16.8.5) to catch `useState` calls and force a rerender manually.

PR #2126, opened 2019-05-15
Fix adapter helper regression

ahuth:
Fixes a regression caused by #2116.

If we specify `enzyme-adapter-react-install 16`, we want the latest version of the adapter supporting React 16 to be installed.

After #2116, however, we ended up getting the earliest version supporting React 16. The fix is to treat the provided version as a range (so 16 is basically `>= 16 < 17`) instead of a version (16 is `16.0.0`).

Unfortunately, pre-release versions complicates this, and this solution right now drops special handling for pre-release versions.

### DIFF
--- a/packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
+++ b/packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
@@ -1,30 +1,36 @@
 import semver from 'semver';
 
-export default function getAdapterForReactVersion(reactVersion) {
-  const normalizedVersion = semver.coerce(reactVersion);
+function removePreRelease(version) {
+  return semver.inc(version, 'patch');
+}
 
-  if (semver.satisfies(normalizedVersion, '^16.4.0-0')) {
+export default function getAdapterForReactVersion(reactVersion) {
+  const versionRange = semver.prerelease(reactVersion)
+    ? removePreRelease(reactVersion)
+    : semver.validRange(reactVersion);
+
+  if (semver.intersects(versionRange, '^16.4.0')) {
     return 'enzyme-adapter-react-16';
   }
-  if (semver.satisfies(normalizedVersion, '~16.3.0-0')) {
+  if (semver.intersects(versionRange, '~16.3.0')) {
     return 'enzyme-adapter-react-16.3';
   }
-  if (semver.satisfies(normalizedVersion, '~16.2')) {
+  if (semver.intersects(versionRange, '~16.2')) {
     return 'enzyme-adapter-react-16.2';
   }
-  if (semver.satisfies(normalizedVersion, '~16.0.0-0 || ~16.1')) {
+  if (semver.intersects(versionRange, '~16.0.0 || ~16.1')) {
     return 'enzyme-adapter-react-16.1';
   }
-  if (semver.satisfies(normalizedVersion, '^15.5.0')) {
+  if (semver.intersects(versionRange, '^15.5.0')) {
     return 'enzyme-adapter-react-15';
   }
-  if (semver.satisfies(normalizedVersion, '15.0.0 - 15.4.x')) {
+  if (semver.intersects(versionRange, '15.0.0 - 15.4.x')) {
     return 'enzyme-adapter-react-15.4';
   }
-  if (semver.satisfies(normalizedVersion, '^0.14.0')) {
+  if (semver.intersects(versionRange, '^0.14.0')) {
     return 'enzyme-adapter-react-14';
   }
-  if (semver.satisfies(normalizedVersion, '^0.13.0')) {
+  if (semver.intersects(versionRange, '^0.13.0')) {
     return 'enzyme-adapter-react-13';
   }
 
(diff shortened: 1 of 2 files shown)


packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
```diff
@@ -1,30 +1,30 @@
 import semver from 'semver';
 
 export default function getAdapterForReactVersion(reactVersion) {
-  const normalizedVersion = semver.coerce(reactVersion);
+  const versionRange = semver.validRange(reactVersion);
 
-  if (semver.satisfies(normalizedVersion, '^16.4.0-0')) {
+  if (semver.intersects(versionRange, '^16.4.0')) {
```
ahuth: Hey @ljharb , I know you said before that handling pre-release versions was important. Unfortunately, I can't figure out a good solution here that handles pre-releases and matches versions appropriately.

What's the use case for handling pre-releases here? I may be missing something, but in my mind since users of this package provide a specific react version (through either a CLI arg or an env variable), they can provide a general version that makes sense for them (ie - 16.5). Even if they're on a pre-release version of React.

ljharb: anybody still depending on a prerelease of react 16 - which includes a number of our packages, including enzyme itself - would be unable to use this helper without that support.

In other words, they'd need to be able to provide a `REACT` value that explicitly installed a prerelease.

ahuth: Okay, so what you're saying is we need to be able to do something like:

```
REACT=16.8.6-apha.666
```

Which would definitely break as a result of this PR. I've been searching all morning and have not found a solution that gives us that while also giving us the correct version for non-pre-release versions. Was hopeful that semver's `includePrelease` option would help, but it doesn't seem to.

Only thing I can think of now is to detect the prelease and have a different set of conditionals for those 😑

ljharb: or `REACT=16.9.0-alpha` etc.

packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
```diff
@@ -1,30 +1,40 @@
 import semver from 'semver';
 
+function isPreRelease(version) {
+  return semver.valid(version) && semver.SemVer(version).prerelease.length > 0;
+}
+
+function removePreRelease(version) {
+  return semver.inc(version, 'patch');
```
ahuth: This works because a pre-release is **_less_** than the normal release it corresponds to. So, `16.8.0-alpha.1` < `16.8.0`

packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
```diff
@@ -1,30 +1,40 @@
...
+}
+
 export default function getAdapterForReactVersion(reactVersion) {
-  const normalizedVersion = semver.coerce(reactVersion);
+  const versionRange = isPreRelease(reactVersion)
+    ? removePreRelease(reactVersion)
+    : semver.validRange(reactVersion);
```
ahuth: Great idea @ljharb about detecting prereleases and handling them as exact versions (which they are).

ahuth:
Alright, should be good to go @ljharb. Or, at least we have the following tested and green:

- pre-release versions
- exact versions
- ranges

packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
```diff
@@ -1,30 +1,40 @@
 import semver from 'semver';
 
+function isPreRelease(version) {
+  return semver.valid(version) && semver.SemVer(version).prerelease.length > 0;
```
ahuth: This is accessing semver internals, which may not be super great. Another option could be trying to regex the version number or something.

ljharb: i note that semver has an `includePrerelease` option that we could probably use in `intersects`, that defaults to false?

Separately, `semver. prerelease()` can be used to determine if it's a prerelease without using any internals.

ahuth: Ah, nice. I missed that one

ahuth: Also, we can definitely use `includePrerelease`, but it doesn't quite do what we want (as far as I can tell, in my testing).

Not sure exactly how `includePrerelease` affects `intersects`, but because a prerelease version is lower than its corresponding non-prerelease, they tend to match a version lower than we intend.

packages/enzyme-adapter-react-helper/src/getAdapterForReactVersion.js
ahuth: Just realized that there probably should've been a `semver.prerelease` check here (because if it's not a prerelease, the action here doesn't make sense.

Could've been something like:

```js
function makeValidRange(version) {
  return semver.prerelease(reactVersion)
    ? semver.inc(version, 'patch') // Remove prerelease part of the version
    : semver.validRange(reactVersion);
}
```

ljharb: Sounds like a great follow up PR :-D

PR #2127, opened 2019-05-17
docs(detach): fix example test

danielsogl:
The jest functions have been changed


ljharb:
I assume you discovered that all of our examples are using chai, not jest matchers :-)

Issue #2009, opened 2019-02-07
With shallow, redux store update does not trigger componentDidUpdate


adngdb:
Hi, 

I'm having trouble with a specific test scenario in my React + redux application. I want to test that a change to the redux store is correctly reflected in my component. For that, in my test I create a shallow component, connected to redux, and then I update the store using an action, then verify that the mounted component has the correct data. Which it doesn't.

Here's a test case to reproduce this: 

```javascript
import * as React from 'react';
import { connect } from 'react-redux';
import { createStore } from 'redux';

import { shallow } from 'enzyme';
import sinon from 'sinon';


const actions = {
    // This would fetch a resource on the network.
    getList: () => {},

    receiveList: list => {
        return {
            type: 'RECEIVE_LIST',
            list,
        };
    },

    // This would fetch a list of resources on the network.
    getObjects: () => {},
};


const reducer = (state = { list: [] }, action) => {
    switch (action.type) {
        case 'RECEIVE_LIST':
            return { ...state, list: action.list };
        default:
            return state;
    }
};


class MyAppBase extends React.Component {
    componentDidMount() {
        this.props.dispatch(actions.getList());
    }

    componentDidUpdate(nextProps) {
        if (nextProps.list !== this.props.list) {
            this.props.dispatch(actions.getObjects(nextProps.list));
        }
    }

    render() {
        return null;
    }
}


const MyApp = connect(state => { return { list: state.list }; })(MyAppBase);



describe('<MyApp>', () => {
    it('fetches objects once list is received', () => {
        // Mock actions that call the network.
        sinon.stub(actions, 'getList');
        actions.getList.returns({type: 'whatever'});
        sinon.stub(actions, 'getObjects');
        actions.getObjects.returns({type: 'whatever'});

        // Create a redux store.
        const store = createStore(reducer);

        // Create the component.
        const wrapper = shallow(<MyApp store={store} />).dive();

        // Verify that `componentDidMount` was called but not `componentDidUpdate`.
        expect(actions.getList.callCount).toEqual(1);
        expect(actions.getObjects.callCount).toEqual(0);

        // Dispatch an action that will change the redux store.
        store.dispatch(actions.receiveList([ 'aa', 'bb' ]));

        wrapper.update();

        // Expect the `componentDidUpdate` method to be called.
        expect(actions.getObjects.callCount).toEqual(1);

        // This doesn't work, for some reason `componentDidUpdate` is never called.
    });
});
```

### Current behavior

[...]

PR #2132, opened 2019-05-22
Failing test for shallow context propagation

tilgovi:
Forgive me if this is not expected behavior.

This PR contains a failing test case that I believe describes the core of the issue at #2009.

### DIFF
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -813,15 +813,21 @@ describe('shallow', () => {
 
         it('propagates updates to rendered children', () => {
           const wrapper = shallow(<TestComponent />, { context: { baz: 'enzyme' } });
+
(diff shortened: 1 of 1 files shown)


tilgovi:
This may be an intended change in v3, with a related description here: https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#calling-props-after-a-state-change

But, I came across this enough times in the issues that I wasn't totally clear what to expect and wanted to close the loop on the comment thread at #2009 by opening a PR as you asked.

ljharb:
Thanks, this is great; I'll look into it.

ljharb:
So, looking at this - as soon as you `.dive()`, the new wrapper is severed from the original one - so changes to the root *don't* affect the newly created wrappers. (this is also true for any new wrappers created from, eg, `find`, as you linked to in the migration guide)

I've updated this PR with the changes needed to make the tests pass, as intended.

PR #2140, opened 2019-05-25
explained why need to re-call .find() after update

### LINKED ISSUE #2139: cached find() result is not updated after re-render
In short: if I put `find()` result into interm variable it is not updated after wrapper is re-rendered. Meanwhile fresh call to `find` returns actual element.

Having really simple component:
```
const List = ({ options = [] }) => 
    <>
       {options.map((i, index) => <span key={index}>{i}</span>)}
    </>

it('updates first item on updating options', () => {
	const options = [1];
	const wrapper = shallow(<List options={options} />);
	const first = wrapper.find('span').slice(0, 1);
	expect(first.props().children).toEqual(1);
	wrapper.setProps({ options: [2] });
	expect(wrapper.find('span').slice(0, 1).props().children).toEqual(2);
	expect(first.props().children).toEqual(2);	
})
```

### Current behavior
Test fails at last statement. Pre-last(with new `find()` call) passes.

### Expected behavior
Both cached wrapper and call to `find()` returns the same `Wrapper` instance

### Your environment
Windows 10, Node 8.12.0

#### API

- [x] shallow
- [ ] mount
- [ ] render

#### Version

| enzyme              |  3.9.0
| react               |  16.8.6
| react-dom           |  16.8.6
| react-test-renderer | 16.8.6
| adapter (below)     | 1.13.1

#### Adapter

- [x] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] others ( )

skyboyer:
This behavior currently is explained only in 2-to-3 version migration guide. People who never used v2 have small chance to get there. 
Since it's related to different methods and both mount/shallow suppose this file is better place for describing that.

Closes #2139.

### DIFF
--- a/docs/common-issues.md
+++ b/docs/common-issues.md
@@ -10,6 +10,51 @@ This could be due to a regression, or the feature is not yet implemented. If you
 certain query syntax, make sure it is implemented first before raising an issue. Here is the list of
 selectors we currently support: https://github.com/airbnb/enzyme/blob/master/docs/api/selector.md
 
+### Nested component may not be updated after wrapper updates
+
+Assume we have a simple component with an `<input />` and a `<button>`. Clicking the `<button>` updates the `<input>`'s value.
+
+However, this test fails:
+```jsx
+const input = wrapper.find('input');
+const button = wrapper.find('button');
+button.prop('onClick')();
+expect(input.prop('value')).to.equal('test');
+```
+fails. While this test:
+```jsx
+const input = wrapper.find('input');
+const button = wrapper.find('button');
+button.prop('onClick')();
+expect(wrapper.find('input').prop('value')).to.equal('test');
+```
+passes.
+
+This is because the wrapper returned by `.find()` (and every other method that produces a new wrapper) is immutable and won't update.
+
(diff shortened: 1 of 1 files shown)


docs/common-issues.md
```diff
@@ -10,6 +10,48 @@ This could be due to a regression, or the feature is not yet implemented. If you
...
+fails. While
+```jsx
+const input = wrapper.find('input');
+const button = wrapper.find('button');
+button.props().onClick();
+expect(wrapper.find('input').props().value).toEqual('test');
+```
+passes.
+
+This is related to both `mount()` and `shallow()`, `find()` and `findWhere` methods.
```
ljharb: It relates to any method that can produce a new wrapper; filter, filterWhere, find, findWhere, dive, mount, shallow, etc. Can we find a way to rephrase this so it's not an explicit list?

skyboyer: Rephrased. Not really sure if it should be `wrapper` or `wrappers` when goes to list/collection.

ljharb: It's a single wrapper, around 0, 1, or more things, so `wrapper` is correct.

skyboyer: is there anything else I should fix?

PR #2142, opened 2019-05-26
Support for an optional `last` argument, to count backward

maraisr:
With this PR I aim to add an optional argument to the `last` _filter_, so address the scenario where you might not wan't the last, but the second last. But without having to keep a count of children, to then use the `.at(index)`.

I see it work like this:

```jsx
const wrapper = shallow(<ul>
	<li>a</li>
	<li>b</li>
	<li>c</li>
	<li>d</li>
</ul>);

expect(wrapper.find('li').last().text()).toEqual('d');
expect(wrapper.find('li').last(-1).text()).toEqual('c');
// or
const children = wrapper.find('li')
expect(children.at(children.length - 2).text()).toEqual('c');
```

### DIFF
--- a/docs/api/ShallowWrapper/last.md
+++ b/docs/api/ShallowWrapper/last.md
@@ -2,6 +2,10 @@
 
 Reduce the set of matched nodes to the last in the set, just like `.at(length - 1)`.
 
+#### Arguments
+
+1. fromIndex (Number [optional]): The index at which item you wan't moving backwards.
+
 
 #### Returns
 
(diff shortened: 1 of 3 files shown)


ljharb:
This already exists - it's called `slice`. You can do `.slice(-1)`, for example.

maraisr:
Oh - in which case this is probably mute. Just thought that `.at` is akin to `:nth-child`, and `.last` was akin to `:nth-last-child`. But thanks. :100:

ljharb:
Since enzyme wrappers are collections, there's not actually a difference between what "last" provides (a wrapper around 1 item) - and what `.slice(-2, 1)` provides (a wrapper around 1 item).

PR #2148, opened 2019-05-30
React hooks test case part 3 -- useCallback

chenesan:
The tests are almost as same as #2144 , but for `useCallback`.

And I saw that shallow renderer doesn't memoize the callback between two rendering even the dependencies not change. So the test in https://github.com/chenesan/enzyme/commit/5cc14bacee8846bb5595936b041b66540528eeae#diff-2e9c9c89f001140a7699819fc2f5d926R1252 will fail and I temporarily skip it. Have opened an issue in react (https://github.com/facebook/react/issues/15774) . If it's confirmed as a bug of shallow renderer then I'll try to fix it in react side.

### DIFF
--- a/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
@@ -29,6 +29,7 @@ import {
   Profiler,
   PureComponent,
   Suspense,
+  useCallback,
   useEffect,
   useLayoutEffect,
   useMemo,
@@ -613,7 +614,7 @@ describeWithDOM('mount', () => {
         );
 
         const context = { name: 'foo' };
-        expect(() => mount(<SimpleComponent />, { context })).to.not.throw();
+        expect(() => mount(<SimpleComponent />, { context })).not.to.throw();
       });
 
       itIf(is('< 16'), 'is introspectable through context API', () => {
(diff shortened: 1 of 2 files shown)


packages/enzyme-test-suite/test/ReactWrapper-spec.jsx
```diff
@@ -1071,6 +1072,44 @@ describeWithDOM('mount', () => {
       const nextValue = wrapper.find(Child).prop('memoized');
       expect(initialValue).to.not.equal(nextValue);
     });
+
+    it('get same callback when using `useCallback` and rerender with same prop in dependencies', () => {
+      function Child() {
+        return false;
```
```suggestion
        return null;
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1247,6 +1248,44 @@ describe('shallow', () => {
       const nextValue = wrapper.find(Child).prop('memoized');
       expect(initialValue).to.not.equal(nextValue);
     });
+
+    it.skip('get same callback when using `useCallback` and rerender with same prop in dependencies', () => {
```
```suggestion
    // pending https://github.com/facebook/react/issues/15774
    it.skip('get same callback when using `useCallback` and rerender with same prop in dependencies', () => {
```

chenesan:
I just sent a PR for this in https://github.com/facebook/react/pull/15843 ; As for bugfix in enzyme side, I'm not sure what the way is -- inherit `ReactShallowRenderer` class and override `_createDispatcher` in https://github.com/facebook/react/blob/master/packages/react-test-renderer/src/ReactShallowRenderer.js#L230 ? That sounds like a big change... but seems a way to fix hooks issue in shallow renderer (for example, `useEffect` won't work in shallow).

ljharb:
We could stub `useCallback` in the shallow renderer, perhaps?

PR #2158, opened 2019-06-09
Wrap `.Invoke()` with act to make trigger rerender by handler less error prone

chenesan:
This PR expose `act` call out of `enzyme-adapter-react-16` and make the `invoke` api call handler with `act`. This will make the added test case passed without explicitly call `act()`. Before this change, we have to call `act()` outside of the `onClick` handler manually, which is easy to forget to do that.

Also update the doc of `invoke` to encourage people to trigger handler with it (so we don't need to explictly call `act()` and `update()`).

### DIFF
--- a/docs/api/ReactWrapper/invoke.md
+++ b/docs/api/ReactWrapper/invoke.md
@@ -1,6 +1,7 @@
 # `.invoke(propName)(...args) => Any`
 
 Invokes a function prop.
+Note that in React 16.8+, `.invoke` will wrap your handler with [`ReactTestUtils.act`](https://reactjs.org/docs/test-utils.html#act) and call `.update()` automatically.
 
 #### Arguments
 
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -495,6 +495,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
           }),
         };
       },
+      ...(is168 && { wrapInvoke: wrapAct }),
     };
   }
 
--- a/packages/enzyme/src/ReactWrapper.js
+++ b/packages/enzyme/src/ReactWrapper.js
@@ -841,7 +841,9 @@ class ReactWrapper {
         throw new TypeError('ReactWrapper::invoke() requires the name of a prop whose value is a function');
       }
       return (...args) => {
-        const response = handler(...args);
+        const response = typeof this[RENDERER].wrapInvoke === 'function'
+          ? this[RENDERER].wrapInvoke(() => handler(...args))
+          : handler(...args);
         this[ROOT].update();
         return response;
       };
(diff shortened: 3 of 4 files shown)


docs/api/ReactWrapper/invoke.md
```diff
@@ -1,6 +1,8 @@
 # `.invoke(propName)(...args) => Any`
 
 Invokes a function prop.
+Note that `.invoke` will wrap your handler with [`ReactTestUtils.act`](https://reactjs.org/docs/test-utils.html#act) and call `.update()` automatically. It would
```
ljharb:
```suggestion
Note that in React 16.8+, `.invoke` will wrap your handler with [`ReactTestUtils.act`](https://reactjs.org/docs/test-utils.html#act) and call `.update()` automatically. This may be more ergonomic than directly invoking a `.props()` function, since you won't have to wrap it in `act` yourself.
```

docs/api/ReactWrapper/invoke.md
```diff
@@ -1,6 +1,8 @@
...
+be less error prone than directly call handler prop with `.props()`.
```
ljharb:
```suggestion
```

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -495,6 +495,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
           }),
         };
       },
+      act: wrapAct,
```
ljharb: this is semver-minor in the adapter

packages/enzyme/src/ReactWrapper.js
```diff
@@ -841,7 +841,8 @@ class ReactWrapper {
         throw new TypeError('ReactWrapper::invoke() requires the name of a prop whose value is a function');
       }
       return (...args) => {
-        const response = handler(...args);
+        const wrapAct = this[RENDERER].act || function wrapAct(fn) { return fn(); };
```
ljharb: instead of the `||` here, let's either add a default `act` to `EnzymeAdapter`, or, let's just avoid the wrapping where it's not needed.

docs/api/ReactWrapper/invoke.md
chenesan: IMO,
```javascript
wrapper.invoke('handler')
expect(...)
```
is better than
```javascript
ReactTestUtils.act(wrapper.prop('handler')());
wrapper.update();
expect(...)
```
if we simply want to trigger handler. Since it's easy to forget (1) the `act` wrap and (2) the `update` call. I think this knowledge may worth a document (or maybe my opinion is wrong).

ljharb: I agree; I folded this line into the one above it.

packages/enzyme/src/ReactWrapper.js
ljharb: ping @chenesan; this remains

chenesan:
@ljharb Just address the review and rebase on master to resolve the conflict :)

packages/enzyme/src/ReactWrapper.js
chenesan: Ah, just changed it... Let's revert it back

chenesan: Sorry for misunderstanding. Just changed the logic so it will only wrap `handler` call with `.act` when `this[RENDERER].act` exists.

PR #2175, opened 2019-06-26
[Docs] `invoke`: Fix missing white space

alecklandgraf:
I noticed the syntax highlighting isn't working in the docs for invoke. Other docs seem to have 
a space after the ```` ```jsx```` 
https://airbnb.io/enzyme/docs/api/ShallowWrapper/invoke.html

image

@milesj

### DIFF
--- a/docs/api/ShallowWrapper/invoke.md
+++ b/docs/api/ShallowWrapper/invoke.md
@@ -15,7 +15,7 @@ This essentially calls wrapper.prop(propName)(...args).
 
 #### Example
 
-```jsx
+```jsx 
 class Foo extends React.Component {
   loadData() {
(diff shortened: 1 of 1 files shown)


alecklandgraf:
will build the docs locally and see if it fixes

alecklandgraf:
In my local docs, both with and without a space rendered as expected

alecklandgraf:
cc @milesj

milesj:
IMO this space shouldn't be necessary. GFM totally works regardless of this space so I suspect it's something else.

ljharb:
This was fixed in #2160; it will be updated on the site with the next release of `enzyme` itself.

PR #2178, opened 2019-06-30
Explicit React CSS selector syntax description

levimeahan:
It is not obvious on reading through the document as-is that the CSS selector syntax works on React component names and their props. While it can be inferred through reading "The Key and Ref Props" section, that section can easily be missed when someone is specifically looking at the lists and examples of selectors to see what can be used. It is much clearer to explicitly state this syntax works on React components.

### DIFF
--- a/docs/api/selector.md
+++ b/docs/api/selector.md
@@ -13,6 +13,7 @@ follows:
 - id syntax (`#foo`, `#foo-bar`, etc.)
 - attribute syntax (`[href="foo"]`, `[type="text"]`, and the other attribute selectors listed [here](https://developer.mozilla.org/en-US/docs/Learn/CSS/Introduction_to_CSS/Attribute_selectors).)
 - universal syntax (`*`)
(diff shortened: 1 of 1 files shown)


docs/api/selector.md
```diff
@@ -13,6 +13,7 @@ follows:
 - id syntax (`#foo`, `#foo-bar`, etc.)
 - attribute syntax (`[href="foo"]`, `[type="text"]`, and the other attribute selectors listed [here](https://developer.mozilla.org/en-US/docs/Learn/CSS/Introduction_to_CSS/Attribute_selectors).)
 - universal syntax (`*`)
+- React component name and props (`Button`, `Button[type="submit"]`, etc)
```
ljharb:
```suggestion
- React component name and props (`Button`, `Button[type="submit"]`, etc) - however, please note that it is strongly encouraged to find by component constructor/function and not by display name.
```

PR #2180, opened 2019-06-30
Add HOCComponent case to 'toTree' function.

psr1919plus21:
Hi!
I write several tests for my component and it's work fine, but when I wrapped my component with some HOC that create function like this:

```
function Foo(props) {
                var _this = _super.call(this, props) || this;
                _this.action = _this.action.bind(_this);
                _this.nodeParams = __assign({}, nodeParams, props.logNode, { events: __assign({}, nodeParams.events, (props.logNode && props.logNode.events)) });
                return _this;
}
```

I faced with error: `Enzyme Internal Error: unknown node with tag 9`.

I add code for this case to `enzyme-adapter-react-16.2` because I have `"enzyme-adapter-react-16": "1.2.0",` in my `package.json` file.

Not sure about that I need to add similar changes for other versions.

1. Can you approve my changes?
2. Can you help me with understanding about other adapter versions?
3. How else I can help? What kind of information I need to provide additionally?

Thanks :)

### DIFF
--- a/packages/enzyme-adapter-react-16.2/src/ReactSixteenTwoAdapter.js
+++ b/packages/enzyme-adapter-react-16.2/src/ReactSixteenTwoAdapter.js
@@ -48,6 +48,7 @@ const HostPortal = 4;
 const HostComponent = 5;
 const HostText = 6;
 const Mode = 11;
+const HOCComponent = 9;
 
 function nodeAndSiblingsArray(nodeWithSibling) {
   const array = [];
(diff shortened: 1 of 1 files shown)


ljharb:
Also, `this` in a function component makes no sense - they shouldn’t have one.

psr1919plus21:
Thanks for answer

I use `"react": "16.8.4"`
`this` in a function component utilizes for logs purposes.

If we talk out of context of my HOC component, it's good idea to handle this case?
Should I use `ContextConsumer` constant and add some regression tests?

ljharb:
React passes `this` as `undefined` (or the global object) to function components, so there's no logging value to be had.

If you're using React 16.4, then changing the 16.2 adapter doesn't seem productive. Can you share more of your code? Specifically, the untranspiled code?

ljharb:
@psr1919plus21 is this PR something you're still interested in completing?

ljharb:
Happy to reopen if there's still value to be had here; closing for now.

PR #2182, opened 2019-07-01
update docs to note future deprecation of .simulate

browne0:
This PR updates the doc to explain that simulate doesn't work as intended. I included a link to the original issue that I opened on Friday.

### DIFF
--- a/book.json
+++ b/book.json
@@ -1,5 +1,5 @@
 {
-  "gitbook": "3.2.2",
+  "gitbook": "3.2.3",
   "title": "Enzyme",
   "description": "React Testing",
   "plugins": [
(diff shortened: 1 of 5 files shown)


book.json
```diff
@@ -1,5 +1,5 @@
 {
-  "gitbook": "3.2.2",
+  "gitbook": "3.2.3",
```
browne0: this was needed to get the docs working, looks like they updated the latest version.

browne0:
@ljharb Could I get a review on this? Thanks.

azangru:
Don't know if it's too much to ask, but I was wondering whether you might include an example of the officially blessed way of testing components' behavior in response to DOM events.

I am particularly curious about DOM-mounted components. It makes total sense to test shallow-rendered components by directly calling functions that they pass to their children as props, but for a fully mounted component, you might want to test it via its interactions with the DOM.

ughitsaaron:
Would it be considered out of scope/desirable to add a warning output to the console when `.simulate` is called? Happy to make a PR to follow up.

ljharb:
@browne0 sorry for the long delay; i'll review this soon.

@azangru that would require a full simulation interface, which I want too, but is way too massive a project to be developed as part of enzyme. It's not as simple as clicks; you'd want to simulate things like "type `foobar` into the text field, hit tab, type `baz`, use the mouse to click a checkbox and then select from a dropdown, and then click into another text field and type `quux`, and then hit enter" - which involves a very large number of events, precisely timed and ordered, with a bunch of faked event objects.

@ughitsaaron no, I'm not interested in spamming warnings on every usage of the method, i'd rather just remove it in v4; if we used react's caching approach to only show the message once, it'll be ignored anyways.

PR #2187, opened 2019-07-09
update to match API

Kalpan-Jasani:
The API code documentation says that .instance() can only be called on the root, but this is not mentioned in the documentation. Therefore, a descriptive line is added at the appropriate place.


ljharb:
This is true for shallow, but not for mount, as i assume you discovered?

PR #2211, opened 2019-08-01
[Docs] Update ShallowWrapper/instance wording to account for hooks

thiskevinwang:
# Description

This updates the wording on https://airbnb.io/enzyme/docs/api/ShallowWrapper/instance.html# to account for React Hooks.


ljharb (approved):
I'm still pretty opposed to the idea of renaming SFCs just because the react team decided to incorrectly name the hook "useState" - functional components are stateless regardless - but this seems like a good change. I've modified it to also update the `mount` docs.

Issue #2098, opened 2019-04-13
Incorrect behaviour when statically rendering components returning strings 


ottosichert:
### Current behavior

Using Enzyme's static `render()` on a component that returns a string returns an empty result or breaks:

#### Setup

```js
import chai from "chai";
import chaiEnzyme from "chai-enzyme";
import Enzyme from "enzyme";
import Adapter from "enzyme-adapter-react-16";

Enzyme.configure({ adapter: new Adapter() });
chai.use(chaiEnzyme());
```

#### Tests

```jsx
import { expect } from "chai";
import { render } from "enzyme";
import React from "react";

const Identity = props => props.children;

describe("Components rendering a string", () => {
  it("should contain this string", () => {
    const wrapper = render(<Identity>foo</Identity>);
    expect(wrapper.text()).to.equal("foo");
  });

  it("should not break static rendering", () => {
    const wrapper = render(<Identity>bar!</Identity>);
    expect(wrapper.text()).to.equal("bar!");
  });

  it("should work as expected when wrapping in a DOM element", () => {
    const wrapper = render(
      <div>
        <Identity>foo</Identity>
      </div>
    );
    expect(wrapper.text()).to.equal("foo");
  });
});
```

#### Results

```
  Components rendering a string
    1) should contain this string
    2) should not break static rendering
    ✓ should work as expected when wrapping in a DOM element

  1 passing (329ms)
  2 failing

  1) Components rendering a string
       should contain this string:

      AssertionError: expected '' to equal 'foo'
      + expected - actual

      +foo

      at Context.equal (test.js:10:31)

  2) Components rendering a string
       should not break static rendering:
     SyntaxError: Unmatched selector: !
      at parse (node_modules/css-what/index.js:85:9)
      at compileUnsafe (node_modules/css-select/lib/compile.js:31:14)
      at select (node_modules/css-select/index.js:18:49)
      at CSSselect (node_modules/css-select/index.js:41:9)
      at initialize.exports.find (node_modules/cheerio/lib/api/traversing.js:40:21)
      at initialize.module.exports (node_modules/cheerio/lib/cheerio.js:88:18)
      at new initialize (node_modules/cheerio/lib/static.js:34:20)
      at initialize (node_modules/cheerio/lib/static.js:31:14)
      at render (node_modules/enzyme/build/render.js:42:39)
      at Context.<anonymous> (test.js:14:21)
```

The results are identical when using class components.

### Expected behavior

All tests should pass

#### API

[...]

PR #2221, opened 2019-08-11
[Fix] `render`: handle Fiber strings and numbers

ottosichert:
- Fix https://github.com/airbnb/enzyme/issues/2098
- Introduce `loadCheerioRoot` util
- Use util in `render`, `mount` and `shallow`
- Added test cases in `test/staticRender-spec.jsx`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_script:
 script:
   - 'if [ -n "${LINT-}" ]; then npm run lint; elif [ -n "${KARMA-}" ]; then npm run test:karma -- --single-run; elif [ -n "${REACT-}" ]; then npm run travis; else false ; fi'
 after_script:
-  - 'if [ -n "${REACT-}" ]; then case "${TRAVIS_NODE_VERSION}" in "8" | "6") ;; *) cat ./coverage/lcov.info | ./node_modules/.bin/coveralls ;; esac ; fi'
+  - 'if [ -n "${REACT-}" ] && [ "${TRAVIS_ALLOW_FAILURE}" != true ]; then case "${TRAVIS_NODE_VERSION}" in "8" | "6") ;; *) cat ./coverage/lcov.info | ./node_modules/.bin/coveralls ;; esac ; fi'
 sudo: false
 matrix:
   fast_finish: true
@@ -38,12 +38,6 @@ matrix:
       env: REACT=16.7 RENDERER=16.8
     - node_js: "lts/*"
       env: REACT=16.7 RENDERER=16.7
-    - node_js: "lts/*"
-      env: REACT=16.6
-    - node_js: "lts/*"
-      env: REACT=16.5
-    - node_js: "lts/*"
-      env: REACT=16.4
     - node_js: "lts/*"
       env: REACT=16.3 ADAPTER=16
     - node_js: "6"
--- a/packages/enzyme-adapter-utils/src/createMountWrapper.jsx
+++ b/packages/enzyme-adapter-utils/src/createMountWrapper.jsx
@@ -48,7 +48,6 @@ export default function createMountWrapper(node, options = {}) {
     constructor(...args) {
       super(...args);
       const { props, wrappingComponentProps, context } = this.props;
-      this.rootFinderInstance = null;
       this.state = {
         mount: true,
         props,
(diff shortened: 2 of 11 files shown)


ottosichert:
Rebased and fixed linter errors

ljharb:
I've rebased again, including my latest changes as well as your utils spec.

ottosichert:
That was quick, thank you very much! I didn't see your approval while I was adding the utils spec. I added them because the coverage went down.

With your latest push everything is as it should be :+1:

ottosichert:
This build will fail: https://travis-ci.org/airbnb/enzyme/jobs/585805429

```
  1) Utils loadCheerioRoot always returns a Cheerio instance:

     AssertionError: expected { Object (options) } to be an instance of 

      at Context.<anonymous> (packages/enzyme-test-suite/test/Utils-spec.jsx:1081:42)
```

I will push a fix shortly

ljharb:
cheerio does need to be a dep of the test suite, but that doesn't fix it. I'll wait for your push.

ottosichert:
I used the public property of the `Cheerio` constructor called `cheerio`, see:

https://github.com/cheeriojs/cheerio/blob/1.0.0-rc.3/lib/cheerio.js#L101
```js
Cheerio.prototype.cheerio = '[cheerio object]';
```

It is only available on Cheerio instances:

```js
$ node
> const cheerio = require('cheerio');
undefined
> cheerio.version
'1.0.0-rc.3'
> cheerio.cheerio
undefined
> cheerio('<div>foo</div>').cheerio
'[cheerio object]'
> cheerio.load('')().cheerio
'[cheerio object]'
> 

```

ljharb:
eesh, i mean, that clearly works but that doesn't seem robust enough to be a valuable test. why was instanceOf(cheerio) failing? it works in the repl.

ottosichert:
The `cheerio` property has been available since the very first release: https://github.com/cheeriojs/cheerio/blob/0.1.0/src/cheerio.coffee#L66
```coffeescript
    cheerio : "0.0.1"   
```

It was later changed to hold the special value `"[cheerio object]"` instead of the version number in https://github.com/cheeriojs/cheerio/commit/41337f9df1b09a27905ff71389ad478c56053c8f

```diff
-    cheerio : '0.4.2',
+    cheerio : '[cheerio object]',
```

As it is also used in `dom-serializer` https://github.com/cheeriojs/dom-serializer/blob/v0.2.1/index.js#L88 I followed the same approach.

However, I didn't look into why the `.instanceOf` didn't work as expected, and I can reproduce the same behaviour in `node`:

```js
> const cheerio = require('cheerio');
undefined
> cheerio.version
'1.0.0-rc.3'
> cheerio.root() instanceof cheerio
true
> cheerio.load('<div>foo</div>', null, false).root() instanceof cheerio
true
> cheerio.load('')('<div>foo</div>') instanceof cheerio
true
> 
```

I will investigate and submit a new version when I found the issue

ljharb:
Thanks! The issue isn't whether the property works in every version of cheerio - this is knowable - but whether any object can mistakenly pass the test by having an arbitrary string property. In tests of our own code, this really isn't a big concern, just a philosophical one :-) really, it's that i'd prefer to at least understand why no better option is available before accepting a non-ideal one.

ottosichert:
What I found out so far is that there are two different **identities** for `cheerio`:
- `require('cheerio')` in `enzyme/build/Utils`
- `import cheerio from 'cheerio'` in `enzyme-test-suite/test/Utils-spec.jsx`

Which means the versions are identical but `instanceOf` and equality check will fail:

```
  describe('loadCheerioRoot', () => {
    it('always returns a Cheerio instance', () => {
      const buildCheerio = loadCheerioRoot().constructor;
      expect(buildCheerio.version).to.equal(cheerio.version);
      expect(String(buildCheerio)).to.equal(String(cheerio));
      expect(buildCheerio).to.equal(cheerio); // fails
      [...]
```

```
  1) Utils loadCheerioRoot always returns a Cheerio instance:

      AssertionError: expected [Function] to equal [Function]
      + expected - actual

      
      at Context.<anonymous> (packages/enzyme-test-suite/test/Utils-spec.jsx:1084:14)
```

This seems to be more of a build setup issue, did you run into this problem before?

ottosichert:
Also reproducible in `node`:

```js
enzyme/packages/enzyme-test-suite$ node
> const cheerio = require('cheerio');
undefined
> const loadCheerioRoot = require('enzyme/build/Utils').loadCheerioRoot;
undefined
> loadCheerioRoot().constructor.version === cheerio.version
true
> String(loadCheerioRoot().constructor) === String(cheerio)
true
> loadCheerioRoot().constructor === cheerio
false
> 
```

ljharb:
gotcha, now that makes sense :-) i'd expect lerna to hoist them up so they're the same version, but either way this is something I can solve. Will fix shortly.

ljharb:
k, i ended up going with your test suggestion ¯\\\_(ツ)_/¯

ottosichert:
@ljharb Thanks for wrapping this up!

PR #2222, opened 2019-08-20
Updating to secure URL

votemike:


votemike:
> this isn't a published package.json, so nobody ever looks at this URL, but might as well update it :-) thanks!

When running `yarn outdated`, each package has its URL displayed.

ljharb:
@votemike yes but the root package.json is never published, so if yarn's showing that URL, it's broken. Perhaps you meant to update https://github.com/airbnb/enzyme/blob/master/packages/enzyme/package.json#L5, and the same line in all the subpackages in this monorepo?

votemike:
Ahh, my mistake.

Issue #2086, opened 2019-04-08
`useEffect` not called when the component is `shallow` renderered


jlandic:
### Current behavior

`useEffect` fonction does not seem to be executed when the component is rendered with `shallow`.

Given the following simple component using `useEffect` hook to call a method given as a prop:

```
import React, { useEffect } from 'react';

export const MyComponent = ({ callback }) => {
  useEffect(() => {
    callback('Say something');
  });

  return <div>Hello world!</div>;
};
```

And the following tests checking that the `callback` function is indeed called accordingly:

```
import React from 'react';
import { shallow, mount } from 'enzyme';
import { MyComponent } from '../MyComponent';

describe('MyComponent', () => {
  describe('shallow render', () => {
    it('should call callback when shallow rendered', () => {
      const callback = jest.fn();
      shallow(<MyComponent callback={callback} />);
      expect(callback).toHaveBeenCalled();
    });
  });

  describe('mount', () => {
    it('should call callback when mounted', () => {
      const callback = jest.fn();
      mount(<MyComponent callback={callback} />);
      expect(callback).toHaveBeenCalled();
    });
  });
});
```

We observe that the method is called as expected when "mounted", but not when using a `shallow` renderer:

```
 FAIL  src/tests/MyComponent.test.js
  MyComponent
    shallow render
      ✕ should call callback when shallow rendered (12ms)
    mount
      ✓ should call callback when mounted (24ms)

  ● MyComponent › shallow render › should call callback when shallow rendered

    expect(jest.fn()).toHaveBeenCalled()

    Expected mock function to have been called, but it was not called.

       8 |       const callback = jest.fn();
       9 |       shallow(<MyComponent callback={callback} />);
    > 10 |       expect(callback).toHaveBeenCalled();
         |                        ^
      11 |     });
      12 |   });
      13 |

      at Object.toHaveBeenCalled (src/tests/MyComponent.test.js:10:24)
```

You may find this [reproducible example here](https://github.com/jlandic/enzyme-shallow-hook-issue).

### Expected behavior

When using `shallow` to render the components, the function passed to `useEffect` should be executed, as it is when using `mount`.

#### API

- [x] shallow
- [ ] mount
- [ ] render

#### Version

| library             | version
| ------------------- | -------
| enzyme              | 3.9.0
| react               | 16.8.6
| react-dom           | 16.8.6
| react-test-renderer | 16.8.6
| adapter (below)     | 

#### Adapter

[...]

PR #2223, opened 2019-08-21
[fix] Allow developing without nvm/root

insidewhy:
See my comments in https://github.com/airbnb/enzyme/issues/2086

I assume that some changes happened to `lerna` which broke the `env.js` script. I assume the people developing on `enzyme` recently are all doing so with `root` access. Naughty!

Anyway, this fixes it. Now I can develop.

### DIFF
--- a/env.js
+++ b/env.js
@@ -11,7 +11,8 @@ const promisify = (fn) => new Promise((res, rej) => {
   fn(done);
 });
 const getFile = (fpath) => promisify((cb) => fs.readFile(fpath, 'utf8', cb));
-// const getFiles = fpath => promisify(cb => fs.readdir(fpath, cb));
+const unlinkFile = (fpath) => promisify(cb => fs.unlink(fpath, cb));
+// const getFiles = (fpath) => promisify(cb => fs.readdir(fpath, cb));
 const getJSON = (fpath) => getFile(fpath).then((json) => JSON.parse(json));
 const writeFile = (fpath, src) => promisify((cb) => {
   console.log('writeFile', fpath, src);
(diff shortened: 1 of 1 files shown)


env.js
```diff
@@ -138,7 +139,9 @@ Promise.resolve()
 
     return writeJSON(adapterPackageJsonPath, adapterJson, true).then(() => Promise.all([
       // npm install the peer deps at the root
-      run('npm', 'i', '--no-save', ...installs),
+      Promise.all(
+        rmrfs.map(s => primraf(s))
+      ).then(() => run('npm', 'i', '--no-save', ...installs)),
```
ohjames: The previous `npm i` in this script readds the `node_modules/npm` link... so it has to be removed before this `npm i --no-save` otherwise the same problem just happens again.

ljharb:
On the contrary; we’re all (me) using nvm, so no root access is required. I’d suggest that approach.

ohjames:
@ljharb I'd suggest just merging this so that people don't *have* to use nvm if they don't want to. I'm just about to push an improvement though, so hang on a few minutes.

ohjames:
Okay all done now.

env.js
```diff
@@ -106,14 +107,12 @@ const packagesToRemove = [
   'create-react-class',
 ].map(s => `./node_modules/${s}`);
 
-const additionalDirsToRemove = [
```
ljharb: If these aren’t both removed, then development fails in Travis, and on windows.

ohjames: @ljharb Cool, thanks. I thought it was maybe an artefact of how lerna used to work. I've added them back now.

ohjames:
Hm... travis is failing on stuff not related to this PR at all. *mega-shrug*. Also noticed the `16` version tests are broke when I run locally, and also in another recent travis build unrelated to my stuff.

ljharb:
I believe the 16.9 release broke some of our tests; I'll rebase and merge this once master is passing again.

ohjames:
@ljharb Cool thanks. I'll resume work on the hooks support once that is done. As of now, I can't get the tests to pass reliably even once I revert all my changes and use the pure github master. I tried them in the node:10, node:11 and node:12 containers so there shouldn't be anything about my system interfering here.

ljharb:
@ohjames the tests are broken on master with the release of react 16.9; all work on enzyme is blocked on fixing that. I'll merge this as soon as I can get master passing again.

ljharb:
@ohjames i've rebased this, but the tests are failing legitimately.

ljharb:
@ohjames are you no longer interested in this PR?

ljharb:
@insidewhy i'm still hoping to merge this, if you'd mind reopening it.

PR #2250, opened 2019-10-02
turn off linebreak-style

phapdinh:
Hi, this removes this error that I was seeing when I was running npm run lint in the repo
enzyme-lint


phapdinh:
@ljharb can we add a .gitattributes file to do this?

ljharb:
Since only Windows users need it, it seems better for Windows users to have their own global gitattributes file.

Issue #2249, opened 2019-10-01
Why is `.wrap()` not included in the documentation?


jpenna:
Sorry for not following the template, but it is a simple request.

Why is `.wrap()` not included in the documentation?

As you can see here #919 at least more 6 people found this method useful. Just for completion it should be in the docs. Even if a note is included against the use of it and a suggestion of how to better write the component.

The author suggested to write the PR, I put myself available to do the same.

### My use case
My use case is a HOC `<Flow>` that receives the content component `<Step>` as a prop. The component `<Container>`, which renders this HOC, renders the `<Step>` conditionally based on the current step of a specific flow. I'm using `useMemo()` to render this component inside the `<Container>` and that's why I need `.wrap()` to wrap the `<Step>` passed as a prop to `<Flow>`. 

I could use a separate function to conditionally render and return `<Step>`, but this would obscure the code and have the same behavior: this function would be wrapped in a `useMemo()` and passed as a prop to `<Flow>`.

ljharb:
Indeed, see https://github.com/airbnb/enzyme/issues/919#issuecomment-297756797 and #620.

It is intentionally undocumented.

I’m not clear tho on why you need wrap. Can you share some code?

jpenna:
#620 is unrelated, no?

This is a simplified version of the code, but the structure is the basically the same.

I used the same names as the example I gave above, so you can follow that to understand what is going on below if needed.

```js
import React, { useState, useMemo } from 'react';
import Flow from 'components/Flow';
import FlowData from 'components/FlowData';
import { statuses } from 'lib/ducks/meta';

import ChoosePathStep from './steps/ChoosePathStep';
import IssuesStep from './steps/IssuesStep';
import UploadingStep from './steps/UploadingStep';

const Container = ({ uploadStatus }) => {
  const [step, setStep] = useState(() => ChoosePathStep);

  let Step = step;
  if (uploadStatus === statuses.LOADING) {
    Step = UploadingStep;
  } else if (uploadStatus === statuses.FAILURE) {
    Step = IssuesStep;
  }

  const MountedStep = useMemo(() => (
    <Step
      me={me}
      setStep={setStep}
    />
  ), [me, setStep]);

  return (
    <>
      <FlowData
        collapseContentGutter
        rounderBorders
        editable
      />
      <Flow
        blankContent={MountedStep}
      />
    </>
  );
};

export default Container;

```

ljharb:
sorry, i meant #920 :-)

[...]

PR #2255, opened 2019-10-05
Add wrapProp() method

jpenna:
#2249

### DIFF
--- a/docs/api/ReactWrapper/wrapProp.md
+++ b/docs/api/ReactWrapper/wrapProp.md
@@ -1,0 +1,60 @@
+# `.wrapProp(propName[, options]) => ReactWrapper`
+
+Returns a new wrapper around the component provided to the original wrapper's prop `propName`.`
+
+#### Arguments
+
+1. `propName` (`String`): Name of the prop to be wrapped
+1. `options` (`Object` [optional]): Will be passed to the renderer constructor.
+   Refer to the [`mount()` options](https://enzymejs.github.io/enzyme/docs/api/mount.html#arguments).
+
+This essentially does:
+
+```jsx
+const Node = () => wrapper.prop('node');
+const node = mount(<Node />);
+```
+
+#### Returns
+
+`ReactWrapper`: A new wrapper that wraps the node from the provided prop.
+
+#### Examples
+
+##### Test Setup
+
+```jsx
+class Inner extends React.Component {
+  render() {
+    return <div className="inner" />;
+  }
+}
+
+class Outer extends React.Component {
+  render() {
+    if (!this.props.renderNode) return <div />;
+    return this.props.node;
+  }
+}
+
+class Container extends React.Component {
+  render() {
+    /*
+     * Just as an example, <Outer> can render or not the provided prop.
+     * Independent of what it does, you want to test the component given to node.
+    */
+    return <Outer renderNode={false} node={<Inner />} />;
+  }
+}
+```
+
+##### Testing with no arguments
+
+```jsx
+const wrapper = mount(<Container />)
+  .find(Outer)
+  .wrapProp('node');
+
+expect(wrapper.find('div').equals(<div className="inner" />)).to.equal(true);
+expect(wrapper.html()).to.equal('<div class="inner"></div>');
+```
--- a/packages/enzyme-test-suite/test/shared/methods/wrapProp.jsx
+++ b/packages/enzyme-test-suite/test/shared/methods/wrapProp.jsx
@@ -1,0 +1,71 @@
+import React from 'react';
+import { expect } from 'chai';
+import wrap from 'mocha-wrap';
+
+import getAdapter from 'enzyme/build/getAdapter';
+
+export default function describeWrapProp({
+  Wrap,
+  WrapRendered,
+  WrapperName,
+  isShallow,
+}) {
+  wrap()
+    .withConsoleThrows()
+    .describe('.wrapProp()', () => {
+      class Inner extends React.Component {
+        render() {
+          return <div className="inner" />;
+        }
+      }
+      class Outer extends React.Component {
+        render() {
+          return <div />;
+        }
+      }
+      class Container extends React.Component {
+        render() {
+          const node = this.props.replace || <Inner />;
+          return <Outer node={node} />;
+        }
+      }
+
+      it('returns a wrapper around the node provided by the given prop', () => {
+        const wrapper = Wrap(<Container />);
+        const wrappedPropWrapper = wrapper.find(Outer).wrapProp('node');
+        expect(wrappedPropWrapper.find('div').equals(<div className="inner" />)).to.equal(true);
+        if (isShallow) expect(wrappedPropWrapper.equals(<div className="inner" />)).to.equal(true);
+      });
+
+      it('throws on a non-string prop name', () => {
+        const wrapper = Wrap(<Container />);
+        expect(() => wrapper.find(Outer).wrapProp([])).to.throw(
+          TypeError,
+          `${WrapperName}::wrapProp(): \`propName\` must be a string`,
+        );
+      });
+
+      it('throws on a missing prop', () => {
+        const wrapper = Wrap(<Container />);
+        expect(() => wrapper.find(Outer).wrapProp('missing')).to.throw(
+          Error,
+          `${WrapperName}::wrapProp(): no prop called "missing" found`,
+        );
+      });
+
+      it('throws on an invalid element prop value', () => {
+        const wrapper = Wrap(<Container replace={() => <div />} />);
+        expect(() => wrapper.find(Outer).wrapProp('node')).to.throw(
+          TypeError,
+          `${WrapperName}::wrapProp(): prop "node" does not contain a valid element`,
+        );
+      });
+
+      wrap()
+        .withOverride(() => getAdapter(), 'wrap', () => undefined)
+        .it('throws with a react adapter that lacks a `.wrap`', () => {
+          const wrapper = Wrap(<Container />);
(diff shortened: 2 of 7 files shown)


jpenna:
I'm thinking if it would make sense to have `wrapProp` receive an optional "wrapper" method, so a `shallow` wrapped component could require a `mount` wrapped component and vice versa. What do you think? @ljharb

docs/api/ReactWrapper/wrapProp.md
```diff
@@ -0,0 +1,60 @@
+# `.wrapProp(propName) => ReactWrapper`
+
+Returns a new wrapper around the component provided to the original wrapper's prop `propName`.`
+
+NOTE: can only be called on wrapper of a single non-DOM component element node.
```
ljharb: this doesn't seem entirely necessary to me - obviously it'd be weird to have a prop that's a react element on a DOM node, but why restrict it?

packages/enzyme-test-suite/test/shared/methods/wrapProp.jsx
```diff
@@ -0,0 +1,84 @@
...
+        render() {
+          const node = this.props.replace || <Inner />;
+          return <Outer node={node} />;
+        }
+      }
+
+      it('returns a wrapper around the node provided by the given prop', () => {
+        const wrapper = Wrap(<Container />);
+        const wrappedPropWrapper = wrapper.find(Outer).wrapProp('node');
+        expect(wrappedPropWrapper.find('div').equals(<div className="inner" />)).to.be.true;
```
ljharb:
```suggestion
        expect(wrappedPropWrapper.find('div').equals(<div className="inner" />)).to.equal(true);
```
noop matchers should be avoided.

packages/enzyme-test-suite/test/shared/methods/wrapProp.jsx
```diff
@@ -0,0 +1,84 @@
...
+        if (isShallow) expect(wrappedPropWrapper.equals(<div className="inner" />)).to.be.true;
```
ljharb:
```suggestion
        if (isShallow) expect(wrappedPropWrapper.equals(<div className="inner" />)).to.equal(true);
```

packages/enzyme/src/ReactWrapper.js
```diff
@@ -886,6 +886,38 @@ class ReactWrapper {
...
+      const props = this.props();
+      if (!has(props, propName)) {
+        throw new Error(`ReactWrapper::wrapProp(): no prop called "${propName}" found`);
+      }
+      const node = props[propName];
+      if (!adapter.isValidElement(node)) {
+        throw new TypeError(`ReactWrapper::wrapProp(): prop "${propName}" does not contain a valid element`);
+      }
+
+      return new ReactWrapper(node);
```
ljharb: what about the options? i'd expect `wrapProp` to take optional `options`.

docs/api/ReactWrapper/wrapProp.md
jpenna: I agree, did it like this because I thought it was a "convention". Will do the change

ljharb: Only on the methods where it makes no sense to call it on a DOM element :-)

packages/enzyme/src/ReactWrapper.js
jpenna: What kind of `options`? 

If you mean `props`, I didn't see how to use them, because you will have a mounted react element as a `prop`, and that's why you want to wrap it, else you use `renderProp`. In this case doesn't make sense to have `props` for the wrapped component.

ljharb: Not props, constructor options.

jpenna: Ah ok! I'll include those

packages/enzyme-test-suite/test/shared/methods/wrapProp.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+      it('throws on an invalid element prop value', () => {
+        const wrapper = Wrap(<Container replace={() => <div />} />);
+        expect(() => wrapper.find(Outer).wrapProp('node')).to.throw(
+          TypeError,
+          `${WrapperName}::wrapProp(): prop "node" does not contain a valid element`,
+        );
+      });
+
+      it('throws on host elements', () => {
```
ljharb: is this test still valid?

jpenna: True. The tests were passing and I had to `build` the files.
I removed the test. There is an error warning because I'm passing a prop to a DOM component and this is a bad practice (https://reactjs.org/warnings/unknown-prop.html).

jpenna:
@ljharb The test error is not related to my PR. Could you rerun it, please?

jpenna:
@ljharb what are these errors?

ljharb:
@jpenna they're linting errors. in this case, within your PR's docs themselves.

jpenna:
Yeah.. It requires functional examples, I'll change them

jpenna:
@ljharb hook tests are failing. Travis checks are failing because of some change of mine?

jpenna:
@ljharb Could you give me some direction?

brybrophy:
Is this PR still being worked on, or has it been abandoned?

jpenna:
I'm still around

GreenGremlin:
Once this is merged, someone will need to update [@types/enzyme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/enzyme/index.d.ts). Is there a process for that?

ljharb:
@GreenGremlin nope, i didn't create them and i don't maintain them (but i'm a reviewer for them if someone puts up a PR)

docs/api/ReactWrapper/wrapProp.md
```diff
@@ -0,0 +1,60 @@
...
+
+##### Testing with no arguments
+
+```jsx
+const wrapper = mount(<Container />)
+  .find(Outer)
+  .wrapProp('node');
+
+expect(wrapper.find('div').equals(<div className="inner" />)).to.equal(true);
+expect(wrapper.html()).to.equal('<div class="inner"></div>');
```
ljharb: let's use `.debug` in these examples, instead of `.html`

ljharb:
It's an interesting idea to produce a shallow wrapper from a mount wrapper or vice versa.

Given that shallow and mount can have different options, how would you suggest doing that?

jpenna:
@ljharb What I was thinking is passing an option with the wrapper method we want to use, this wouldn't bring any dependency loop issues and it's quite clear. Could it have any side effects? 

The signature would be like: 

```js
import { shallow, mount } from 'enzyme';
const wrapper = shallow(<Component />);
wrapper.wrapProp('renderedButton', { wrapper: mount });
```

And the other way around for `mount`. If it is passed `shallow` to a shallow wrapper, then it should work the same as if the prop wasn't passed at all.

ljharb:
@jpenna not a bad idea, but it raises a few questions. Can they pass `render`? What if they pass a *different* function than shallow/mount/render, what happens? What about the options (second argument) to shallow/mount? For example, `mount` may need `attachTo`, but the original shallow wrapper wouldn't have that option.

jpenna:
@ljharb Good points. 
`render` should work too, it wouldn't make sense to have only `mount` and `shallow` indeed.

We could have 2 props for this, similar to `wrappingComponent` and `wrappingComponentProps` approach. Would that work? `wrapper` and `wrapperProps`? Props validation would happen in the wrapper directly, and we don't "propagate" props, so they have to be passed every time you use this method.

A different function than mount/shallow/render could error out or give a warning. I'm thinking about the warning in case they want to use a custom function/another lib to build a component, just for extensibility, but I don't know if this would make sense...

ljharb:
Another alternative would be to add a `prop` option to `.render`/`.shallow`/`.mount`, but i do like `wrapProp` :-/

jpenna:
Yeah.. I don't think it is as straightforward as the method call though. Do you have concerns on using a `wrapper` prop with the `wrapProp` method?

Now that I'm thinking about it, the `wrapProp` method could replace `renderProp`, no?

ljharb:
`renderProp` is for a prop that's a function that produces jsx - `wrapProp` is for a prop that *is* the jsx. I think it's important to keep these separate, for clarity.

ljharb:
@jpenna are you no longer interested in this PR? Despite no decision having been made yet, i think it's still useful to have this open.

jpenna:
@ljharb I thought we wouldn't work on this anymore, sorry. I'm using `react-testing-library`, so I'm not very much willing to put my time on `enzyme` at the time being, thank you for you support on this PR though!

jpenna:
@ljharb I'm transfering this fork to you, so you have better control over it's life cycle

jpenna:
@ljharb I'm cleaning up the list of PRs I have opened and there isn't anything for me to do here anymore. Can we close this PR and maybe you create a new one referencing it? Or is there another option? Thanks

jpenna:
@ljharb Can you create a new PR based on this one please? I want to delete this fork

ljharb:
@jpenna Once the fork is deleted, the PR is unrecoverable, and there's not much point opening a new PR (and permanently adding a new ref to the repo) until someone's available to work on it. I'd very much prefer to leave it open, but that's up to you.

jpenna:
Ah, didn't know that. I won't delete the repo then, but can you create another PR referencing this one?

ljharb:
Yes, if its presence in your "pulls" list is that problematic - but i'd have to reopen it every time i needed to update it, and i'd need to keep both PRs in sync to avoid polluting the repo with an excess ref :-/

PR #2259, opened 2019-10-10
`mount`: ensure the root’s `ref` prop gets attached to the actual root

### LINKED ISSUE #2253: Ref incorrectly set for HTMLElement
### Current behavior
In react,
Refs for node elements are the underlying DOM HTMLElement, 
Refs for Components are the Component instance. 

When using Enzyme and mounting an html node element, the ref is not an HTMLElement

### Expected behavior
The expected behaviour is to be on par with React, it shouldn't behave differently, the ref should be the element itself.

### Your environment
Node 12.4
Mac OSx Sierra

#### API
- mount

#### Version

| enzyme              | 3.8.0
| react               | 16.8.2
| react-dom           | 16.8.2
| react-test-renderer | 
| adapter (below)     | 1.9.1

#### Adapter
- enzyme-adapter-react-16

```
    mount(<div  
      ref={(el) => {  
        console.log(el);  
        el.removeEventListener('click'); // exception, not a function  
      }}  
    >  
    </div>);  
```

ljharb:
Fixes #2253.

After merging, this will need adapter-utils bumped and published, and then every adapter's minimum utils version bumped, and then they'll also need a publish.

### DIFF
--- a/packages/enzyme-adapter-react-14/src/ReactFourteenAdapter.js
+++ b/packages/enzyme-adapter-react-14/src/ReactFourteenAdapter.js
@@ -110,7 +110,7 @@ class ReactFourteenAdapter extends EnzymeAdapter {
             wrappingComponentProps: options.wrappingComponentProps,
             props,
             context,
-            ...(ref && { ref }),
+            ...(ref && { refProp: ref }),
           };
           const ReactWrapperComponent = createMountWrapper(el, { ...options, adapter });
           const wrappedEl = React.createElement(ReactWrapperComponent, wrapperProps);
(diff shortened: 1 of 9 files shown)


eps1lon:
Isn't this exactly what I wanted to do in #2101 originally which wasn't deemed backwards compat safe?

ljharb:
@eps1lon you're right that it's trying to solve the same problem, and in a very similar way - i believe I did it in a backwards compatible way here, and I admit I forgot #2101 existed when working on this :-/

However, #2101 may be worth rebasing to better handle React 0.13, if you're willing.

eps1lon:
> However, #2101 may be worth rebasing to better handle React 0.13, if you're willing.

I won't have the time for that. But thanks for releasing this. Allowed us to remove a couple of seemingly useless fragments :+1:

PR #2262, opened 2019-10-16
Fix typos

mvasin:


packages/enzyme/src/ShallowWrapper.js
```diff
@@ -933,7 +933,7 @@ class ShallowWrapper {
   }
 
   /**
-   * Whether or not one of the given react elements exists in the shallow render tree.
+   * Whether or not one of the given react elements exist in the shallow render tree.
```
ljharb: this one is correct as-is; "one … exists"

mvasin: Indeed.

mvasin:
@ljharb Please have a look again.

PR #2296, opened 2019-11-26
Inherit root context information in ShallowWrapper::shallow()

### LINKED ISSUE #2290: ShallowWrapper::dive fails when diving into a memoized component inside a context provider
### Current behavior
Calling
```jsx
const SampleContextA = React.createContext<T>(true);

const MyComponent = () => <div />;
const MyComponentMemoized = React.memo(MyComponent);

enzyme
  .shallow(
    <SampleContextA.Provider value={false}>
      <MyComponentMemoized />
    </SampleContextA.Provider>
  )
  .dive();
```
throws with `TypeError: ShallowWrapper::dive() can only be called on components`. If you replace the provider wrapper with just a plain wrapper, e.g.
```jss
const Wrapper = (props) => React.Children.only(props.children);
```
it renders fine. It also works fine in you use MyComponent unmemoized.

### Expected behavior
This code should return a ShallowWrapper to the contents of `<MyComponent />` instead of throwing.

#### API

- [X] shallow
- [ ] mount
- [ ] render

#### Version

| enzyme              |  3.10.0
| react               |  16.12.0
| react-dom           | 16.12.0
| react-test-renderer |  16.12.0
| adapter (below)     | 1.15.1

#### Adapter

- [X] enzyme-adapter-react-16
- [ ] others ( ) 

### Notes
I believe this is similar to but distinct from #2103. That issue addressed memoized components being wrapped by regular custom components.

moward:
Fixes #2290

### DIFF
--- a/packages/enzyme/src/ShallowWrapper.js
+++ b/packages/enzyme/src/ShallowWrapper.js
@@ -369,6 +369,20 @@ function makeShallowOptions(nodes, root, passedOptions, wrapper) {
 }
 
 
+function makeInheritedChildOptions(wrapper, options = {}) {
+  const childOptions = {
+    ...wrapper[OPTIONS],
+    ...options,
+    context: options.context || {
+      ...wrapper[OPTIONS].context,
+      ...wrapper[ROOT][CHILD_CONTEXT],
+    },
+  };
+  privateSet(childOptions, PROVIDER_VALUES, wrapper[ROOT][PROVIDER_VALUES]);
+  return childOptions;
+}
+
+
 /**
  * @class ShallowWrapper
  */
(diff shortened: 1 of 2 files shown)


moward:
@ljharb FYI

moward:
@ljharb could you please take a look? 😁

packages/enzyme/src/ShallowWrapper.js
```diff
@@ -1683,7 +1698,7 @@ class ShallowWrapper {
    * @param {Object} options
    * @returns {ShallowWrapper}
    */
-  dive(options = {}) {
+  dive(options) {
```
ljharb: the options are optional, so the default value is correct (this change would change the length of the function)
```suggestion
  dive(options = {}) {
```

Issue #2200, opened 2019-07-22
React.Lazy always returning the Fallback value


philearley:
### Current behaviour

When writing a test to validate my component is rendered, I am only ever seeing the fallback value.
I am unsure as to what I am doing incorrect or if this is an issue.

```
import React, { Suspense } from 'react';

const LazyComponent = React.lazy(() => import('./lazy-component-file'));

const ComponentWrapper = () => {
  return (
    <Suspense fallback='...'>
      <LazyComponent />
    </Suspense>
  );
};

export default ComponentWrapper;

....

import { shallow } from 'enzyme';
import React from 'react';

import ComponentWrapper from './component-wrapper';

test('renders the suspense component', () => {
  const wrapper = shallow(<ComponentWrapper />);
  expect(wrapper.find('LazyComponent').length).toEqual(0); // Passes
});

test('renders the LazyComponent', () => {
  const wrapper = shallow(<ComponentWrapper />, { suspenseFallback: false });
  expect(wrapper.find('LazyComponent').length).toEqual(1); // Errors
});

```

### Expected behaviour

I've tried exporting the `LazyComponent` and doing:
`expect(wrapper.find(LazyComponent).length).toEqual(1);`

But my issue is flipped on its head and `wrapper.find(LazyComponent).length` is always 1.

#### API

- [x] shallow
- [ ] mount
- [ ] render

#### Version

| library             | version
| ------------------- | -------
| enzyme              | 3.10.0
| react               | 16.8.6
| react-dom           | 16.8.6
| react-test-renderer | 16.8.6
| adapter (below)     | 1.14.0

#### Adapter

- [x] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-16.3
- [ ] enzyme-adapter-react-16.2
- [ ] enzyme-adapter-react-16.1
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-15.4
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] enzyme-adapter-react-helper
- [ ] others ( )

ljharb:
By default, `Lazy` in `shallow` renders the child, but if you pass `{ suspenseFallback: true }` as the second argument to `shallow`, it should render the fallback. 

You can get more info by using `wrapper.debug()`.

philearley:
Yeah I was trying to use `wrapper.debug()` to see the output.

The suspenseFallback option seems to not be making a difference to what is output.

```
[...]

PR #2328, opened 2020-02-08
Fix suspense fallback is invalid

lamhieu-vk:
Resolved https://github.com/airbnb/enzyme/issues/2200

I updated `enzyme-adapter-react-16` to fix issue suspense fallback is invalid. I handled to always wrap `React.Suspense` component outside of `React.Lazy` and `React.Fallback` components in any case.

I passed all tests in master branch and local test by below code:
```js
test('renders suspenseFallback=true', () => {
  const el = <ComponentWrapper />;
  const wrapper = shallow(el, { suspenseFallback: true });
  expect(wrapper.find(Fallback).length).toEqual(1);
  expect(wrapper.find(LazyComponent).length).toEqual(0);
});

test('renders suspenseFallback=false', () => {
  const el = <ComponentWrapper />;
  const wrapper = shallow(el, { suspenseFallback: false });
  expect(wrapper.find(Fallback).length).toEqual(0);
  expect(wrapper.find(LazyComponent).length).toEqual(1);
});
```
Code cloned from issue to reproduce, it's failed in master branch and passed in this branch.

Note: Lazy component will always display as `<lazy />` in debug and If you want to expect this component, should be use `.find(Component)` instead of `.find('ComponentName')`.

Thanks for review!

### DIFF
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -120,6 +120,22 @@ function unmemoType(type) {
   return isMemo(type) ? type.type : type;
 }
 
+function checkIsSuspenseAndCloneElement(el, { suspenseFallback }) {
+  if (!isSuspense(el)) {
+    return el;
+  }
+
+  let { children } = el.props;
+
+  if (suspenseFallback) {
+    const { fallback } = el.props;
+    children = replaceLazyWithFallback(children, fallback);
+  }
+
+  const FakeSuspenseWrapper = (props) => React.createElement(el.type, { ...el.props, ...props }, children);
+  return React.createElement(FakeSuspenseWrapper, null, children);
+}
+
 function elementToTree(el) {
   if (!isPortal(el)) {
     return utilElementToTree(el, elementToTree);
@@ -588,6 +604,22 @@ class ReactSixteenAdapter extends EnzymeAdapter {
       return wrappedComponent;
     };
 
+    const renderElement = (elConfig, ...rest) => {
+      const renderedEl = renderer.render(elConfig, ...rest);
+
+      const typeIsExisted = !!(renderedEl && renderedEl.type);
+      if (is166 && typeIsExisted) {
+        const clonedEl = checkIsSuspenseAndCloneElement(renderedEl, { suspenseFallback });
+
+        const elementIsChanged = clonedEl.type !== renderedEl.type;
+        if (elementIsChanged) {
+          return renderer.render({ ...elConfig, type: clonedEl.type }, ...rest);
+        }
+      }
+
+      return renderedEl;
+    };
+
     return {
       render(el, unmaskedContext, {
         providerValues = new Map(),
@@ -602,7 +634,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
             (props) => props.children,
             el.type,
           );
-          return withSetStateAllowed(() => renderer.render({ ...el, type: MockProvider }));
+          return withSetStateAllowed(() => renderElement({ ...el, type: MockProvider }));
         } else if (isContextConsumer(el)) {
           const Provider = adapter.getProviderFromConsumer(el.type);
           const value = providerValues.has(Provider)
@@ -612,22 +644,15 @@ class ReactSixteenAdapter extends EnzymeAdapter {
             (props) => props.children(value),
             el.type,
           );
-          return withSetStateAllowed(() => renderer.render({ ...el, type: MockConsumer }));
+          return withSetStateAllowed(() => renderElement({ ...el, type: MockConsumer }));
         } else {
           isDOM = false;
           let renderedEl = el;
           if (isLazy(renderedEl)) {
             throw TypeError('`React.lazy` is not supported by shallow rendering.');
           }
-          if (isSuspense(renderedEl)) {
-            let { children } = renderedEl.props;
-            if (suspenseFallback) {
-              const { fallback } = renderedEl.props;
-              children = replaceLazyWithFallback(children, fallback);
-            }
-            const FakeSuspenseWrapper = () => children;
-            renderedEl = React.createElement(FakeSuspenseWrapper, null, children);
-          }
+
+          renderedEl = checkIsSuspenseAndCloneElement(renderedEl, { suspenseFallback });
           const { type: Component } = renderedEl;
 
           const context = getMaskedContext(Component.contextTypes, unmaskedContext);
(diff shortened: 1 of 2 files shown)


packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -120,6 +120,26 @@ function unmemoType(type) {
...
 
+function transformSuspenseElement(el, { suspenseFallback }) {
+  let { children } = el.props;
+
+  if (suspenseFallback) {
+    const { fallback } = el.props;
+    children = replaceLazyWithFallback(children, fallback);
+  }
+
+  const FakeSuspenseWrapper = (props) => React.createElement(el.type, { ...el.props, ...props }, children);
```
ljharb: this seems like it might break class components that use private fields, or functional components that use hooks.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -256,8 +276,9 @@ function toTree(vnode) {
         rendered: childrenToTree(node.child),
       };
     }
-    case FiberTags.Lazy:
+    case FiberTags.Lazy: {
```
ljharb: the curlies are only needed when there's block-scoped variables inside the case; let's revert this.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -528,7 +549,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
       }
       if (lastComponent !== Component) {
         if (isStateful(Component)) {
-          wrappedComponent = class extends Component {}; // eslint-disable-line react/prefer-stateless-function
+          wrappedComponent = class extends Component { }; // eslint-disable-line react/prefer-stateless-function
```
ljharb: please revert unrelated formatting changes
```suggestion
          wrappedComponent = class extends Component {}; // eslint-disable-line react/prefer-stateless-function
```

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -588,6 +609,22 @@ class ReactSixteenAdapter extends EnzymeAdapter {
       return wrappedComponent;
     };
 
+    const renderElement = (elConfig, ...rest) => {
+      const renderedEl = renderer.render(elConfig, ...rest);
+
+      const typeIsExisted = !!(renderedEl && renderedEl.type);
```
ljharb: when would this not be true?

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -1722,7 +1722,7 @@ describe('shallow', () => {
     });
 
     it('finds LazyComponent when render component wrapping lazy component', () => {
-      const LazyComponent = lazy(fakeDynamicImport(DynamicComponent));
+      const LazyComponent = lazy(() => fakeDynamicImport(DynamicComponent));
```
ljharb: why are these test changes needed?

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
lamhieu-vk: I reverted it!

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
lamhieu-vk: I reverted it!

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
lamhieu-vk: Yep, sometimes `type` in Element is `undefined ` with `$$typeof` is `(react.element)`. I'm not sure I know which cases it can be `undefined`.

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
lamhieu-vk: I changes to like user cases example in [react docs](https://reactjs.org/docs/code-splitting.html#reactlazy).

```js
const OtherComponent = React.lazy(() => import('./OtherComponent'));
```

> React.lazy takes a function that must call a dynamic import(). This must return a Promise which resolves to a module with a default export containing a React component.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
lamhieu-vk: In all cases `el.type` is `react.Suspense`, the wrapper of `lazy` and `fallback` inside. I just create it after changed children inside.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
ljharb: We've got a bunch of existing helpers for this kind of thing; is there one of those we could use here?

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
ljharb: Perhaps it would be cleaner to inline `transformSuspenseElement`, since it's only used once, and then it couldn't possibly be invoked when `el` isn't a Suspense element?

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
ljharb: I'm concerned about changing existing tests; instead, let's leave these unchanged, and also *add* test cases with the thunk?

lamhieu-vk: I viewed `lazy` test of react [here](https://github.com/facebook/react/blob/b979db4e7215957f03c4221622f0b115a868439a/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js#L37). The old test is not correct with user case of react `lazy` but I don't know it still passed. However, I think we should update it!

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
lamhieu-vk: I don't see any helper to check is existed type in element. Do you want to me create a helper to use it?

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
lamhieu-vk: I updated it, make inline `transformSuspenseElement` function. Thanks for feedback!

lamhieu-vk:
hi @ljharb , could you please give me a feedback?

lamhieu-vk:
hi @ljharb, PR is ready merge!

cascornelissen:
Looks like this change broke the `.instance()` method on a `shallow()` rendered component for us when the root component is a `React.Suspense`.

```
const wrapper = shallow(<ComponentWithReactSuspense />);
wrapper.instance(); // Always null
```

Any ideas?

PR #2329, opened 2020-02-08
Format all `.babelrc` files to off warning

lamhieu-vk:
💃

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,10 +1,8 @@
 {
   "presets": ["airbnb"],
   "plugins": [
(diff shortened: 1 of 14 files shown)


lamhieu-vk:
hi @ljharb , I know files don't have to be strictly json but I think bringing it back to json format will make it look better 💃 If you think we don't need it, please close PR ^^.

ljharb:
In general, aesthetic preferences are up to the maintainers of a project, and aren't really appropriate things for contributors to suggest. Thanks for the PR!

PR #2447, opened 2020-09-16
Improve README remove duplicated parentheses

OtacilioN:
Is there a particular reason for using duplicate parentheses in the documentation? Or it is just a minor typo?

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,11 +135,11 @@ describe('<MyComponent />', () => {
   });
 
   it('renders children when passed in', () => {
-    const wrapper = shallow((
+    const wrapper = shallow(
       <MyComponent>
         <div className="unique" />
       </MyComponent>
-    ));
+    );
     expect(wrapper.contains(<div className="unique" />)).to.equal(true);
   });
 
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
It's quite intentional, since the "trailing comma" rule requires a trailing comma when they're omitted, but when they're included, the single jsx argument is treated as a single token, and no trailing comma is required.

OtacilioN:
I don't know if I got the reason about using duplicated parentheses, but it can be a little bit confusing, specially to beginners. 

That's no possibly to maintain the trailing coma and also remove the internal parentheses?

ljharb:
Not without violating the lint rule.

OtacilioN:
Okay got it, sorry for the pull request, to me this duplicated parentheses was very confusing.

Since it's intentional, can I close the PR? Or can we change de lint rule?

ljharb:
You'd have to change the rule inside eslint core itself first (which would be great). I think the PR can be closed, yes.

Issue #2367, opened 2020-04-07
Please some information about find() as a snapshot to docs


Adrianjewell91:
### Current behavior

Hello, I would simply like to request that the documentation for ```find()``` be updated to reflect its snapshot-nature. For example, in a situation where clicking a button updates state under the hood... 

```
   // some set up code goes here. 

    const deleteButton = wrapper.find('[data-enzyme-id="someid"]').first();
    deleteButton.simulate('click');

    expect(wrapper.find("#modal").first().props().isOpen).toBe(true)

```

the above code will work, but the below will not: 

```
    const modals = wrapper.find("#modal");
    const deleteButton = wrapper.find('[data-enzyme-id="someid"]').first();
    deleteButton.simulate('click');

    expect(modals.first().props().isOpen).toBe(true)

```

Presumably, this is because the objects that are returned by find() do not reflect the actual DOM tree, but a snapshot thereof. 

It would be great to see this mentioned briefly in the documentation.  Please suggest a way to explain this in a documentation-appropriate manner. Thank you!

ljharb:
This applies not just to find, but to every enzyme API - you always have to re-traverse from the root.

It's described here: https://github.com/enzymejs/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#element-referential-identity-is-no-longer-preserved but it might be useful to summarize and link to it on all the relevant methods' docs.

Adrianjewell91:
Thank you for the reference!

cadence09:
I found the example for Calling props() after a state change in  [is](https://github.com/enzymejs/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#element-referential-identity-is-no-longer-preserved)  is incorrect. The first test statement should fail in v3, but it passed. I wonder would that be caused by the mount() method is full DOM rendering.

cadence09:
I want to work on this issue and summarize it and link it to all relevant methods.

ljharb:
@cadence09 thanks, that would be great!

It's entirely possible the migration guide, written for v3.0.0, has become incorrect over time; it's important we identify what may have inadvertently broken vs what was a bug that's been fixed :-)

cadence09:
[...]

PR #2450, opened 2020-09-20
Update the API docs

cadence09:
Regarding the conversation from #2367 , I open the PR for the changed I made for refining some text, typo errors, and failing tests.

### DIFF
--- a/docs/api/ReactWrapper/equals.md
+++ b/docs/api/ReactWrapper/equals.md
@@ -21,7 +21,11 @@ like the one passed in.
 
 
 ```jsx
-const wrapper = mount(<MyComponent />);
+function MyComponent() {
+  return <div className="foo bar" />;
+}
+
+const wrapper = mount(<MyComponent />).childAt(0);
 expect(wrapper.equals(<div className="foo bar" />)).to.equal(true);
 ```
 
--- a/docs/api/ReactWrapper/render.md
+++ b/docs/api/ReactWrapper/render.md
@@ -29,5 +29,6 @@ function Bar() {
 
 ```jsx
 const wrapper = mount(<Bar />);
-expect(wrapper.find(Foo).render().find('.in-foo')).to.have.lengthOf(1);
+expect(wrapper.find('.in-foo')).to.have.lengthOf(1);
+expect(wrapper.render().find('.in-foo')).to.have.lengthOf(1);
 ```
--- a/docs/api/ReactWrapper/simulate.md
+++ b/docs/api/ReactWrapper/simulate.md
@@ -71,12 +71,12 @@ const wrapper = mount((
   />
 ));
 
-expect(wrapper.find('input').at(0).prop('value')).toEqual(10);
-expect(wrapper.find('input').at(1).prop('value')).toEqual(20);
+expect(wrapper.find('input').at(0).props().value).to.equal(10);
+expect(wrapper.find('input').at(1).props().value).to.equal(20);
 wrapper.find('input').at(0).simulate('change', { target: { name: 'width', value: 50 } });
 wrapper.find('input').at(1).simulate('change', { target: { name: 'height', value: 70 } });
-expect(testState.width).toEqual(50);
-expect(testState.height).toEqual(70);
+expect(testState.width).to.equal(50);
+expect(testState.height).to.equal(70);
 ```
 
 #### Common Gotchas
--- a/docs/api/ReactWrapper/type.md
+++ b/docs/api/ReactWrapper/type.md
@@ -17,7 +17,7 @@ If it's `null`, it will be `null`. It must be a single-node wrapper.
 function Foo() {
   return <div />;
 }
-const wrapper = mount(<Foo />);
+const wrapper = mount(<Foo />).childAt(0);
 expect(wrapper.type()).to.equal('div');
 ```
 
@@ -38,7 +38,8 @@ function Foo() {
   return <Bar />;
 }
 const wrapper = mount(<Foo />);
-expect(wrapper.type()).to.equal(Bar);
+expect(wrapper.type()).to.equal(Foo);
+expect(wrapper.childAt(0).type()).to.equal(Bar);
 ```
 
 ```jsx
--- a/docs/api/ShallowWrapper/render.md
+++ b/docs/api/ShallowWrapper/render.md
@@ -30,5 +30,5 @@ function Bar() {
 ```jsx
 const wrapper = shallow(<Bar />);
 expect(wrapper.find('.in-foo')).to.have.lengthOf(0);
-expect(wrapper.find(Foo).render().find('.in-foo')).to.have.lengthOf(1);
+expect(wrapper.render().find('.in-foo')).to.have.lengthOf(1);
 ```
--- a/docs/api/ReactWrapper/every.md
+++ b/docs/api/ReactWrapper/every.md
@@ -33,5 +33,4 @@ expect(wrapper.find('.foo').every('.bar')).to.equal(false);
 #### Related Methods
 
 - [`.someWhere(predicate) => Boolean`](someWhere.md)
-- [`.every(selector) => Boolean`](every.md)
 - [`.everyWhere(predicate) => Boolean`](everyWhere.md)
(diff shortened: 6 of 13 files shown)


docs/api/ReactWrapper/equals.md
```diff
@@ -22,7 +22,7 @@ like the one passed in.
 
 ```jsx
 const wrapper = mount(<MyComponent />);
-expect(wrapper.equals(<div className="foo bar" />)).to.equal(true);
+expect(wrapper.find("div").equals(<div className="foo bar" />)).to.equal(true);
```
ljharb, line 25:
```suggestion
expect(wrapper.find('div').equals(<div className="foo bar" />)).to.equal(true);
```

docs/api/ReactWrapper/render.md
```diff
@@ -29,5 +29,5 @@ function Bar() {
 
 ```jsx
 const wrapper = mount(<Bar />);
-expect(wrapper.find(Foo).render().find('.in-foo')).to.have.lengthOf(1);
```
ljharb, line 32: i'm not sure about this one - it seems useful to show that you can call `.render()` on any sub-wrapper.

docs/api/ReactWrapper/simulate.md
```diff
@@ -71,12 +71,12 @@ const wrapper = mount((
   />
 ));
 
-expect(wrapper.find('input').at(0).prop('value')).toEqual(10);
-expect(wrapper.find('input').at(1).prop('value')).toEqual(20);
+expect(wrapper.find('input').at(0).props().value).to.equal(10);
+expect(wrapper.find('input').at(1).props().value).to.equal(20);
```
ljharb, line 75: the change from `toEqual` to `.to.equal` is good; i'm not sure about whether it's better to use `.props().x` or `.prop('x')` here. is there a reason you made the choice?

docs/api/ReactWrapper/type.md
```diff
@@ -18,7 +18,7 @@ function Foo() {
   return <div />;
 }
 const wrapper = mount(<Foo />);
-expect(wrapper.type()).to.equal('div');
+expect(wrapper.find("div").type()).to.equal('div');
```
ljharb, line 21:
```suggestion
expect(wrapper.find('div').type()).to.equal('div');
```

ljharb: altho this might be more useful:
```suggestion
expect(wrapper.type()).to.equal(Foo);
expect(wrapper.find('div').type()).to.equal('div');
```

docs/api/ReactWrapper/type.md
```diff
@@ -38,7 +38,7 @@ function Foo() {
   return <Bar />;
 }
 const wrapper = mount(<Foo />);
-expect(wrapper.type()).to.equal(Bar);
+expect(wrapper.find(Bar).type()).to.equal(Bar);
```
ljharb, line 41:
```suggestion
expect(wrapper.type()).to.equal(Foo);
expect(wrapper.find(Bar).type()).to.equal(Bar);
```

docs/api/ShallowWrapper/render.md
```diff
@@ -30,5 +30,5 @@ function Bar() {
 ```jsx
 const wrapper = shallow(<Bar />);
 expect(wrapper.find('.in-foo')).to.have.lengthOf(0);
-expect(wrapper.find(Foo).render().find('.in-foo')).to.have.lengthOf(1);
+expect(wrapper.render().find('.in-foo')).to.have.lengthOf(1);
```
ljharb, line 33: same question here

ljharb (changes requested):
Thanks, this looks great! Just a few comments.

docs/api/ReactWrapper/render.md
cadence09: `.render() `can call on any sub-wrapper, but `.render() `seems doesn't return the Bar component's highest `<div> `as well as `Foo `component's. I used `HTML() `on `wrapper.render()`and `wrapper.find(Foo).render()`, it didnt' render the first `<div>` and `<Foo> `component didn't render anything HTML output, so the test failed and didn't find the `.in-foo`. I put a wrapping `<div>` around the `.in-foo`on my end, it did rendered the `.in-foo `classname. It seems calling `.render()` may need to have a wrapping tag around the HTML.

ljharb: hmm - I would expect `wrapper.find(Foo).render()` to be the same as `render(<Foo />)`. Is it not?

cadence09: I tested the above example with the following test, the `wrapper.find(Foo).render()` was same as `render(<Foo />)` and `wrapper.find(Foo).render()` had one length, but it didn't find the `.in-foo` node.
```
it('is wrapper.find(Foo).render() == render(<Foo />) ',()=>{
                const wrapper = mount(<Bar />);
                const renderWrapper=render(<Foo/>)
                console.log("a",wrapper.find(Foo).render().html())  // not outputing HTML 
                expect(wrapper.find(Foo).render()).toEqual(renderWrapper)   //test pass
                expect(wrapper.find(Foo).render()).toHaveLength(1)   // test pass
                 expect(wrapper.find(Foo).render().find(".in-foo")).toHaveLength(1)  // test failed
                
            })
```

ljharb: These examples are actually what's used in the tests: https://github.com/enzymejs/enzyme/blob/master/packages/enzyme-test-suite/test/shared/methods/render.jsx#L13-L42

Based on your last test, this is actually correct - if it *is* `in-foo`, `.find` will only search the children, and thus won't find any.

cadence09: If `.find` will only search the children in this case, then should we make .in-foo in the `.render `example as children like the below example?
```	
function Foo( ) {
  return (
     <div>
         <div className="in-foo" />
     </div>
);
}
```

docs/api/ReactWrapper/simulate.md
cadence09: The reason I made the choice is `.props()` return the object, and `.props().x` return the value.

`.props("x")` return `{ name: 'width', value: 10, onChange: [Function: onChange] }`.
`.prop().x`  return 10

docs/api/ReactWrapper/render.md
ljharb: `.filter` or `.is` could work

cadence09: Based on your suggestion, I modified the test to this. Is it better?
`expect(wrapper.find(Foo).render().filter(".in-foo")).to.Have.Length(1);`

docs/api/ShallowWrapper/render.md
cadence09: For this method, should I also use `.filter` to modify it in the same way as the ` .render ` in ReactWrapper ? Like this?
`expect(wrapper.find(Foo).render().filter(".in-foo")).to.Have.Length(1)`

ljharb: That seems great to me - ideally the examples in both the shallow and mount docs are as close to identical as possible, just like the test cases :-)

docs/api/ReactWrapper/render.md
ljharb: Sure, that works! Would it still work if the `.filter` came before the `.render()`?

cadence09: when the `.filter `came before the `.render()`,  it didn't search the children and didn't find any nodes, so the test failed. I used `.children()` to allow the `.filter` to search the matched node in the children. It passed. Does this look good? What do you think? 
`expect(wrapper.find(Foo).filter(".in-foo").render()).to.Have.Length(1);  //test failed. Error: HTML means to run on 1 node but 0 node found`
` expect(wrapper.find(Foo).children().filter(".in-foo").render()).to.Have.Length(1);  // test passed`

ljharb (approved):
Thanks, this looks great :-) I replaced the finds with `.childAt(0)` - to me that reads with more precision regarding the differences between shallow and mount.

PR #2482, opened 2020-12-07
[Fix] Ensure recursive calls to displayNameOfNode uses the adapter's version of the method

### LINKED ISSUE #2481: When using React.memo(React.forwardRef(Component)) displayName will be inaccurate
### Current behavior

When shallow rendering a component that has an inner displayName set wrapped in `memo` and `forwardRef` the `displayName` will be inaccurate.

Discussion in material-ui https://github.com/mui-org/material-ui/issues/23831

Upstream behavior added in React: https://github.com/facebook/react/pull/18925

So given the following:

```javascript
const Foo = () => <div />;
Foo.displayName = 'CustomWrapper';

const MemoForwardFoo = React.memo(React.forwardRef(Foo));
expect(adapter.displayNameOfNode(<MemoForwardFoo />)).to.equal('Memo(ForwardRef(CustomWrapper))');
```

Currently : `Memo([object object])`

### Expected behavior

From the React work, I would expect this to be `CustomWrapper`, but within Enzyme, it seems more like `Memo(ForwardRef(CustomWrapper))` is the expectation

#### API

- [x] shallow
- [ ] mount
- [ ] render

#### Version

| enzyme              |  3.11.0
| react               |  >= 16.6
| react-dom           | >= 16.6
| react-test-renderer |  16.14.0
| adapter (below)     | 

#### Adapter

- [x] enzyme-adapter-react-16
- [ ] enzyme-adapter-react-15
- [ ] enzyme-adapter-react-14
- [ ] enzyme-adapter-react-13
- [ ] others ( )

mejackreed:
Fixes #2481

This is my first contribution to enzyme so happy to update/modify this PR as needed.

### DIFF
--- a/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
+++ b/packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
@@ -863,6 +863,7 @@ class ReactSixteenAdapter extends EnzymeAdapter {
   displayNameOfNode(node) {
     if (!node) return null;
     const { type, $$typeof } = node;
+    const adapter = this;
 
     const nodeType = type || $$typeof;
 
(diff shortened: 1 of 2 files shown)


packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -886,13 +887,13 @@ class ReactSixteenAdapter extends EnzymeAdapter {
       case ContextProvider || NaN: return 'ContextProvider';
       case Memo || NaN: {
         const nodeName = displayNameOfNode(node);
```
mejackreed, line 889: One thing that wasn't clear to me was when/why we should use the adapter here or not.

ljharb: i think if we did, it would be an infinite loop.

ljharb (approved):
Thanks, this seems great.

PR #2500, opened 2021-02-08
fix invoke prop docs

yiminghe:


ljharb (approved):
Thanks, this is a nice improvement

PR #2507, opened 2021-03-14
Add support for passing context to React class based components via contextType

unverbraucht:
Based on the patches posted in https://github.com/enzymejs/enzyme/issues/2189#issuecomment-796416083 by @forivall (thanks!).

Adds changes to ReactSixteenAdapter and simple test case.

### DIFF
--- a/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
+++ b/packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
@@ -666,6 +666,34 @@ describe('shallow', () => {
           expect(consumer.text()).to.equal('foo');
         });
       });
+
+      describe('shallow() on Provider and Consumer through .contextType', () => {
+
+        const { Provider } = React.createContext('howdy!');
+
+        class OuterComponent extends React.Component {
+          render() {
+            return (
+              <Provider value="foo"><InnerComponent /></Provider>
+            );
+          }
+        }
+
+        class InnerComponent extends React.Component {
+          render() {
+            return this.context;
+          }
+        }
+
+        InnerComponent.contextType = Provider;
+
+        it('works on a Provider', () => {
+          const wrapper = shallow(<OuterComponent />);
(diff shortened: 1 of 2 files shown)


unverbraucht:
Hmm, the new unit tests are failing, but they're working locally here. I'll try to reproduce this with the respective build configurations

unverbraucht:
My bad. The tests are expected to fail because they use the unpatched react-shallow-renderer :)

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -666,6 +666,48 @@ describe('shallow', () => {
...
+            return this.context;
+          }
+        }
+
+        InnerComponent.contextType = Provider;
+
+        it('works on a Provider', () => {
+          const wrapper = shallow(<OuterComponent />);
+          const provides = wrapper.find(Provider).dive();
+          const provider = provides.find(InnerComponent).shallow();
```
ljharb, line 702:
```suggestion
          const provider = provides.find(InnerComponent).dive();
```
any reason this wouldn't work?

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -666,6 +666,48 @@ describe('shallow', () => {
           expect(consumer.text()).to.equal('foo');
         });
       });
+
+      describe('shallow() on Provider and Consumer through .contextType', () => {
+
+        const { Provider } = React.createContext('howdy!');
```
ljharb, line 672:
```suggestion
      describeIf(is('>= 16.3'), 'shallow() on Provider and Consumer through .contextType', () => {

        const { Provider } = createContext('howdy!');
```

ljharb (changes requested):
Could we also add a test that changes the provided value, and ensures that the new value also shows up?

ljharb:
We'll also need the same changes in the 16.3 adapter, I suspect.

PR #2509, opened 2021-03-25
[Docs] `shallow`: fix variable name in `prop` example

igor-anosov:
Changed variable in test. I think the previous one was wrong


ljharb (approved):
Thanks!

PR #2525, opened 2021-06-15
Fix `simulateError()` on Memo component

henryqdineen:
We are seeing `TypeError: Cannot read property 'displayNameOfNode' of undefined` errors and I believe I tracked it down to a regression introduced in https://github.com/enzymejs/enzyme/pull/2482. The `displayNameOfNode()` method  now references `this` but with `simulateError()` it ends up being called unbound. My fix is to just always bind `adapter.displayNameOfNode` so `this` is always `adapter`.

Thanks!


ljharb (approved):
Looks great; tests fail without your fix, and pass with them. Thanks!

PR #2430, opened 2020-08-12
[New] add `enzyme-adapter-react-17`

layershifter:
Fixes #2429.

PR #2531, opened 2021-07-17
Add  Unofficial Adapter Package for React 17

MenaiAla:


ljharb (changes requested):
Absolutely not - adding it here would make it official, and imply that any support is provided for it, which none is.

#2430 tracks the official adapter.

PR #2554, opened 2022-03-20
Add contextType support

pablopalacios:
@ljharb I've rebased #2507 , applied your suggestions and added test cases for dynamic context values. Unfortunately, the test which uses `wrappingComponent` does not pass, it keeps rendering the initial context value. The context is correctly updated in the `RootFinder` though, but it's not propagated down the tree.

I tried my best to understand what is happening but I don't understand yet how things work internally yet. If you have any hint to point me in the right direction, I'll gladly move this PR forward.

Thank you for all the great work  :)


ljharb:
@pablopalacios thanks, but opening an additional PR is hugely disruptive - now both PRs must stay open forever, manually kept in sync, until both can be landed. In the future, please post a link to your branch on the original PR, and I can pull in your changes.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
```diff
@@ -690,7 +690,13 @@ class ReactSixteenAdapter extends EnzymeAdapter {
           renderedEl = transformSuspense(renderedEl, renderedEl, { suspenseFallback });
           const { type: Component } = renderedEl;
 
-          const context = getMaskedContext(Component.contextTypes, unmaskedContext);
+          let context;
+          if (Component.contextType) {
+            const Provider = adapter.getProviderFromConsumer(Component.contextType);
```
ljharb, line 695: what happens in React when `.contextType` isn't set to a proper `Consumer`?

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -666,6 +666,78 @@ describe('shallow', () => {
           expect(consumer.text()).to.equal('foo');
         });
       });
+
+      describeIf(is('>= 16.3'), 'shallow() on Provider and Consumer through .contextType', () => {
+
```
ljharb, line 671:
```suggestion
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -666,6 +666,78 @@ describe('shallow', () => {
...
+          }
+        }
+
+        InnerComponent.contextType = Provider;
+
+        describe('rendering the Provider directly', () => {
+          it('renders initial context value', () => {
+            const wrapper = shallow(<OuterComponent value="foo" />);
+            const provides = wrapper.find(Provider).shallow();
+            const provider = provides.find(InnerComponent).shallow();
```
ljharb, line 704:
```suggestion
            const provider = provides.find(InnerComponent).dive();
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -666,6 +666,78 @@ describe('shallow', () => {
...
+
+            expect(provider.text()).to.equal('foo');
+          });
+
+          it('renders updated context value', () => {
+            const wrapper = shallow(<OuterComponent value="foo" />);
+            wrapper.setProps({ value: 'bar' });
+            const provides = wrapper.find(Provider).shallow();
+            const provider = provides.find(InnerComponent).shallow();
```
ljharb, line 713:
```suggestion
            const provider = provides.find(InnerComponent).dive();
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -666,6 +666,78 @@ describe('shallow', () => {
...
+
+            expect(provider.text()).to.equal('bar');
+          });
+        });
+
+        describe('rendereing the Provider through wrappingComponent', () => {
```
ljharb, line 719:
```suggestion
        describe('rendering the Provider through wrappingComponent', () => {
```

ljharb (changes requested):
- https://github.com/enzymejs/enzyme/pull/2507#issuecomment-810529544 (same changes needed in the 16.3 adapter)
 - https://github.com/enzymejs/enzyme/pull/2507#pullrequestreview-624639101 (add a test that changes the provided value, and ensures that the new value also shows up)

ljharb:
I'll take a closer look at test failures when I have time to do so.

pablopalacios:
@ljharb I can close this PR and post the branch in the other one, no problem. Regarding applying the changes in the 16.3 adapter: does it make sense? React just added support for `contextType` in version 16.6 (see: https://reactjs.org/blog/2018/10/23/react-v-16-6.html) or am I missing something?

ljharb:
Please don't; once a PR is opened its a permanent ref on the repo, so it needs to stay open until it's merged.

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
pablopalacios: What do you mean with a proper consumer? A consumer without a parent provider? It should display the default provider value.

ljharb:
hmm, if contextType is only added in 16.6 then you’re right, it should only go in the 16 adapter

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -666,6 +666,77 @@ describe('shallow', () => {
           expect(consumer.text()).to.equal('foo');
         });
       });
+
+      describeIf(is('>= 16.3'), 'shallow() on Provider and Consumer through .contextType', () => {
```
ljharb, line 670:
```suggestion
      describeIf(is('>= 16.6'), 'shallow() on Provider and Consumer through .contextType', () => {
```

packages/enzyme-test-suite/test/ShallowWrapper-spec.jsx
```diff
@@ -666,6 +666,77 @@ describe('shallow', () => {
...
+
+          it('renders updated context value', () => {
+            const wrapper = shallow(<InnerComponent />, {
+              wrappingComponent: WrappingComponent,
+              wrappingComponentProps: { value: 'foo' },
+            });
+            const wrappingComponent = wrapper.getWrappingComponent();
+            wrappingComponent.setProps({ value: 'bar' });
+            wrapper.update();
+            expect(wrapper.text()).to.equal('bar');
```
pablopalacios, line 736: @ljharb this is the only test that fails.

pablopalacios:
@ljharb I finally had time to take a look a this one again. I was able to make the missing test to pass. Could you review it again?

packages/enzyme-adapter-react-16/src/ReactSixteenAdapter.js
ljharb: like, for example, what if `Component.contextType` is a random value - some object, or function, or primitive? What exact check does React do?

pablopalacios:
@ljharb the tests are expected to fail since they use a version of `react-shallow-renderer` that does not have support for `contextType`. We need to merge [this PR](https://github.com/enzymejs/react-shallow-renderer/pull/100) first.

ljharb:
gotcha, thanks, i'll take a look at that one.

ljharb:
@pablopalacios actually enzyme doesn't use `react-shallow-renderer` _at all_. Is there no way to fix it prior to taking on that dependency?

pablopalacios:
@ljharb it depends on react-test-renderer which depends on react-shallow-renderer.

ljharb:
We use v16 of react-test-renderer which does not depend on that - only v17+ does.

PR #2561, opened 2022-06-13
Bump cheerio to @1.0.0-rc.11

vinodkumarsharma276:
Due to recent security vulnerability in nth-checkv1.2.0 which is fetched transitively from enzyme --> cheerio --> css-select --> .... --> nth-checkv1.2.0.

cherrio@1.0.0-rc.11 removes dependency of css-select which ultimately removes dependency of nth-check

### DIFF
--- a/packages/enzyme/package.json
+++ b/packages/enzyme/package.json
@@ -39,7 +39,7 @@
   "license": "MIT",
   "dependencies": {
     "array.prototype.flat": "^1.2.4",
-    "cheerio": "=1.0.0-rc.3",
+    "cheerio": "=1.0.0-rc.11",
     "enzyme-shallow-equal": "^1.0.4",
     "function.prototype.name": "^1.1.4",
     "has": "^1.0.3",


ljharb:
It's worth noting that this is not actually a vulnerability, it's a false positive, at least for enzyme's use case.

packages/enzyme/package.json
```diff
@@ -39,7 +39,7 @@
   "license": "MIT",
   "dependencies": {
     "array.prototype.flat": "^1.2.4",
-    "cheerio": "=1.0.0-rc.3",
+    "cheerio": "=1.0.0-rc.11",
```
ljharb, line 42:
```suggestion
    "cheerio": "^1.0.0-rc.11",
```
i suspect this will break tests, which is why it's pinned to rc3.

ljharb (changes requested):

tbowmo:
cheerio 1.0.0 is now released, so perhaps (if anyone ever comes around) it should be updated to that package

ljharb:
Due to engine requirements, we may never be able to update to it. We’re far more likely to drop the render API.

PR #2571, opened 2022-10-28
[Docs] docs/readmefix : update : grammar fix

rksp25:
hi @zeke @robertknight @roberto Please review my PR


ljharb (changes requested):
These changes don’t fix any grammar - it is correct as-is - and i don’t think they’re readability improvements.

PR #2600, opened 2023-08-03
refactor: flatten the try catch block.

mrdulin:


ljharb:
Thanks! Could you explain why this is an improvement?

mrdulin:
@ljharb Sorry, I don't know what kind of explaination you want. I think I already did. And the code explains itself.

ljharb:
Why is a flattened try/catch block better?

PR #2602, opened 2023-10-11
Fix typo in ShallowWrapper and ReactWrapper docs: MyCStatefullass -> Stateful

josephfrazier:


ljharb (approved):
lol, find/replace strikes again

josephfrazier:
heh, I see it's in ReactWrapper too: https://github.com/enzymejs/enzyme/blob/3a7701ca24a02d55a07596676705c5cdc1a72472/docs/api/ReactWrapper/instance.md?plain=1#L41

Want me to fix that here, or in another PR, or will you take care of it?

ljharb:
In the same PR, please :-) i'll get it if you don't, but please feel free to cover it in the meantime

josephfrazier:
done!

hiSandog:
Nice small docs fix. Since the same typo appears in both `ShallowWrapper` and `ReactWrapper`, it is worth keeping both changes in this PR rather than splitting them. A quick `grep -R "MyCStatefullass" docs packages` before merge would also make sure the typo is not present in another generated/example page.

PR #2615, opened 2026-06-04
[spam]

txmc8a-del:
This workflow triggers Datadog Synthetic tests on pushes and pull requests to the master branch, utilizing the Datadog API and Application Key stored as GitHub secrets.


ljharb (changes requested):
no idea why we'd want this

